PR #17, opened 2021-12-20
re-write Riemann solver to use HLLC-LM

BenWibking:
implement HLLC-LM Riemann solver and use it by default. Ref: Journal of Computational Physics 423 (2020) 109762

Does not appear to have enough dissipation for very low-amplitude linear waves, as used in the HydroWave test.


BenWibking:
With 3d Sedov test, there are instabilities in low-Mach regions of the flow, at least with PVRS states/wavespeeds. Need to try again with Roe wavespeeds.

BenWibking:
Still does not work with linear waves. Very strange!

PR #27, opened 2022-02-14
add auxiliary internal energy equation for high-Mach flows

BenWibking:
Based on the algorithm described in [1] with the eta value from [2]. However, the velocity divergence term is computed following Appendix B of [3], rather than the original method used in [1].

[1] https://ui.adsabs.harvard.edu/abs/2008ASPC..385..273L/abstract
[2] https://flash-x.github.io/Flash-X-docs/Hydro.html
[3] https://ui.adsabs.harvard.edu/abs/2017ApJ...834..144S/abstract


BenWibking:
Run memory checker on HydroHighMach -- somehow getting different L1 norms on CPU and GPU

**Fixed**: there were uninitialized variables in the reference solution multifab.

PR #47, opened 2022-05-25
Use 5th-order hybrid PPM/WENO reconstruction

BenWibking:
This implements the 5th-order, extrema-preserving, hybrid PPM-WENO reconstruction method of Rider, Greenough, and Kamm (2007): https://ui.adsabs.harvard.edu/abs/2007JCoPh.225.1827R/abstract

The number of ghost zones required is the same as PPM (4, if shock flattening is enabled).

Note that only the reconstruction is 5th-order. The overall order of accuracy in space in still 2nd order.

Depends on https://github.com/quokka-astro/quokka/pull/74.


BenWibking:
HydroBlast3D test does not conserve energy.

Is this because of the pressure floor? Or a bug in the flux correction? This issue blocks merging into development.

PR #53, opened 2022-06-17
Refactor simulation initialisation

AstroKriel:
Moved most of the initialisation implementation to AMRSimulation.hpp + user should implement preInitC and setInitC funcs. The user should also specify what physics modules the simulation uses (variables are only defined if physics module is used).


BenWibking:
> Moved most of the initialisation implementation to AMRSimulation.hpp + user should implement preInitC and setInitC funcs. The user should also specify what physics modules the simulation uses (variables are only defined if physics module is used).

As we discussed, let's split this into three separate pull requests:

1. `Physics_Traits`, indexing changes, turning off radiation/hydro with `if constexpr`, etc.
2. Changes to how we set the initial conditions for simulations
3. Face-centered MultiFabs

Changing state_new_ to state_new_cc_ etc. could go in either 2 or 3, whichever is easier.

BenWibking:
FYI- I've added code development guidelines for Quokka here: https://github.com/BenWibking/quokka/blob/development/CONTRIBUTING.md

AstroKriel:
Thanks for linking me the contribution guideline-doc, Ben!

I will separate those changes into different branches and then make a pull-request for each.

Cheers, Neco

PR #62, opened 2022-06-27
add passive scalar test problem

BenWibking:
The previous PR left out a test problem to verify that passive scalars worked, this adds it. (This also fixes subtle undefined behavior in how passive scalars were treated in the Riemann solver, which only affected CPU-only builds using gcc 10+.)

A contact wave (and passive scalar) are advected across a 1D periodic box four times, and the initial and final conditions are compared to see how accurate they were both advected.

Note: with the default PPM reconstruction, some staircasing artefacts are visible in the passive scalar. This can be eliminated by using the hybrid PPM/WENO method, but that method may not be appropriate for multiphysics problems (the numerical dissipation actually might be _too_ small).


BenWibking:
There is either undefined behavior or a gcc compiler bug. All CPU tests built with `AMReX_SPACEDIM=1` pass with gcc 8.5.0 and 9.5.0 (and with Clang 14.0.5), but the PassiveScalar test fails (on CPU only!) with gcc 10.3.0, 11.1.0, 11.2.0, or 12.1.0.

No asserts are triggered when PassiveScalar is built in debug mode with any of these compilers. The only warnings that appear are for unrecognized pragmas (that are intended to suppress warnings when compiling with nvcc).

**Tests pass when gcc 10.3.0 is used with -O0, gcc 11.2.0 is used with -O0 or -O1, or when gcc 12.1.0 is used with -O0.** The PassiveScalar test fails at -O2 or above for gcc 11.2.0 and at -O1 or above for gcc 10.3.0 and 12.1.0.

BenWibking:
I think gcc 10+ is optimizing out the call to the initial constructor, since it sees the array elements are unused until the assignment from an initializer list (which does not actually populate all of the array elements...):

```
    // initialize all components to zero
    quokka::valarray<double, fluxdim> D_L(0.);
    quokka::valarray<double, fluxdim> D_R(0.);
    quokka::valarray<double, fluxdim> D_star(0.);

    // N.B.: quokka::valarray is written to allow assigning <= fluxdim
    // components, so this works even if there are more components than
    // enumerated in the initializer list
    if constexpr (DIR == FluxDir::X1) {
      D_L = {0., 1., 0., 0., u_L, 0., 0.};
      D_R = {0., 1., 0., 0., u_R, 0., 0.};
      D_star = {0., 1., 0., 0., S_star, 0., 0.};
    } else if constexpr (DIR == FluxDir::X2) {
      D_L = {0., 0., 1., 0., u_L, 0., 0.};
      D_R = {0., 0., 1., 0., u_R, 0., 0.};
      D_star = {0., 0., 1., 0., S_star, 0., 0.};
    } else if constexpr (DIR == FluxDir::X3) {
      D_L = {0., 0., 0., 1., u_L, 0., 0.};
      D_R = {0., 0., 0., 1., u_R, 0., 0.};
      D_star = {0., 0., 0., 1., S_star, 0., 0.};
    }
```

PR #68, opened 2022-07-02
[WIP] overlapping communication/computation with inner-outer updates

BenWibking:
Test suite does not pass.


src/RadhydroSimulation.hpp
```diff
@@ -469,152 +474,411 @@ void RadhydroSimulation<problem_t>::advanceHydroAtLevel(int lev, amrex::Real tim
...
-							  << i
-							  << ", ncells = "
-							  << redoFlag.sum<amrex::RunOn::Device>(0)
-							  << "\n";
+	#pragma omp parallel sections
+		{
+			#pragma omp section
+			{
+				// update ghost zones [intermediate stage stored in state_new_]
+				fillBoundaryConditions(state_new_[lev], state_new_[lev], lev, time + dt_lev);
```
BenWibking, line 660: Because state_new_[lev] is used for both holding the result of the stage 1 update (and therefore filling the boundary conditions for stage 2) and writing the result of the stage 2 update, there is a race condition introduced here. For non-overlapping updates, this is not a problem, since the hydro update is done all at once for each FAB. When overlapping updates are used, FABs are only partially updated before they are (partially) overwritten. Assuming that the inner updates run first, this gives an inconsistent initial condition for the outer updates.

This can be fixed by allocated a separate amrex::Multifab to store the result of the stage 1 calculation and calling fillBoundaryConditions with that Multifab.

src/RadhydroSimulation.hpp
```diff
@@ -469,152 +474,411 @@ void RadhydroSimulation<problem_t>::advanceHydroAtLevel(int lev, amrex::Real tim
...
-	fillBoundaryConditions(state_old_[lev], state_old_[lev], lev, time);
+  std::chrono::steady_clock::time_point begin = std::chrono::steady_clock::now();
+
+  amrex::MFIter::allowMultipleMFIters(true);
+
+  #pragma omp parallel sections
+  {
+    #pragma omp section
+    {
+      AMREX_CUDA_SAFE_CALL(cudaSetDevice(amrex::Gpu::Device::deviceId()));
```
BenWibking: There are AMReX wrapper functions that handle all 3 GPU architectures via ifdefs. If there's a function that doesn't have an AMReX wrapper for this, let me know.

src/RadhydroSimulation.hpp
```diff
@@ -469,152 +474,411 @@ void RadhydroSimulation<problem_t>::advanceHydroAtLevel(int lev, amrex::Real tim
...
+            incrementFluxRegisters(iter, fr_as_crse, fr_as_fine, expandedFluxes, lev,
+                      fluxScaleFactor * dt_lev);
+          }
+        }
+      }
+   }
+
+    #pragma omp section
+    {
+      AMREX_CUDA_SAFE_CALL(cudaSetDevice(amrex::Gpu::Device::deviceId()));
```
BenWibking: Same issue as above -- we want Quokka to work out of the box on AMD and Intel GPUs, so we need to replace this with an AMReX portable wrapper function, if it exists.

BenWibking:
@dliptai There is a race condition introduced in the second stage of the RK integrator -- can you fix this?

src/RadhydroSimulation.hpp
```diff
@@ -469,152 +474,411 @@ void RadhydroSimulation<problem_t>::advanceHydroAtLevel(int lev, amrex::Real tim
 		fluxScaleFactor = 1.0;
 	}
 
-	// update ghost zones [old timestep]
-	fillBoundaryConditions(state_old_[lev], state_old_[lev], lev, time);
+  std::chrono::steady_clock::time_point begin = std::chrono::steady_clock::now();
+
+  amrex::MFIter::allowMultipleMFIters(true);
+
+  #pragma omp parallel sections
```
BenWibking, line 526: Is there a good reason to use sections? I think this can be replaced with `#pragma omp parallel num_threads(2)`, and then within this construct, we can use an if statement to check the thread ID with `omp_get_thread_num()`.

This will enable us to use MPI_THREAD_FUNNELED (which has better performance) instead of MPI_THREAD_MULTIPLE. (See https://cvw.cac.cornell.edu/hybrid/overlapcommwork)

BenWibking (commented):
I've added some comments on the implementation. @markkrumholz can you take a look at this as well to see if you have anything to add?

BenWibking:
FYI- Github CI doesn't run on this PR due to the merge conflict. This is a design flaw in the CI. If you are able to fix the merge conflict, then I can run the CI tests on Github Actions.

src/RadhydroSimulation.hpp
```diff
@@ -514,61 +519,325 @@ void RadhydroSimulation<problem_t>::advanceHydroAtLevel(int lev, amrex::Real tim
...
+      AMREX_ASSERT(!state_old_[lev].contains_nan()); // check ghost cells
+
+      for (amrex::MFIter iter(state_new_[lev]); iter.isValid(); ++iter) {
+        const amrex::Box &realZones = iter.validbox(); // 'validbox' == exclude ghost zones
+
+          //6 should be 2*ndims
+        for (int i=1; i<=AMREX_SPACEDIM*2; ++i) {
+          amrex::Box computeRange = realZones;
+
+          computeRange.grow(nghost_);
```
BenWibking, line 546: I think this line is the source of all the test failures. If you print out each outer box with `amrex::Print() << computeRange << std::endl;` on a 1D test problem, you should see that this screws up all the indices of the outer boxes.

BenWibking: On my fork, I just deleted this line, and almost everything works. There are still test failures for problems that have blocking factors that are less than `2*nghosts_`, which is an edge case that this code ignores.

BenWibking:
This doesn't compile unless I change `device_id` to `amrex::Gpu::Device::deviceId()`.

conradtchan:
> This doesn't compile unless I change `device_id` to `amrex::Gpu::Device::deviceId()`.

Ah, you're right. I'd just copied those lines from elsewhere in the code, but didn't get up to compiling and testing because my understanding was that the development would happen on your branch instead.

The fixes here are unfinished; sorry about that.

BenWibking:
> > This doesn't compile unless I change `device_id` to `amrex::Gpu::Device::deviceId()`.
> 
> Ah, you're right. I'd just copied those lines from elsewhere in the code, but didn't get up to compiling and testing because my understanding was that the development would happen on your branch instead.

Yes, that's fine. Just wanted to check whether you meant it to call that, or this was actually another variable defined in some code that didn't get committed.

BenWibking:
This version doesn't give any cuda context errors, but my version of inner-outer updates without omp sections does. All the other differences don't appear relevant. This is very odd.

Edit: the solution was to set OMP_NUM_THREADS=2.

PR #69, opened 2022-07-04
[WIP] inner-outer updates with communication overlap

BenWibking:
Split each RK stage into "inner" and "outer" updates on each FAB, where the inner updates are done on regions defined by `iter.validbox().ngrow(-nghost_)`. Then we can effectively overlap MPI communication and computation.

Because the inner updates do not need ghost cells, they can be computed immediately. The outer updates must wait until the ghost cell communication and unpacking has completed. Since GPU compute kernels do not run on multiple streams in practice (too many registers are used), only the communication (including pack/unpack) and inner updates are run concurrently. Once those complete, the outer updates are done. Finally, the flux registers are updated.

**Note: First order flux correction is removed from this version of the code**, since it prevents all of the kernels from being launched asynchronously within an MFIter loop. It can be moved to a separate MFIter, analogous to how the flux register update is done.


BenWibking:
Ok, it works now in 1D and 2D (except for HydroHighMach) without AMR.

For some reason, YAFluxRegister does not work correctly with split updates, so an additional 3 MultiFabs must be allocated to store the fluxes. Once this workaround is added to RadhydroSimulation, AMR should work as well.

Edit: AMR now works. The fluxes are temporarily stored on MultiFabs allocated during each level update.

BenWibking:
TODO: figure out why HydroHighMach test fails.

BenWibking:
Runs clean under Debug+ASAN, test suite passes. Unclear why GPU runs are crashing randomly with memory address errors.

BenWibking:
Correct GPU results depend on https://github.com/AMReX-Codes/amrex/pull/2871

BenWibking:
Error on Gadi nodes that does not appear on local GPU node:
```
  what():  CUDA error 201 in file /g/data/jh2/bw0729/quokka/extern/amrex/Src/Base/AMReX_GpuDevice.cpp line 645: invalid device context
[gadi-gpu-v100-0096:35415] *** Process received signal ***
[gadi-gpu-v100-0096:35415] Signal: Aborted (6)
[gadi-gpu-v100-0096:35415] Signal code:  (-6)
[gadi-gpu-v100-0096:35415] [ 0] /lib64/libpthread.so.0(+0x12ce0)[0x14771c2adce0]
[gadi-gpu-v100-0096:35415] [ 1] /lib64/libc.so.6(gsignal+0x10f)[0x14771b12ca9f]
[gadi-gpu-v100-0096:35415] [ 2] /lib64/libc.so.6(abort+0x127)[0x14771b0ffe05]
[gadi-gpu-v100-0096:35415] [ 3] /g/data/jh2/bw0729/spack/opt/spack/linux-rocky8-cascadelake/gcc-11.1.0/gcc-11.3.0-geikkjqs2k2apo5ooftqf4xsmhsj4773/lib64/libstdc++.so.6(+0xa97da)[0x14771bd2a7da]
[gadi-gpu-v100-0096:35415] [ 4] /g/data/jh2/bw0729/spack/opt/spack/linux-rocky8-cascadelake/gcc-11.1.0/gcc-11.3.0-geikkjqs2k2apo5ooftqf4xsmhsj4773/lib64/libstdc++.so.6(+0xb4fea)[0x14771bd35fea]
[gadi-gpu-v100-0096:35415] [ 5] /g/data/jh2/bw0729/spack/opt/spack/linux-rocky8-cascadelake/gcc-11.1.0/gcc-11.3.0-geikkjqs2k2apo5ooftqf4xsmhsj4773/lib64/libstdc++.so.6(+0xb4049)[0x14771bd35049]
[gadi-gpu-v100-0096:35415] [ 6] /g/data/jh2/bw0729/spack/opt/spack/linux-rocky8-cascadelake/gcc-11.1.0/gcc-11.3.0-geikkjqs2k2apo5ooftqf4xsmhsj4773/lib64/libstdc++.so.6(__gxx_personality_v0+0x87)[0x14771bd35777]
[gadi-gpu-v100-0096:35415] [ 7] /g/data/jh2/bw0729/spack/opt/spack/linux-rocky8-cascadelake/gcc-11.1.0/gcc-11.3.0-geikkjqs2k2apo5ooftqf4xsmhsj4773/lib64/libgcc_s.so.1(+0x11524)[0x14771b4b4524]
[gadi-gpu-v100-0096:35415] [ 8] /g/data/jh2/bw0729/spack/opt/spack/linux-rocky8-cascadelake/gcc-11.1.0/gcc-11.3.0-geikkjqs2k2apo5ooftqf4xsmhsj4773/lib64/libgcc_s.so.1(_Unwind_RaiseException+0x311)[0x14771b4b4bc1]
[gadi-gpu-v100-0096:35415] [ 9] /g/data/jh2/bw0729/spack/opt/spack/linux-rocky8-cascadelake/gcc-11.1.0/gcc-11.3.0-geikkjqs2k2apo5ooftqf4xsmhsj4773/lib64/libstdc++.so.6(__cxa_throw+0x3c)[0x14771bd3629c]
[gadi-gpu-v100-0096:35415] [10] ./build/src/HydroBlast3D/test_hydro3d_blast[0x40cc42]
[gadi-gpu-v100-0096:35415] [11] ./build/src/HydroBlast3D/test_hydro3d_blast[0x5f9c60]
[gadi-gpu-v100-0096:35415] [12] ./build/src/HydroBlast3D/test_hydro3d_blast[0x5a49c4]
[gadi-gpu-v100-0096:35415] [13] ./build/src/HydroBlast3D/test_hydro3d_blast[0x47b354]
[gadi-gpu-v100-0096:35415] [14] /g/data/jh2/bw0729/spack/opt/spack/linux-rocky8-cascadelake/gcc-11.1.0/gcc-11.3.0-geikkjqs2k2apo5ooftqf4xsmhsj4773/lib64/libgomp.so.1(+0x1d596)[0x14771b6d9596]
[gadi-gpu-v100-0096:35415] [15] /lib64/libpthread.so.0(+0x81cf)[0x14771c2a31cf]
[gadi-gpu-v100-0096:35415] [16] /lib64/libc.so.6(clone+0x43)[0x14771b117dd3]
[gadi-gpu-v100-0096:35415] *** End of error message ***
scripts/gpu_wrapper_sedov_256.sh: line 4: 35415 Aborted                 ./build/src/HydroBlast3D/test_hydro3d_blast tests/blast_unigrid_256.in
```

BenWibking:
Reference: https://developer.nvidia.com/blog/cuda-pro-tip-always-set-current-device-avoid-multithreading-bugs/

BenWibking:
The solution is to set the number of omp threads _before_ initializing AMReX:
```
  // Ensure the OpenMP runtime starts 2 threads (used for inner-outer updates)
  //  NOTE: This *must* be called before AMReX is initialized
  const int nthreads = 2;
  omp_set_num_threads(nthreads);
```

BenWibking:
For HydroBlast3D with a 256^3 uniform grid, the non-overlapping reference run achieves 80.1 Mupdates/GPU/s on 1 node (4 V100 GPUs). This version of inner-outer overlapping achieves 79.7 Mupdates/GPU/s, so it's 0.6% slower on a single node.

On 2 nodes (8 GPUs), non-overlapping gets 72.2 Mupdates/GPU/s, while the overlapping version gets 71.4 Mupdates/GPU/s, which is 1.1% slower.

On 8 nodes (32 GPUs) with a 1024^3 grid, non-overlapping gets 83.4 Mupdates/GPU/s, while overlapping (nwidth=4) gets 82.9 Mupdates/GPU/s, which is 0.6% slower. (These were built with -O3.) However, **overlapping with nwidth=32 achieves 88.9 Mupdates/GPU/s, which is 6.6% faster than the reference implementation**.

There is likely to be a larger performance benefit for problems with radiation and/or AMR.

BenWibking:
Modifying the kernels to do something like this should improve performance:

overlap_kernel_modification

PR #83, opened 2022-07-26
initialise auxiliary gas internal energy in test problems

AstroKriel:
Fixed the value that gas internal energy was being initialized to. It is now Eint = Etot - Ekin, where `Ekin = (px*px + py*py + pz*pz)/(2*rho)`.


src/RadForce/test_radiation_force.cpp
```diff
@@ -133,11 +133,12 @@ void RadhydroSimulation<TubeProblem>::setInitialConditionsAtLevel(int lev) {
       state(i, j, k, RadSystem<TubeProblem>::x2RadFlux_index) = 0;
       state(i, j, k, RadSystem<TubeProblem>::x3RadFlux_index) = 0;
 
-      state(i, j, k, RadSystem<TubeProblem>::gasDensity_index) = rho;
       state(i, j, k, RadSystem<TubeProblem>::x1GasMomentum_index) = rho * vel;
       state(i, j, k, RadSystem<TubeProblem>::x2GasMomentum_index) = 0;
       state(i, j, k, RadSystem<TubeProblem>::x3GasMomentum_index) = 0;
+      state(i, j, k, RadSystem<TubeProblem>::gasDensity_index) = rho;
       state(i, j, k, RadSystem<TubeProblem>::gasEnergy_index) = 0;
+      state(i, j, k, RadSystem<TubeProblem>::gasInternalEnergy_index) = -((rho * vel)*(rho * vel))/(2*rho);
```
BenWibking: the gas internal energy should not be initialized to be < 0. however, in this case, the problem is isothermal, so it won't affect the solution. here we should just set it to zero.

src/RadForce/test_radiation_force.cpp
```diff
@@ -192,10 +193,11 @@ AMRSimulation<TubeProblem>::setCustomBoundaryConditions(
     consVar(i, j, k, RadSystem<TubeProblem>::x3RadFlux_index) = 0;
 
     consVar(i, j, k, RadSystem<TubeProblem>::gasDensity_index) = rho;
+    consVar(i, j, k, RadSystem<TubeProblem>::gasEnergy_index) = 0;
+    consVar(i, j, k, RadSystem<TubeProblem>::gasInternalEnergy_index) = -((rho * vel)*(rho * vel))/(2*rho);
```
BenWibking: just set it to zero here. the gas internal energy can't be negative (physically). this problem is isothermal, so it doesn't use either the gas energy or gas internal energy variables.

src/RadhydroShock/test_radhydro_shock.cpp
```diff
@@ -68,7 +68,7 @@ template <> struct RadSystem_Traits<ShockProblem> {
 template <> struct HydroSystem_Traits<ShockProblem> {
   static constexpr double gamma = gamma_gas;
   static constexpr bool reconstruct_eint = true;
-  static constexpr int nscalars = 1;       // number of passive scalars
+  static constexpr int nscalars = 0;       // number of passive scalars
```
BenWibking: I set this to 1 just to make sure that having a passive scalar doesn't mess up the hydro solver. So this is important to keep as a test.

src/RadhydroShell/test_radhydro_shell.cpp
```diff
@@ -232,6 +232,10 @@ void RadhydroSimulation<ShellProblem>::setInitialConditionsAtLevel(int lev) {
       state(i, j, k, RadSystem<ShellProblem>::x1RadFlux_index) = Frad_xyz;
       state(i, j, k, RadSystem<ShellProblem>::x2RadFlux_index) = Frad_xyz;
       state(i, j, k, RadSystem<ShellProblem>::x3RadFlux_index) = Frad_xyz;
+
+      state(i, j, k, RadSystem<ShellProblem>::gasEnergy_index) = 0.;
```
BenWibking: This is not correct. The gas density and energy are set a few lines above in this file. They are not zero.

src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp
```diff
@@ -71,7 +71,7 @@ template <> struct RadSystem_Traits<ShockProblem> {
 template <> struct HydroSystem_Traits<ShockProblem> {
 	static constexpr double gamma = gamma_gas;
 	static constexpr bool reconstruct_eint = true;
-  	static constexpr int nscalars = 0;       // number of passive scalars
+  static constexpr int nscalars = 0;       // number of passive scalars
```
BenWibking: is there a reason for the whitespace change here?

src/simulation.hpp
```diff
@@ -943,6 +943,9 @@ void AMRSimulation<problem_t>::MakeNewLevelFromScratch(
   // set state_new_[lev] to desired initial condition
   setInitialConditionsAtLevel(level);
 
+  amrex::IntVect iv(AMREX_D_DECL(19, 0, 5));
+  print_state(state_new_[level], iv);
```
BenWibking: can you delete this?

tests/highmach_output.csv
```diff
@@ -0,0 +1,129 @@
+# x density velocity pressure
```
BenWibking: can you remove this file from the PR?

BenWibking (changes requested):
Thanks for fixing these issues. There are a few changes that need to be made. I've added comments to the source. Let me know if you have questions.

src/RadhydroShock/test_radhydro_shock.cpp
BenWibking: But we should initialize the scalar to some constant value for this problem. It could be zero, or one, or pi.

src/RadhydroShell/test_radhydro_shell.cpp
AstroKriel: On lines 236 and 237 the gas energy and density are being set to zero, no? Namely:
state(i, j, k, RadSystem<ShellProblem>::gasEnergy_index) = 0.;
state(i, j, k, RadSystem<ShellProblem>::gasDensity_index) = 0.;

BenWibking: Those are lines you added, right?

The density is set on line https://github.com/BenWibking/quokka/blob/e6391c605b55afbce73f194c86114bb98a0e54e9/src/RadhydroShell/test_radhydro_shell.cpp#L224

BenWibking: The total gas energy is set here:
https://github.com/BenWibking/quokka/blob/e6391c605b55afbce73f194c86114bb98a0e54e9/src/RadhydroShell/test_radhydro_shell.cpp#L228

src/RadTube/test_radiation_tube.cpp
```diff
@@ -173,6 +174,7 @@ AMRSimulation<TubeProblem>::setCustomBoundaryConditions(
     const double Ekin = 0.5 * (x1Mom * x1Mom) / rho0;
     consVar(i, j, k, RadSystem<TubeProblem>::gasEnergy_index) = Egas + Ekin;
     consVar(i, j, k, RadSystem<TubeProblem>::gasDensity_index) = rho0;
+    consVar(i, j, k, RadSystem<TubeProblem>::gasInternalEnergy_index) = (Egas + Ekin) - (x1Mom*x1Mom)/(2*rho0);
```
BenWibking: This is not necessary. You are adding the kinetic energy with Ekin and the just subtracting it again. You can just set this to Egas.

src/RadTube/test_radiation_tube.cpp
```diff
@@ -195,6 +197,7 @@ AMRSimulation<TubeProblem>::setCustomBoundaryConditions(
     const double Ekin = 0.5 * (x1Mom * x1Mom) / rho1;
     consVar(i, j, k, RadSystem<TubeProblem>::gasEnergy_index) = Egas + Ekin;
     consVar(i, j, k, RadSystem<TubeProblem>::gasDensity_index) = rho1;
+    consVar(i, j, k, RadSystem<TubeProblem>::gasInternalEnergy_index) = (Egas + Ekin) - (x1Mom*x1Mom)/(2*rho1);
```
BenWibking: Just set this to Egas.

src/RadhydroShell/test_radhydro_shell.cpp
BenWibking: Delete these lines.

BenWibking (changes requested):

src/hydro_system.hpp
```diff
@@ -138,7 +138,8 @@ class HydroSystem : public HyperbolicSystem<problem_t> {
   // C++ does not allow constexpr to be uninitialized, even in a templated
   // class!
   static constexpr double gamma_ = HydroSystem_Traits<problem_t>::gamma;
-  static constexpr double cs_iso_ = HydroSystem_Traits<problem_t>::cs_isothermal;
+  static constexpr double cs_iso_ =
```
BenWibking: since there's no substantive changes to this file, can you revert the style changes for this file?

BenWibking:
The CI is failing only due to a bug in SonarCloud: https://community.sonarsource.com/t/sonarcloud-pull-request-analysis-fon-github-error/29290. It looks like the tests actually pass.

(Note to self: possible solutions are here https://github.com/Checkmarx/kics/pull/3765 and possibly here: https://docs.sonarcloud.io/improving/pull-request-analysis/)

src/RadhydroShell/test_radhydro_shell.cpp
AstroKriel: Fixed.

src/RadTube/test_radiation_tube.cpp
AstroKriel: Fixed.

src/RadTube/test_radiation_tube.cpp
AstroKriel: Fixed.

src/hydro_system.hpp
AstroKriel: Done.

BenWibking (approved):

PR #88, opened 2022-07-27
add Physics_Traits compile-time options

AstroKriel:
Physics modules (i.e., hydro, radiation, etc) are now enabled at compile time and should be defined by the user via the Physics_Traits struct.


src/Cooling/test_cooling.cpp
```diff
@@ -39,7 +39,16 @@ template <> struct HydroSystem_Traits<CoolingTest> {
...
   static constexpr bool reconstruct_eint = true;
-  static constexpr int nscalars = 0;       // number of passive scalars
+};
+
+template <> struct Physics_Traits<CoolingTest> {
+  static constexpr bool is_hydro_enabled = true;
+  static constexpr bool is_radiation_enabled = false;
+  static constexpr bool is_mhd_enabled = false;
+  static constexpr bool is_primordial_chem_enabled = false;
+  static constexpr bool is_metalicity_enabled = false;
```
BenWibking: Can we combine these two options into `is_chemistry_enabled`? We will have several types of chemistry, not just primordial chemistry. The use of passive scalars for a metallicity field should be controlled by a supernova feedback option, but let's leave that for a future PR that adds that physics as well.

src/Cooling/test_cooling.cpp
```diff
@@ -39,7 +39,16 @@ template <> struct HydroSystem_Traits<CoolingTest> {
   static constexpr double gamma = 5. / 3.; // default value
   // if true, reconstruct e_int instead of pressure
   static constexpr bool reconstruct_eint = true;
-  static constexpr int nscalars = 0;       // number of passive scalars
+};
+
+template <> struct Physics_Traits<CoolingTest> {
+  static constexpr bool is_hydro_enabled = true;
+  static constexpr bool is_radiation_enabled = false;
+  static constexpr bool is_mhd_enabled = false;
```
BenWibking: Can you make it clear in the comments that setting `is_mhd_enabled` to true doesn't actually do anything yet? Alternatively, we could just add `is_mhd_enabled` in a future MHD-specific PR.

BenWibking (changes requested):
Looks good. Just a request for two changes: i) change the chemistry option to `is_chemistry_enabled` and delete the metallicity option for now, and ii) make it clear that the chemistry and MHD options don't do anything yet.

src/Cooling/test_cooling.cpp
BenWibking: You could add a `static_assert` in the RadhydroSimulation constructor to ensure that `is_chemistry_enabled` and `is_mhd_enabled` are set to false, just to make sure users don't try to activate them yet.

AstroKriel: Yes, I think it would make sense to remove all references to is_mhd_enabled, and then I will add it in the mhd-branch

AstroKriel: Also, yep, I'll add the assert.

AstroKriel: Done.

src/Cooling/test_cooling.cpp
AstroKriel: Sounds good. Done.

src/Cooling/test_cooling.cpp
```diff
@@ -44,8 +44,7 @@ template <> struct HydroSystem_Traits<CoolingTest> {
 template <> struct Physics_Traits<CoolingTest> {
   static constexpr bool is_hydro_enabled = true;
   static constexpr bool is_radiation_enabled = false;
-  static constexpr bool is_mhd_enabled = false;
-  static constexpr bool is_primordial_chem_enabled = false;
+  static constexpr bool is_chemistry_enabled = false;
   static constexpr bool is_metalicity_enabled = false;
```
BenWibking: Can you remove the `is_metalicity_enabled` variable? I think we want to approach this somewhat differently. The user can add metallicity fields by setting `numPassiveScalars` for now.

BenWibking (changes requested):
Just one change requested.

src/Cooling/test_cooling.cpp
AstroKriel: Yes, that should be removed now. At the moment only is_hydro_enabled, is_radiation_enabled, is_chemistry_enabled, and numPassiveScalars are defined in Physics_Traits.

BenWibking (approved):

PR #98, opened 2022-08-02
Fix periodicity bug in the rad-hydro-shell simulation

AstroKriel:
Removed duplicate and incorrect state variable initialization, and fixed the incorrect periodicity set in the problem parameter file.


BenWibking:
Can you get rid of the unrelated changes?

PR #99, opened 2022-08-02
Added include guards for the physics-trait header files

AstroKriel:
Added include gaurds for the physics-trait (i.e., physics_numVars.hpp and physics_info.hpp) header files.


BenWibking:
It looks like there are some unrelated changes in here. Can you remove those?

PR #106, opened 2022-08-09
Refactor simulation initialisation

AstroKriel:
Boilerplate initialization code has been moved to the AMRSimulation class. Two functions are exposed to the user, first, which allows the user to perform a computation before initialization (for example to read in a table), and a second is called while the initialization code iterates over the simulation domain. These functions have empty implementations by default, so the user will need to implement them when relevant.


src/grid.hpp
```diff
@@ -0,0 +1,25 @@
+#include <AMReX.H>
```
BenWibking, line 4: This needs an include guard.

src/grid.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include <AMReX_Array.H>
+#include <AMReX_Array4.H>
+#include <AMReX_Box.H>
+#include <AMReX_REAL.H>
+
+enum class centering { cc=0, fc, ec };
```
BenWibking: put the enums and the struct inside a `namespace quokka { .. }` block

src/simulation.hpp
```diff
@@ -942,6 +969,9 @@ void AMRSimulation<problem_t>::MakeNewLevelFromScratch(
   // set state_new_[lev] to desired initial condition
   setInitialConditionsAtLevel(level);
 
+  amrex::IntVect iv(AMREX_D_DECL(19, 0, 5));
+  print_state(state_new_[level], iv);
+
```
BenWibking: Delete this debugging code.

BenWibking (changes requested):
Looks good overall. The main thing is I think we want to put the `grid` struct inside the `quokka` namespace. This means it will be referenced with `quokka::grid` instead of just `grid`.

src/simulation.hpp
AstroKriel: Ugh, I thought I had already deleted this. Done now (again).

src/grid.hpp
AstroKriel: Done.

src/grid.hpp
AstroKriel: Done.

BenWibking (approved):

BenWibking:
@AstroKriel can you take a look at the build errors?

AstroKriel:
> @AstroKriel can you take a look at the build errors?

Yep, fixed the errors. The errors were because I forgot to add the quokka namespace call to all references of the grid-struct. I have checked everything compiles fine now.

BenWibking:
> > @AstroKriel can you take a look at the build errors?
> 
> Yep, fixed the errors. The errors were because I forgot to add the quokka namespace call to all references of the grid-struct. I have checked everything compiles fine now.

The GPU build is still failing. Can you take a look at that?

AstroKriel:
> > > @AstroKriel can you take a look at the build errors?
> > 
> > 
> > Yep, fixed the errors. The errors were because I forgot to add the quokka namespace call to all references of the grid-struct. I have checked everything compiles fine now.
> 
> The GPU build is still failing. Can you take a look at that?

It looks like the build error occurred because the reference to the state array was stored in grid_vec[...].array, and that vector lives on the CPU. There isn't an operator implemented to access vector elements stored on the CPU from the GPU, and that is why the error was being thrown. I fixed this by extracting the reference to the state variable and passing that reference directly to computeWaveSolution(...) instead of accessing the reference via the grid_vec[...] vector in the function call.

BenWibking:
> It looks like the build error occurred because the reference to the state array was stored in grid_vec[...].array, and that vector lives on the CPU. There isn't an operator implemented to access vector elements stored on the CPU from the GPU, and that is why the error was being thrown. I fixed this by extracting the reference to the state variable and passing that reference directly to computeWaveSolution(...) instead of accessing the reference via the grid_vec[...] vector in the function call.

Yeah, the fact that `std::vector` is not usable from the GPU requires a lot of workarounds in general. This looks like a reasonable way to do it. I'll re-run the GPU tests and merge if it passes.

BenWibking:
@AstroKriel something is still not right. GPU build still fails.

AstroKriel:
Ah yep, it looks like the state reference (ie grid_vec[...].array) needs to be extracted outside all the ParFor loops, because when compiled on GPU, the same problem as before occurs. I'll fix this and see if the problem persists / if there are other problems.

BenWibking:
@AstroKriel there are GPU build failures in the HydroBlast3D problem.

PR #110, opened 2022-09-02
add Rayleigh-Taylor test problems

BenWibking:
This adds the 2D multimode test problem from the Athena Code Test Page (https://www.astro.princeton.edu/~jstone/Athena/tests/rt/rt.html) and a 3D generalization. The gravitational force is added with Strang splitting.

In the nonlinear regime, the mixing length may be compared to laboratory experiments.


BenWibking:
Density slice for 3D problem (128^3):
AMR0000

PR #121, opened 2022-09-28
enable gpu-aware MPI by default

BenWibking:
It turns out GPU-aware MPI works great, as long as managed memory is disabled and all CUDA devices are visible from each MPI process. These conditions mean that with OpenMPI, the `cuda_ipc` transport will be used.

On-node scaling is now **~99%** efficient on Delta, Gadi.

_Note_: this will need to be re-tested for Setonix and Frontier.

For AMD devices, see also: https://docs.amd.com/bundle/AMD-Instinct-MI250-High-Performance-Computing-and-Tuning-Guide-v5.3/page/GPU-Enabled_MPI.html


BenWibking:
For the HydroBlast3D problem (with `max_grid_size` set to $256^3$), we get:

1 A100: `217.6 Mupdates/s` ( $256^3$ problem)
4 A100s / 1 node: `864.4 Mupdates/s` ( $512^3$ problem)

for a scaling efficiency of >99%.

BenWibking:
GPU tests are failing because `fextract` does not work if MultiFabs are not accessible from the host.

PR #125, opened 2022-10-07
Add ability to create and work with face-centred quantities

AstroKriel:
Introduced infrastructure to create and initialize face-centered (fc) quantities. Additionally fixed a bug in the initialization phase where the vector of grid elements being passed to the user-init function would store the incorrect indexRange reference. I also generalized some of the FillPatch functions to work with both cc and fc quantities.


BenWibking:
Ok, this is too big to review in a single PR. Can you split this up into two PRs, i) for the changes in the problem initialization and ii) for everything else? The second PR shouldn't have to modify the problem files.

Can we consider keeping `state_new_` and `state_old_` as the cell-centered MultiFab variable names? This will prevent having to change (and me having to review) every single problem file. At least, let's make that a separate PR. Otherwise, I can't tell what changes are significant and what changes are just renaming variables by reading the diff.

BenWibking:
I think it would be helpful to have the FillPatch changes as a separate PR. Those will require careful testing to make sure they are working correctly.

AstroKriel:
Yeah, I can break these up into different PRs. Just so you know, ther are only two changes in the problem files: (1) they now specify that mhd is off, and (2) a very minor change to the user-init func, where the func now works on a single grid-element, rather a grid-vector.

AstroKriel:
I will create a PR for renaming variables (e.g., state -> state_cc, etc), which we can merge before this PR, and that way most of the minor changes will go away. Does that sound alright?

BenWibking:
> Yeah, I can break these up into different PRs. Just so you know, ther are only two changes in the problem files: (1) they now specify that mhd is off, and (2) a very minor change to the user-init func, where the func now works on a single grid-element, rather a grid-vector.

That's a good point. Can the changes to the problem files go in their own PR? Is `is_mhd_enabled` used anywhere else?

BenWibking:
> I will create a PR for renaming variables (e.g., state -> state_cc, etc), which we can merge before this PR, and that way most of the minor changes will go away. Does that sound alright?

That sounds good to me.

AstroKriel:
Cool! I will do that. 

> Is is_mhd_enabled used anywhere else?

Yep, outside of being referenced once to create fc-variable names in the RadHydro class constructor, the `is_mhd_enabled` only appears in the problem files to turn mhd off.

BenWibking:
@AstroKriel somehow the PR to rename the variables confused the merge algorithm. Can you fix the merge conflicts here?

AstroKriel:
Yep, and done. There weren't too many merge conflicts. It looks like the merge algorithm got confused by some of the variable name changes.

BenWibking:
Would you be able to separate this into two PRs, one for the problem initialization changes, and one with the rest of the changes (mostly in AMRSimulation and RadhydroSimulation)?

AstroKriel:
Sure thing.

AstroKriel:
I will create a new PR to deal with the changed input argument to the user initialization function: `setInitialConditionsOnGrid`.

AstroKriel:
I have created a test problem `FCQuantities` to check that face-centered variables can be written to/read from chk-files. The only cell-centered quantities that exist are the hydro ones, and those quantities are initialized the same as the `HydroWave` problem. Still to be done is for me to write a function that checks the read-in quantities are equal to what is expected. At the moment I have just confirmed this by plotting the data.

BenWibking:
Is the GPU node working again? We need to run the GPU tests once more before merging.

AstroKriel:
Yep, it's working again, now. I'll check the test suite runs as expected on GPU, in the morning.

BenWibking:
Can you check why the builds are failing?

src/FCQuantities/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_executable(test_fc_quantities ../main.cpp ../MFInterpolater.cpp test_fc_quantities.cpp)
+
+if(AMReX_GPU_BACKEND MATCHES "CUDA")
+    setup_target_for_cuda_compilation(test_fc_quantities)
+endif()
+
+add_test(NAME FCQuantities COMMAND test_fc_quantities hydro_wave.in WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
```
BenWibking: You should create a new parameter file for this problem, rather than re-using `hydro_wave.in`. The parameters might change.

src/FCQuantities/test_fc_quantities.cpp
```diff
@@ -0,0 +1,136 @@
+//==============================================================================
+// Copyright 2022 Neco Kriel.
+// Released under the MIT license. See LICENSE file included in the GitHub repo.
+//==============================================================================
+/// \file test_mhd_prelim.cpp
```
BenWibking: Update the name of the file here.

src/FCQuantities/test_fc_quantities.hpp
```diff
@@ -0,0 +1,20 @@
+#ifndef TEST_MHD_PRELIM_HPP_ // NOLINT
+#define TEST_MHD_PRELIM_HPP_
```
BenWibking: Update the name of the file used in the source.

src/RadhydroSimulation.hpp
```diff
@@ -259,6 +251,40 @@ template <typename problem_t> class RadhydroSimulation : public AMRSimulation<pr
...
+  if constexpr (Physics_Traits<problem_t>::is_radiation_enabled) {
+    std::vector<std::string> radNames = {"radEnergy", "x-RadFlux", "y-RadFlux", "z-RadFlux"};
+    componentNames_cc_.insert(componentNames_cc_.end(), radNames.begin(), radNames.end());
+    ncomp_cc_ += radNames.size();
+  }
+
+  // face-centred
+  // add mhd state variables
+  if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
+    componentNames_fc_.push_back({"MagEnergy"});
```
BenWibking: Why is this `MagEnergy`? There should be `AMREX_SPACEDIM` components and they should be labeled, e.g., `x-bfield`, `y-bfield`, `z-bfield` or something similar.

src/RadhydroSimulation.hpp
```diff
@@ -350,11 +374,11 @@ auto RadhydroSimulation<problem_t>::computeExtraPhysicsTimestep(int const level)
 	return std::numeric_limits<amrex::Real>::max();
 }
 
-#if !defined(NDEBUG)
-#define CHECK_HYDRO_STATES(mf) checkHydroStates(mf, __FILE__, __LINE__)
-#else
-#define CHECK_HYDRO_STATES(mf) 
-#endif
+// #if !defined(NDEBUG)
```
BenWibking: Since this works now, please change this back to what it was before.

src/RadhydroSimulation.hpp
```diff
@@ -577,17 +603,31 @@ void RadhydroSimulation<problem_t>::FillPatch(int lev, amrex::Real time,
   amrex::Vector<amrex::Real> ftime;
 
   if (lev == 0) {
-    // in this case, should return either state_new_cc_[lev] or state_old_cc_[lev]
-    GetData(lev, time, fmf, ftime);
+    // in this case, should return either state_new_[lev] or state_old_[lev]
+    if (cen == quokka::centering::cc) {
+      GetCCData(lev, time, fmf, ftime);
```
BenWibking: There is a lot of code duplication here. Can you add a centering argument to `GetData` instead, and call either `GetCCData` or `GetFCData` based on that argument?

src/grid.hpp
```diff
@@ -10,6 +12,7 @@
 namespace quokka {
   enum class centering { cc=0, fc, ec };
   enum class direction { na=-1, x, y, z };
+  const std::array<const char, 4> face_dir_name{"xyz"};
```
BenWibking: This should be an array of `std::string`s, not an array of `char`s.

src/mhd_system.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include "valarray.hpp"
+#include "simulation.hpp"
+
+/// Class for a MHD system of conservation laws
+template <typename problem_t> class MHDSystem {
+public:
+  static constexpr int nvar_ = Physics_NumVars::numMHDVars; // number of quantities per face-centering
+
+  enum varIndex {
+    energy_index = Physics_Indices<problem_t>::mhdFirstIndex,
```
BenWibking: Why is this labeled `energy_index`? It's not an energy.

src/physics_numVars.hpp
```diff
@@ -1,9 +1,12 @@
 #ifndef PHYSICS_NUMVARS_HPP_ // NOLINT
 #define PHYSICS_NUMVARS_HPP_
 
-template <typename problem_t> struct Physics_NumVars {
+struct Physics_NumVars {
+  // cell-centred
   static const int numHydroVars = 6;
   static const int numRadVars = 4;
+  // face-centred (specify the number of quantities stored per spatial dimension, i.e., each face-centering)
+  static const int numMHDVars = 1;
```
BenWibking: This should be set equal to the total number of magnetic field variables, i.e. `AMREX_SPACEDIM`. Otherwise it's too confusing.

src/simulation.hpp
```diff
@@ -161,6 +170,8 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore {
   template <typename PreInterpHook, typename PostInterpHook>
   void fillBoundaryConditions(amrex::MultiFab &S_filled, amrex::MultiFab &state,
                               int lev, amrex::Real time,
+                              amrex::Vector<amrex::BCRec> &BCs,
```
BenWibking: I don't think it's necessary to pass in both the boundary conditions and whether it's face-centered or cell-centered. It should be able to access the appropriate boundary conditions object based on whether it's face- or cell-centered.

src/simulation.hpp
```diff
@@ -1024,23 +1028,45 @@ void AMRSimulation<problem_t>::MakeNewLevelFromCoarse(
     const amrex::DistributionMapping &dm) {
   BL_PROFILE("AMRSimulation::MakeNewLevelFromCoarse()");
 
-  const int ncomp = state_new_cc_[level - 1].nComp();
-  const int nghost = state_new_cc_[level - 1].nGrow();
+  const int ncomp_cc = ncomp_cc_; // state_new_cc_[level - 1].nComp();
```
BenWibking: Let's keep this as `state_new_cc_[level - 1].nComp()`. It makes it clear that the number of components is the same as the level it's copying the data from.

src/simulation.hpp
```diff
@@ -1024,23 +1028,45 @@ void AMRSimulation<problem_t>::MakeNewLevelFromCoarse(
...
+  const int ncomp_fc = ncomp_fc_;
```
BenWibking: Let's make this `state_new_fc_[level - 1].nComp()`, for the same reason as above.

src/simulation.hpp
```diff
@@ -1024,23 +1028,45 @@ void AMRSimulation<problem_t>::MakeNewLevelFromCoarse(
...
+  const int nghost = nghost_; // state_new_cc_[level - 1].nGrow();
```
BenWibking: We might want to have a different number of ghosts for cell-centered and face-centered MultiFabs, so there should be `nghost_cc = state_new_cc_[level - 1].nGrow()` and `nghost_fc = state_new_fc_[level - 1].nGrow()`.

src/simulation.hpp
```diff
@@ -1052,26 +1078,49 @@ void AMRSimulation<problem_t>::RemakeLevel(
...
-  const int nghost = state_new_cc_[level].nGrow();
-
-  amrex::MultiFab new_state(ba, dm, ncomp, nghost);
-  amrex::MultiFab old_state(ba, dm, ncomp, nghost);
-  amrex::MultiFab max_signal_speed(ba, dm, 1, nghost);
-
-  FillPatch(level, time, new_state, 0, ncomp, FillPatchType::fillpatch_function);
+  const int ncomp_cc = ncomp_cc_; // state_new_cc_[level].nComp();
+  const int ncomp_fc = ncomp_fc_;
+  const int nghost = nghost_; // state_new_cc_[level].nGrow();
```
BenWibking: Please keep this in the same form as before. There is no reason to change it.

src/simulation.hpp
```diff
@@ -1052,26 +1078,49 @@ void AMRSimulation<problem_t>::RemakeLevel(
...
+  FillPatch(level, time, old_state, 0, ncomp_cc, quokka::centering::cc,
+            quokka::direction::na, FillPatchType::fillpatch_function); // also necessary
   std::swap(new_state, state_new_cc_[level]);
   std::swap(old_state, state_old_cc_[level]);
+
+  amrex::MultiFab max_signal_speed(ba, dm, 1, nghost);
   std::swap(max_signal_speed, max_signal_speed_[level]);
 
+  amrex::Array<amrex::MultiFab, AMREX_SPACEDIM> tmp_state_old_fc;
+  amrex::Array<amrex::MultiFab, AMREX_SPACEDIM> tmp_state_new_fc;
```
BenWibking: These should not be named `tmp_`. They are not temporary! When `std::swap` is called, they become owned by `AMRSimulation` and live as long as `AMRSimulation` does (or until they are `swap`'ed again).

src/simulation.hpp
```diff
@@ -1052,26 +1078,49 @@ void AMRSimulation<problem_t>::RemakeLevel(
...
+    // define
+    for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+      tmp_state_new_fc[idim] = amrex::MultiFab(
+          amrex::convert(ba, amrex::IntVect::TheDimensionVector(idim)), dm,
+          ncomp_fc, nghost);
+      tmp_state_old_fc[idim] = amrex::MultiFab(
+          amrex::convert(ba, amrex::IntVect::TheDimensionVector(idim)), dm,
+          ncomp_fc, nghost);
+    }
+    // TODO(neco): fill
```
BenWibking: What is this comment for? If there's something missing here, please be specific as to what remains to be implemented.

src/simulation.hpp
```diff
@@ -1133,17 +1191,31 @@ void AMRSimulation<problem_t>::FillPatch(int lev, amrex::Real time,
   amrex::Vector<amrex::Real> ftime;
 
   if (lev == 0) {
-    // in this case, should return either state_new_cc_[lev] or state_old_cc_[lev]
-    GetData(lev, time, fmf, ftime);
+    // in this case, should return either state_new_[lev] or state_old_[lev]
+    if (cen == quokka::centering::cc) {
+      GetCCData(lev, time, fmf, ftime);
```
BenWibking: There is a lot of code duplication here. I think it could be cleaner to create a `GetData` function that would call either `GetCCData` or `GetFCData` as appropriate.

src/simulation.hpp
```diff
@@ -1155,34 +1227,89 @@ void AMRSimulation<problem_t>::MakeNewLevelFromScratch(
...
 
   if (level > 0 && (do_reflux != 0)) {
     flux_reg_[level] = std::make_unique<amrex::YAFluxRegister>(
         ba, boxArray(level - 1), dm, DistributionMap(level - 1), Geom(level),
-        Geom(level - 1), refRatio(level - 1), level, ncomp);
+        Geom(level - 1), refRatio(level - 1), level, ncomp_cc);
   }
 
-  // set state_new_cc_[lev] to desired initial condition
-  setInitialConditionsAtLevel(level);
```
BenWibking, line 1237: Why did you get rid of `setInitialConditionsAtLevel`? It was a perfectly good function, it seemed to me.

src/simulation.hpp
```diff
@@ -1213,7 +1343,11 @@ void AMRSimulation<problem_t>::fillBoundaryConditions(amrex::MultiFab &S_filled,
     amrex::Vector<amrex::Real> coarseTime;
 
     // returns old state, new state, or both depending on 'time'
-    GetData(lev - 1, time, coarseData, coarseTime);
+    if (cen == quokka::centering::cc) {
+      GetCCData(lev - 1, time, coarseData, coarseTime);
```
BenWibking: A lot of code duplication can be eliminated by creating a `GetData` function that calls either `GetCCData` or `GetFCData` as appropriate.

src/simulation.hpp
```diff
@@ -1191,6 +1318,9 @@ void AMRSimulation<problem_t>::fillBoundaryConditions(amrex::MultiFab &S_filled,
                                                       amrex::MultiFab &state,
                                                       int const lev,
                                                       amrex::Real const time,
+                                                      amrex::Vector<amrex::BCRec> &BCs,
```
BenWibking: The code can be simplified if this argument is removed. This function can always access the appropriate boundary conditions as class member variables.

src/simulation.hpp
```diff
@@ -1314,14 +1445,21 @@ void AMRSimulation<problem_t>::FillPatchWithData(
...
+                                               quokka::direction dir) {
   BL_PROFILE("AMRSimulation::FillCoarsePatch()");
 
   AMREX_ASSERT(lev > 0);
 
   amrex::Vector<amrex::MultiFab *> cmf;
   amrex::Vector<amrex::Real> ctime;
-  GetData(lev - 1, time, cmf, ctime);
+  if (cen == quokka::centering::cc) {
+    GetCCData(lev - 1, time, cmf, ctime);
```
BenWibking: Same comment as before.

src/simulation.hpp
```diff
@@ -1369,6 +1505,40 @@ void AMRSimulation<problem_t>::GetData(int lev, amrex::Real time,
...
+  datatime.clear();
+
+  const amrex::Real teps =
+      (tNew_[lev] - tOld_[lev]) * 1.e-3; // generous roundoff error threshold
+
+  int dim = static_cast<int>(dir);
+
+  if (time > tNew_[lev] - teps &&
+      time < tNew_[lev] + teps) { // if time == tNew_[lev] within roundoff
+    data.push_back(&state_new_fc_[lev][dim]);
```
BenWibking: This is the wrong level of abstraction to put this at. Put the code that picks out the appropriate MultiFab, e.g., either `state_new_fc_[lev][dim]` or `state_new_cc_[lev]` in a new `GetData` function.

src/simulation.hpp
```diff
@@ -1394,24 +1576,61 @@ auto AMRSimulation<problem_t>::PlotFileName(int lev) const -> std::string {
...
+                       state_fc[boxidx](i, j + 1, k, srccomp_start + icomp));
+          }
+          if constexpr (AMREX_SPACEDIM > 2) {
+            state_cc[boxidx](i, j, k, dstcomp_start + icomp) =
+                0.5 * (state_fc[boxidx](i, j, k, srccomp_start + icomp) +
+                       state_fc[boxidx](i, j, k + 1, srccomp_start + icomp));
+          }
+        }
+      });
+}
```
BenWibking, line 1677: The MultiFab version of ParallelFor does not synchronize after it ends. You need to call `amrex::Gpu::streamSynchronize()` here to make sure the operation completed.

See the example here: https://github.com/WeiqunZhang/amrex-microbenchmarks/blob/561aae6b962f0d39ac0861b572bb8697c92a37af/aos_simple/main.cpp#L13

src/simulation.hpp
```diff
@@ -1394,24 +1576,61 @@ auto AMRSimulation<problem_t>::PlotFileName(int lev) const -> std::string {
...
     -> amrex::MultiFab {
   // Combine state_new_cc_[lev] and derived variables in a new MF
   int comp = 0;
   const int nGrow = state_new_cc_[lev].nGrow(); // workaround Ascent bug
-  const int nCompState = state_new_cc_[lev].nComp();
+  const int nCompState_cc = ncomp_cc_; // state_new_cc_[lev].nComp();
+  const int nCompState_fc = ncomp_fc_;
   const int nCompDeriv = derivedNames_.size();
-  const int nCompPlotMF = nCompState + nCompDeriv;
+  const int nCompPlotMF = nCompState_cc + (AMREX_SPACEDIM * nCompState_fc) + nCompDeriv;
```
BenWibking: `nCompState_fc` should just have the `AMREX_SPACEDIM` factor absorbed into it. It's too confusing otherwise.

src/simulation.hpp
```diff
@@ -1394,24 +1576,61 @@ auto AMRSimulation<problem_t>::PlotFileName(int lev) const -> std::string {
...
+  auto const &state_fc = mf_fc.const_arrays();
+  amrex::ParallelFor(
+      mf_cc, amrex::IntVect(AMREX_D_DECL(nGrow, nGrow, nGrow)),
+      [=] AMREX_GPU_DEVICE(int boxidx, int i, int j, int k) {
+        for (int icomp = 0; icomp < srccomp_total; ++icomp) {
+          state_cc[boxidx](i, j, k, dstcomp_start + icomp) =
+              0.5 * (state_fc[boxidx](i, j, k, srccomp_start + icomp) +
+                     state_fc[boxidx](i + 1, j, k, srccomp_start + icomp));
+          if constexpr (AMREX_SPACEDIM > 1) {
+            state_cc[boxidx](i, j, k, dstcomp_start + icomp) =
```
BenWibking: This does not make sense. You cannot average a vector into a scalar. You need as many cell-centered components as face-centered components. Only the centering is different, not the number of components.

src/simulation.hpp
```diff
@@ -1517,8 +1737,14 @@ template <typename problem_t> void AMRSimulation<problem_t>::WritePlotFile() con
   amrex::Vector<amrex::MultiFab> mf = PlotFileMF();
   amrex::Vector<const amrex::MultiFab *> mf_ptr = amrex::GetVecOfConstPtrs(mf);
   amrex::Vector<std::string> varnames;
-  varnames.insert(varnames.end(), componentNames_cc_.begin(),
-                  componentNames_cc_.end());
+  varnames.insert(varnames.end(), componentNames_cc_.begin(), componentNames_cc_.end());
+  if constexpr (Physics_Indices<problem_t>::nvarTotal_fc > 0) {
+    for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+      for (int icomp = 0; icomp < ncomp_fc_; ++icomp) {
```
BenWibking: This code becomes simpler if you absorb the `AMREX_SPACEDIM` factor into `ncomp_fc_`.

BenWibking (changes requested):
I've added some comments.

The most important things are: i) I think some of the code in simulation.hpp can be simplified a bit, particularly the parts involving `GetDataCC` and `GetDataFC`; ii) The averaging from face-centered to cell-centered variables is not correct.

AstroKriel:
Thanks for the comments. I will try and address them all tomorrow.

src/FCQuantities/test_fc_quantities.cpp
```diff
@@ -0,0 +1,136 @@
...
+constexpr double P0 =
+    1.0 / HydroSystem<FCQuantities>::gamma_; // background pressure
+constexpr double v0 = 0.;                   // background velocity
+constexpr double amp = 1.0e-6;              // perturbation amplitude
+
+AMREX_GPU_DEVICE void computeWaveSolution(
+    int i, int j, int k, amrex::Array4<amrex::Real> const &state,
+    amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &dx,
+    amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &prob_lo) {
+  const amrex::Real x_L = prob_lo[0] + (i + amrex::Real(0.0)) * dx[0];
```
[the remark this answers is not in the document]
AstroKriel: To be consistent with the rest of the codebase, I will leave the cast to amrex::Real as is. What do you think, @BenWibking?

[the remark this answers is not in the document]
BenWibking: That's true. I followed the AMReX style in using the C-style casts here, which is not the style preferred by `clang-tidy`. That was probably not a good idea in light of modern C++ practice, and I think it's slightly clearer to use `static_cast`. It's up to you.

src/simulation.hpp
BenWibking: The issue is that we want to avoid code duplication between the cell-centered and face-centered version of this function. Currently, the face-centered and cell-centered versions actually have different behavior, since I have updated the roundoff criteria for the cell-centered version since you started working on this pull request.

src/FCQuantities/test_fc_quantities.cpp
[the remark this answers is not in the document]
AstroKriel: Okay, I will try and keep clang-tidy happy, then, and use the modern C++ casting-style.

BenWibking:
I think we can get this merged faster if we split the changes to `Physics_Traits` for the problem files (the changes that add `is_mhd_enabled = false` for each problem) into a separate PR.

src/FCQuantities/test_fc_quantities.cpp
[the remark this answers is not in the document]
AstroKriel: Once I have made these changes, I am gonna create a separate PR to handle these, so the thousands of instances of amrex::Real(bla), don't plague this PR.

src/RadhydroSimulation.hpp
AstroKriel: You are right, this should not be "MagEnergy" (magnetic energy), because it doesn't have units of energy. Instead,  this is just the amplitude component of the magnetic field vector, in each direction. I have renamed this quantity to "magField", to reflect this.

Now, the reason why there is only a single quantity named, rather than an x-, y-, and possible z-component, is because this quantity will need to be created for each dimension. Here I am just specifying which quantity needs to be defined for magnetic fields to exist. The number of vector components are being interpreted and handled by the code. We can discuss this more tomorrow if you like.

src/grid.hpp
AstroKriel: Why is that?

AstroKriel: I am not sure why I chose to initialize it as an array of `char`s, but I am now defining it to be a `const std::string`. However, clang-tidy is complaining about `Initialization of 'face_dir_name' with static storage duration may throw an exception that cannot be caught clang-tidy(cert-err58-cpp)`. I am a little confused what it means, though. Let's chat in the meeting tomorrow.

src/physics_numVars.hpp
AstroKriel: Let's discuss it tomorrow.

src/simulation.hpp
AstroKriel: Sounds good.

src/simulation.hpp
AstroKriel: Sounds good.

src/simulation.hpp
AstroKriel: I must have missed this. By the looks of it, I still need to implement the FillPatch functionality for fc quantities.

src/simulation.hpp
AstroKriel: Okay, I will see what I can do.

src/simulation.hpp
AstroKriel: It just seemed like an extra function call, but now that the code for initializing at a level has kind of blown up and dominated this outer function, I think it is a good idea to bring this function back, to encapsulate all this code in it.

src/FCQuantities/test_fc_quantities.cpp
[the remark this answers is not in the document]
BenWibking: Don't worry about converting any old instances. Just the new lines of code you have written.

BenWibking:
`clang-tidy-review` only runs on the lines of code that have changed. Don't try to change the code style in the entire codebase, just use the modern code style in the lines of code changed in this PR.

If you want to see the output on your local machine before pushing the commit, you can try using this python script that attempts to do the same thing: https://clang.llvm.org/extra/doxygen/clang-tidy-diff_8py_source.html (save this script, then run `git diff -U0 HEAD^ | clang-tidy-diff.py -p1`)

BenWibking:
The `FillPatch`/`fillBoundaryConditions` changes can go in another PR. The rest of the code is more straightforward and can be merged before the code for boundary conditions filling.

src/simulation.hpp
AstroKriel: I have replaced the tmp_ qualifier in the name with int_, to reflect that the variables are not temporary (in memory), but rather an intermediate variable.

markkrumholz (commented):
Hi Neco, Ben asked me to review these changes before they come to him, since going back and forth not in real time is taking up a lot of his time. Since I'm coming up to speed on this PR, and have not followed the details up to this point, can we set up a time for you to walk me through them? That way I can get my head around the entire PR, rather than trying to look file by file and back out the overall architecture of the changes you're making.

src/simulation.hpp
BenWibking: My preference would be to just follow the same naming convention as the other MultiFabs here for consistency, i.e. use `state_old_fc` and `state_new_fc` without the trailing underscore.

src/simulation.hpp
BenWibking: I think this is only remaining thing from my code review. Otherwise, it looks good to me.

BenWibking:
I had some time to look over the changes today and it looks good to me. The only thing remaining from my original code review comments is to add `amrex::Gpu::streamSynchronize()` after computing the cell-centered average.

I still need to manually test input/output of face-centered fields on my machine.

BenWibking:
FYI- I've merged in the changes from the Poisson gravity PR. These affected `RemakeLevel`, but it should be fixed now.

I realized it's not actually necessary to call `FillPatch` on the old time step when remaking the level. By avoiding this, it reduces communication and improves performance a non-negligible amount for AMR simulations, so I've removed it from the code.

BenWibking:
sorry, the GitHub permissions model doesn't quite work right for user-owned repositories. migrating quokka next week to the quokka-astro organisation, which should fix `/azp run`.

AstroKriel:
> sorry, the GitHub permissions model doesn't quite work right for user-owned repositories. migrating quokka next week to the quokka-astro organisation, which should fix `/azp run`.

No worries!

AstroKriel:
Looks like there is a problem with the 3D shock cloud problem. Something about quantities being defined but not initialized. I will have a closer look, now.

BenWibking:
It looks like the GPU run failed to compile because of the new SphericalCollapse problem doesn't have all of the Physics_Traits that you added: https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=1431&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=d735d76d-5905-5d5d-cb1c-7198326fbfa2

Can you add `is_mhd_enabled` to that problem?

AstroKriel:
Yep, will do.

BenWibking:
You should be able to `/azp run` now. Can you give it a try?

AstroKriel:
Yep, I will try once I push this fix to the SphericalCollapse problem. I am just compiling everything in 3D to make sure the compilation works before I push.

BenWibking:
the change in the repo URL broke azure pipelines. working on a fix

AstroKriel:
Oh, okay. No stress!

BenWibking:
Can you try `/azp run` again?

BenWibking:
ok, try one more time?

BenWibking:
it's running!

AstroKriel:
[Its alive!](https://www.youtube.com/watch?v=QuoKNZjr8_U)

BenWibking:
@markkrumholz Did you want to review this in detail? It looks good to me, except I haven't yet had a chance to manually examine the face-centered outputs.

markkrumholz:
I thought we had decided that I would look at Neco's PRs going forward. It's a bit hard for me to step in and make sense of what is going on, since I have not been following this PR from the start.

AstroKriel:
Hi all, I realized I have not yet implemented a check that the chk file is being read correctly. I will add this check.

markkrumholz:
This check that Neco mentioned is the last thing to put in the PR. It's just an automated test to verify that when you write a checkpoint and then read it back it, you get back something bitwise identical.

BenWibking:
> I thought we had decided that I would look at Neco's PRs going forward. It's a bit hard for me to step in and make sense of what is going on, since I have not been following this PR from the start.

Yes, I wasn't sure whether that would apply to this PR for this reason. What do you want to do for this PR?

markkrumholz:
I can look over the test that Neco is adding now, since that is fairly self-contained, if you have looked over the rest of the monster PR. I'll take a first look at future PRs, which will hopefully be a bit more bite-sized.

src/FCQuantities/test_fc_quantities.cpp
```diff
@@ -124,13 +160,23 @@ auto problem_main() -> int {
...
-  sim_read.stopTime_ = 1.0;
-  sim_read.maxTimesteps_ = 2e4;
-  sim_read.setInitialConditions();
+  amrex::Vector<amrex::Array<amrex::MultiFab, AMREX_SPACEDIM>> const& state_old_fc_write = sim_write.getOldMF_fc();
+  amrex::Vector<amrex::Array<amrex::MultiFab, AMREX_SPACEDIM>> const& state_new_fc_write = sim_write.getNewMF_fc();
+
+  RadhydroSimulation<FCQuantities> sim_restart(BCs_cc);
+  sim_restart.cflNumber_ = 0.1;
+  sim_restart.stopTime_ = 1.0;
+  sim_restart.maxTimesteps_ = 2e4;
```
BenWibking: If this just tests the checkpoint files, does this need to run for more than 1 timestep?

markkrumholz: In fact, does it need to run for any time steps at all? Why not just initialize, write a checkpoint, read it back in, then compare, without ever doing any evolution. That is also safer, because at some point we are going to finish the MHD and then evolution will actually do something, so that the state after evolving for even one time step won't be bitwise identical to the initial state.

BenWibking: Yes, agreed. I think this should just remain an I/O test problem. We can create a new test problem for an Alfven wave.

AstroKriel: Yep, that would be the plan.

markkrumholz (approved):
OK, this looks good to me now.

BenWibking (approved):

BenWibking:
@AstroKriel You should be able to merge this yourself as soon as the tests pass, since the review requirement has been satisfied.

PR #129, opened 2022-10-10
Update input argument to user initialisation function

AstroKriel:
Updated the input argument passed to the user-defined initialization function.

This function now only accepts a single grid element. The grid element stores information about a (segment of a) state-array: its centering (e.g., cell-centered), the index-range corresponding with the part of the domain that data can be written to (part of the domain that should be initialized), etc.


.gitignore
```diff
@@ -7,7 +7,6 @@ docs/doxygen/
 tests/plt*
 tests/chk*
 tests/*.csv
-tests/plt*
```
BenWibking: Can you get rid of this change? Otherwise it looks fine to me.

.gitignore
```diff
@@ -4,10 +4,8 @@ compile_commands.json
 codeql-db/
 build/
 docs/doxygen/
-tests/plt*
```
BenWibking: Now there is an extra change here. I think you meant to revert the earlier change, but that's not what happened.

AstroKriel: Did that fix it?

BenWibking: No... When you click on 'Files changed' on the PR, you should not see the .gitignore file there. You can check this yourself: https://github.com/BenWibking/quokka/pull/129/files. If it worked, the .gitignore file just won't be here.

BenWibking: The easiest way to fix it is to checkout this file from before you changed it: https://stackoverflow.com/questions/215718/how-do-i-reset-or-revert-a-file-to-a-specific-revision
and then make a new commit.

BenWibking:
The `.gitignore` file is still present in the diff. Can you get rid of it?

AstroKriel:
`tests/plt*` will then appear twice in the gitignore (which seems to be the case in your development branch, already: see lines 7 and 10 in https://github.com/BenWibking/quokka/blob/development/.gitignore). I can add the duplicate if you want.

AstroKriel:
I can get rid of the file from the diff, but then I would need to add the duplicate entry. I can do that, but I just now realized why I removed it (I was confused before about why there was a change in the .gitignore. I thought maybe I changed something by mistake), so I point that out.

BenWibking:
> I can get rid of the file from the diff, but then I would need to add the duplicate entry. I can do that, but I just now realized why I removed it (I was confused before about why there was a change in the .gitignore. I thought maybe I changed something by mistake), so I point that out.

Yes, it's a bug that should be fixed, but not in this PR.

AstroKriel:
I think the diff algorithm is getting confused... why is it detecting that I deleted the last line, and then added it again?

BenWibking:
LGTM. I'll merge once it passes the GPU tests.

BenWibking:
Can you take a look at the arm64 test failures? Those should work.

AstroKriel:
It is a memory error. Everything compiles fine, but there is a problem with the startup phase, where the gpu is running out of memory. The code requests 25 GB when only 22 GB is available.

Here is the full error message:
`CUDA initialized with 1 GPU per MPI rank; 1 GPU(s) used in total
  what():  Out of gpu memory. Free: 22235774976 Asked: 25567444992`
  
 Have you seen this before? Do you know what might be causing it?

BenWibking:
> It is a memory error. Everything compiles fine, but there is a problem with the startup phase, where the gpu is running out of memory. The code requests 25 GB when only 22 GB is available.
> 
> Here is the full error message: `CUDA initialized with 1 GPU per MPI rank; 1 GPU(s) used in total terminate called after throwing an instance of 'std::runtime_error' what(): Out of gpu memory. Free: 22235774976 Asked: 25567444992`
> 
> Have you seen this before? Do you know what might be causing it?

That's for the GPU tests, not the arm64 tests. That is caused by someone else running on the GPU at the same time, so that's not a concern. I'm more concerned about the arm64 tests, which should always pass.

BenWibking:
> It is a memory error. Everything compiles fine, but there is a problem with the startup phase, where the gpu is running out of memory. The code requests 25 GB when only 22 GB is available.
> 
> Here is the full error message: `CUDA initialized with 1 GPU per MPI rank; 1 GPU(s) used in total terminate called after throwing an instance of 'std::runtime_error' what(): Out of gpu memory. Free: 22235774976 Asked: 25567444992`
> 
> Have you seen this before? Do you know what might be causing it?

FYI- The reason it's 25 GB is that, by default, AMReX allocates 75% of each GPU's memory on startup to reduce memory allocation overhead.

AstroKriel:
I knew AMReX allocated a good percentage of the GPU memory at startup to save on having to reallocate new memory as soon as memory requirements grew, but I didn't realize GPUs had that much memory. I guess that isn't surprising, but I was surprised, lol.

AstroKriel:
When I try and build the code on my M1 Mac, I also run into issues. From a first look, I think the issue might lie on the AMReX side of things. I will have another look at the error messages, but they are a little hard to understand. If I can't figure them out, let's discuss them in our meeting at the end of the week.

BenWibking:
> When I try and build the code on my M1 Mac, I also run into issues. From a first look, I think the issue might lie on the AMReX side of things. I will have another look at the error messages, but they are a little hard to understand. If I can't figure them out, let's discuss them in our meeting at the end of the week.

Can you copy them here?

AstroKriel:
I cannot make much of the error messages, but here is the output from the build and compilation steps copied into a txt file:

[error.txt](https://github.com/BenWibking/quokka/files/9761815/error.txt)

BenWibking:
> I cannot make much of the error messages, but here is the output from the build and compilation steps copied into a txt file:
> 
> [error.txt](https://github.com/BenWibking/quokka/files/9761815/error.txt)

These errors are very different from the CI errors. This looks like a bug in the XCode tools that Apple ships.

Can you try the workaround described here: https://stackoverflow.com/questions/73714336/xcode-update-to-version-2395-ld-compile-problem-occurs-computedatomcount-m

BenWibking:
Alternatively, can you build inside your Linux virtual machine?

AstroKriel:
Wait, what changed? Why are the Arm64 release test working now?

BenWibking:
> Wait, what changed? Why are the Arm64 release test working now?

I merged the latest changes from development into this branch, but I didn't re-run the arm64 tests. I don't see where it shows they are working now.

BenWibking:
It's still failing: https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=670&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=6fd035ca-753b-50ba-730b-d42543fbef5a

AstroKriel:
Oh, I see; I mistook the tests that automatically run when a new change has been pushed to a PR for the Azure Pipeline tests.

BenWibking:
Ok, it looks like it's fixed now. There were some memory issues with the initialization, but for some reason it didn't trigger an error until now.

PR #135, opened 2022-10-18
use MultiFab ParallelFor for hydro

### LINKED ISSUE #45: combine small FABs in a single kernel launch
Since CUDA stream parallelism does not work in practice (too many registers are used for multiple streams to run concurrently), we need to combine several small FABs within each MFIter iteration so that we get >= 128^3 cells per kernel. Otherwise, GPU threads/blocks are just sitting idle.

For single-level simulations, we can just increase the blocking factor to 128. However, this is not viable for AMR simulations, where the typical blocking_factor is 32 (ideally would be 16).

See https://github.com/AMReX-Codes/amrex/discussions/2638 for suggested implementation strategy.

BenWibking:
This rewrites the hydro driver to use the MultiFab version of ParallelFor for each kernel. This means that only one kernel is launched for all boxes on the GPU and processes every box at once. This should reduce kernel launch overhead substantially for small boxes and/or when using AMR.

This improves performance for the ShockCloud AMR problem by ~10%.

Closes https://github.com/BenWibking/quokka/issues/45.

### DIFF
--- a/src/AdvectionSimulation.hpp
+++ b/src/AdvectionSimulation.hpp
@@ -301,7 +301,7 @@ void AdvectionSimulation<problem_t>::advanceSingleTimestepAtLevel(int lev, amrex
 		if (do_reflux) {
 #ifdef USE_YAFLUXREGISTER
 			// increment flux registers
-			//incrementFluxRegisters(iter, fr_as_crse, fr_as_fine, fluxArrays, lev, fluxScaleFactor * dt_lev);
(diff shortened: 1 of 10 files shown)


BenWibking:
~~TODO: rewrite flux register incrementing code~~

BenWibking:
~~This PR may (somehow?) break GPU-aware MPI with AMReX.~~

Edit: just an unfortunate coincidence that two clusters failed due to configuration changes at the same time.

PR #146, opened 2022-10-25
Rename BCs-variable to indicate centering

AstroKriel:
Future-proofing the naming convention of boundary condition (BCs) variables. The variable name now reflects the fact that the quantities that the BCs are responsible for are cell-centered (cc). This is so that when face-centered (fc) variables are introduced, there is no confusion when trying to distinguish between cc and fc BC-variables.


BenWibking:
`.gitignore` was accidentally checked in. Otherwise LGTM.

PR #149, opened 2022-10-27
retry hydro step on post-hoc CFL violation

### LINKED ISSUE #101: check for CFL violation at the end of the hydro advance
Due to nonlinear effects, it's possible for the CFL stable timestep to be violated over the coarse of a timestep, even if the timestep was computed immediately prior to the hydro update. We can prevent this from causing the simulation to crash by checking whether the CFL timestep was violated post-hoc and (if needed) redoing the update after halving the timestep.

See, e.g.: https://github.com/AMReX-Astro/Castro/blob/f41722cc24da0843210337258b8e6e8d7c2ceaf5/Source/hydro/Castro_hydro.cpp#L234

BenWibking:
If the actual timestep taken by a hydro advance is greater than the CFL timestep as computed from the post-advance state, then re-do the hydro advance with a halved timestep.

Closes https://github.com/BenWibking/quokka/issues/101.

### DIFF
--- a/src/RadhydroSimulation.hpp
+++ b/src/RadhydroSimulation.hpp
@@ -749,7 +749,8 @@ auto RadhydroSimulation<problem_t>::isCflViolated(int lev, amrex::Real time,
 	const amrex::Real dt_cfl = cflNumber_ * (dx_min / max_signal);
 
 	// check whether dt_actual > dt_cfl (CFL violation)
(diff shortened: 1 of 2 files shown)


BenWibking:
~~Depends on fix to https://github.com/AMReX-Codes/amrex/issues/3000.~~

Not a bug.

PR #153, opened 2022-10-29
add Hancock predictor-corrector integrator

BenWibking:
This implements the Hancock predictor-corrector time integrator [1][2]. When used with PLM reconstruction (`hydro.reconstruction_order = 2`), we obtain the MUSCL-Hancock method (which itself is a simplification of the MUSCL method first presented in [3]).

This method is equivalent to the VL2 integrator implemented in Athena.

[1] [van Albada et al. (1982)](https://ui.adsabs.harvard.edu/abs/1982A%26A...108...76V/abstract)
[2] [Falle (1991)](https://ui.adsabs.harvard.edu/abs/1991MNRAS.250..581F/abstract)
[3] [van Leer (1979)](https://ui.adsabs.harvard.edu/abs/1979JCoPh..32..101V/abstract)

### DIFF
--- a/src/RadhydroSimulation.hpp
+++ b/src/RadhydroSimulation.hpp
@@ -189,7 +189,12 @@ template <typename problem_t> class RadhydroSimulation : public AMRSimulation<pr
 				 amrex::YAFluxRegister *fr_as_crse,
 				 amrex::YAFluxRegister *fr_as_fine);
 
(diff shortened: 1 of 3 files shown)


BenWibking:
This doesn't seem to improve stability over RK2, so there's not much use for it. Closing for now.

PR #160, opened 2022-11-01
add random blast ISM example

BenWibking:
This simulates the explosion of supernovae in a homogeneous ISM.

The existing carbuncle correction is not sufficient to prevent strong carbuncles from appearing when cooling is enabled.


BenWibking:
To illustrate the carbuncle issue, this is after the first blast:

density00380
scalar00380
temp00380

BenWibking:
We need to deposit the energy with a smoothing kernel, e.g. Wendland C2 or similar.

PR #186, opened 2022-11-17
add exact riemann solver

BenWibking:
This implements an exact Riemann solver for ideal gases following Chapter 4 of Toro (1998).

Cases that generate vacuum are not handled. Isothermal EOS is not handled.

_Caution: carbuncles and grid alignment effects can be more pronounced than with the HLLC solver!_


BenWibking:
Does not pass the HydroBlast3D test for some reason.

PR #188, opened 2022-11-28
update build instructions in README

BenWibking:
Updates build instructions in README.md.

- Using `ninja` to build instead of `make` is recommended.
- Added subsection listing dependencies (including optional dependencies)
- Removed old note about Intel classic compilers being broken.
- Removed old note about CUDA 11.6 being broken.
- Added notes on building for AMD GPUs.
- Added notes on building for Intel GPUs.
- Added notes on building a specific test problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 **The Quokka methods paper is now available: https://arxiv.org/abs/2110.01792**
 
-**NOTE: Documentation is still a work in progress. Please see the sections "Quickstart" and "Running on GPUs" below. You can start a [Discussion](https://github.com/BenWibking/quokka/discussions) for technical support, or open an [Issue](https://github.com/BenWibking/quokka/issues) for any bug reports.**
+**Please see the sections "Quickstart" and "Running on GPUs" below. You can start a [Discussion](https://github.com/BenWibking/quokka/discussions) for technical support, or open an [Issue](https://github.com/BenWibking/quokka/issues) for any bug reports.**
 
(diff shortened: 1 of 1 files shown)


markkrumholz:
Just looking over these changes. Do you require ninja to build now, or can the code be built with just cmake? The readme says that ninja is optional, but then the instructions all seem to assume it is available. It seems like you might want to provide separate instructions for building with ninja+cmake versus just cmake, in order to avoid confusion.

BenWibking:
> Just looking over these changes. Do you require ninja to build now, or can the code be built with just cmake? The readme says that ninja is optional, but then the instructions all seem to assume it is available. It seems like you might want to provide separate instructions for building with ninja+cmake versus just cmake, in order to avoid confusion.

Either `ninja` or `make` is required in addition to CMake. It can be built with either, but I wanted to recommend building with `ninja` since it is often significantly faster. It can also be installed with `pip`, so not having it installed shouldn't be an issue. Should I just remove "optional"?

markkrumholz:
I would err on the side of not making people install additional stuff, even if it is easy. Thus I would suggest writing this as something like that ninja is the recommended build system, and give the directions for that first, and then give alternate directions for installing via cmake. If you don't want to continue supporting cmake into the future you could even note that cmake support may be removed in future releases, but for now I would prefer leaving both options available unless there is a really compelling reason to drop cmake now.

BenWibking:
Ok, that makes sense.

BenWibking:
@markkrumholz how is this version?

markkrumholz:
OK, much clearer now. Looks good.

markkrumholz (approved):

PR #189, opened 2022-12-01
Add more notes about GPU-aware MPI to README

BenWibking:
Explicitly note that GPU-aware MPI support is required in the Dependencies section of the README.


markkrumholz (approved):
Given the confusion we seem to have caused at NCI, it is probably good to make this extremely explicit in the README. Happy with this change.

BenWibking:
> Given the confusion we seem to have caused at NCI, it is probably good to make this extremely explicit in the README. Happy with this change.

Yes, agreed. Thanks for reviewing it at lightspeed!

PR #193, opened 2022-12-05
Adding Fixed potential, SN feedback in test suite with enforced limits

aditivijayan:
Added MetalAdvectionProblem to the test suite.


BenWibking:
The title is rather confusing. I think it would be clearer if the title of this PR should just be that this adds the metal advection problem to the code.

markkrumholz:
I was holding off on commenting on this PR until we get the previous one sorted, since some stuff is going to have to migrate from this one into that one, and there should be some general restructuring. Basically, I think there should be 3 distinct PRs: one for enforcing limits, one for fixed potential gravity, and one for the SN problem setup.

PR #195, opened 2022-12-06
Add HLLD solver for ideal MHD

AstroKriel:
Ported the [Athena++ HLLD](https://github.com/PrincetonUniversity/athena/blob/mhd4_3D/src/hydro/rsolvers/mhd/hlld.cpp) solver with a general EOS, to Quokka.


AstroKriel:
\azp run

src/HLLD.hpp
```diff
@@ -0,0 +1,343 @@
...
+#include <AMReX.H>
+#include <AMReX_REAL.H>
+
+#include "ArrayView.hpp"
+#include "HydroState.hpp"
+#include "valarray.hpp"
+
+namespace quokka::Riemann
+{
+#define DELTA 1.0e-4
```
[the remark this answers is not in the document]
BenWibking: This is a good suggestion -- can you change this to a constexpr variable?

src/HLLD.hpp
```diff
@@ -0,0 +1,343 @@
...
+
+template <class T> constexpr auto SQUARE(const T x) -> T { return x * x; }
+
+template <int N_scalars> auto FastMagnetosonicSpeed(double gamma, quokka::HydroState<N_scalars> const state, const double bx) -> double
```
BenWibking: This will not compile on GPU as-is. You need to add `AMREX_FORCE_INLINE AMREX_GPU_DEVICE` annotations to this function, just like the HLLD() function.

src/HLLD.hpp
```diff
@@ -0,0 +1,343 @@
...
+	double gp = gamma * state.P;
+	double bx_sq = bx * bx;
+	double byz_sq = state.by * state.by + state.bz * state.bz;
+	double b_sq = bx_sq + byz_sq;
+	double bgp_p = b_sq + gp;
+	double bgp_m = b_sq - gp;
+	return std::sqrt(0.5 * (bgp_p + std::sqrt(bgp_m * bgp_m + 4.0 * gp * byz_sq)) / state.rho);
+}
+
+inline auto GetWeightForCT(double dflx, double rho_L, double rho_R, double dx, double dt) -> double
```
BenWibking: This needs to be annotated with `AMREX_FORCE_INLINE AMREX_GPU_DEVICE`.

src/HLLD.hpp
```diff
@@ -0,0 +1,343 @@
...
+{
+	double v_over_c = (1024.0) * dt * dflx / (dx * (rho_L + rho_R));
```
BenWibking: Do you know where the factor of `1024.0` comes from?

src/HLLD.hpp
```diff
@@ -0,0 +1,343 @@
...
+		f_x.rho = f_R.rho + u_star_R.rho;
+		f_x.mx = f_R.mx + u_star_R.mx;
+		f_x.my = f_R.my + u_star_R.my;
+		f_x.mz = f_R.mz + u_star_R.mz;
+		f_x.E = f_R.E + u_star_R.E;
+		f_x.by = f_R.by + u_star_R.by;
+		f_x.bz = f_R.bz + u_star_R.bz;
+	}
+
+	F_hydro = {f_x.rho, f_x.mx, f_x.my, f_x.mz, f_x.E, 0}; // Eint=0 for now
```
BenWibking: Can you mark this with a `// TODO(neco): ` comment? Passive scalars also need to be implemented in the future.

BenWibking (changes requested):
A few minor changes requested.

src/HLLD.hpp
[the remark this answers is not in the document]
BenWibking: Also, it looks like you changed this to `1e-4` instead of `1e-8` as in the Athena++ version. What is the reason for this change?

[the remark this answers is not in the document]
AstroKriel: The Athena++ `master` branch uses `1e-4`, whereas the `mhd4` branch uses `1e-8`. I was looking at the `master` branch when I ported the HLLD solver. I will update this `DELTA` value to `1e-8`.

https://github.com/PrincetonUniversity/athena/blob/master/src/hydro/rsolvers/mhd/hlld.cpp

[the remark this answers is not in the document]
BenWibking: Ok, this may require some experimentation to see what is a robust value once we have some test problems.

BenWibking:
@markkrumholz The "new" Athena++ method does _not_ use the electric fields returned from the Riemann solver. The old GS05 method did, which is why it is still there in both branches of the code.

To summarize:
* GS05 used the electric fields returned by the Riemann solver directly to obtain the node-centered electric fields (see `Field::ComputeCornerE`). The numerical diffusion in the magnetic induction equation depends on the adopted Riemann solver.
* The newer method (originally from Londrillo and Del Zanna 2004) uses a completely different reconstruction method to do this (see `Field::ComputeCornerE_UCT4`). The numerical diffusion in the induction equation depends only on the reconstruction method used for the magnetic field.

For Athena++, both functions are implemented in here: https://github.com/PrincetonUniversity/athena/blob/mhd4_3D/src/field/calculate_corner_e.cpp

For PLUTO, both functions are implemented here: http://plutocode.ph.unito.it/Doxygen/API-Reference_Guide/ct__emf__average_8c.html

There is also a detailed comparison of these two schemes (and others) by Mignone & Del Zanna (https://ui.adsabs.harvard.edu/abs/2021JCoPh.42409748M/abstract). They call GS05 the "CT-Contact" scheme and the the newer method the "UCT-HLL" scheme.

BenWibking:
@AstroKriel The "new" CT method is described in section 4: https://ui.adsabs.harvard.edu/abs/2018JCoPh.375.1365F/abstract

AstroKriel:
Thanks @BenWibking!

src/HLLD.hpp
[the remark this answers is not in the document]
AstroKriel: Done.

src/HLLD.hpp
AstroKriel: Not yet. I will get back to you on this.

BenWibking: If you haven't figured this out from reading the code, you could open an issue on the Athena++ Github and ask about this.

BenWibking: Actually, I realized this is only needed for the GS05/CT-Contact scheme. My preference would be to just delete it, since we're not planning to implement this scheme.

src/HLLD.hpp
```diff
@@ -0,0 +1,346 @@
...
+		f_x.my = f_R.my + u_star_R.my;
+		f_x.mz = f_R.mz + u_star_R.mz;
+		f_x.E = f_R.E + u_star_R.E;
+		f_x.by = f_R.by + u_star_R.by;
+		f_x.bz = f_R.bz + u_star_R.bz;
+	}
+
+	F_hydro = {f_x.rho, f_x.mx, f_x.my, f_x.mz, f_x.E, 0}; // TODO(neco): Eint=0 for now. Also need to add pscalars.
+	Efield_y = -f_x.by;
+	Efield_z = f_x.bz;
```
BenWibking: My preference would be to delete the Efield_y and Efield_z variables, since they are only used by the GS05/CT-Contact scheme.

AstroKriel: I don't think that is quite right. It looks like the electric field quantities derived in the HLLD Riemann solver is also used in the FS17 method. See for example the 1D update step (near the start of the function; lines 245-248) in `ComputeCornerE_UCT4`.

https://github.com/PrincetonUniversity/athena/blob/mhd4_3D/src/field/calculate_corner_e.cpp

AstroKriel:
Yes, if we want to have 4th order accuracy then we will need to implement this other method that you've pointed out. I think we will still need the electric field terms in order to compute the EMF (even with the new method), though.

src/HLLD.hpp
BenWibking: That is only done in 1D and 2D for the components that are not actually affected by the emf update. In 3D, the electric fields from the Riemann solver are never used.

BenWibking: In the 1D and 2D cases, I am almost certain that the electric fields can just be copied without involving the Riemann solver at all.

BenWibking:
> Yes, if we want to have 4th order accuracy then we will need to implement this other method that you've pointed out. I think we will still need the electric field terms in order to compute the EMF (even with the new method), though.

That is true, but the practical issue is not 4th order accuracy. PS found that UCT-HLL gave the best results with AMR even when using a second-order method. You should ask him for details. I think there are some tests on the old ORION wiki as well (may not exist anymore).

The reference for the 2nd order version of the UCT-HLL scheme is: https://ui.adsabs.harvard.edu/abs/2004JCoPh.195...17L/abstract

I don't think we need the electric field terms, see my comments above.

BenWibking:
This is what PS wrote on the old ORION wiki (located here, but it requires PS to create an account for you: http://orionwiki.berkeley.edu/doku.php?id=mhd):

> There are different choices on how to average the EMF in CT scheme, including “ARITHMETIC”, “UCT0”, “UCT_CONTACT”, and “UCT_HLL”. UCT_HLL gives the best result in all the test but it requires CFL number < 0.35. ARITHMETIC provides a simple and robust average scheme for driven turbulence simulation.

AstroKriel:
@BenWibking can you resolve your review comment? All the sub-comments have been addressed.

BenWibking (approved):

PR #196, opened 2022-12-06
create EOS and EOS_Traits classes

BenWibking:
This creates `EOS<problem_t>` and `EOS_Traits<problem_t>` classes.

These are currently used to compute gas temperature as a function of internal energy and density and vice versa within the radiation matter-energy exchange. The hydro solver also obtains the value of `gamma` and the value of the isothermal sound speed from `EOS_Traits<problem_t>`.

### DIFF
--- a/src/CloudyCooling.hpp
+++ b/src/CloudyCooling.hpp
@@ -70,7 +70,7 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto cloudy_cooling_function(Real const
 {
 	// interpolate cooling rates from Cloudy tables
 	const Real rhoH = rho * cloudy_H_mass_fraction; // mass density of H species
-	const Real nH = rhoH / hydrogen_mass_cgs_;
+	const Real nH = rhoH / quokka::hydrogen_mass_cgs;
 	const Real log_nH = std::log10(nH);
 	const Real log_T = std::log10(T);
 
@@ -94,7 +94,7 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto cloudy_cooling_function(Real const
 
 	// compute electron density
 	// N.B. it is absolutely critical to include the metal contribution here!
-	double n_e = (rho / hydrogen_mass_cgs_) * (1.0 - mu * (X + Y / 4. + Z / mean_metals_A)) / (mu - (electron_mass_cgs / hydrogen_mass_cgs_));
+	double n_e = (rho / quokka::hydrogen_mass_cgs) * (1.0 - mu * (X + Y / 4. + Z / mean_metals_A)) / (mu - (electron_mass_cgs / quokka::hydrogen_mass_cgs));
 	// the approximation for the metals contribution to e- fails at high densities (~1e3 or higher)
 	n_e = std::max(n_e, 1.0e-4 * nH);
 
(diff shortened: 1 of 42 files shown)


BenWibking:
Some 2D and 3D test problems still need to be fixed to work with this. No need to review yet.

markkrumholz:
Design looks reasonable, but I think that mean molecular weight shoudl default to NaN, so that you can only compute a temperature if you set it explicitly. The problem is that there's no reasonable default which will be good for all the types of applications that you would do with quokka -- you'd want different defaults for molecular gas, atomic gas, and ionised gas. Safer just to have it be undefined, so unless the user explicitly overrides it in their problem, you can't derive the temperature and you get NaNs.

markkrumholz:
The other design change I'd suggest is not to use gamma directly from EOS traits, but instead to add a function that returns gamma to the EOS class, which by default just returns the gamma in EOS traits. The advantage is that this lets users who want to overwrite the gamma function in the EOS class with their own non-constant gamma in a problem file, if they want to do something more sophisticated. (I know that the Riemann solver doesn't properly include non-constant gamma at this point, but one could still interpret this as having a variable gamma just with lagging, so that gamma is effectively held frozen during the hydro update, but is allowed to change between steps.) Moving gamma from being a constant to a function will also make it easier in the future if we want to expand to a full non-constant gamma. Thus I suggest adding a public function to the EOS class of the form 

`[[nodiscard]] AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE static auto ComputeGamma(amrex::Real rho, amrex::Real Eint) -> amrex::Real;`

with a default implementation that just returns `EOS_Traits<problem_t>::gamma`, and then calling this function any time you need a value of gamma in the Riemann solver.

BenWibking:
> The other design change I'd suggest is not to use gamma directly from EOS traits, but instead to add a function that returns gamma to the EOS class, which by default just returns the gamma in EOS traits. The advantage is that this lets users who want to overwrite the gamma function in the EOS class with their own non-constant gamma in a problem file, if they want to do something more sophisticated. (I know that the Riemann solver doesn't properly include non-constant gamma at this point, but one could still interpret this as having a variable gamma just with lagging, so that gamma is effectively held frozen during the hydro update, but is allowed to change between steps.) Moving gamma from being a constant to a function will also make it easier in the future if we want to expand to a full non-constant gamma. Thus I suggest adding a public function to the EOS class of the form
> 
> `[[nodiscard]] AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE static auto ComputeGamma(amrex::Real rho, amrex::Real Eint) -> amrex::Real;`
> 
> with a default implementation that just returns `EOS_Traits<problem_t>::gamma`, and then calling this function any time you need a value of gamma in the Riemann solver.

Is there anywhere outside of the Riemann solver this would be useful? The issue is that the wavespeeds currently used by the Riemann solver only make sense for a gas where gamma is constant everywhere. For a general EOS, we have to use a more diffusive and more general wavespeed estimate. In that case, I think we only need to compute the pressure as a function of internal energy and density.

BenWibking:
> Design looks reasonable, but I think that mean molecular weight shoudl default to NaN, so that you can only compute a temperature if you set it explicitly. The problem is that there's no reasonable default which will be good for all the types of applications that you would do with quokka -- you'd want different defaults for molecular gas, atomic gas, and ionised gas. Safer just to have it be undefined, so unless the user explicitly overrides it in their problem, you can't derive the temperature and you get NaNs.

Ok, done.

markkrumholz:
I can think of at least one other use case where you'd want to know gamma, or at least c_v, which is if you're doing a chemistry network where you're also evolving the temperature along with the chemistry. Depending on your solver you might need to know gamma. But I agree that is a hypothetical, as none of our chemistry modules do that now, so I don't feel really strongly about making gamma a function if you would prefer not to at this point.

BenWibking:
> I can think of at least one other use case where you'd want to know gamma, or at least c_v, which is if you're doing a chemistry network where you're also evolving the temperature along with the chemistry. Depending on your solver you might need to know gamma. But I agree that is a hypothetical, as none of our chemistry modules do that now, so I don't feel really strongly about making gamma a function if you would prefer not to at this point.

I'm going to defer this for when we need it. You can already get $\rho c_v = d{E_{int}}/d{T}$ from the function `EOS::ComputeEintTempDerivative`.

BenWibking:
@markkrumholz Can you approve?

markkrumholz (approved):

psharda:
Not sure if this is the same, but we will ultimately use a variable gamma for primordial chemistry, where gamma is calculated based on density, temperature and mass fraction of H2.

PR #197, opened 2022-12-07
Read cc and fc ncomp from Physics_Indices

AstroKriel:
Removed the member variable `ncomp_cc_` from `AMRSimulation` and updated all references to this member variable to instead query `Physics_Indices<problem>::nvarTotal_cc`, which computes at compile time how many cc and fc quantities need to be defined for a particular problem configuration (see `Physics_Traits`).


markkrumholz (approved):

AstroKriel:
\azp run

BenWibking:
> \azp run

You have to use a forward-slash:
/azp run

PR #200, opened 2022-12-13
fix build error with CUDA 12

BenWibking:
This changes the way `AMRSimulation<problem_t>::InterpHookNone` is used in `AdvectionSimulation` to fix a build error that appears with CUDA 12.


BenWibking:
The GPU build is currently failing because Sven is running something on the GPU, which kills performance and causes the tests to time out.

markkrumholz (commented):
No idea why the old version would fail, but as far as I can tell the new version should be 100% functionally identical, so looks fine to me.

markkrumholz (approved):
Looks fine to me.

PR #201, opened 2022-12-16
disable readability-function-cognitive-complexity warning

BenWibking:
This disables the `readability-function-cognitive-complexity` warning in the clang-tidy settings file [1].

We have a lot of kernels that are very long, and splitting them up into separate functions in order to avoid this warning does not make sense.

[1] https://clang.llvm.org/extra/clang-tidy/checks/readability/function-cognitive-complexity.html


BenWibking:
@markkrumholz Can you approve this? This will make future PR reviews slightly easier.

markkrumholz (approved):

PR #202, opened 2022-12-16
add uniform cooling box test

BenWibking:
This adds a cooling test with tabulated cooling for a uniform box of gas. This can be used to test the tabulated cooling machinery in isolation from hydrodynamics.


markkrumholz (commented):
The test itself loosk fine, but calling the problem test_box is a bit non-descriptive. Maybe change to test_cooling_amr or something like that?

BenWibking:
> The test itself loosk fine, but calling the problem test_box is a bit non-descriptive. Maybe change to test_cooling_amr or something like that?

Ok, that makes sense. What about `test_cooling_onezone`?

markkrumholz:
> > The test itself loosk fine, but calling the problem test_box is a bit non-descriptive. Maybe change to test_cooling_amr or something like that?
> 
> Ok, that makes sense. What about `test_cooling_onezone`?

Yes, that name sounds fine.

BenWibking:
I don't think there's a use case for this test anymore. Closing for now.

PR #203, opened 2022-12-17
add flowchart to docs

BenWibking:
This adds a SVG flowchart generated with Google Drawings that illustrates the functions called during the main timestepping loop.

Preview: https://github.com/quokka-astro/quokka/blob/35f854081e93f266c9e53335e6a93000e0647898/docs/flowchart.rst


markkrumholz (approved):
Looks good to me.

PR #206, opened 2022-12-19
Changed box coordinates for the SphericalCollapse test problem

psharda:
Changed box coordinates for the test problem such that the center is at (0,0,0)


BenWibking:
Can you provide some motivation or justification for doing this?

markkrumholz:
I'll let Piyush respond, but I find the changed coordinate system more intuitive, so it's justified if only for that reason.

BenWibking (approved):

PR #209, opened 2022-12-20
allow simulation to continue if retries exceed max

BenWibking:
This adds a runtime option to prevent the timestep retry mechanism from being triggered when first-order flux correction (FOFC) fails to avoid non-positive-density cells in a given RK stage. This also uses a wider, more diffusive stencil for the P dV term (for the internal energy equation) when a bad cell is produced.

This should allow simulations to continue even if first-order flux correction fails to produce a positive-density state, rather than failing due to exceeding the maximum number of timestep retries. This may happen if vacuum is the physical solution in some region of the flow.

### DIFF
--- a/src/RadhydroSimulation.hpp
+++ b/src/RadhydroSimulation.hpp
@@ -104,13 +104,12 @@ template <typename problem_t> class RadhydroSimulation : public AMRSimulation<pr
 	bool computeReferenceSolution_ = false;
 	amrex::Real errorNorm_ = NAN;
 	amrex::Real pressureFloor_ = 0.;
-	int fofcMaxIterations_ =
-	    3; // maximum number of flux correction iterations -- only 1 is needed in almost all cases, but in rare cases a second iteration is needed
 
 	int integratorOrder_ = 2;	       // 1 == forward Euler; 2 == RK2-SSP (default)
 	int reconstructionOrder_ = 3;	       // 1 == donor cell; 2 == PLM; 3 == PPM (default)
 	int radiationReconstructionOrder_ = 3; // 1 == donor cell; 2 == PLM; 3 == PPM (default)
 	int useDualEnergy_ = 1;		       // 0 == disabled; 1 == use auxiliary internal energy equation (default)
+	int abortOnFofcFailure_ = 1;	       // 0 == keep going, 1 == abort hydro advance if FOFC fails
 
 	amrex::Long radiationCellUpdates_ = 0; // total number of radiation cell-updates
 
--- a/src/RadhydroSimulation.hpp
+++ b/src/RadhydroSimulation.hpp
@@ -856,13 +856,13 @@ auto RadhydroSimulation<problem_t>::advanceHydroAtLevel(amrex::MultiFab &state_o
 			HydroSystem<problem_t>::PredictStep(stateOld, stateNew, rhs, dt_lev, ncompHydro_, redoFlag);
 
 			amrex::Gpu::streamSynchronizeAll(); // just in case
-			int ncells_bad = redoFlag.sum(0);
+			amrex::Long ncells_bad = redoFlag.sum(0);
 			if (ncells_bad > 0) {
 				// FOFC failed
 				if (Verbose()) {
 					amrex::Print() << "[FOFC-1] failed for " << ncells_bad << " cells on level " << lev << "\n";
 				}
-				if (abortOnFofcFailure_) {
+				if (abortOnFofcFailure_ != 0) {
 					return false;
 				}
 			}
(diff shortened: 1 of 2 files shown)


markkrumholz:
Physics-wise this looks fine, but I'm wondering if we should switch the default behavior for failed FOFC to continuing rather than aborting. In fact, even if FOFC fails, the density and internal energy floors should catch and fix the problem. Thus in the majority of real calculations, the "right" behavior to producing a negative density state even after FOFC is just to floor and move on, not to abort. I would suggest that should be the default.

BenWibking:
> Physics-wise this looks fine, but I'm wondering if we should switch the default behavior for failed FOFC to continuing rather than aborting. In fact, even if FOFC fails, the density and internal energy floors should catch and fix the problem. Thus in the majority of real calculations, the "right" behavior to producing a negative density state even after FOFC is just to floor and move on, not to abort. I would suggest that should be the default.

The reason I was hesitant to do this is because by default, it won't immediately abort, it will just retry with a halved timestep and check again. There might be some problems for which this is sufficient, and in those cases, I think it would give a more correct solution than imposing a floor.

markkrumholz:
I agree it is better to try again if a shorter time step. Maybe I misunderstood the control flow: what I was thinking was that the proper action is to first try FOFC, then try again with a shorter time step, but limit the number of retries with shorter time steps allowed so that you don't get the time step falling to zero. It is only a failure if you reach that maximum number of retries. Am I misunderstanding what you're calling a "failure" of FOFC?

BenWibking:
> I agree it is better to try again if a shorter time step. Maybe I misunderstood the control flow: what I was thinking was that the proper action is to first try FOFC, then try again with a shorter time step, but limit the number of retries with shorter time steps allowed so that you don't get the time step falling to zero. It is only a failure if you reach that maximum number of retries. Am I misunderstanding what you're calling a "failure" of FOFC?

No, that is what is happening in Aditi's simulations (and, in very rare cases, in my cloud simulations). FOFC fails on each re-try, and then it hits the maximum number of retries, and then fails because it's hit the maximum without the hydro advance returning 0. So we could disable the 'feature' where it crashes if it hits the maximum and just have it continue, but it may also hit the maximum because the CFL condition has been violated at the end of the timestep, in which case I think it would be dangerous to allow the simulation to continue. Currently, we don't distinguish between these cases in the retry mechanism.

markkrumholz:
I see. I had misunderstood the control flow. OK, then how about this proposal: the default behavior is that, if you hit the maximum number of retries, you immediately invoke the EnforceLimits function, then declare it a success. This should separate the cases of failure to FOFC not working and failure due to CFL being violated because the velocity went up.

BenWibking:
> I see. I had misunderstood the control flow. OK, then how about this proposal: the default behavior is that, if you hit the maximum number of retries, you immediately invoke the EnforceLimits function, then declare it a success. This should separate the cases of failure to FOFC not working and failure due to CFL being violated because the velocity went up.

I don't quite follow. How does this distinguish the two cases?

markkrumholz:
I guess I am still not understanding how your control flow works. Let me abstract away from how the code is currently structured, and describe what I see as the correct default behavior from the standpoint of physics. This is:

1) First attempt the step as normal, with high order fluxes.
2) If (1) fails for some cells, retry with first-order fluxes.
3) If (2) fails for some cells, reduce time step and try again.
4) if (3) fails, go back to (3) with an even smaller time step and try again, up to some maximum number of attempts.
5) If negative densities or energies persist after (4), then apply the floors.

This seems to me like the correct default behavior. Is that reasonably implementable?

BenWibking:
> I guess I am still not understanding how your control flow works. Let me abstract away from how the code is currently structured, and describe what I see as the correct default behavior from the standpoint of physics. This is:
> 
> 1. First attempt the step as normal, with high order fluxes.
> 2. If (1) fails for some cells, retry with first-order fluxes.
> 3. If (2) fails for some cells, reduce time step and try again.
> 4. if (3) fails, go back to (3) with an even smaller time step and try again, up to some maximum number of attempts.
> 5. If negative densities or energies persist after (4), then apply the floors.
> 
> This seems to me like the correct default behavior. Is that reasonably implementable?

Yes, this is easily implementable. In it's most easily implementable form, it would also keep the simulation going (after applying the floors) in the case where the CFL timestep is violated post-hoc even after step 4.

markkrumholz:
OK, well that seems fine with me. If after repeated attempts to reduce the time step the CFL keeps being violated by an increase in velocity occuring mid-time step, then clearly something strange is going on (e.g., in an MHD sim the density is trying to go to zero, so Alfven speed is diverging), and floors / ceilings are probably the right solution.

BenWibking:
The implementation issue is that `advanceHydroAtLevel` returns `false` when _either_:

1. FOFC does not keep all densities positive (https://github.com/quokka-astro/quokka/blob/19012b4b98b30d5402eb85b3b26d3896ad9fe133/src/RadhydroSimulation.hpp#L922)
2. The CFL condition is violated based on the post-advance state (https://github.com/quokka-astro/quokka/blob/19012b4b98b30d5402eb85b3b26d3896ad9fe133/src/RadhydroSimulation.hpp#L945)

BenWibking:
> OK, well that seems fine with me. If after repeated attempts to reduce the time step the CFL keeps being violated by an increase in velocity occuring mid-time step, then clearly something strange is going on (e.g., in an MHD sim the density is trying to go to zero, so Alfven speed is diverging), and floors / ceilings are probably the right solution.

Ok, we can do that. I think the danger is that it might also indicate a bug somewhere, either in the initial conditions or problem-specific boundary conditions or source terms, which would be harder to detect if this were the default.

markkrumholz:
I see. Well, I see no reason they need to be separated, in the sense that I think it is fine to use the algorithm I described above in response to either a negative density or a repeated violation of the CFL.

markkrumholz:
I understand that flooring rather than aborting does allow bugs to go undetected, but, as we are discovering in Aditi's runs, in a production calculation some sort of flooring is often going to be required. I think it should therefore be on by default.

BenWibking:
Ok, that is a good point. I'll turn it on by default.

markkrumholz (approved):

PR #210, opened 2022-12-20
avoid using inf for limits

BenWibking:
This uses `std::numeric_limits<double>::max()` instead of using `inf` for the default ceilings.

This is because `inf` incurs a performance penalty [1]. The behavior should be unchanged, since no floating-point value can be larger than `std::numeric_limits<double>::max()` except for `inf` itself.

[1] https://randomascii.wordpress.com/2012/05/20/thats-not-normalthe-performance-of-odd-floats/ (This also seems to be true on NVIDIA GPUs in my tests, but I haven't found a reference for this.)


markkrumholz (approved):
I agree this has no effect on functionality, and if it speeds up the code, that's fine.

PR #211, opened 2022-12-20
fix publish docs action

BenWibking:
This uses GitHub-native mechanisms to deploy to GitHub Pages: i) it uploads a build artefact that contains a *.tar file with the static HTML files for the documentation, ii) it triggers deployment to the `github-pages` environment if pushed to `development` on the official quokka-astro/quokka repository.

Due to the repository permissions, no other branch (or repository) is allowed to publish files to `quokka-astro.github.io`, which is why there are errors below about deployment failures. (Hopefully, these will not occur once this is merged into development.)

This will hopefully fix deployment to GitHub Pages, which is currently broken.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -69,5 +69,6 @@ jobs:
     needs: build
     steps:
       - name: Deploy to GitHub Pages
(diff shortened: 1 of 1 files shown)


markkrumholz (approved):
Hopefully this works!

PR #214, opened 2022-12-21
Increased pressure for spherical collapse gravity test

psharda:
Increased the pressure from 1e-3 to 1e-1 so that the Jeans length at t=0 is resolved by ~4 cells in the gravity test simulation with one level of AMR. If we do not increase the pressure, the Jeans length is too small and the test simulation has to be performed at a much higher resolution.


BenWibking (commented):
This should be fine, but what does the profile look like? Does it agree with the analytic solution just as well as before?

psharda:
Do you mean this? 
gravtest_quokka_fracerror
gravtest2_quokka

psharda:
I put this movie in #quokka-shakka but maybe it is worth attaching it to this PR if we need to reference it in the future. The radial profiles are very flat in the center now, and that bump we saw before is not present.

https://user-images.githubusercontent.com/34922596/208942812-17f57ffb-a2a4-4583-b582-973f92ac5333.mp4

BenWibking:
Yes, this looks essentially perfect. Very flat! Yes, let's keep this here for reference, since our Slack messages expire. I'll approve.

BenWibking (dismissed):

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "extern/grackle_data_files"]
 	path = extern/grackle_data_files
 	url = https://github.com/grackle-project/grackle_data_files.git
+[submodule "extern/Microphysics"]
+	path = extern/Microphysics
+	url = https://github.com/AMReX-Astro/Microphysics
```
BenWibking, line 12: This should go in a separate branch / separate PR.

psharda:
Oops, Microphysics got attached to this PR because it was not merged and I pushed Microphysics as a submodule.

PR #215, opened 2022-12-21
Increased pressure for spherical collapse gravity test

psharda:
Increased the pressure from 1e-3 to 1e-1 so that the Jeans length at t=0 is resolved by ~4 cells in the gravity test simulation with one level of AMR. If we do not increase the pressure, the Jeans length is too small and the test simulation has to be performed at a much higher resolution.

Attached plots showing the results of this test. The numerical solution matches the analytic solution very well, and the radial profile of density is very flat.

https://user-images.githubusercontent.com/34922596/208946113-61f2c7c3-1e46-415c-bf1f-4dc1fddb3347.mp4

gravtest_quokka_fracerror
gravtest2_quokka


BenWibking (dismissed):

psharda:
also increased the number of steps from 5000 to 5000 as the simulation takes more than 5000 steps to reach densities where the jeans length is resolved by 1 cell.

BenWibking (approved):

PR #216, opened 2022-12-21
added Microphysics as a submodule for Quokka

psharda:
This PR will add Microphysics as a submodule for Quokka.


BenWibking:
Ah, we actually need to wait for this: https://github.com/AMReX-Astro/Microphysics/issues/1103

psharda:
Right, so how do we do this

psharda:
Also, Mike will merge my cmake PR in `Microphysics/development`. So, it will not be available in the `main` branch of Microphysics (the one we are adding as a submodule in quokka) until the start of the next month (when he merges `development` into `main`

BenWibking:
I would add your Microphysics branch as a submodule, then create a new test problem in Quokka to build with the network.

BenWibking:
That's ok, we can point the submodule to any Microphysics branch we want.

BenWibking:
We'll need to add the same Microphysics *.cpp files to the Quokka CMakeLists.txt file that you added to the unit test CMakeLists.txt:
```
add_executable(test_microphysics
    main.cpp
    ${CMAKE_SOURCE_DIR}/interfaces/eos_data.cpp
    ${CMAKE_SOURCE_DIR}/interfaces/network_initialization.cpp
    ${CMAKE_SOURCE_DIR}/EOS/primordial_chem/actual_eos_data.cpp
    ${CMAKE_BINARY_DIR}/extern_parameters.cpp
    )
```

So it might make sense to create (in the Microphysics root CMakeLists.txt) a variable to store this list of files that we can then reference in the Quokka CMakeLists.txt.

psharda:
> We'll need to add the same Microphysics *.cpp files to the Quokka CMakeLists.txt file that you added to the unit test CMakeLists.txt:
> 
> ```
> add_executable(test_microphysics
>     main.cpp
>     ${CMAKE_SOURCE_DIR}/interfaces/eos_data.cpp
>     ${CMAKE_SOURCE_DIR}/interfaces/network_initialization.cpp
>     ${CMAKE_SOURCE_DIR}/EOS/primordial_chem/actual_eos_data.cpp
>     ${CMAKE_BINARY_DIR}/extern_parameters.cpp
>     )
> ```
> 
> So it might make sense to create (in the Microphysics root CMakeLists.txt) a variable to store this list of files that we can then reference in the Quokka CMakeLists.txt.

But we use these files in the burn_cell_primordial_chem test, and not in Microphysics root. If we create a variable for these files in Microphysics root, can we just call it in CMakeLists for burn_cell_primordial_chem?

psharda:
> That's ok, we can point the submodule to any Microphysics branch we want.

Added the specific branch in the submodule call.

BenWibking:
> > We'll need to add the same Microphysics *.cpp files to the Quokka CMakeLists.txt file that you added to the unit test CMakeLists.txt:
> > ```
> > add_executable(test_microphysics
> >     main.cpp
> >     ${CMAKE_SOURCE_DIR}/interfaces/eos_data.cpp
> >     ${CMAKE_SOURCE_DIR}/interfaces/network_initialization.cpp
> >     ${CMAKE_SOURCE_DIR}/EOS/primordial_chem/actual_eos_data.cpp
> >     ${CMAKE_BINARY_DIR}/extern_parameters.cpp
> >     )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So it might make sense to create (in the Microphysics root CMakeLists.txt) a variable to store this list of files that we can then reference in the Quokka CMakeLists.txt.
> 
> But we use these files in the burn_cell_primordial_chem test, and not in Microphysics root. If we create a variable for these files in Microphysics root, can we just call it in CMakeLists for burn_cell_primordial_chem?

Yes, it can be used in both the unit test CMakeLists.txt and in the Quokka CMakeLists.txt. A win-win :)

BenWibking:
Quokka will also need a variable for the include directories. That can be re-used by the unit test CMakeLists.txt as well.

psharda:
> > > We'll need to add the same Microphysics *.cpp files to the Quokka CMakeLists.txt file that you added to the unit test CMakeLists.txt:
> > > ```
> > > add_executable(test_microphysics
> > >     main.cpp
> > >     ${CMAKE_SOURCE_DIR}/interfaces/eos_data.cpp
> > >     ${CMAKE_SOURCE_DIR}/interfaces/network_initialization.cpp
> > >     ${CMAKE_SOURCE_DIR}/EOS/primordial_chem/actual_eos_data.cpp
> > >     ${CMAKE_BINARY_DIR}/extern_parameters.cpp
> > >     )
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > So it might make sense to create (in the Microphysics root CMakeLists.txt) a variable to store this list of files that we can then reference in the Quokka CMakeLists.txt.
> > 
> > 
> > But we use these files in the burn_cell_primordial_chem test, and not in Microphysics root. If we create a variable for these files in Microphysics root, can we just call it in CMakeLists for burn_cell_primordial_chem?
> 
> Yes, it can be used in both the unit test CMakeLists.txt and in the Quokka CMakeLists.txt. A win-win :)

I did this, created a variable called `primordial_chem_sources` for all these files in Microphysics root, and then called them in the burn_cell_primordial_chem's CMakeLists using `add_executable(test ${primordial_chem_sources})`. I get this error during `make`: 

Undefined symbols for architecture x86_64:
"_main", referenced from:
    implicit entry/start for main executable
ld: symbol(s) not found for architecture x86_64

BenWibking:
You still need to include `main.cpp` within the unit test `add_executable` command.

psharda:
In the burn_cell_primordial_chem CMakeLists, I did `add_executable(test_burn_cell_primordial_chem main.cpp ${primordial_chem_sources})`

At the end of root CMakeLists, I define:
set(primordial_chem_sources
    ${CMAKE_SOURCE_DIR}/interfaces/eos_data.cpp
    ${CMAKE_SOURCE_DIR}/interfaces/network_initialization.cpp
    ${CMAKE_SOURCE_DIR}/EOS/primordial_chem/actual_eos_data.cpp
    ${CMAKE_BINARY_DIR}/extern_parameters.cpp
   )
   
   
I now get all the unreferenced errors  :  "_temperature", referenced from:
      burn_cell_c() in main.cpp.o

psharda:
Tyring it with `target_sources` and `source_group`

psharda:
Nope, still get all the linking errors.

psharda:
All the solutions I find on stackoverflow suggest using `file(GLOB...)` to store multiple source files in the same variable. But then `GLOB` is not recommended by cmake developers.

BenWibking:
> All the solutions I find on stackoverflow suggest using `file(GLOB...)` to store multiple source files in the same variable. But then `GLOB` is not recommended by cmake developers.

It should, in principle, work the same whether it's in a variable or directly listed, e.g.: https://stackoverflow.com/questions/37225229/adding-sources-to-a-list-variable-in-cmake-vs-directly-in-add-executable.  Can you copy and paste the differences between the verbose make outputs?

BenWibking:
Maybe there's some subtle issue with variable scope. Can you try defining the variable in the `burn_cell_primordial_chem/CMakeLists.txt` itself just to make sure that works?

psharda:
> Maybe there's some subtle issue with variable scope. Can you try defining the variable in the `burn_cell_primordial_chem/CMakeLists.txt` itself just to make sure that works?

yeah that works.

BenWibking:
> > Maybe there's some subtle issue with variable scope. Can you try defining the variable in the `burn_cell_primordial_chem/CMakeLists.txt` itself just to make sure that works?
> 
> yeah that works.

Gahh, ok. Some weird CMake thing is going on then when it's defined in a different file :(

Unfortunately, I think we have to figure out what it is, since we need the ability to do this to integrate with the existing way Quokka builds are done.

psharda:
I posted a question on SO

psharda:
There is a `PARENT_SCOPE` option for setting variables in cmake that can make the variable available one directory up from where it's defined. But, it does not seem to work two (or more) directories up.

BenWibking:
> There is a `PARENT_SCOPE` option for setting variables in cmake that can make the variable available one directory up from where it's defined. But, it does not seem to work two (or more) directories up.

This is the opposite case, no? I think the variable should be defined in the root CMakeLists.txt, and then used from the (nested) subdirectories.

psharda:
I might have found a solution!

psharda:
Okay, it works now. I learned that the content in CMakeLists has to be ordered? I was using `set` after `add_subdirectory` in the root CMakeLists, which is why the variable being set was not visible in the child scope (aka the subdirectory). Once I set the variable first and then added the subdirectory, it was available in the CMakeLists in burn_cell_primordial_chem

psharda:
Why do 2 of the tests fail?

BenWibking:
> Why do 2 of the tests fail?

I broke the test machine that the arm64 tests run on :( I'll fix it.

psharda:
Haha no worries. 
What else needs to be done before we can approve this PR?

BenWibking:
Can you check what commit the Microphysics submodule points to? On the GitHub 'files changed' tab, it shows a commit from before your Microphysics PR was merged, so CMakeLists.txt is not present there.

BenWibking:
Nothing other than checking the commit hash for this PR. We can do another PR to make changes to the CMakeLists.txt files in Quokka, as well as adding a test problem in Quokka.

psharda:
How about now?

BenWibking:
Looks good now.

BenWibking (approved):

PR #217, opened 2022-12-23
use Roe-average wavespeeds for HLLC

BenWibking:
This uses the Roe-average-based wavespeeds for the HLLC solver as recommended by Batten et al. (1997) [1].

These wavespeeds are an improvement over the current estimates because they are i) slightly more diffusive, and therefore more robust and ii) in principle, compatible with a general EOS [2][3][4].

[1] https://ui.adsabs.harvard.edu/abs/1997SJSC...18.1553B/abstract
[2] https://ui.adsabs.harvard.edu/abs/1990JCoPh..87....1L/abstract
[3] https://doi.org/10.1016/S0045-7825(97)00228-4
[4] https://ui.adsabs.harvard.edu/abs/1988JCoPh..74..382G/abstract


markkrumholz (approved):
Looks good to me.

PR #221, opened 2022-12-27
revert to first-order space+time flux when RK2 flux fails

### LINKED ISSUE #218: first-order flux correction should revert to forward Euler
The short story is that the current way we do flux correction doesn't guarantee positivity for the result of the second (or later, if any) stages. To guarantee positivity, we need to revert to first-order in both space *and* time.

First-order flux correction, as originally proposed by [1], uses the VL2 integrator, and as such can simply revert to the fluxes produced in the predictor step. For RK2 integration, things are less clear, since it is a multi-stage integrator where each stage is a convex combination of the current-stage fluxes and the previous stage state values. However, RK2 can be written as a convex combination of the first- and second-stage _fluxes_ alone. Then for fluxes adjacent to problem cells, we can replace the high-order combination with the first-order (both space and time) flux. This is conceptually straightforward, but requires an additional set of flux multifabs for temporary storage during the hydro update.

A similar first-order correction has been described by [2], but they apply the correction to the cell state, rather than to the fluxes, so it is not conservative.

[1] https://ui.adsabs.harvard.edu/abs/2009ApJ...691.1092L/abstract
[2] https://doi.org/10.1016/S0045-7825(97)00228-4
[3] Linde, Timur, Philip Roe. Robust Euler Codes. 13th Computational Fluid Dynamics Conference, 2098 (1997)
[4] Perthame, B., Shu, CW. On positivity preserving finite volume schemes for Euler equations . Numer. Math. 73, 119–130 (1996). https://doi.org/10.1007/s002110050187

BenWibking:
This improves the first-order flux correction so that it reverts to forward Euler with a donor cell flux (i.e., Godunov's method) for cells that produce negative densities with the RK2+PPM/PLM update. This replaces the fluxes surrounding problem cells, so that the overall method remains conservative.

This causes a tiny difference in the output of the HydroHighMach test. The solution looks no better or worse than before by eye, but the error tolerance is just slightly too stringent for it to pass, so it's updated here.

Fixes https://github.com/quokka-astro/quokka/issues/218.


markkrumholz (dismissed):

BenWibking:
@markkrumholz here's the first FOFC fix. The boundary box consistency will go in the next PR.

BenWibking:
@markkrumholz All tests now pass. Please review at your convenience.

markkrumholz (approved):

PR #223, opened 2023-01-07
adjust multigrid parameters for gravity solve

BenWibking:
Uses a slightly less strict tolerance for the multigrid solve for self-gravity and uses first-order operator splitting, rather than Strang splitting.

On CPUs, this reduces the walltime spent in the gravity solve from ~90-95% to ~25% for the spherical collapse test.

On GPUs, the walltime is *strongly* dependent on the number of cells per GPU and the GPU architecture. On MI250X, good performance requires at least $256^3$ cells *per GCD*. Then, about ~70% of walltime is in the gravity solve (or equivalently, gravity is ~2.3x slower than hydro).


BenWibking:
@markkrumholz Fixed a merge conflict. Can you re-approve?

markkrumholz (approved):

PR #227, opened 2023-01-15
Timed output

aditivijayan:
This modification allows users to create data dumps at equal time intervals. Checkpoints are created when plt files are stored.


aditivijayan (commented):
Changes suggested in hydro_system.hpp are superfluous, I believe.

src/CMakeLists.txt
```diff
@@ -126,5 +126,3 @@ add_subdirectory(ODEIntegration)
 add_subdirectory(Cooling)
 add_subdirectory(ShockCloud)
 add_subdirectory(PassiveScalar)
-add_subdirectory(FCQuantities)
-add_subdirectory(SphericalCollapse)
```
BenWibking: Can you revert this change? These should not be modified

src/hydro_system.hpp
```diff
@@ -939,4 +939,4 @@ void HydroSystem<problem_t>::ComputeFluxes(amrex::MultiFab &x1Flux_mf, amrex::Mu
 	});
 }
 
-#endif // HYDRO_SYSTEM_HPP_
+#endif // HYDRO_SYSTEM_HPP_
```
BenWibking: Can you revert this change? There's no reason to modify this file.

src/simulation.hpp
```diff
@@ -703,6 +708,14 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 			WritePlotFile();
 		}
 
+		// Writing Plot files at time intervals
+		double next_plot_file_time = next_plot_file_num * plotTimeInterval_;
+		if (plotTimeInterval_ > 0 && ((next_plot_file_time > cur_time) && (next_plot_file_time < (cur_time + dt_[0])))) {
```
BenWibking: I think there might be a floating-point roundoff issue here. Is the last condition necessary?

src/simulation.hpp
```diff
@@ -703,6 +708,14 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
...
+			next_plot_file_num += 1;
+			WritePlotFile();
+			WriteCheckpointFile();
```
BenWibking, line 724: The checkpoint files can be very large. I think we should not automatically output the checkpoint file at the same time. Perhaps make that a separate option?

src/simulation.hpp
```diff
@@ -703,6 +708,14 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 			WritePlotFile();
 		}
 
+		// Writing Plot files at time intervals
+		double next_plot_file_time = next_plot_file_num * plotTimeInterval_;
+		if (plotTimeInterval_ > 0 && ((next_plot_file_time > cur_time) && (next_plot_file_time < (cur_time + dt_[0])))) {
+			next_plot_file_num += 1;
```
BenWibking: I think this might be clearer if you change it to `next_plot_file_time += plotTimeInterval_`. Then `next_plot_file_num` is not needed.

BenWibking (changes requested):
There are some minor comments. Also, can you revert the changes to the amrex submodule, hydro_system.hpp and src/CMakeLists.txt? It looks like those were added by mistake.

src/simulation.hpp
aditivijayan: Do you mean- `(next_plot_file_time < (cur_time + dt_[0])`? Yes, we need to make sure that `next_plot_file_time` is within dt of `cut_time`. Because if `next_plot_file_time` is say cur_time + 3dt, then it would write 3 files instead of 1.

src/simulation.hpp
aditivijayan: I can do this.

src/simulation.hpp
BenWibking: Could you change the latter two conditions to just `next_plot_file_time <= cur_time`?

src/simulation.hpp
aditivijayan: Sure thing.

src/hydro_system.hpp
BenWibking: You can checkout the last commit that modified this file with
```
git checkout `git --no-pager log -1 --pretty=tformat:"%h" -- src/hydro_system.hpp` -- src/hydro_system.hpp
```

BenWibking: Ah, oops, that won't work. This should checkout the previous version of the file:
```
git checkout `git --no-pager log -1 --skip 1 --pretty=tformat:"%h" -- src/hydro_system.hpp` -- src/hydro_system.hpp
```

BenWibking: Hm, ok, the previous commit doesn't seem to be the same as that on the development branch. It might be better to just checkout the version of this file from the development branch, e.g.:
```
git checkout development -- src/hydro_system.hpp
```

BenWibking:
Something is still not right with hydro_system.hpp. Do you have the latest version pulled locally? You might have to checkout development with `git checkout development` and then `git pull` to make sure you have the latest version of the development branch.

aditivijayan:
> Something is still not right with hydro_system.hpp. Do you have the latest version pulled locally? You might have to checkout development with `git checkout development` and then `git pull` to make sure you have the latest version of the development branch.

Locally, it shows that everything is up to date with the development branch. But I noticed that the declaration of AddinternalpDV was different. So, I've just copy-pasted the right one from the development branch. Hopefully, it should work now.

BenWibking:
> Locally, it shows that everything is up to date with the development branch. But I noticed that the declaration of AddinternalpDV was different. So, I've just copy-pasted the right one from the development branch. Hopefully, it should work now.

Hmm, it's not showing that when I look at the changes here: https://github.com/quokka-astro/quokka/pull/227/files.

aditivijayan:
I am really perplexed by this because I haven't touched that part of the code at all. I've just copied the entire function from the development branch now.

BenWibking:
Ok, that was strange. It looks like it's fixed now.

Can you set the amrex submodule to the same commit used on the development branch?

aditivijayan:
Ohhh.. I don't know how to do that.

BenWibking:
> Ohhh.. I don't know how to do that.

You can check the current commit of the amrex submodule by looking at the development branch for Quokka here: https://github.com/quokka-astro/quokka/tree/development/extern.

Then, on your local machine, go into the submodule directory:
`cd extern/amrex`

Checkout this commit:
`git checkout ae58672`

Go back to the main git repository:
`cd ../..`

`git status` should now show that the submodule has been updated. You can then commit the change to the submodule with `git add extern/quokka && git commit -m 'revert amrex submodule'`.

aditivijayan:
Thanks, Ben!

BenWibking (approved):

BenWibking:
I'm running the GPU tests now. If those pass, it can be merged.

aditivijayan:
Great! Thanks, Ben!

PR #228, opened 2023-01-15
add multidimensional artificial viscosity

BenWibking:
Adds an artificial viscosity term to `ComputeFluxes` following [1]. This may still be necessary for some difficult multidimensional problems.

[1] https://ui.adsabs.harvard.edu/abs/1984JCoPh..54..174C/abstract


markkrumholz (approved):

markkrumholz (approved):
Looks like the build failure on the last attempt was a network issue. Triggering another try.

PR #248, opened 2023-04-10
Disable parallel builds in CI

BenWibking:
For non-GPU builds run on virtual machines, only allow building one file at a time. This prevents the operating system from killing the compiler when it runs out of memory when building the chemistry network.

Added the `merge_group` trigger to the CMake build action so that it runs during the merge queue (see https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions).

This also updates the `clang-tidy-review` action to use `ZedThree/clang-tidy-review@v0.12.2` and ensure that the `python3-dev`, `python3-numpy`, and `python3-matplotlib` packages are installed before running `clang-tidy`. This is necessary for the CMake build to work with the chemistry network enabled.

### DIFF
--- a/.ci/azure-pipelines-aarch64-debug.yml
+++ b/.ci/azure-pipelines-aarch64-debug.yml
@@ -18,7 +18,7 @@ steps:
 - task: CMake@1
   displayName: 'Build Quokka'
   inputs:
-    cmakeArgs: '--build .'
+    cmakeArgs: '--build . --parallel 1'
 
 - task: CMake@1
   displayName: 'Run CTest'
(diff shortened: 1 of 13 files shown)


BenWibking:
`ZedThree/clang-tidy-review@v0.10.1` appears to be broken.

BenWibking:
`clang-tidy-review` Docker build fails :(

markkrumholz (dismissed):

BenWibking:
@markkrumholz There was a merge queue problem. Can you re-approve?

markkrumholz (approved):

PR #252, opened 2023-04-17
Optimize load

aditivijayan:
This modification ensures that in case of restart with larger cores boxarrays are readjusted to reflect new load.


BenWibking:
It looks like the build is failing due to a bug that was in an old version of AMReX. Can you use the same AMReX submodule version that is used by the current development branch?

aditivijayan:
I am unsure what this error is since it doesn't show up when I compile on gadi.

PR #263, opened 2023-04-18
Patched microphysics to quokka, and added primordial chem test

### LINKED ISSUE #104: add primordial chemistry network from Microphysics
Piyush has successfully used the AMReX-Astro/Microphysics integrator to integrate a primordial chemistry network. This needs to be called for each cell during the Strang-split update. A suitable test problem needs to be written.

See https://github.com/AMReX-Astro/Microphysics/pull/962.

psharda:
Continuation of #241 . Closes #104


src/Chemistry.hpp
```diff
@@ -0,0 +1,121 @@
+#ifndef CHEMISTRY_HPP_ // NOLINT
+#define CHEMISTRY_HPP_
+//==============================================================================
+// TwoMomentRad - a radiation transport library for patch-based AMR codes
+// Copyright 2020 Benjamin Wibking.
+// Released under the MIT license. See LICENSE file included in the GitHub repo.
+//==============================================================================
+/// \file CloudyCooling.hpp
```
BenWibking: Can you change this to `Chemistry.hpp`?

src/Chemistry.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include "burn_type.H"
+#include "burner.H"
+#include "eos.H"
+#include "extern_parameters.H"
+
+namespace quokka::chemistry
+{
+template <typename problem_t> void computeChemistry(amrex::MultiFab &mf, const Real dt_in, const Real max_density_allowed)
+{
+	const Real dt = dt_in;
```
BenWibking: This was a workaround only needed for old CUDA versions. You can get rid of this line and change `dt_in` to `dt`.

src/Chemistry.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+			for (int nn = 0; nn < NumSpec; ++nn) {
+				inmfracs[nn] /= insum;
+				// update the number densities with conserved mass fractions
+				chemstate.xn[nn] = inmfracs[nn] * chemstate.rho / spmasses[nn];
+			}
+
+			// update the number density of electrons due to charge conservation
+			chemstate.xn[0] = -chemstate.xn[3] - chemstate.xn[7] + chemstate.xn[1] + chemstate.xn[12] + chemstate.xn[6] + chemstate.xn[4] +
+					  chemstate.xn[9] + 2.0 * chemstate.xn[11];
```
BenWibking, line 102: Is there a way to update the electrons without hard-coding the species numbers and electrons per species? If so, then this function would work for any chemistry network, not just primordial chemistry.

src/Chemistry.hpp
```diff
@@ -0,0 +1,121 @@
...
+{
+	const Real dt = dt_in;
+
+	for (amrex::MFIter iter(mf); iter.isValid(); ++iter) {
+		const amrex::Box &indexRange = iter.validbox();
+		auto const &state = mf.array(iter);
+
+		amrex::ParallelFor(indexRange, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+			const Real rho = state(i, j, k, HydroSystem<problem_t>::density_index);
+			const Real Eint = state(i, j, k, HydroSystem<problem_t>::internalEnergy_index);
```
BenWibking: You should actually compute the internal energy by subtracting the kinetic energy from the total energy. This internal energy is an auxiliary variable that is only used within the hydro solver.

src/Chemistry.hpp
```diff
@@ -0,0 +1,121 @@
...
+			for (int nn = 0; nn < NumSpec; ++nn) {
+				inmfracs[nn] /= insum;
+				// update the number densities with conserved mass fractions
+				chemstate.xn[nn] = inmfracs[nn] * chemstate.rho / spmasses[nn];
+			}
+
+			// get the updated Eint
+			eos(eos_input_rt, chemstate);
+
+			state(i, j, k, HydroSystem<problem_t>::internalEnergy_index) = chemstate.e;
```
BenWibking: You need to compute dE and add this to *both* the total energy variable and the auxiliary internal energy variable. See how this is done in cooling.hpp for reference.

src/PrimordialChem/test_primordial_chem.cpp
```diff
@@ -0,0 +1,295 @@
...
+	AMREX_GPU_MANAGED amrex::Real primary_species_5;
+	AMREX_GPU_MANAGED amrex::Real primary_species_6;
+	AMREX_GPU_MANAGED amrex::Real primary_species_7;
+	AMREX_GPU_MANAGED amrex::Real primary_species_8;
+	AMREX_GPU_MANAGED amrex::Real primary_species_9;
+	AMREX_GPU_MANAGED amrex::Real primary_species_10;
+	AMREX_GPU_MANAGED amrex::Real primary_species_11;
+	AMREX_GPU_MANAGED amrex::Real primary_species_12;
+	AMREX_GPU_MANAGED amrex::Real primary_species_13;
+	AMREX_GPU_MANAGED amrex::Real primary_species_14;
```
BenWibking: The `AMREX_GPU_MANAGED` annotation is not necessary here. Can you remove it?

src/PrimordialChem/test_primordial_chem.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+	// loop over the grid and set the initial condition
+	amrex::ParallelFor(indexRange, [=] AMREX_GPU_DEVICE(int i, int j, int k) {
+		Real const x = prob_lo[0] + (i + static_cast<Real>(0.5)) * dx[0];
+		if constexpr (AMREX_SPACEDIM >= 2) {
+			Real const y = prob_lo[1] + (j + static_cast<Real>(0.5)) * dx[1];
+		}
+		if constexpr (AMREX_SPACEDIM == 3) {
+			Real const z = prob_lo[2] + (k + static_cast<Real>(0.5)) * dx[2];
+		}
```
BenWibking: Do you use `x`, `y`, or `z` anywhere? If not, please delete.

BenWibking (changes requested):
The chemistry update in Chemistry.hpp needs to be changed to correctly use and set both the total energy variable and the internal energy variable (see comments).

The rest are just style changes or unused variables.

BenWibking:
`clang-tidy-review` seems to be posting comments to the old PR... can you check those?

src/Chemistry.hpp
psharda: Can we leave this like it for now and come back to it when we generalize to other chem networks?

BenWibking: Ok. Can you add a comment, e.g. `// TODO(psharda): generalize this to other chem networks`?

src/Chemistry.hpp
psharda: This was problematic in FLASH, because, often when shocks were present, subtraction of internal energy from the total energy resulted in very small or even negative internal energies, depending on the precision with which they were stored and subtraction was done. This would then make the chemistry barf. 

Since quokka stores the momentum (or is it specific momentum?)?

BenWibking: This is solved in Quokka by evolving both total and auxiliary internal energy variables. At the end of the hydro update, if the total minus kinetic is negative, it's replaced with the auxiliary internal + kinetic. So at the start of the chemistry update, it should always be positive. (If it's not, something went wrong at the end of the last timestep.)

Quokka stores $\rho v_i$ for $i = 0, 1, 2$.

BenWibking: The issue with negative internal energies is *not* due to floating-point precision, but due to discretization error. At high Mach numbers, the spatial resolution requirements formally increase as $\mathcal{M}$ (see https://arxiv.org/abs/0704.2608).

BenWibking:
Actually, it looks like `std::array<N>` should work on GPU (according to this: https://github.com/NVIDIA/libcudacxx/issues/51).

Can you replace the arrays you create, e.g.`Real[N]` etc., with that?

psharda:
> Actually, it looks like `std::array<N>` should work on GPU (according to this: [NVIDIA/libcudacxx#51](https://github.com/NVIDIA/libcudacxx/issues/51)).
> 
> Can you replace the arrays you create, e.g.`Real[N]` etc., with that?

So, like `std::array<real, 14>N = {-1};`?

BenWibking:
> > Actually, it looks like `std::array<N>` should work on GPU (according to this: [NVIDIA/libcudacxx#51](https://github.com/NVIDIA/libcudacxx/issues/51)).
> > Can you replace the arrays you create, e.g.`Real[N]` etc., with that?
> 
> So, like `std::array<real, 14>N = {-1};`?

Yes, see examples here: https://en.cppreference.com/w/cpp/container/array

src/Chemistry.hpp
psharda: I just realised something: in Microphysics, the internal energy I calculate is the specific internal energy. I find it the way FLASH finds it (using equation 16.6 of https://flash.rochester.edu/site/flashcode/user_support/flash4_ug_4p7/node108.html). FLASH also used specific internal energy, and the kinetic energy in FLASH was `ekin = 0.5*(vx^2 + vy^2 + vz^2)`. So, energies in FLASH are energy per unit mass `(km^2 s^-2)`.

I just checked `hydro_system.hpp` in Quokka, and Quokka uses  `ekin = 0.5*rho*(vx^2 + vy^2 + vz^2)`. So, Quokka energies are `(g/cm^3)*(km^2 s^-2)`. This means there will be an inconsistency between internal energy calculated by chemistry (via microphysics) and hydro (in quokka)?

BenWibking: You have to convert between specific internal energy (SIE) and internal energy by dividing/multiplying by the gas density.

BenWibking: In the code, I have distinguished between SIE and internal energy by using lowercase `e` for SIE and uppercase `Eint` for internal energy (where $E_{int} = \rho e$).

psharda: Yeah. I was wondering if I should calculate Eint in microphysics in the same units as quokka. But all microphysics EOS use specific internal energy and we want to keep all Microphysics EOS consistent. So using a conversion factor between microphysics EOS and quokka EOS seems like the best choice.

BenWibking: Yes, I agree.

psharda:
@BenWibking I parmparse two variables in `RadhydroSimulation.hpp`:

```
#ifdef PRIMORDIAL_CHEM
	// set chemistry runtime parameters
	{
		amrex::ParmParse hpp("primordial_chem");
		hpp.query("enabled", enableChemistry_);
		hpp.query("max_density_allowed", max_density_allowed);
	}
#endif
```

However they remain unused; this is the logfile of the tests:

```
Unused ParmParse Variables:
  [TOP]::primordial_chem.enabled(nvals = 1)  :: [1]
  [TOP]::primordial_chem.max_density_allowed(nvals = 1)  :: [3e-6]
```

Why are they unused? I use them below in `RadhydroSimulation.hpp` to compute chemistry...

BenWibking:
> @BenWibking I parmparse two variables in `RadhydroSimulation.hpp`:
> 
> ```
> #ifdef PRIMORDIAL_CHEM
> 	// set chemistry runtime parameters
> 	{
> 		amrex::ParmParse hpp("primordial_chem");
> 		hpp.query("enabled", enableChemistry_);
> 		hpp.query("max_density_allowed", max_density_allowed);
> 	}
> #endif
> ```
> 
> However they remain unused; this is the logfile of the tests:
> 
> ```
> Unused ParmParse Variables:
>   [TOP]::primordial_chem.enabled(nvals = 1)  :: [1]
>   [TOP]::primordial_chem.max_density_allowed(nvals = 1)  :: [3e-6]
> ```
> 
> Why are they unused? I use them below in `RadhydroSimulation.hpp` to compute chemistry...

Somehow this block of code is not running. Are you certain that `PRIMORDIAL_CHEM` is defined?

psharda:
> Somehow this block of code is not running. Are you certain that `PRIMORDIAL_CHEM` is defined?

I mean, yes? I defined it in root CMakeLists, and then I used `add_compile_definitions(PRIMORDIAL_CHEM)` if `PRIMORDIAL_CHEM` is set to true (which it is, for the primordialchem test)

psharda:
Ah, in primordialchem test CmakeLists, to set `PRIMORDIAL_CHEM` to true, I am using `set(PRIMORDIAL_CHEM true)`. But, should it be `set(${PRIMORDIAL_CHEM} true)`, given that it is defined as a boolean variable in root CMakeLists??

BenWibking:
> > Somehow this block of code is not running. Are you certain that `PRIMORDIAL_CHEM` is defined?
> 
> I mean, yes? I defined it in root CMakeLists, and then I used `add_compile_definitions(PRIMORDIAL_CHEM)` if `PRIMORDIAL_CHEM` is set to true (which it is, for the primordialchem test)

Can you add an `#error Error message here` inside the `#ifdef`: https://stackoverflow.com/questions/2124968/gcc-pragma-to-stop-compilation?

psharda:
Or maybe I am not actually calling the `addStrangSplitSourcesWithBuiltin` routine in the test

psharda:
> Can you add an `#error Error message here` inside the `#ifdef`: https://stackoverflow.com/questions/2124968/gcc-pragma-to-stop-compilation?

No error message displayed, and the test passes. But, the variables are being parmpased. So the test is never going into `addStrangSplitSourcesWithBuiltin` function. I need to explicitly call it somewhere?

BenWibking:
> > Can you add an `#error Error message here` inside the `#ifdef`: https://stackoverflow.com/questions/2124968/gcc-pragma-to-stop-compilation?
> 
> No error message displayed, and the test passes. But, the variables are being parmpased. So the test is never going into `addStrangSplitSourcesWithBuiltin` function. I need to explicitly call it somewhere?

No, it means that `PRIMORDIAL_CHEM` is not defined and the variables are therefore not being parmparsed. You do not explicitly need to call this function -- it is already called within RadhydroSimulation.hpp.

psharda:
Hmm. Then this is a cmake issue. I tried a few different things but nothing worked. Will try more tomorrow.

psharda:
Done. The chemistry _actually_ runs now :)

psharda:
Ugh ffff, GPU builds fail even with `ncpus=48` and `mem=192GB` on Gadi because chemistry is now actually being compiled for PrimordialChem test. Error:

```
LLVM ERROR: out of memory
nvcc error   : 'cicc' died due to signal 6
```

I guess I will have to decrease CPUs I am using. Currently, I use `ninja -j 10`

psharda:
build on CPU (in 3D) works fine. Primordialchem test is built and passes too, but some other tests time out:

```
          5 - HydroBlast3D (Timeout)
          7 - HydroLeblanc (Timeout)
          8 - HydroShocktube (Timeout)
         24 - RadhydroShock (Timeout)
         25 - RadhydroShockCGS (Timeout)
         27 - PassiveScalar (Timeout)
         29 - SphericalCollapse (Timeout)

```

psharda:
Nope, even with no parallelisation, GPU builds fail on Gadi because of memory error. We will have to understand how can we free some memory or allocate more? https://stackoverflow.com/questions/65468120/llvm-error-out-of-memory-how-to-fix-this-error-if-its-fixable

BenWibking:
> Nope, even with no parallelisation, GPU builds fail on Gadi because of memory error. We will have to understand how can we free some memory or allocate more? https://stackoverflow.com/questions/65468120/llvm-error-out-of-memory-how-to-fix-this-error-if-its-fixable

Unfortunately, the only solution is to figure out how to use less memory when compiling. The Microphysics unit test does build for GPU though, right? If so, then I think the solution to move the code from Chemistry.hpp -> Chemistry.cpp so it will compile separately from the rest of Quokka. If not, then the solution is to remove the Jacobian from the code and use finite differences or autodiff.

psharda:
Do you mean move everything that is in Chemistry.hpp to Chemistry.cpp?

BenWibking:
> Do you mean move everything that is in Chemistry.hpp to Chemistry.cpp?

Yes. Then remove the `template <typename problem_t>` from the function, so it just looks like this:
```
void computeChemistry(amrex::MultiFab &mf, const Real dt, const Real max_density_allowed)
```

psharda:
By the way, if I moved everything that is related to Microphysics from `src/CMakeLists` to `src/PrimordialChem/CMakeLists` (e.g., instead of `include_directories(all_of_the_Microphysics_stuff)` in `src/CMakeLists`, I do `target_include_directories(primordialchem PUBLIC all_of_the_Microphysics_stuff)`), why are Microphysics directories still being included while compiling other tests? For example, after this change, the code fails for different tests as follows:

```
[307/406] Building CUDA object src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o
FAILED: src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o
/apps/cuda/12.0.0/bin/nvcc -forward-unknown-to-host-compiler -DHAVE_PYTHON -DNAUX_NET -DSTRANG -I. -I../extern/Microphysics/util -I../extern/Microphysics/util/gcem/include -I../extern/Microphysics/integration/VODE -I../extern/Microphysics/integration/utils -I../extern/Microphysics/integration -I../extern/Microphysics/interfaces -I../extern/Microphysics/EOS -I../extern/Microphysics/EOS/primordial_chem -I../extern/Microphysics/networks/primordial_chem -I../extern/Microphysics/networks -I../extern/Microphysics/constants -I../extern/Microphysics/networks/general_null -I/home/100/ps3459/.local/lib/python3.9/site-packages/numpy/core/include -I../src -I../extern/amrex/Src/Base -I../extern/amrex/Src/Base/Parser -I../extern/amrex/Src/Boundary -I../extern/amrex/Src/AmrCore -I../extern/amrex/Src/LinearSolvers/MLMG -I../extern/amrex/Src/LinearSolvers/OpenBC -Iamrex -I../extern/fmt/include -O3 -DNDEBUG --generate-code=arch=compute_70,code=[compute_70,sm_70] --generate-code=arch=compute_80,code=[compute_80,sm_80] -Wno-psabi --fmad=false --expt-relaxed-constexpr --expt-extended-lambda -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored -maxrregcount=255 -Xcudafe --display_error_number --Wext-lambda-captures-this --use_fast_math --generate-line-info --source-in-ptx -Xcompiler -pthread -std=c++17 -MD -MT src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o -MF src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o.d -x cu -dc ../src/RadBeam/test_radiation_beam.cpp -o src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o
```

Why does the code still include Microphysics directories in the radiation beam test (or any other test for that matter)?

BenWibking:
> By the way, if I moved everything that is related to Microphysics from `src/CMakeLists` to `src/PrimordialChem/CMakeLists` (e.g., instead of `include_directories(all_of_the_Microphysics_stuff)` in `src/CMakeLists`, I do `target_include_directories(primordialchem PUBLIC all_of_the_Microphysics_stuff)`), why are Microphysics directories still being included while compiling other tests? For example, after this change, the code fails for different tests as follows:
> 
> ```
> [307/406] Building CUDA object src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o
> FAILED: src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o
> /apps/cuda/12.0.0/bin/nvcc -forward-unknown-to-host-compiler -DHAVE_PYTHON -DNAUX_NET -DSTRANG -I. -I../extern/Microphysics/util -I../extern/Microphysics/util/gcem/include -I../extern/Microphysics/integration/VODE -I../extern/Microphysics/integration/utils -I../extern/Microphysics/integration -I../extern/Microphysics/interfaces -I../extern/Microphysics/EOS -I../extern/Microphysics/EOS/primordial_chem -I../extern/Microphysics/networks/primordial_chem -I../extern/Microphysics/networks -I../extern/Microphysics/constants -I../extern/Microphysics/networks/general_null -I/home/100/ps3459/.local/lib/python3.9/site-packages/numpy/core/include -I../src -I../extern/amrex/Src/Base -I../extern/amrex/Src/Base/Parser -I../extern/amrex/Src/Boundary -I../extern/amrex/Src/AmrCore -I../extern/amrex/Src/LinearSolvers/MLMG -I../extern/amrex/Src/LinearSolvers/OpenBC -Iamrex -I../extern/fmt/include -O3 -DNDEBUG --generate-code=arch=compute_70,code=[compute_70,sm_70] --generate-code=arch=compute_80,code=[compute_80,sm_80] -Wno-psabi --fmad=false --expt-relaxed-constexpr --expt-extended-lambda -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored -maxrregcount=255 -Xcudafe --display_error_number --Wext-lambda-captures-this --use_fast_math --generate-line-info --source-in-ptx -Xcompiler -pthread -std=c++17 -MD -MT src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o -MF src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o.d -x cu -dc ../src/RadBeam/test_radiation_beam.cpp -o src/RadBeam/CMakeFiles/test_radiation_beam.dir/test_radiation_beam.cpp.o
> ```
> 
> Why does the code still include Microphysics directories in the radiation beam test (or any other test for that matter)?

That's odd. Did you try deleting the build directory and re-`cmake`-ing from scratch?

psharda:
ah nvm, I figured it out

Now Microphysics is only included in the primordial chem test.

BenWibking:
Can you adjust the timeout for the arm64 builds? Those are failing due to timeout. Those _might_ also work with `--parallel 2`.

psharda:
> > Do you mean move everything that is in Chemistry.hpp to Chemistry.cpp?
> 
> Yes. Then remove the `template <typename problem_t>` from the function, so it just looks like this:
> 
> ```
> void computeChemistry(amrex::MultiFab &mf, const Real dt, const Real max_density_allowed)
> ```

I get this error:

```
/Users/psharda/Documents/quokka/src/RadhydroSimulation.hpp:438:11: error: no member named 'chemistry' in namespace 'quokka'
                quokka::chemistry::computeChemistry(state, dt, max_density_allowed);
                ~~~~~~~~^
```

BenWibking:
In C++, you have to declare the function in the header file: https://stackoverflow.com/a/25274411/1549283

psharda:
> In C++, you have to declare the function in the header file: https://stackoverflow.com/a/25274411/1549283

Yeah just saw CloudyCooling and realised.

psharda:
Is this not right? This is Chemistry.hpp now:
 
```
#ifndef CHEMISTRY_HPP_ // NOLINT
#define CHEMISTRY_HPP_
//==============================================================================
// TwoMomentRad - a radiation transport library for patch-based AMR codes
// Copyright 2020 Benjamin Wibking.
// Released under the MIT license. See LICENSE file included in the GitHub repo.
//==============================================================================
/// \file Chemistry.hpp
/// \brief Defines methods for integrating primordial chemical network using Microphysics
///

#include "AMReX.H"
#include "AMReX_BLassert.H"
#include "AMReX_Extension.H"
#include "AMReX_GpuQualifiers.H"

namespace quokka::chemistry
{

void computeChemistry(amrex::MultiFab &mf, const Real dt, const Real max_density_allowed);

}

#endif // CHEMISTRY_HPP_

```

BenWibking:
That looks right. If you are using VSCode, it will usually tell you with red squiggles if it's wrong before you compile it.

psharda:
```
/Users/psharda/Documents/quokka/src/Chemistry.hpp:20:65: error: unknown type name 'Real'; did you mean 'amrex::Real'?
void computeChemistry(amrex::MultiFab &mf, const Real dt, const Real max_density_allowed);
                                                                ^~~~
                                                                amrex::Real
/Users/psharda/Documents/quokka/extern/amrex/Src/Base/AMReX_REAL.H:69:11: note: 'amrex::Real' declared here
    using Real = amrex_real;
          ^
```

BenWibking:
Yes, this is a standard C++ thing: https://jmmv.dev/2013/12/header-files-avoid-c-using-directives.html

psharda:
I replaced `const Real dt` with `amrex::Real dt`. But now I get more errors from the function:

```
/Users/psharda/Documents/quokka/src/Chemistry.hpp:21:30: error: no type named 'MultiFab' in namespace 'amrex'
void computeChemistry(amrex::MultiFab &mf, amrex::Real dt, amrex::Real max_density_allowed);
                      ~~~~~~~^
/Users/psharda/Documents/quokka/src/Chemistry.cpp:38:48: error: use of undeclared identifier 'problem_t'
                        const Real rho = state(i, j, k, HydroSystem<problem_t>::density_index);
                                                                    ^
```

BenWibking:
Ah... since it's no longer a template <problem_t> function, `problem_t` doesn't exist anymore. I think we have to split the current function into an outer function in `Chemistry.hpp` that assigns values to/from the MultiFabs (which can use `problem_t`) and an inner function in `Chemistry.cpp` that is just called to do the burn.

BenWibking:
Otherwise, we just keep everything as it was in `Chemistry.hpp` but delete the Jacobian.

psharda:
I want to try moving the actual burn to Chemistry.cpp first because using the Jacobians would be really nice.

So, the only thing we need to move out of Chemistry.hpp is this line: `burner(chemstate, dt);` right? 

One side question: do we need `set(PRIMORDIAL_CHEM true)` anymore, since we use `target_compile_options(test_primordial_chem PUBLIC PRIMORDIAL_CHEM)` to `#define PRIMORDIAL_CHEM`? I am thinking we don't.

BenWibking:
> I want to try moving the actual burn to Chemistry.cpp first because using the Jacobians would be really nice.
> 
> So, the only thing we need to move out of Chemistry.hpp is this line: `burner(chemstate, dt);` right?
> 
I think that's right.

> One side question: do we need `set(PRIMORDIAL_CHEM true)` anymore, since we use `target_compile_options(test_primordial_chem PUBLIC PRIMORDIAL_CHEM)` to `#define PRIMORDIAL_CHEM`? I am thinking we don't.

Yeah, agree we don't need that anymore.

BenWibking:
I was thinking we could also get around this by changing how Quokka does the indexing, but we can't do this, since Physics_Traits only works because it is templated, and it is needed to determine what the indices actually are. So we are stuck with these awkward workarounds :/

psharda:
> I was thinking we could also get around this by changing how Quokka does the indexing, but we can't do this, since Physics_Traits only works because it is templated, and it is needed to determine what the indices actually are. So we are stuck with these awkward workarounds :/

Naive question: is the main issue the number of lines in `actual_rhs.H`, or the size of the file? If I reduced the number of lines in this file from 30000 to 300, but kept the content the same, would it compile faster? What I mean is, say, here is some code with line numbers:

```
1: a = b + 
2:       c - d/10 + e*7
3:       + 9 - 19
4:       - 90
....   
10000:    -300
```

and I reduced it to 1 line:

```
1: a = b + c - d/10 + e*7 + 9 -19 - 90 .......... -300
```

would it compile faster?

BenWibking:
> > I was thinking we could also get around this by changing how Quokka does the indexing, but we can't do this, since Physics_Traits only works because it is templated, and it is needed to determine what the indices actually are. So we are stuck with these awkward workarounds :/
> 
> Naive question: is the main issue the number of lines in `actual_rhs.H`, or the size of the file? If I reduced the number of lines in this file from 30000 to 300, but kept the content the same, would it compile faster? What I mean is, say, here is some code with line numbers:
> 
> ```
> 1: a = b + 
> 2:       c - d/10 + e*7
> 3:       + 9 - 19
> 4:       - 90
> ....   
> 10000:    -300
> ```
> 
> and I reduced it to 1 line:
> 
> ```
> 1: a = b + c - d/10 + e*7 + 9 -19 - 90 .......... -300
> ```
> 
> would it compile faster?

No. The issue is with the intrinsic complexity of the code, not the formatting.

psharda:
@BenWibking pretty sure the cmake test is failing because of a silly bug in the way I have moved the burner from Chemistry.hpp to Chemistry.cpp. Can you please check? The way `burner(chemstate, dt)` works is that it takes `chemstate` (which is an object of type `burn_t`, and stores all the mass fractions, density, temp, eint) as an input, updates it post burn and returns the updated state. In fortran, this would be called an INOUT variable.

BenWibking:
I've exceeded my time on external projects this week (since I'm not funded to work on Quokka in my current position), but will be able to take a look at the start of next week.

BenWibking:
> @BenWibking pretty sure the cmake test is failing because of a silly bug in the way I have moved the burner from Chemistry.hpp to Chemistry.cpp. Can you please check? The way `burner(chemstate, dt)` works is that it takes `chemstate` (which is an object of type `burn_t`, and stores all the mass fractions, density, temp, eint) as an input, updates it post burn and returns the updated state. In fortran, this would be called an INOUT variable.

In C++, you should be able to use a reference in the same way that you would use an `INOUT` variable in Fortran. All you have to do is add an ampersand before the variable name in the function signature: `burn_t &chemstate`. Now it's a reference, which is modified in-place.

src/Chemistry.hpp
```diff
@@ -0,0 +1,132 @@
...
+			chemstate.rho = rho;
+			chemstate.e = Eint / rho;
+
+			// call the EOS to set initial internal energy e
+			eos(eos_input_re, chemstate);
+
+			// do the actual integration
+			// do it in .cpp so that it is not built at compile time for all tests
+			// which would otherwise slow down compilation due to the large RHS file
+			void chemburner(burn_t chemstate, const Real dt);
```
BenWibking: When calling a function in C++, you can't include the return type or the types of the parameters. This is only done when declaring/defining the function.

src/Chemistry.cpp
```diff
@@ -0,0 +1,19 @@
...
+///
+
+#include "Chemistry.hpp"
+#include "burn_type.H"
+#include "burner.H"
+
+namespace quokka::chemistry
+{
+
+void chemburner(burn_t chemstate, const Real dt) { return burner(chemstate, dt); }
```
BenWibking: You need to add `AMREX_GPU_DEVICE` before `void chemburner(...)` so that it can be called on the GPU.

src/Chemistry.hpp
psharda: Do you mean changing `void chemburner(burn_t chemstate, const Real dt);` to `void chemburner(chemstate, dt);` when the function `chemburner` is called in `Chemistry.hpp`? If I do that, I get an error (will dig it out). Keeping it the same, the code compiles.

psharda:
`Chemistry.cpp` fails to compile on GPUs due to `LLVM error: out of memory; nvcc error   : 'cicc' died due to signal 6` 

Maybe trying a different set of modules is worth it? I currently use `intel-mkl/2020.3.304` `cuda/12.0.0` `gcc/11.1.0` I tried using `intel-compiler-llvm/2023.0.0` instead of `intel-mkl/2020.3.304` but it doesn't work even with CPUs.

Alternatively, some StackO posts suggest using `lld` or `gold` as linkers with LLVM instead of the default `ld` as they are faster and use less memory. I am not sure if linkers using less memory will help build Chemistry.cpp though?

An easier solution could be to use destructors in `actual_rhs.H`, which should free memory once jacobians have been calculated and stored in `chemstate`. I am looking into how can I use them, let me know if you have used them before.

psharda:
Okay, I might have found a way out...

src/Chemistry.hpp
BenWibking: Function calls are always like this: `chemburner(chemstate, dt);`

BenWibking:
> `Chemistry.cpp` fails to compile on GPUs due to `LLVM error: out of memory; nvcc error : 'cicc' died due to signal 6`
> 
> Maybe trying a different set of modules is worth it? I currently use `intel-mkl/2020.3.304` `cuda/12.0.0` `gcc/11.1.0` I tried using `intel-compiler-llvm/2023.0.0` instead of `intel-mkl/2020.3.304` but it doesn't work even with CPUs.
> 
I don't think this will help.

> Alternatively, some StackO posts suggest using `lld` or `gold` as linkers with LLVM instead of the default `ld` as they are faster and use less memory. I am not sure if linkers using less memory will help build Chemistry.cpp though?
> 
I suspect it is running out of memory when building the individual translation unit, not when linking, so I don't think this will solve the problem.

> An easier solution could be to use destructors in `actual_rhs.H`, which should free memory once jacobians have been calculated and stored in `chemstate`. I am looking into how can I use them, let me know if you have used them before.
This will not solve the problem. The memory usage that is relevant is that used by the compiler, which is almost entirely determined by the complexity of the code it is trying to compile.

It should be possible in principle to compile this, since it worked for the Microphysics unit test. This suggests the compiler is trying to compile more than just the network when trying to compile Chemistry.cpp. However, it may be easier to just get rid of the analytic Jacobian.

psharda:
Victory! I reduced the RHS file for primordial chem from 30k to 6k lines of code, and file size reduced from 9.8 MB to 241 kB! No more memory issues, fast compilation, and we can continue to use the analytic jacobian.

psharda:
All 1D tests pass on CPU and GPU, but the 3D primordial chem test fails. I suspect this again has to do with my `chemburner` function not being called properly in 3D within `Chemistry.hpp`

BenWibking:
> Victory! I reduced the RHS file for primordial chem from 30k to 6k lines of code, and file size reduced from 9.8 MB to 241 kB! No more memory issues, fast compilation, and we can continue to use the analytic jacobian.

Wow, how did you do that?

psharda:
> > Victory! I reduced the RHS file for primordial chem from 30k to 6k lines of code, and file size reduced from 9.8 MB to 241 kB! No more memory issues, fast compilation, and we can continue to use the analytic jacobian.
> 
> Wow, how did you do that?

`sympy.cse` !! I didn't know sympy has it until today.

src/Chemistry.hpp
psharda: I get this error:

```
/g/data1b/jh2/ps3459/quokka/src/Chemistry.hpp:77:25: error: there are no arguments to ‘chemburner’ that depend on a template parameter, so a declaration of ‘chemburner’ must be available [-fpermissive]
   77 |                         chemburner(chemstate, dt);

```

psharda:
I just unresolved a conversation and inserted a comment. Is it visible to you?

BenWibking:
> > > Victory! I reduced the RHS file for primordial chem from 30k to 6k lines of code, and file size reduced from 9.8 MB to 241 kB! No more memory issues, fast compilation, and we can continue to use the analytic jacobian.
> > 
> > 
> > Wow, how did you do that?
> 
> `sympy.cse` !! I didn't know sympy has it until today.

_That_ would explain it. The compiler also tries to perform this optimization, but it is very stupid about it. Nice work!

src/Chemistry.hpp
BenWibking: Ah, this is because chemburner must be 'declared' before it is 'defined'. You can fix this by copying the function signature `void chemburner(burn_t chemstate, Real dt);` and pasting it near the beginning of Chemistry.hpp.

BenWibking:
> I just unresolved a conversation and inserted a comment. Is it visible to you?

Yes, replied above.

src/Chemistry.hpp
BenWibking: See https://stackoverflow.com/questions/1410563/what-is-the-difference-between-a-definition-and-a-declaration

psharda:
Avatar GPU build times out. Which `.yml` file is for the avatar GPU?

Also, it should be possible to enable parallel build again I think.

BenWibking:
> Avatar GPU build times out. Which `.yml` file is for the avatar GPU?
> 
> Also, it should be possible to enable parallel build again I think.

Yeah, you can try re-enabling parallelism. It's this one: https://github.com/quokka-astro/quokka/blob/development/.ci/azure-pipelines.yml

BenWibking:
I think it's timing out because the timeout limit for the job as a whole (default: 60 min) overrides the timeout limits for the individual tasks: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?tabs=yaml&view=azure-devops#timeouts

psharda:
> https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?tabs=yaml&view=azure-devops#timeouts

Yeah. I was confused by this because the timeout is set to 120 minutes for `cmake`. There is no `job` in the `.yml` file, so how do we override the default time for the entire process?

BenWibking:
> > https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?tabs=yaml&view=azure-devops#timeouts
> 
> Yeah. I was confused by this because the timeout is set to 120 minutes for `cmake`. There is no `job` in the `.yml` file, so how do we override the default time for the entire process?

You have to create a `job` section and then move the `tasks` inside it.

BenWibking:
See this as an example: https://github.com/quokka-astro/quokka/blob/development/.ci/azure-pipelines-aarch64.yml

psharda:
GPU tests pass too! The only one that remains queued is `debug arm64 gcc` but otherwise this PR can be reviewed now.

src/Chemistry.hpp
```diff
@@ -0,0 +1,135 @@
...
+			chemstate.rho = rho;
+			chemstate.e = Eint / rho;
+
+			// call the EOS to set initial internal energy e
+			eos(eos_input_re, chemstate);
+
+			// do the actual integration
+			// do it in .cpp so that it is not built at compile time for all tests
+			// which would otherwise slow down compilation due to the large RHS file
+			AMREX_GPU_DEVICE chemburner(chemstate, dt);
```
BenWibking: `AMREX_GPU_DEVICE` is not needed here.

src/RadhydroSimulation.hpp
```diff
@@ -94,6 +95,8 @@ template <typename problem_t> class RadhydroSimulation : public AMRSimulation<pr
 	SimulationData<problem_t> userData_;
 
 	int enableCooling_ = 0;
+	int enableChemistry_ = 0;
+	Real max_density_allowed = NAN;
```
BenWibking: This should probably be set to `std::numeric_limits<amrex::Real>::max()`

CMakeLists.txt
```diff
@@ -57,5 +57,9 @@ endif()
 
 add_subdirectory(${QuokkaCode_SOURCE_DIR}/extern/amrex ${QuokkaCode_BINARY_DIR}/amrex)
 add_subdirectory(${QuokkaCode_SOURCE_DIR}/extern/fmt ${QuokkaCode_BINARY_DIR}/fmt)
+
+#add compiler definitions again because
+#they do not carry forward from Microphysics
+add_definitions(-DNAUX_NET -DSTRANG)
```
BenWibking, line 65: Is there a way to avoid hard-coding this? Can this be defined as a variable in Microphysics?

psharda: For our purposes (ISM chemistry), it looks like we will always use these. Maybe we can come back to it later when we add more networks.

BenWibking (changes requested):
I added a few code style comments, and un-resolved a few clang-tidy messages that I think should be fixed.

The main thing is that there needs to be a check that the answer is 'correct'. It probably makes sense to evolve the simulation until it reaches chemical equilibrium, then you can compare the values with the equilibrium abundances. In terms of implementing that in Quokka, you can follow the example here: https://github.com/quokka-astro/quokka/blob/eb935c48193c5785c30de31c867af899f5b189b3/src/HydroContact/test_hydro_contact.cpp#L86.

src/RadhydroSimulation.hpp
psharda: So, like this: `Real max_density_allowed = std::numeric_limits<amrex::Real>::max()` ?

CMakeLists.txt
BenWibking: Ok, can you add a comment explaining what these definitions do / why we need them enabled?

psharda:
> I added a few code style comments, and un-resolved a few clang-tidy messages that I think should be fixed.
> 
> The main thing is that there needs to be a check that the answer is 'correct'. It probably makes sense to evolve the simulation until it reaches chemical equilibrium, then you can compare the values with the equilibrium abundances. In terms of implementing that in Quokka, you can follow the example here:
> 
> https://github.com/quokka-astro/quokka/blob/eb935c48193c5785c30de31c867af899f5b189b3/src/HydroContact/test_hydro_contact.cpp#L86
> 
> .

I will look at the clang-tidy comments again. Some of them were outdated. For several remaining clang-tidy comments, the code was borrowed from existing test problems - why were they not fixed in the existing tests? Clang-tidy should have provided the same comments in those tests too.

For the second part, I am not sure what you mean by chemical equilibrium - this is fundamentally non-equilibrium chemistry. But if we want to compare with, say, results of the Microphysics unit test, it would require increasing the density by hand as we did in the unit test (otherwise the chemistry will not evolve). The implementation would then be identical to what we have in the unit test. At this point, I would prefer finishing off this PR without this check and opening a new PR for it, because we have reached a level where a basic primordial chem implementation exists and works well on CPUs and GPUs. It also gets hard to manage bulky PRs with lots of commits, and we already closed one PR to begin the current one.

BenWibking:
> For the second part, I am not sure what you mean by chemical equilibrium - this is fundamentally non-equilibrium chemistry. But if we want to compare with, say, results of the Microphysics unit test, it would require increasing the density by hand as we did in the unit test (otherwise the chemistry will not evolve). The implementation would then be identical to what we have in the unit test. At this point, I would prefer finishing off this PR without this check and opening a new PR for it, because we have reached a level where a basic primordial chem implementation exists and works well on CPUs and GPUs. It also gets hard to manage bulky PRs with lots of commits, and we already closed one PR to begin the current one.

It could go in a separate PR. I'm not sure what you mean by 'fundamentally non-equilibrium' - is the equilibrium timescale longer than a Hubble time or something? Otherwise, we should be able to evolve until the values reach a steady state.

BenWibking:
> I will look at the clang-tidy comments again. Some of them were outdated. For several remaining clang-tidy comments, the code was borrowed from existing test problems - why were they not fixed in the existing tests? Clang-tidy should have provided the same comments in those tests too.

I didn't install clang-tidy as a pull request action until recently, so it has only reviewed PRs from the past few months or so. Anything written before then wasn't reviewed by clang-tidy.

psharda:
> > I will look at the clang-tidy comments again. Some of them were outdated. For several remaining clang-tidy comments, the code was borrowed from existing test problems - why were they not fixed in the existing tests? Clang-tidy should have provided the same comments in those tests too.
> 
> I didn't install clang-tidy as a pull request action until recently, so it has only reviewed PRs from the past few months or so. Anything written before then wasn't reviewed by clang-tidy.

Okay, all clang-tidy issues are now resolved. The last few open ones (since yesterday) were changes I had already implemented, so I resolved them too.

psharda:
> > For the second part, I am not sure what you mean by chemical equilibrium - this is fundamentally non-equilibrium chemistry. But if we want to compare with, say, results of the Microphysics unit test, it would require increasing the density by hand as we did in the unit test (otherwise the chemistry will not evolve). The implementation would then be identical to what we have in the unit test. At this point, I would prefer finishing off this PR without this check and opening a new PR for it, because we have reached a level where a basic primordial chem implementation exists and works well on CPUs and GPUs. It also gets hard to manage bulky PRs with lots of commits, and we already closed one PR to begin the current one.
> 
> It could go in a separate PR. I'm not sure what you mean by 'fundamentally non-equilibrium' - is the equilibrium timescale longer than a Hubble time or something? Otherwise, we should be able to evolve until the values reach a steady state.

I mean 1.) the rate of forward and reverse reactions can be quite different. The ratio of the forward to the backward reaction can be as high as 10^(30). Chemical equilibrium would demand the two rates be equal. 2.) timescale for chemical composition to change is >= the freefall timescale.

BenWibking (approved):

PR #269, opened 2023-04-26
Optimize load

aditivijayan:
I think I've updated extern folder. I hope this works.


aditivijayan (commented):
Looks good to me.

BenWibking:
The code looks fine to me, but the modules in the extern folder still do not match those on the development branch.

Issue #278, opened 2023-05-03
Add support for a general EOS


BenWibking:
This requires the ability to compute:
* temperature $T(\rho, e)$
* pressure $P(\rho, e)$
* specific internal energy $e(\rho, P)$
* adiabatic sound speed $a^2 = (\partial P / \partial \rho)_s$
* additional thermodynamic derivatives _(needed to calculate the Roe-averaged wavespeeds used by the Riemann solver; see eqs. 4.9-4.12 and Appendix A of [3])_
  * $\partial P / \partial \rho$
  * $\partial P / \partial e$

From these, we can derive:
* isentropic compressibility $K_s = -(1/V) (\partial V / \partial P)_s = \rho a^2$ (used in shock flattening)

[1] Glaister, P. Journal of Computational Physics, Volume 74, Issue 2, p. 382-408. https://ui.adsabs.harvard.edu/abs/1988JCoPh..74..382G/abstract

[3] Kershaw et al., Computer Methods in Applied Mechanics and Engineering, Volume 158, Issues 1–2, 1998,
p. 81-116. https://doi.org/10.1016/S0045-7825(97)00228-4

BenWibking:
The fundamental derivative $\mathcal{G}$ is also useful for constructing a nonlinear wavespeed estimate [2], since it is proportional to $(\partial^2 P / \partial \rho^2)_s$.

[2] https://doi.org/10.1016/S0045-7930(98)00033-4

psharda:
Continuing from #271 , we can now include gamma_law EOS from Microphysics. So, now we have to see if gamma_law can compute all the things you list above. I don't know what was the conclusion of your conversation with Mike on this? Are you or Mike going to modify gamma_law EOS to compute everything listed above?

BenWibking:
Well, I'm not sure what the conclusion was either... 😆

It looks like the gamma_law EOS (https://github.com/AMReX-Astro/Microphysics/blob/main/EOS/gamma_law/actual_eos.H) computes all the derivatives that are needed, yes.

It might be necessary to check that the behavior is consistent for other EOSes. For instance, Mike said that `state.cs` is *not* set to $(\partial P / \partial \rho)_s$ for the `helmholtz` equation of state. So it might require some careful checking to make sure that everything is defined as expected in all cases, in case someone decides to use a stellar matter EOS in Quokka for some reason :)

psharda:
[...]

PR #280, opened 2023-05-08
Call Microphysics for all EOS computations in all problems

### LINKED ISSUE #271: Make EOS consistent between hydro and chemistry
In simulations with hydro+chemistry, we should use the same EOS for hydro and chemistry. This is currently not the case. For primordial chem, we are using the EOS in Microphysics whereas for hydro we are using the one in Quokka Physics Traits.

psharda:
fixes #271 and #278


psharda:
@BenWibking I pushed my latest working version of Quokka here. I am only building three tests for now, to see if Microphysics is being called in the right way for all the tests. Once we converge on the changes, I will accordingly update the rest of the tests. On CPUs, it seems to be working fine.

The key issue I have now is that there will be a lot repetition of code and compilation, so we should look into ways to reduce that.

BenWibking:
There is a division-by-zero error triggered here (https://github.com/quokka-astro/quokka/actions/runs/4949568537/jobs/8851912358?pr=280#step:8:20):
```
quokka/extern/Microphysics/EOS/eos_composition.H:73:20: runtime error: division by zero
```

psharda:
Pushed through another working version of the code (still building only three tests). Removed EOS_Traits, which produced some errors. I have tried to solve them. Among other things, I am still unsure about how to best deal with `mean_molecular_weight` and `cs_isothermal` that were included in EOS_Traits.

src/Chemistry.hpp
```diff
@@ -6,9 +6,11 @@
 // Released under the MIT license. See LICENSE file included in the GitHub repo.
 //==============================================================================
 /// \file Chemistry.hpp
-/// \brief Defines methods for integrating primordial chemical network using Microphysics
+/// \brief Defines methods for integrating chemical networks using Microphysics
 ///
 
+#ifdef CHEM
```
psharda, line 12: wrapped up the entire chemistry stuff in `#ifdef CHEM`

src/CloudyCooling.hpp
```diff
@@ -265,7 +268,7 @@ template <typename problem_t> void computeCooling(amrex::MultiFab &mf, const Rea
 			const Real Egas = state(i, j, k, HydroSystem<problem_t>::energy_index);
 
 			const Real Eint = RadSystem<problem_t>::ComputeEintFromEgas(rho, x1Mom, x2Mom, x3Mom, Egas);
-			const Real gamma = quokka::EOS_Traits<problem_t>::gamma;
+			const Real gamma = eos_gamma; // quokka::EOS_Traits<problem_t>::gamma;
```
psharda: `eos_gamma` is defined within gmma_law

But, primordial_chem does not have a single, constant gamma, so there is no `eos_gamma` defined for it. So, I have removed building CloudyCooling.cpp with primordial_chem

src/EOS.hpp
```diff
@@ -8,23 +8,29 @@
...
+#include "burn_type.H"
+#include "eos.H"
+#include "extern_parameters.H"
 
 namespace quokka
 {
 static constexpr double boltzmann_constant_cgs = 1.380658e-16; // cgs
 static constexpr double hydrogen_mass_cgs = 1.6726231e-24;     // cgs
+static constexpr double cs_isothermal = NAN;		       // only used when gamma = 1
+static constexpr double mean_molecular_weight = NAN;
```
psharda: I don't know what to do about `cs_isothermal` and `mean_molecular_weight`, so I moved them from EOS_Traits to the namespace itself.

src/HydroShocktubeCMA/test_hydro_shocktube_cma.cpp
```diff
@@ -13,6 +13,7 @@
 #include <unordered_map>
 
 #include "AMReX_BC_TYPES.H"
+#include "AMReX_ParmParse.H"
```
psharda, line 16: need parmparse to read in the modified value of `eos_gamma` for this test. Default value is 5/3, but for this test we use 1.4 to be consistent with Plewa & Muller 1999

src/HydroShocktubeCMA/test_hydro_shocktube_cma.cpp
```diff
@@ -48,6 +61,14 @@ template <> struct Physics_Traits<ShocktubeProblem> {
...
 };
 
+// template <> void RadhydroSimulation<ShocktubeProblem>::preCalculateInitialConditions()
+//{
+//	// initialize microphysics routines
+//	init_extern_parameters();
+//	eos_init();
+//	network_init();
+// }
+
```
psharda, line 70: initializing microphysics routines in RadhydroSimulation.hpp now

src/PrimordialChem/CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
-add_executable(test_primordial_chem test_primordial_chem.cpp ../main.cpp ../CloudyCooling.cpp ../GrackleDataReader.cpp ../Chemistry.cpp ${primordial_chem_sources})
-target_compile_definitions(test_primordial_chem PUBLIC PRIMORDIAL_CHEM) #this will add #define PRIMORDIAL_CHEM
+set(microphysics_network_name "primordial_chem") #this will override network_name to primordial_chem for this directory only
+setup_target_for_microphysics_compilation(${microphysics_network_name} "${CMAKE_CURRENT_BINARY_DIR}/")
+
+add_executable(test_primordial_chem test_primordial_chem.cpp ../main.cpp ../GrackleDataReader.cpp ../Chemistry.cpp ${primordial_chem_sources})
+
+include_directories(${primordial_chem_dirs} "${CMAKE_CURRENT_BINARY_DIR}/" "includes/extern_parameters.H" "includes/network_properties.H")
+target_compile_definitions(test_primordial_chem PUBLIC CHEM) #this will add #define CHEM
```
psharda, line 8: removed CloudyCooling.cpp from primordial chem test, because of the non-constant gamma issue

src/RadhydroSimulation.hpp
```diff
@@ -44,14 +44,22 @@
 #include "AMReX_YAFluxRegister.H"
 
 #include "Chemistry.hpp"
+
+#ifndef CHEM
 #include "CloudyCooling.hpp"
+#endif
+
```
psharda, line 51: wrapped up CloudyCooling in #ifndef CHEM: now, we either compile CloudyCooling OR Chemistry, but not both.

src/EOS.hpp
BenWibking: `mean_molecular_weight` is only used to compute the temperature, so we should replace wherever it's used with calls to the EOS temperature function.

`cs_isothermal` is only used by the Riemann solver for tests that assume an isothermal equation of state. I'm not sure of the best way to replace this in the code.

psharda:
> There is a division-by-zero error triggered here (https://github.com/quokka-astro/quokka/actions/runs/4949568537/jobs/8851912358?pr=280#step:8:20):
> 
> ```
> quokka/extern/Microphysics/EOS/eos_composition.H:73:20: runtime error: division by zero
> ```

This error didn't pop up on my Mac for some reason. This is coming from the fact that we do not define any species while calling `gamma_law` EOS from Quokka, so the sum of mass fractions is 0. In Microphysics, `gamma_law` is always called with some network (that is not `general_null`) that has a certain number of species.

psharda:
One way I can resolve this issue is by creating an empty network in Microphysics, and using that with gamma_law EOS for Quokka tests. This also needs edits to `write_network.py` file, so I need to discuss it with Mike

psharda:
I don't know why the shocktube test is failing now. It seems to return `success=false`. We may need to look into it together.

psharda:
still fails: https://github.com/quokka-astro/quokka/actions/runs/4962088034/jobs/8879707458?pr=280

psharda:
We need to wait for Mike to get back to us on whether he is going to implement functionality to have an empty network with no EOS, as otherwise there's several changes in Microphysics objects we will have to introduce.

psharda:
Continued in #311

PR #283, opened 2023-05-12
treat all compiler warnings as errors

BenWibking:
This turns on `-Werror` when compiling the CI tests. This means that the code must compile without warnings to pass the CI tests.

### DIFF
--- a/.ci/azure-pipelines-aarch64-debug.yml
+++ b/.ci/azure-pipelines-aarch64-debug.yml
@@ -20,7 +20,7 @@ jobs:
   - task: CMake@1
     displayName: 'Configure CMake'
     inputs:
-      cmakeArgs: '.. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DAMReX_SPACEDIM=1'
+      cmakeArgs: '.. -DCMAKE_BUILD_TYPE=Debug -DWARNINGS_AS_ERRORS=ON -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DAMReX_SPACEDIM=1'
 
   - task: CMake@1
     displayName: 'Build Quokka'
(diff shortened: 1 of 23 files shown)


BenWibking:
@psharda This line is causing the CI to fail: https://github.com/psharda/Microphysics/blob/main/EOS/primordial_chem/actual_eos.H#L164 because there is a potential uninitialized use on lines 257, 258, and 261.

Can you initialize `temp`, `dens`, and `eint` to `NAN`?

psharda:
> @psharda This line is causing the CI to fail: https://github.com/psharda/Microphysics/blob/main/EOS/primordial_chem/actual_eos.H#L164 because there is a potential uninitialized use on lines 257, 258, and 261.
> 
> Can you initialize `temp`, `dens`, and `eint` to `NAN`?

Done. But the error seemed to come from HydroShocktubeCMA test, which doesn't use this file; it uses gamma_law/actual_eos.H: https://github.com/quokka-astro/quokka/actions/runs/4962088034/jobs/8879707458?pr=280

BenWibking:
Depends on https://github.com/AMReX-Astro/Microphysics/pull/1186.

psharda:
@BenWibking one GPU test failed.

BenWibking:
It looks like a CMake bug, sigh. (https://gitlab.kitware.com/cmake/cmake/-/issues/22541)

BenWibking:
I will figure out a workaround.

BenWibking:
> It looks like a CMake bug, sigh. (https://gitlab.kitware.com/cmake/cmake/-/issues/22541)

Sorry, I mean CMake "feature" :)

BenWibking:
I've disabled the warnings-as-errors option for the GPU build. This should catch all warnings produced when doing a CPU-only build on all platforms, but will not catch a warning that _only_ appears when building for GPU. This is still a significant improvement on the current state of the code, which ignores warnings in all cases.

psharda (approved):

PR #287, opened 2023-05-15
optimise load

### LINKED ISSUE #242: regrid level 0 on restart
We need to regrid level 0 when restarting, since users may restart with more MPI ranks / processors than in the previous run.

aditivijayan:
Fixes https://github.com/quokka-astro/quokka/issues/242.

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -1957,8 +1957,15 @@ template <typename problem_t> void AMRSimulation<problem_t>::ReadCheckpointFile(
 		// face-centred
 		if constexpr (Physics_Indices<problem_t>::nvarTotal_fc > 0) {
 			for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
-				amrex::VisMF::Read(state_new_fc_[lev][idim], amrex::MultiFabFileFullPrefix(lev, restart_chkfile, "Level_",
+				if (lev == 0) {
+					amrex::MultiFab tmp;
+					amrex::VisMF::Read(tmp, amrex::MultiFabFileFullPrefix(lev, restart_chkfile, "Level_",
 													   std::string("Face_") + quokka::face_dir_str[idim]));
+					state_new_fc_[0][idim].ParallelCopy(tmp, 0, 0, Physics_Indices<problem_t>::nvarTotal_cc, nghost_cc_, nghost_cc_);
+		    	} else {
+					amrex::VisMF::Read(state_new_fc_[lev][idim], amrex::MultiFabFileFullPrefix(lev, restart_chkfile, "Level_",
+													   std::string("Face_") + quokka::face_dir_str[idim]));
+				}		
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/simulation.hpp
```diff
@@ -1926,7 +1947,13 @@ template <typename problem_t> void AMRSimulation<problem_t>::ReadCheckpointFile(
...
+		if (lev == 0) {
+			amrex::MultiFab tmp;
+			amrex::VisMF::Read(tmp, amrex::MultiFabFileFullPrefix(lev, restart_chkfile, "Level_", "Cell"));
+			state_new_cc_[0].ParallelCopy(tmp, 0, 0, Physics_Indices<problem_t>::nvarTotal_cc, nghost_cc_, nghost_cc_);
+		} else {
+			amrex::VisMF::Read(state_new_cc_[lev], amrex::MultiFabFileFullPrefix(lev, restart_chkfile, "Level_", "Cell"));
+		}
 		// face-centred
 		if constexpr (Physics_Indices<problem_t>::nvarTotal_fc > 0) {
 			for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
```
BenWibking, line 1959: I realized that what you've done immediately above for the cell-centered MultiFabs also needs to be done for the face-centered MultiFabs. Otherwise, it will result in an inconsistent simulation when run with magnetic fields.

BenWibking (changes requested):
The code looks good, but I realized that the same thing needs to be done for the face-centered MultiFabs. Otherwise the code will crash if it's restarted with a run that has magnetic fields.

src/simulation.hpp
aditivijayan: Will do.

src/simulation.hpp
```diff
@@ -1926,12 +1947,26 @@ template <typename problem_t> void AMRSimulation<problem_t>::ReadCheckpointFile(
...
-				amrex::VisMF::Read(state_new_fc_[lev][idim], amrex::MultiFabFileFullPrefix(lev, restart_chkfile, "Level_",
-													   std::string("Face_") + quokka::face_dir_str[idim]));
+				if (lev == 0) {
+					amrex::MultiFab tmp;
+					amrex::VisMF::Read(tmp, amrex::MultiFabFileFullPrefix(lev, restart_chkfile, "Level_",
+											      std::string("Face_") + quokka::face_dir_str[idim]));
+					state_new_fc_[0][idim].ParallelCopy(tmp, 0, 0, Physics_Indices<problem_t>::nvarTotal_cc, nghost_cc_, nghost_cc_);
```
BenWibking: This should use `Physics_Indices<problem_t>::nvarPerDim_fc` and `nghost_fc_` instead.

BenWibking (changes requested):
The `ParallelCopy` done on the face-centered MultiFabs needs to use the number of face-centered variables and ghost cells, rather than the cell-centered numbers.

BenWibking (approved):

BenWibking:
Looks good to me. Assuming the tests pass, I'll merge.

PR #295, opened 2023-05-23
Using CMake object library to compile files needed by all tests

psharda:
Currently, we have to include `main.cpp` `CloudyCooling.cpp` and `GrackleDataReader.cpp` by hand in every test problem, via `add_executable()`. So, when we compile Quokka, these three files are being compiled multiple times - once each for every test problem.

This PR removes duplicate compilation of these files by compiling them only once as part of an object library that we can link to each test problem. On GPUs, this is made possible by using `setup_target_for_cuda_compilation` function in amrex.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,4 +10,4 @@
 [submodule "extern/Microphysics"]
 	path = extern/Microphysics
 	url = https://github.com/psharda/Microphysics
-	branch = networks-changes-dev
+	branch = development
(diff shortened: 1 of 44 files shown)


psharda:
For a 3D + GPU build on Gadi, it reduces the number of files to be compiled from 414 to 288

psharda:
Several tests use `fextract.cpp` and `interpolate.cpp` in addition to the files included in `QuokkaObj` in this PR, so we could also reduce duplicate compilation for these files in the same way. This is not too important I guess, but as we add more test problems to Quokka, this will help compiling and building the code faster (for debug, tests, etc.).

BenWibking (approved):
Assuming tests pass, approve! This is quite nice.

BenWibking:
Actually, can you add a description to this PR? (You should be able to edit it with the "three dots" button above the "no description provided" box.) In case we forget in the future why this PR was useful, it would be good to add 1-2 sentences explaining why this helps and what this does.

PR #296, opened 2023-05-23
Compute massScalars and use them in EOS routines

psharda:
This PR continues the groundwork laid down in #291 to use massScalars while computing Eint from Tgas or vice-versa. Now, it adds a function that can be called to obtain values of massScalars, and pass them on to EOS.

I also replaced the word massFractions with massScalars since the latter is more appropriate given we store partial densities and not mass fractions.

### DIFF
--- a/src/EOS.hpp
+++ b/src/EOS.hpp
@@ -82,9 +82,8 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeTgasFromEin
 		}
 
 		eos(eos_input_re, chemstate);
-		amrex::Real Tgas = chemstate.T;
+		Tgas = chemstate.T;
 	} else {
-		amrex::Real Tgas = NAN;
 		if constexpr (gamma_ != 1.0) {
 			const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
 			Tgas = Eint / (rho * c_v);
@@ -120,9 +119,8 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintFromTga
 		}
 
 		eos(eos_input_rt, chemstate);
-		amrex::Real const Eint = chemstate.e * chemstate.rho;
+		Eint = chemstate.e * chemstate.rho;
 	} else {
-		amrex::Real Eint = NAN;
 		if constexpr (gamma_ != 1.0) {
 			const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
 			Eint = rho * c_v * Tgas;
@@ -157,9 +155,8 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintTempDer
 		}
 
 		eos(eos_input_rt, chemstate);
-		amrex::Real const dEint_dT = chemstate.dedT * chemstate.rho;
+		dEint_dT = chemstate.dedT * chemstate.rho;
 	} else {
-		amrex::Real dEint_dT = NAN;
 		if constexpr (gamma_ != 1.0) {
 			const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
 			dEint_dT = rho * c_v;
--- a/src/EOS.hpp
+++ b/src/EOS.hpp
@@ -63,7 +63,7 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeTgasFromEin
     -> amrex::Real
 {
 	// return temperature for an ideal gas
-	amrex::Real Tgas;
+	amrex::Real Tgas = NAN;
 
 	if constexpr (PRIMORDIAL_CHEM_ENABLED) {
 		burn_t chemstate;
@@ -99,7 +99,7 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintFromTga
     -> amrex::Real
 {
 	// return internal energy density for a gamma-law ideal gas
-	amrex::Real Eint;
+	amrex::Real Eint = NAN;
 
 	if constexpr (PRIMORDIAL_CHEM_ENABLED) {
 		burn_t chemstate;
@@ -137,7 +137,7 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintTempDer
     -> amrex::Real
 {
 	// compute derivative of internal energy w/r/t temperature
-	amrex::Real dEint_dT;
+	amrex::Real dEint_dT = NAN;
 
 	if constexpr (PRIMORDIAL_CHEM_ENABLED) {
 		burn_t chemstate;
--- a/src/EOS.hpp
+++ b/src/EOS.hpp
@@ -62,11 +62,11 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeTgasFromEin
 
 #ifdef PRIMORDIAL_CHEM
 	burn_t chemstate;
-        chemstate.rho = rho;
-        chemstate.e = Eint / rho;
-        eos(eos_input_re, chemstate); // this will cause an error when primordial chem is run with hydro, because we also need to input values of the mass scalars in
-            // chemstate.xn
-        amrex::Real Tgas = chemstate.T;
+	chemstate.rho = rho;
+	chemstate.e = Eint / rho;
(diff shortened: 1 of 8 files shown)


psharda:
@BenWibking I forgot what was the next step :D 

We now want to use this function wherever `hydro_system.hpp` and `radiation_system.hpp` call `computeEintFromTgas` and `computeTgasFromEint` right?

BenWibking:
Yes, more generally we want to replace any direct usage of `gamma` with a call to the appropriate EOS function.

psharda:
Why do I get this error:

```
/g/data1b/jh2/ps3459/quokka/src/EOS.hpp:118:23: error: ‘Tgas’ was not declared in this scope
  118 |         chemstate.T = Tgas;

```

psharda:
and this one too @BenWibking 

```
/Users/runner/work/quokka/quokka/src/EOS.hpp:94:22: error: type 'std::optional<amrex::GpuArray<amrex::Real, nmscalars_>>' (aka 'optional<GpuArray<double, nmscalars_>>') does not provide a subscript operator
                chemstate.xn[nn] = massScalars[nn] / spmasses[nn]; // massScalars are partial densities (massFractions * rho)
                                   ^~~~~~~~~~~ ~~
```

psharda:
> 

Okay I think I solved this one. The one above it remains.

psharda:
> and this one too @BenWibking
> 
> ```
> /Users/runner/work/quokka/quokka/src/EOS.hpp:94:22: error: type 'std::optional<amrex::GpuArray<amrex::Real, nmscalars_>>' (aka 'optional<GpuArray<double, nmscalars_>>') does not provide a subscript operator
>                 chemstate.xn[nn] = massScalars[nn] / spmasses[nn]; // massScalars are partial densities (massFractions * rho)
>                                    ^~~~~~~~~~~ ~~
> ```


Okay I think I solved this one. The one above it still remains.

psharda:
okay solved the other one too

psharda:
@BenWibking is this a good chunk of changes to be approved separately, or should I add more stuff in this PR?

src/EOS.hpp
```diff
@@ -49,43 +55,107 @@ template <typename problem_t> class EOS
 
 template <typename problem_t>
 AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeTgasFromEint(amrex::Real rho, amrex::Real Eint,
-										  std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massFractions)
+										  std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars)
     -> amrex::Real
 {
 	// return temperature for an ideal gas
+
+#ifdef PRIMORDIAL_CHEM
```
BenWibking: Can you use `if constexpr (PRIMORDIAL_CHEM)` instead?

src/EOS.hpp
```diff
@@ -49,43 +55,107 @@ template <typename problem_t> class EOS
...
 }
 
 template <typename problem_t>
 AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintFromTgas(amrex::Real rho, amrex::Real Tgas,
-										  std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massFractions)
+										  std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars)
     -> amrex::Real
 {
 	// return internal energy density for a gamma-law ideal gas
+#ifdef PRIMORDIAL_CHEM
```
BenWibking: Use `if constexpr` instead.

src/EOS.hpp
```diff
@@ -49,43 +55,107 @@ template <typename problem_t> class EOS
...
 }
 
 template <typename problem_t>
 AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintTempDerivative(const amrex::Real rho, const amrex::Real /*Tgas*/,
-											std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massFractions)
+											std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars)
     -> amrex::Real
 {
 	// compute derivative of internal energy w/r/t temperature
+#ifdef PRIMORDIAL_CHEM
```
BenWibking: `if constexpr` is preferred.

BenWibking (changes requested):
Looks good, but I strongly prefer `if constexpr` over `ifdef` if possible. The code formatting is easier to read and the compiler can analyze it for possible bugs (whereas ifdefs are invisible to the compiler).

psharda:
Does `if constexpr (PRIMORDIAL_CHEM)` need an end, like `#ifdef` has `#endif`? Also, is there an 'else' associated with `if constexpr`?

BenWibking:
> Does `if constexpr (PRIMORDIAL_CHEM)` need an end, like `#ifdef` has `#endif`? Also, is there an 'else' associated with `if constexpr`?

`if constexpr` works just like an `if` statement in C++. The only difference is that the keyword `constexpr` immediately follows the `if` that begins the statement. Everything else is the same - there are curly braces and `else` statements just as before, e.g.:
```
if constexpr (condition) {
  return true;
} else {
  return false;
}
```

psharda:
The latest commit will fail like this:

```
/g/data1b/jh2/ps3459/quokka/src/EOS.hpp:92:16: error: ‘Tgas’ was not declared in this scope
   92 |         return Tgas;

```

which I don't follow, because Tgas is initialized in both the if and else parts of `constexpr (PRIMORDIAL_CHEM)` so it should already be declared, no?

psharda:
ah nvm

src/EOS.hpp
```diff
@@ -49,42 +59,111 @@ template <typename problem_t> class EOS
...
+
+		if (massScalars.has_value()) {
+			const auto &massArray = massScalars.value();
+			for (int nn = 0; nn < nmscalars_; ++nn) {
+				chemstate.xn[nn] = massArray[nn] / spmasses[nn]; // massScalars are partial densities (massFractions * rho)
+			}
+		}
+
+		eos(eos_input_re, chemstate);
+		amrex::Real Tgas = chemstate.T;
```
BenWibking: This `Tgas` shadows the `Tgas` in the enclosing scope. So the values of Tgas returned will always be `NAN`. You can fix this by just changing this line to `Tgas = chemstate.T;`.

src/EOS.hpp
```diff
@@ -49,42 +59,111 @@ template <typename problem_t> class EOS
...
+	} else {
+		amrex::Real Tgas = NAN;
```
BenWibking: Same issue as above. This can be fixed by deleting this line.

src/EOS.hpp
```diff
@@ -49,42 +59,111 @@ template <typename problem_t> class EOS
...
+
+		if (massScalars.has_value()) {
+			const auto &massArray = massScalars.value();
+			for (int nn = 0; nn < nmscalars_; ++nn) {
+				chemstate.xn[nn] = massArray[nn] / spmasses[nn]; // massScalars are partial densities (massFractions * rho)
+			}
+		}
+
+		eos(eos_input_rt, chemstate);
+		amrex::Real const Eint = chemstate.e * chemstate.rho;
```
BenWibking: This value of `Eint` shadows the `Eint` defined in the enclosing scope. So this function will always return `NAN`. You can fix this by changing this line to just `Eint = chemstate.e * chemstate.rho;`.

src/EOS.hpp
```diff
@@ -49,42 +59,111 @@ template <typename problem_t> class EOS
...
+	} else {
+		amrex::Real Eint = NAN;
```
BenWibking: Same issue as above. Just get rid of this line.

src/EOS.hpp
```diff
@@ -15,6 +15,16 @@
...
+#include "burn_type.H"
+#include "eos.H"
+#include "extern_parameters.H"
+
+#ifdef PRIMORDIAL_CHEM
+constexpr bool PRIMORDIAL_CHEM_ENABLED = true;
+#else
+constexpr bool PRIMORDIAL_CHEM_ENABLED = false;
+#endif
+
```
BenWibking, line 27: Since this code is not specific to primordial chemistry, can you change the variable names to just `CHEMISTRY_ENABLED` or `CHEM_ENABLED`?

BenWibking (changes requested):

src/EOS.hpp
psharda: In an earlier PR, we decided to continue using PRIMORDIAL_CHEM instead of CHEM. Once we are finished with all the code development for Pop III simulations, I will then systematically go through each file and make the changes. So, it's better to keep it PRIMORDIAL_CHEM for now and make the change everywhere at the same time in the end .

BenWibking: Assume that I don't remember any earlier PRs. My memory is wiped when I click 'merge' :)

BenWibking (approved):
This looks good to me. If the tests pass, I'm happy to merge.

psharda:
Ready!

PR #298, opened 2023-05-24
include massScalars in call to EOS functions in hydro and radiation

psharda:
We now include massScalars while calling EOS functions in `hydro_system.hpp` and `radiation_system.hpp`

### DIFF
--- a/src/EOS.hpp
+++ b/src/EOS.hpp
@@ -82,9 +82,8 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeTgasFromEin
 		}
 
 		eos(eos_input_re, chemstate);
-		amrex::Real Tgas = chemstate.T;
+		Tgas = chemstate.T;
 	} else {
-		amrex::Real Tgas = NAN;
 		if constexpr (gamma_ != 1.0) {
 			const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
 			Tgas = Eint / (rho * c_v);
@@ -120,9 +119,8 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintFromTga
 		}
 
 		eos(eos_input_rt, chemstate);
-		amrex::Real const Eint = chemstate.e * chemstate.rho;
+		Eint = chemstate.e * chemstate.rho;
 	} else {
-		amrex::Real Eint = NAN;
 		if constexpr (gamma_ != 1.0) {
 			const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
 			Eint = rho * c_v * Tgas;
@@ -157,9 +155,8 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintTempDer
 		}
 
 		eos(eos_input_rt, chemstate);
-		amrex::Real const dEint_dT = chemstate.dedT * chemstate.rho;
+		dEint_dT = chemstate.dedT * chemstate.rho;
 	} else {
-		amrex::Real dEint_dT = NAN;
 		if constexpr (gamma_ != 1.0) {
 			const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
 			dEint_dT = rho * c_v;
--- a/src/EOS.hpp
+++ b/src/EOS.hpp
@@ -63,7 +63,7 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeTgasFromEin
     -> amrex::Real
 {
 	// return temperature for an ideal gas
-	amrex::Real Tgas;
+	amrex::Real Tgas = NAN;
 
 	if constexpr (PRIMORDIAL_CHEM_ENABLED) {
 		burn_t chemstate;
@@ -99,7 +99,7 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintFromTga
     -> amrex::Real
 {
 	// return internal energy density for a gamma-law ideal gas
-	amrex::Real Eint;
+	amrex::Real Eint = NAN;
 
 	if constexpr (PRIMORDIAL_CHEM_ENABLED) {
 		burn_t chemstate;
@@ -137,7 +137,7 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEintTempDer
     -> amrex::Real
 {
 	// compute derivative of internal energy w/r/t temperature
-	amrex::Real dEint_dT;
+	amrex::Real dEint_dT = NAN;
 
 	if constexpr (PRIMORDIAL_CHEM_ENABLED) {
 		burn_t chemstate;
--- a/src/EOS.hpp
+++ b/src/EOS.hpp
@@ -62,11 +62,11 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeTgasFromEin
 
 #ifdef PRIMORDIAL_CHEM
 	burn_t chemstate;
-        chemstate.rho = rho;
-        chemstate.e = Eint / rho;
-        eos(eos_input_re, chemstate); // this will cause an error when primordial chem is run with hydro, because we also need to input values of the mass scalars in
-            // chemstate.xn
-        amrex::Real Tgas = chemstate.T;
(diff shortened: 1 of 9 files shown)


psharda:
Looks like the EOS functions (compute Tgas from Eint, vice-versa, dEint/dTgas) are only used in two files that are 'common' to all tests - `hydro_system.hpp` and `radiation_system.hpp`

psharda:
@BenWibking 
How can I call a function defined in `hydro_system.hpp` in `radiation_system.hpp`? I tried using `HydroSystem<problem_t>::ComputeMassScalars` but I get an error:

```
/g/data1b/jh2/ps3459/quokka/src/radiation_system.hpp:901:36: error: ‘HydroSystem’ was not declared in this scope; did you mean ‘RadSystem’?
  901 |                 auto massScalars = HydroSystem<problem_t>::ComputeMassScalars(consPrev, i, j, k);

```

BenWibking:
You could try adding it as a header include. It's possible that will cause a circular dependency, which requires moving things to different files to avoid circular includes.

psharda:
> You could try adding it as a header include. It's possible that will cause a circular dependency, which requires moving things to different files to avoid circular includes.

Yes. This circular dependency wont be a problem if I define ComputeMassScalars in radiation_system.hpp, as this file is included in hydro_system.hpp but not vice-versa.

BenWibking:
> > You could try adding it as a header include. It's possible that will cause a circular dependency, which requires moving things to different files to avoid circular includes.
> 
> Yes. This circular dependency wont be a problem if I define ComputeMassScalars in radiation_system.hpp, as this file is included in hydro_system.hpp but not vice-versa.

That should work. Slightly confusing, so might be helpful to add a comment explaining why it's there.

psharda:
Okay, I moved the declaration and definition of `ComputeMassScalars` from hydro_system to radiation_system. I also had to define a `scalar0_index` in radiation_system's conserve variables, otherwise this doesn't work. Please check that this is defined correctly. I did not include a mirror of scalar0_index in primitive variables, like you do in hydro_system.

I now get this error on the type of array - this was not a problem when `ComputeMassScalars` was in hydro_system

```
/g/data1b/jh2/ps3459/quokka/src/radiation_system.hpp:558:48: error: cannot convert ‘const quokka::Array4View<const double, FluxDir::X1, void>’ to ‘const amrex::Array4<const double>&’
  558 |         auto massScalars_L = ComputeMassScalars(consVar, i - 1, j, k);
      |                              ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
/g/data1b/jh2/ps3459/quokka/src/radiation_system.hpp:519:123: note:   initializing argument 1 of ‘static amrex::GpuArray<double, RadSystem<problem_t>::nmscalars_> RadSystem<problem_t>::ComputeMassScalars(const amrex::Array4<const double>&, int, int, int) [with problem_t = SuOlsonProblemCgs]’
  519 | AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto RadSystem<problem_t>::ComputeMassScalars(amrex::Array4<const amrex::Real> const &cons, int i, int j, int k)
      |                                                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
```

BenWibking:
Defining the scalar index twice is not ideal. However, let's put that aside for the moment.

The issue is that in this case `consVar` is a different type of array. It's not an `amrex::Array4<Real>` that it is expecting. There is a general solution to this problem, which is to allow it to accept either type of array, by using C++20 concepts (https://en.cppreference.com/w/cpp/language/constraints).

psharda:
I tried using concepts but it is getting very messy... 
do you have a suitable example?

also, we need to raise cxx standard to 20 for concepts, which raises warnings in other files (simulation.hpp)

BenWibking:
Ah wait, we can actually do this without needing to use concepts. You can just make it a template function:
```
template <typename ArrayType>
auto GetScalars(ArrayType const &arr, int i, int j, int k) -> amrex::GpuArray<amrex::Real, nmscalars_> {
  // ...
}
```

psharda:
> Ah wait, we can actually do this without needing to use concepts. You can just make it a template function:
> 
> ```
> template <typename ArrayType>
> auto GetScalars(ArrayType const &arr, int i, int j, int k) -> amrex::GpuArray<amrex::Real, nmscalars_> {
>   // ...
> }
> ```

What about the declaration of the function that goes in `HyperbolicSystem`? How do I change that? If I just use ArrayType there, it throws an error `ArrayType not declared`. So the solution I tried was to define template ArrayType for RadSystem, but then it leads to all sorts of problems with two templates (problem_t and ArrayType) defined for RadSystem whereas most of its functions only use problem_t....

BenWibking:
> > Ah wait, we can actually do this without needing to use concepts. You can just make it a template function:
> > ```
> > template <typename ArrayType>
> > auto GetScalars(ArrayType const &arr, int i, int j, int k) -> amrex::GpuArray<amrex::Real, nmscalars_> {
> >   // ...
> > }
> > ```
> 
> What about the declaration of the function that goes in `HyperbolicSystem`? How do I change that? If I just use ArrayType there, it throws an error `ArrayType not declared`. So the solution I tried was to define template ArrayType for RadSystem, but then it leads to all sorts of problems with two templates (problem_t and ArrayType) defined for RadSystem whereas most of its functions only use problem_t....

This is a bit confusing because class member functions can have two template types associated with them: i) the template type of the class they are a member of, and ii) their own template type. These are actually completely independent. So you can have a template function that is a member of a template class like this (there are some examples in the code of this):
```
template <typename ClassTemplateType>
template <typename TemplateType>
void MyClass<ClassTemplateType>::MyFunction(TemplateType input_array) { /* ... */ }
```

BenWibking:
Here's an example of this: https://github.com/quokka-astro/quokka/blob/4baa62394d42a35c4000c5912258317baa2ccd43/src/AdvectionSimulation.hpp#L382

psharda:
Thanks. I tried something (see the latest changes in `radiation_system.hpp`) but I get the error:

```
/g/data1b/jh2/ps3459/quokka/src/radiation_system.hpp:120:19: error: ‘ArrayType’ has not been declared
  120 |         template <ArrayType>

``` 

Now, I see that some functions in RadSystem use an additional template called `template<FluxDir DIR>`, and this seems to be the thing that I need for ComputeMassScalars. However, `FluxDir` is defined as a class in `ArrayView.hpp` which is why it works. But, my `ArrayType` template is not defined, which is why I guess this error comes?

psharda:
@BenWibking I made some more modifications. Now, the only thing that remains is declaring 'ArrayType'. How/where do I do that?

BenWibking:
> @BenWibking I made some more modifications. Now, the only thing that remains is declaring 'ArrayType'. How/where do I do that?

You don't need to declare `ArrayType`. In fact, you don't need to specify the type at all, since it should be able to deduce the type automatically from the type of the first argument. You can just call the function normally.

psharda:
> > @BenWibking I made some more modifications. Now, the only thing that remains is declaring 'ArrayType'. How/where do I do that?
> 
> You don't need to declare `ArrayType`. In fact, you don't need to specify the type at all, since it should be able to deduce the type automatically from the type of the first argument. You can just call the function normally.

Yes, I just realised that after I texted.

Okay, code compiles, finally!

psharda:
Should I add more stuff to this PR @BenWibking ? I am thinking this is a small and good chunk to be reviewed separately, but I can add more if you want.

BenWibking:
> Should I add more stuff to this PR @BenWibking ? I am thinking this is a small and good chunk to be reviewed separately, but I can add more if you want.

Ahh no, don't add more. I can review this as-is.

BenWibking (approved):

BenWibking:
Looks good to me, assuming all the tests pass.

psharda:
With this PR, quokka needs more time on avatargpu's.... 60 minutes are not enough. I guess we increase timeout to 120 minutes again?

BenWibking:
> With this PR, quokka needs more time on avatargpu's.... 60 minutes are not enough. I guess we increase timeout to 120 minutes again?

This PR should not cause a significant increase in runtime. I suspect someone else is using the node or the filesystem is broken.

psharda:
> > With this PR, quokka needs more time on avatargpu's.... 60 minutes are not enough. I guess we increase timeout to 120 minutes again?
> 
> This PR should not cause a significant increase in runtime. I suspect someone else is using the node or the filesystem is broken.

Could be, I was tracking the progress of the test and it was quite weird. For 6 minutes, it didn't seem to do anything...
I will `/azp run` again

BenWibking:
> > > With this PR, quokka needs more time on avatargpu's.... 60 minutes are not enough. I guess we increase timeout to 120 minutes again?
> > 
> > 
> > This PR should not cause a significant increase in runtime. I suspect someone else is using the node or the filesystem is broken.
> 
> Could be, I was tracking the progress of the test and it was quite weird. For 6 minutes, it didn't do anything... I will `/azp run` again

Aha. That almost always indicates a filesystem problem.

PR #311, opened 2023-06-13
Two EOS compilation for Quokka + fix build on cray systems

psharda:
Continuation of #280


psharda:
Primordial chem test fails to run. The problem is that right now, primordial chem test includes both network_properties.H from gamma_law EOS and primordial_chem EOS. We only want to include the primordial chem EOS based network_properties file.

I haven't tested max's new structs yet, the focus is on getting Quokka to build with multiple EOSes at the same time.

psharda:
I can't see a way out except for removing `include_directories(${gamma_law_dirs})` from `src/CMakeLists` and adding `target_include_directories(${gamma_law_dirs})` for all quokka tests (except primordial chem).

psharda:
> I can't see a way out except for removing `include_directories(${gamma_law_dirs})` from `src/CMakeLists` and adding `target_include_directories(${gamma_law_dirs})` for all quokka tests (except primordial chem).

This doesn't work either.

psharda:
okay, @BenWibking , this should now hopefully build

psharda:
The two key changes I did are: 1.) we need to de-link `QuokkaObj`  from primordial chem test, because this is built with `gamma_law` source files. An alternate way would be to link QuokkaObj  individually to all targets, but then it defeats the purpose of storing `main.cpp, cloudycooling.cpp and grackledatareader.cpp`  in an object library in the first place. So, we just de-link this object from primordial chem test, and re-compile the three .cpp files for this test.

2.) I could not find a way to only include `primordial_chem_dirs`  for primordial chem test. No matter what I tried, `gamma_law_dirs`  need to be included, or you run into linking and undefined reference errors. So, the trick was to use the `BEFORE`  keyword of `include_directories`  to make sure `primordial_chem_dirs` get priority, and they override `gamma_law_dirs`, for the primordial chem test. This is necessary to link the correct `extern_parameters.H` and `network_properties.H` files with primordial chem test.

3.) The corresponding `CMakeLists.txt` file in Microphysics was also updated (in the `track-max-has-xn` branch.

BenWibking:
FYI- I need to figure out how to add a CI build for the specific ROCm versions available on Setonix before merging this.

psharda:
> FYI- I need to figure out how to add a CI build for the specific ROCm versions available on Setonix before merging this.

Makes sense. Note that I have not yet tested Max's actual changes to Microphysics structs. So far, I have just been able to ensure we can build quokka with two EOS. Once these tests pass on avatarGPU, I will then attempt to validate his changes

psharda:
Hmm, gpu ran out of memory. Don't know if this could be because other pipelines (debug and release) were also running with the 3d gpu nvcc one?

BenWibking:
> Hmm, gpu ran out of memory. Don't know if this could be because other pipelines (debug and release) were also running with the 3d gpu nvcc one?

No, they run on different machines.

BenWibking:
This almost always means someone else is trying to use the machine at the same time:
```
  what():  Out of gpu memory. Free: 23572905984 Asked: 25559728128
```
it's usually Sven.

psharda:
Damnit Sven!

psharda:
@BenWibking I also checked that this PR will still build fine with the change proposed in #316

psharda:
First step is to restructure Microphysics Cmakelists. This will be done by https://github.com/AMReX-Astro/Microphysics/pull/1237

I can forsee a problem when we compile Quokka after this Microphysics PR is merged. I will work on a solution once the microphysics PR is merged.

psharda:
> First step is to restructure Microphysics Cmakelists. This will be done by [AMReX-Astro/Microphysics#1237](https://github.com/AMReX-Astro/Microphysics/pull/1237)
> 
> I can forsee a problem when we compile Quokka after this Microphysics PR is merged. I will work on a solution once the microphysics PR is merged.

Nope, no problems if we remove `eos_composition.H` from primordial chem EOS directory. This will be accomplished by https://github.com/AMReX-Astro/Microphysics/pull/1238

psharda:
ready for review @BenWibking

src/CMakeLists.txt
```diff
@@ -90,13 +96,12 @@ include(CTest)
...
-    setup_target_for_cuda_compilation(QuokkaObj)
-endif()
-
-link_libraries(QuokkaObj)
+#we don't use it anymore because it gives issues on Cray systems
+#add_library(QuokkaObj OBJECT ${QuokkaObjSources} ${gamma_law_sources})
+#if(AMReX_GPU_BACKEND MATCHES "CUDA")
+#    setup_target_for_cuda_compilation(QuokkaObj)
+#endif()
+#link_libraries(QuokkaObj)
```
BenWibking, line 104: Can you delete the commented-out code? It will still be in the git history, but it makes it cluttered to read.

src/PrimordialChem/CMakeLists.txt
```diff
@@ -1,6 +1,20 @@
...
+include_directories(BEFORE ${primordial_chem_dirs} "${CMAKE_CURRENT_BINARY_DIR}/" "includes/extern_parameters.H" "includes/network_properties.H")
+
+add_executable(test_primordial_chem test_primordial_chem.cpp ../main.cpp ../GrackleDataReader.cpp ../CloudyCooling.cpp ../Chemistry.cpp ${primordial_chem_sources})
 target_compile_definitions(test_primordial_chem PUBLIC PRIMORDIAL_CHEM) #this will add #define PRIMORDIAL_CHEM
 
+#de-link QuokkaObj from this target to avoid multiple definitions of same variables
+#this is also why we build main.cpp CloudyCooling.cpp GrackleDataReader.cpp again for this target
+#get_target_property(tlink_libs test_primordial_chem LINK_LIBRARIES)
+#list(REMOVE_ITEM tlink_libs QuokkaObj)
+#set_property(TARGET test_primordial_chem PROPERTY LINK_LIBRARIES ${tlink_libs})
```
BenWibking, line 15: I think it would be good to delete this. We can always go back to look at the old PRs or search through the repo.

BenWibking (changes requested):
Only nitpick is deleting the commented-out, unused old code.

Otherwise, it looks good to me assuming the GPU tests pass.

src/CMakeLists.txt
psharda: Yes, I thought about that. But I would like to keep it in the file for now until the Setonix build is successful. I know that it will be stored in the git history, but with so many PRs and commits in each PRs I lose track of which exact commit/PR was some info present in. For example, I worked on the `setup_target_for_microphysics_compilation` function a while ago, and it took some man hours to dig the right PR/commit out. And these extra linking/de-linking commands took me hours to work through.

I want to revisit this object library issue once we build it fine on Cray and Setonix. There is also some other old commented out code in this file, so we can make a quick, separate PR later on to de clutter the file

src/PrimordialChem/CMakeLists.txt
psharda: Same response as above.

psharda:
Just a note for us to keep in mind: I haven't yet tested Max's new eos struct, and whether it works in Quokka. That will be the next step, in a new PR.

BenWibking (approved):

PR #318, opened 2023-06-15
Update dependencies_hip.sh to use ROCm 5.4.3

BenWibking:
This changes the HIP CI action to use ROCm 5.4.3 rather than the latest ROCm version (currently 5.5.1), since this is the latest version available on Setonix and Frontier.


BenWibking:
For some reason, ROCm 5.2 fails to install as a dependency on the virtual machine, so this is using 5.4.3 instead.

psharda (approved):

PR #319, opened 2023-06-15
Run only one test per commit

psharda:
#317 used 2 cores to run azure pipelines, but they trigger 2 tests each, so each commit is being tested four times. This PR can ensure each commit only triggers one test but uses both cores.

### DIFF
--- a/.ci/azure-pipelines-aarch64-debug.yml
+++ b/.ci/azure-pipelines-aarch64-debug.yml
@@ -30,7 +30,7 @@ jobs:
   - task: CMake@1
     displayName: 'Run CTest'
     inputs:
-      cmakeArgs: '-E chdir . ctest -T Test --output-on-failure'
+      cmakeArgs: '-E chdir . ctest -j 2 -T Test --output-on-failure'
 
   - task: PublishTestResults@2
     inputs:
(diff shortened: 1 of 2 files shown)


BenWibking:
The arm64 debug CI build failed.

.ci/azure-pipelines-aarch64-debug.yml
```diff
@@ -22,15 +22,11 @@ jobs:
     inputs:
       cmakeArgs: '.. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DAMReX_SPACEDIM=1'
 
-  - task: CMake@1
+  - script: cmake --build .
     displayName: 'Build Quokka'
-    inputs:
-      cmakeArgs: '--build .'
```
BenWibking, line 28: Is this change necessary?

.ci/azure-pipelines-aarch64-debug.yml
```diff
@@ -22,15 +22,11 @@ jobs:
...
 
-  - task: CMake@1
+  - script: ctest -j 2 -I 1,1 -T Test --output-on-failure
```
BenWibking, line 28: Can you change the arguments in `cmakeArgs` below instead?

psharda:
Doesn't look like we need it anymore, since recent pushes to other open PRs only trigger two tests per commit (as before).

PR #321, opened 2023-06-19
update Setonix CMake settings

### LINKED ISSUE #312: add conditional includes for std::filesystem
When building with `hipcc`, we have to `#include <experimental/filesystem>`, but otherwise we just `#include <filesystem>`.

In C++17, we can do this:
```
#if __has_include(<filesystem>)
#include <filesystem>
#elif __has_include(<experimental/filesystem>)
#include <experimental/filesystem>
namespace std
{
namespace filesystem = experimental::filesystem;
}
#endif
```

BenWibking:
Adds Bash scripts to use ROCm 5.3.0 and ROCm 5.4.3. Also adds a conditional include for `<filesystem>`.

The recommended build procedure on Setonix is:
```
source scripts/setonix-rocm5.4.3.profile
mkdir build; cd build
cmake -C ../cmake/setonix.profile ..
make -j16
```

Then a single-node test job can be run with:
```
cd ..
sbatch scripts/setonix-1node.submit
```

Fixes https://github.com/quokka-astro/quokka/issues/312.

### DIFF
--- a/cmake/setonix.cmake
+++ b/cmake/setonix.cmake
@@ -1,11 +1,8 @@
 # host configuration
 # run with `cmake -C host_config.cmake ..` from inside build directory
 
-set(CMAKE_C_COMPILER "cc" CACHE PATH "")
-set(CMAKE_CXX_COMPILER "CC" CACHE PATH "")
 set(AMReX_GPU_BACKEND HIP CACHE STRING "")
 set(AMReX_AMD_ARCH gfx90a CACHE STRING "") # MI250X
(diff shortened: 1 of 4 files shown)


BenWibking:
@psharda @markkrumholz Could I get a review on this PR from one of you?

psharda:
not that it would matter, but just doing azp run to make sure changes to simulation.hpp build fine on avatar

BenWibking:
> not that it would matter, but just doing azp run to make sure changes to simulation.hpp build fine on avatar

definitely always good to test!

psharda (approved):

Issue #26, opened 2022-02-09
nvcc internal compiler error when calling template parameter function from inside device lambda


BenWibking:
This code

```
template <typename problem_t>
template <typename F>
void HyperbolicSystem<problem_t>::PredictStep(
    arrayconst_t &consVarOld, array_t &consVarNew,
    std::array<arrayconst_t, AMREX_SPACEDIM> fluxArray, const double dt_in,
    amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dx_in, amrex::Box const &indexRange,
    const int nvars, F&& isStateValid, amrex::Array4<int> const &redoFlag)
{
	auto const dt = dt_in;
	auto const dx = dx_in[0];
	auto const x1Flux = fluxArray[0];
#if (AMREX_SPACEDIM >= 2)
	auto const dy = dx_in[1];
	auto const x2Flux = fluxArray[1];
#endif
#if (AMREX_SPACEDIM == 3)
	auto const dz = dx_in[2];
	auto const x3Flux = fluxArray[2];
#endif

	amrex::ParallelFor(
	    indexRange, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
			for (int n = 0; n < nvars; ++n) {
				consVarNew(i, j, k, n) =
				consVarOld(i, j, k, n) +
				(AMREX_D_TERM( (dt / dx) * (x1Flux(i, j, k, n) - x1Flux(i + 1, j, k, n)),
							+ (dt / dy) * (x2Flux(i, j, k, n) - x2Flux(i, j + 1, k, n)),
							+ (dt / dz) * (x3Flux(i, j, k, n) - x3Flux(i, j, k + 1, n))
							));
			}

			// check if state is valid -- flag for re-do if not
			if (!isStateValid(consVarNew, i, j, k)) {
				redoFlag(i, j, k) = quokka::redoFlag::redo;
			} else {
				redoFlag(i, j, k) = quokka::redoFlag::none;
			}
	    });
}
```

triggers the following nvcc compiler bug:

```
[137/219] Building CUDA object src/Advection/CMakeFiles/test_advection.dir/test_advection.cpp.o
FAILED: src/Advection/CMakeFiles/test_advection.dir/test_advection.cpp.o 
[...]

PR #25, opened 2022-02-09
implement first-order flux correction (FOFC) for hydro

BenWibking:
Replaces high-order hydro fluxes with donor cell fluxes when the high-order fluxes causes a pressure or density positivity failure. Should be virtually identical to the algorithm used in Athena (not Athena++), but simpler implementation.

This is only for hydro states, but will be extended to the radiation subsystem in a future PR.

PR #24, opened 2022-02-02
add optically-thin cooling (radiative, collisional, etc.)

BenWibking:
Add optically-thin cooling assuming ionization equilibrium in the presence of a UV background.

PR #324, opened 2023-06-22
Use microphysics EOS in Quokka's EOS.hpp for all quokka tests + use amu instead of hydrogen mass throughout Quokka

### LINKED ISSUE #325: Why is the mean molecular weight not consistent between all tests?
Most tests that use `EOSTraits mean_molecular_weight` define it as a mass quantity, with units of mass. However, three tests (RadPulse, RadStreaming, RadhydroShock) define it as a dimensionless number of order unity. Why? 

```
ps3459@gadi-login-04:/g/data1b/jh2/ps3459/quokka/src>grep -rnw . -e 'mean_molecular_weight'
./RadhydroShell/test_radhydro_shell.cpp:48:	static constexpr double mean_molecular_weight = 2.2 * m_H;
./RadStreaming/test_radiation_streaming.cpp:26:	static constexpr double mean_molecular_weight = 1.0;
./RayleighTaylor3D/test_hydro3d_rt.cpp:27:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./HydroSMS/test_hydro_sms.cpp:26:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./HydroContact/test_hydro_contact.cpp:26:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./RadhydroShock/test_radhydro_shock.cpp:65:	static constexpr double mean_molecular_weight = mu;
./FCQuantities/test_fc_quantities.cpp:30:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./RadPulse/test_radiation_pulse.cpp:30:	static constexpr double mean_molecular_weight = 1.0;
./ODEIntegration/test_ode.cpp:19:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./PassiveScalar/test_scalars.cpp:28:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./PrimordialChem/test_primordial_chem.cpp:46:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./RadhydroShockCGS/test_radhydro_shock_cgs.cpp:68:	static constexpr double mean_molecular_weight = m_H;
./HydroLeblanc/test_hydro_leblanc.cpp:33:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./RadTube/test_radiation_tube.cpp:40:	static constexpr double mean_molecular_weight = mu;
./RadMarshakAsymptotic/test_radiation_marshak_asymptotic.cpp:29:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./RadBeam/test_radiation_beam.cpp:36:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./HydroWave/test_hydro_wave.cpp:26:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./RadTophat/test_radiation_tophat.cpp:41:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./RadShadow/test_radiation_shadow.cpp:37:	static constexpr double mean_molecular_weight = 10. * quokka::hydrogen_mass_cgs;
./HydroShuOsher/test_hydro_shuosher.cpp:28:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
./Cooling/test_cooling.cpp:38:	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
```

psharda:
@BenWibking , look at this test change I made to `EOS.hpp`: for tests other than primordial chem, now, I want to use microphysics EOS to `ComputeTgasFromEint`. I create a burn_t object, add the density, Eint and mean_molecular_weight to it, and then use the `eos(eos_input_re, ...)` call to find the temperature. 

While compiling a test that actually uses `ComputeTgasFromEint` (e.g., HydroShocktubeCMA), I get these errors:

```
[ 38%] Linking CXX executable test_hydro_shocktube_cma
CMakeFiles/test_hydro_shocktube_cma.dir/test_hydro_shocktube_cma.cpp.o: In function `HydroSystem<ShocktubeProblem>::EnforceLimits(double, double, double, double, double, amrex::MultiFab&)::{lambda(int, int, int, int)#1}::operator()(int, int, int, int) const':
test_hydro_shocktube_cma.cpp:(.text._ZZN11HydroSystemI16ShocktubeProblemE13EnforceLimitsEdddddRN5amrex8MultiFabEENKUliiiiE_clEiiii[_ZZN11HydroSystemI16ShocktubeProblemE13EnforceLimitsEdddddRN5amrex8MultiFabEENKUliiiiE_clEiiii]+0x22f): undefined reference to `EOSData::initialized'
test_hydro_shocktube_cma.cpp:(.text._ZZN11HydroSystemI16ShocktubeProblemE13EnforceLimitsEdddddRN5amrex8MultiFabEENKUliiiiE_clEiiii[_ZZN11HydroSystemI16ShocktubeProblemE13EnforceLimitsEdddddRN5amrex8MultiFabEENKUliiiiE_clEiiii]+0x252): undefined reference to `eos_rp::eos_assume_neutral'
test_hydro_shocktube_cma.cpp:(.text._ZZN11HydroSystemI16ShocktubeProblemE13EnforceLimitsEdddddRN5amrex8MultiFabEENKUliiiiE_clEiiii[_ZZN11HydroSystemI16ShocktubeProblemE13EnforceLimitsEdddddRN5amrex8MultiFabEENKUliiiiE_clEiiii]+0x26a): undefined reference to `EOSData::mine'
test_hydro_shocktube_cma.cpp:(.text._ZZN11HydroSystemI16ShocktubeProblemE13EnforceLimitsEdddddRN5amrex8MultiFabEENKUliiiiE_clEiiii[_ZZN11HydroSystemI16ShocktubeProblemE13EnforceLimitsEdddddRN5amrex8MultiFabEENKUliiiiE_clEiiii]+0x278): undefined reference to `EOSData::maxe'
```

I think the reason behind these errors is that I haven't initialized the microphysics eos and extern_parameters. In primordial chem test, I do `init_extern_parameters()` and `eos_init()` to initialize them. My question is, where should I carry out this initialization for other tests? In their respective `test.cpp` files? Or, in `EOS.hpp` itself?


BenWibking:
This is a linking issue. It looks like the necessary EOS .cpp files aren't being linked into the binary, so these functions don't exist when it tries to link it together.

BenWibking:
Hopefully it's just a missing *.cpp file. However, it *can* be something more exotic, in which case it is unfortunately necessary to understanding how the linker works in some detail.

If you are in the latter situation, I would recommend reading this: https://www.lurklurk.org/linkers/linkers.html.

psharda:
Ah, bug on my part. Fixed it. _Now_ I get the error I was talking about above, when I run the tests:

```
"HydroShocktubeCMA" start time: Jun 22 18:10 CEST
Output:
----------------------------------------------------------
AMReX (23.06-25-g239d4d899334) initialized
Using default ref_ratio = 2 at all levels
Successfully read inputs file ...

[Warning] [Performance] The maximum grid size (32) is too small for reasonable performance. It should be 128 (or greater) when running on GPUs, and 64 (or $

[Level 0 step 1] ADVANCE with time = 0 dt = 0.0004952075711
  what():  EOS: not initialized
```

So, I need `eos_init()`. But instead of doing this in every `test.cpp`, can we initialize microphysics EOS in `EOS.hpp` directly?

BenWibking:
> So, I need `eos_init()`. But instead of doing this in every `test.cpp`, can we initialize microphysics EOS in `EOS.hpp` directly?

This should probably be done in the constructor of `RadhydroSimulation`.

psharda:
> This should probably be done in the constructor of `RadhydroSimulation`.

can you be more specific :D

BenWibking:
Here: https://github.com/quokka-astro/quokka/blob/158a3e6c02b1f809429cf9759321554afa7a8d52/src/RadhydroSimulation.hpp#L125

BenWibking:
> Here:
> 
> https://github.com/quokka-astro/quokka/blob/158a3e6c02b1f809429cf9759321554afa7a8d52/src/RadhydroSimulation.hpp#L125

This code should actually be cleaned up a bit to avoid duplicating it twice for the cases with/without face-centered fields.

psharda:
Cool, thanks. We need to initialize EOS with some small value of density and temperature. Does quokka already store an epsilon density/temperature?

psharda:
Initialized EOS. Onto the next error:

```
  what():  gamma_const cannot be < 0
[gadi-login-02:1955070] *** Process received signal ***
[gadi-login-02:1955070] Signal: Aborted (6)
[gadi-login-02:1955070] Signal code:  (-6)
[gadi-login-02:1955070] [ 0] /lib64/libpthread.so.0(+0x12cf0)[0x7f185e3ebcf0]
[gadi-login-02:1955070] [ 1] /lib64/libc.so.6(gsignal+0x10f)[0x7f185e061aff]
[gadi-login-02:1955070] [ 2] /lib64/libc.so.6(abort+0x127)[0x7f185e034ea5]
[gadi-login-02:1955070] [ 3] /apps/gcc/11.1.0/lib64/libstdc++.so.6(+0xa951a)[0x7f185ec3d51a]
[gadi-login-02:1955070] [ 4] /apps/gcc/11.1.0/lib64/libstdc++.so.6(+0xb4dda)[0x7f185ec48dda]
[gadi-login-02:1955070] [ 5] /apps/gcc/11.1.0/lib64/libstdc++.so.6(+0xb4e45)[0x7f185ec48e45]
[gadi-login-02:1955070] [ 6] /apps/gcc/11.1.0/lib64/libstdc++.so.6(+0xb50d9)[0x7f185ec490d9]
[gadi-login-02:1955070] [ 7] /g/data1b/jh2/ps3459/quokka/build/src/HydroShocktubeCMA/test_hydro_shocktube_cma[0x40eff2]
[gadi-login-02:1955070] [ 8] /g/data1b/jh2/ps3459/quokka/build/src/HydroShocktubeCMA/test_hydro_shocktube_cma[0x43072f]
[gadi-login-02:1955070] [ 9] /g/data1b/jh2/ps3459/quokka/build/src/HydroShocktubeCMA/test_hydro_shocktube_cma[0x41e6e7]
[gadi-login-02:1955070] [10] /lib64/libc.so.6(__libc_start_main+0xe5)[0x7f185e04dd85]
[gadi-login-02:1955070] [11] /g/data1b/jh2/ps3459/quokka/build/src/HydroShocktubeCMA/test_hydro_shocktube_cma[0x42129e]
[gadi-login-02:1955070] *** End of error message ***
```

BenWibking:
> Cool, thanks. We need to initialize EOS with some small value of density and temperature. Does quokka already store an epsilon density/temperature?

No. That's interesting... what does it use it for?

psharda:
> > Cool, thanks. We need to initialize EOS with some small value of density and temperature. Does quokka already store an epsilon density/temperature?
> 
> No. That's interesting... what does it use it for?

To initialize EOS. If you recall, we also use it in primordial chem test in Quokka.

BenWibking:
> > > Cool, thanks. We need to initialize EOS with some small value of density and temperature. Does quokka already store an epsilon density/temperature?
> > 
> > 
> > No. That's interesting... what does it use it for?
> 
> To initialize EOS. If you recall, we also use it in primordial chem test in Quokka.

I mean: what does it actually do with those values?

psharda:
Sets a bunch of EOS min/max parameters, and sets `EOSData::initialized = true;`

eos_init is defined in Microphysics/interfaces/eos.H

BenWibking:
Ok. I think `small_dens` and `small_temp` should be ParmParse parameters.

src/EOS.hpp
```diff
@@ -84,8 +84,14 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeTgasFromEin
 	Tgas = chemstate.T;
 #else
 	if constexpr (gamma_ != 1.0) {
-		const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
-		Tgas = Eint / (rho * c_v);
+		// const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
+		// Tgas = Eint / (rho * c_v);
+		burn_t chemstate;
```
psharda: In line 89 above, instead of using the `burn_t` object, we should now be able to use the `chem_eos_t` struct that Max created in https://github.com/AMReX-Astro/Microphysics/pull/1199

But, I get an error if I use this struct in Quokka. I include eos.H above, which in turn includes eos_type.H which is where this new struct is defined. 
@BenWibking

psharda:
Okaay, Max's chem_eos_t works well.
BUT
I just realised that `gamma_law` uses `m_nucleon = 1 / Avagadro's number` while calculating Tgas, Eint or dEint / d Tgas. This makes sense for nuclear astrophysics stuff, but not for us. For example given Eint and rho, gamma_law does:

```
state.T = state.e * state.mu * m_nucleon * (eos_gamma - 1.0) / C::k_B;
```

whereas Quokka does:
```
Tgas = (Eint/rho) * mean_molecular_weight_ * (gamma_ - 1.0) / boltzmann_constant_;
```
, so there is an additional factor of m_nucleon in gamma_law

BenWibking:
> Okaay, Max's chem_eos_t works well. BUT I just realised that `gamma_law` uses `m_nucleon = 1 / Avagadro's number` while calculating Tgas, Eint or dEint / d Tgas. This makes sense for nuclear astrophysics stuff, but not for us. For example given Eint and rho, gamma_law does:
> 
> ```
> state.T = state.e * state.mu * m_nucleon * (eos_gamma - 1.0) / C::k_B;
> ```
> 
> whereas Quokka does:
> 
> ```
> Tgas = (Eint/rho) * mean_molecular_weight_ * (gamma_ - 1.0) / boltzmann_constant_;
> ```
> 
> , so there is an additional factor of m_nucleon in gamma_law

Is `state.mu` the dimensionless mean molecular weight? If so, then these expressions may be the same.

psharda:
TODO @BenWibking : insert `EOS_Traits` in `RadSuOlSon` test.

psharda:
Okay, all tests should pass now except for `RadhydroShockCGS`. The test seems to work fine but the relative tolerance at the end is ~0.007, which is higher than the accepted threshold of 0.005. Not sure why is the error higher, I am missing something subtle perhaps.

BenWibking:
> Okay, all tests should pass now except for `RadhydroShockCGS`. The test seems to work fine but the relative tolerance at the end is ~0.007, which is higher than the accepted threshold of 0.005. Not sure why is the error higher, I am missing something subtle perhaps.

That is... odd. If you force Microphysics to use the same numerical values of $k_B$ and $m_H$ that we had used previously, does it pass?

BenWibking:
Also, can you post the plot that test problem autogenerates?

psharda:
[Uploading radshock_cgs_temperature.pdf…]()

psharda:
[radshock_cgs_temperature.pdf](https://github.com/quokka-astro/quokka/files/11887887/radshock_cgs_temperature.pdf)

psharda:
image

psharda:
Microphysics uses `m_nucleon = 1 / n_Avagadro`, but this is not strictly equal to the mass of hydrogen atom, `m_H`. If the code unit I use is `m_nucleon` instead of `m_H`, this test passes.

Should I also make the code_units in other tests `m_nucleon` instead of `m_H`, since then the code unit will be _exactly_ cancelled out by `m_nucleon` in microphysics eos?

BenWibking:
> Microphysics uses `m_nucleon = 1 / n_Avagadro`, but this is not strictly equal to the mass of hydrogen atom, `m_H`. If the code unit I use is `m_nucleon` instead of `m_H`, this test passes.
> 
> Should I also make the code_units in other tests `m_nucleon` instead of `m_H`, since then the code unit will be _exactly_ cancelled out by `m_nucleon` in microphysics eos?

Ah, ok. This is because (prior to 2019), 1/N_A was *exactly* the atomic mass unit in grams. I think astrochemistry just pretends the atomic mass unit is identical to the mass of the hydrogen atom.

However, the 2019 redefinition of SI units makes the situation more complicated: https://en.wikipedia.org/wiki/Dalton_(unit)#2019_redefinition_of_the_SI_base_units.

psharda:
Woah, that is some serious change!

I just realised, if in Quokka we used `1/n_A` in place of `m_H` everywhere, we don't the extra code unit we defined in EOS_Traits. For each test, we can then just multiply the `mean_molecular_weight_` by `n_A` before we pass it to Microphysics, and it will work for all tests including dimensionless ones. This makes the required modifications much simpler!

psharda:
So, is it fine if we used `1/n_A` in place of hydrogen mass everywhere in Quokka?

BenWibking:
> So, is it fine if we used `1/n_A` in place of hydrogen mass everywhere in Quokka?

In some places we actually want $m_H$ and in some places we actually want 1 amu. For $\mu$, we want 1 amu, but I would not do a wholesale find and replace. It might not make a difference practically speaking, but it will make some things very confusing.

I don't see how this will fix the dimensionless tests without an extra unit definition somewhere.

psharda:
> > So, is it fine if we used `1/n_A` in place of hydrogen mass everywhere in Quokka?
> 
> In some places we actually want mH and in some places we actually want 1 amu. For μ, we want 1 amu, but I would not do a wholesale find and replace. It might not make a difference practically speaking, but it will make some things very confusing.
> 
> I don't see how this will fix the dimensionless tests without an extra unit definition somewhere.

Because the code unit for the dimensionless tests is also 1/n_A, so that it cancels out with m_nucleon in Microphysics. If you look at the RadPulse test and compare its code unit to, say, RadhydroShell test, they are basically the same, (1/n_A versus m_H).

psharda:
Waiting for https://github.com/AMReX-Astro/Microphysics/pull/1240 before I continue with this PR.

psharda:
@BenWibking RadPulse test will not pass because the relative error is now 0.0078, which is higher than the accepted tolerance of 0.005.

BenWibking:
> @BenWibking RadPulse test will not pass because the relative error is now 0.0078, which is higher than the accepted tolerance of 0.005.

Before the issue was `RadhydroShockCGS`: https://github.com/quokka-astro/quokka/pull/324#issuecomment-1610209230. Is it a different problem now?

BenWibking:
If RadhydroShockCGS is actually the issue, then I think it can be fixed by modifying these two lines:
https://github.com/quokka-astro/quokka/blob/1b572e70c1ab71d40e8c7f8650d9c90f4fcbe0df/src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp#L37
https://github.com/quokka-astro/quokka/blob/1b572e70c1ab71d40e8c7f8650d9c90f4fcbe0df/src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp#L68

src/EOS.hpp
```diff
@@ -35,6 +36,7 @@ template <typename problem_t> struct EOS_Traits {
 	static constexpr double cs_isothermal = NAN; // only used when gamma = 1
 	static constexpr double mean_molecular_weight = NAN;
 	static constexpr double boltzmann_constant = boltzmann_constant_cgs;
+	static constexpr double hydrogen_mass_code_units = hydrogen_mass_cgs;
```
BenWibking: We should not use this anymore. We always need the atomic mass unit in code units.

src/EOS.hpp
```diff
@@ -84,8 +87,15 @@ AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeTgasFromEin
...
 #else
 	if constexpr (gamma_ != 1.0) {
-		const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
-		Tgas = Eint / (rho * c_v);
+		// const amrex::Real c_v = boltzmann_constant_ / (mean_molecular_weight_ * (gamma_ - 1.0));
+		// Tgas = Eint / (rho * c_v);
+		chem_eos_t estate;
+		estate.rho = rho;
+		estate.e = Eint / rho;
+		estate.mu = mean_molecular_weight_ / hydrogen_mass_code_units_;
```
BenWibking: This should be the atomic mass unit instead.

src/RadMarshak/test_radiation_marshak.cpp
```diff
@@ -34,6 +34,7 @@ template <> struct quokka::EOS_Traits<SuOlsonProblem> {
 	static constexpr double mean_molecular_mass = 1.0;
 	static constexpr double boltzmann_constant = 1.0;
 	static constexpr double gamma = 5. / 3.;
+	static constexpr double hydrogen_mass_code_units = 1.0 / C::n_A;
```
BenWibking: Now that we've updated Microphysics to the post-2019 SI definitions, `1/N_A` should not be used anywhere.

src/RadStreaming/test_radiation_streaming.cpp
```diff
@@ -26,6 +26,7 @@ template <> struct quokka::EOS_Traits<StreamingProblem> {
 	static constexpr double mean_molecular_weight = 1.0;
 	static constexpr double boltzmann_constant = 1.0;
 	static constexpr double gamma = 5. / 3.;
+	static constexpr double hydrogen_mass_code_units = 1.0 / C::n_A;
```
BenWibking: `1/N_A` should not be used anywhere.

src/RadhydroShock/test_radhydro_shock.cpp
```diff
@@ -65,6 +65,7 @@ template <> struct quokka::EOS_Traits<ShockProblem> {
 	static constexpr double mean_molecular_weight = mu;
 	static constexpr double boltzmann_constant = k_B;
 	static constexpr double gamma = gamma_gas;
+	static constexpr double hydrogen_mass_code_units = 1.0 / C::n_A;
```
BenWibking: `1/N_A` should not be used.

src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp
```diff
@@ -68,6 +68,7 @@ template <> struct quokka::EOS_Traits<ShockProblem> {
 	static constexpr double mean_molecular_weight = m_H;
```
BenWibking: This should be set equal to the atomic mass unit.

src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp
```diff
@@ -68,6 +68,7 @@ template <> struct quokka::EOS_Traits<ShockProblem> {
...
 	static constexpr double boltzmann_constant = k_B;
 	static constexpr double gamma = gamma_gas;
+	static constexpr double hydrogen_mass_code_units = 1.0 / C::n_A;
```
BenWibking: This should be set equal to the atomic mass unit.

src/EOS.hpp
```diff
@@ -25,8 +26,8 @@
 
 namespace quokka
 {
-static constexpr double boltzmann_constant_cgs = 1.380658e-16; // cgs
-static constexpr double hydrogen_mass_cgs = 1.6726231e-24;     // cgs
+static constexpr double boltzmann_constant_cgs = C::k_B;     // cgs
+static constexpr double hydrogen_mass_cgs = C::m_p + C::m_e; // cgs
```
BenWibking: `hydrogen_mass_cgs` should not be defined. We don't need it for the EOS.

BenWibking (changes requested):

src/RadPulse/test_radiation_pulse.cpp
```diff
@@ -30,6 +31,7 @@ template <> struct quokka::EOS_Traits<PulseProblem> {
 	static constexpr double mean_molecular_weight = 1.0;
 	static constexpr double boltzmann_constant = (2. / 3.);
 	static constexpr double gamma = 5. / 3.;
+	static constexpr double hydrogen_mass_code_units = quokka::hydrogen_mass_cgs;
```
BenWibking: It should be possible to fix the test failure by changing to the atomic mass unit instead.

src/EOS.hpp
psharda: So, wherever we are using the hydrogen mass, we replace it by the atomic mass? @BenWibking

src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp
psharda: yes this is on my TODO list

psharda:
> If RadhydroShockCGS is actually the issue, then I think it can be fixed by modifying these two lines:
> 
> https://github.com/quokka-astro/quokka/blob/1b572e70c1ab71d40e8c7f8650d9c90f4fcbe0df/src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp#L37
> 
> 
> https://github.com/quokka-astro/quokka/blob/1b572e70c1ab71d40e8c7f8650d9c90f4fcbe0df/src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp#L68

So do we replace `m_H` by the atomic mass (`C::m_u`) instead?

src/EOS.hpp
BenWibking: Yes, I think so. There might be one or two exceptions, but probably easier to just change it everywhere and see if any of the tests fail.

BenWibking:
> > If RadhydroShockCGS is actually the issue, then I think it can be fixed by modifying these two lines:
> > https://github.com/quokka-astro/quokka/blob/1b572e70c1ab71d40e8c7f8650d9c90f4fcbe0df/src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp#L37
> > 
> > https://github.com/quokka-astro/quokka/blob/1b572e70c1ab71d40e8c7f8650d9c90f4fcbe0df/src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp#L68
> 
> So do we replace `m_H` by the atomic mass (`C::m_u`) instead?

Yes, that should fix it.

src/EOS.hpp
BenWibking: Actually yes, I think it should be changed everywhere. I thought the opacity was defined by H atom somewhere, but I couldn't find it.

psharda:
@BenWibking done. Only the RadhydroShockCGS test fails.

src/CloudyCooling.hpp
```diff
@@ -81,7 +81,7 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto cloudy_cooling_function(Real const
 {
 	// interpolate cooling rates from Cloudy tables
 	const Real rhoH = rho * cloudy_H_mass_fraction; // mass density of H species
-	const Real nH = rhoH / quokka::hydrogen_mass_cgs;
+	const Real nH = rhoH / C::m_u;
```
BenWibking: I forgot about this part of the code. When computing the hydrogen number density, we do need to use the hydrogen mass, not amu.

src/CloudyCooling.hpp
```diff
@@ -102,7 +102,7 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto cloudy_cooling_function(Real const
 
 	// compute electron density
 	// N.B. it is absolutely critical to include the metal contribution here!
-	double n_e = (rho / quokka::hydrogen_mass_cgs) * (1.0 - mu * (X + Y / 4. + Z / mean_metals_A)) / (mu - (electron_mass_cgs / quokka::hydrogen_mass_cgs));
+	double n_e = (rho / C::m_u) * (1.0 - mu * (X + Y / 4. + Z / mean_metals_A)) / (mu - (electron_mass_cgs / C::m_u));
```
BenWibking: This part should be hydrogen mass.

src/CloudyCooling.hpp
```diff
@@ -129,13 +129,13 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto ComputeEgasFromTgas(double rho, do
 {
 	// convert Egas (internal gas energy) to temperature
 	const Real rhoH = rho * cloudy_H_mass_fraction;
-	const Real nH = rhoH / quokka::hydrogen_mass_cgs;
+	const Real nH = rhoH / C::m_u;
```
BenWibking: This should actually be hydrogen mass.

src/CloudyCooling.hpp
BenWibking: Actually, this is fine as-is.

src/CloudyCooling.hpp
```diff
@@ -157,12 +157,12 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto ComputeTgasFromEgas(double rho, do
 
 	// solve for temperature given Eint (with fixed adiabatic index gamma)
 	const Real rhoH = rho * cloudy_H_mass_fraction;
-	const Real nH = rhoH / quokka::hydrogen_mass_cgs;
+	const Real nH = rhoH / C::m_u;
```
BenWibking: This should be hydrogen mass.

src/ODEIntegration/test_ode.cpp
```diff
@@ -15,13 +15,13 @@
 using amrex::Real;
 
 constexpr double Tgas0 = 6000.;				  // K
-constexpr double rho0 = 0.01 * quokka::hydrogen_mass_cgs; // g cm^-3
+constexpr double rho0 = 0.01 * C::m_u; // g cm^-3
```
BenWibking: This should be $m_H$.

src/ODEIntegration/test_ode.cpp
```diff
@@ -33,7 +33,7 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto cooling_function(Real const rho, R
 	// use fitting function from Koyama & Inutsuka (2002)
 	Real gamma_heat = 2.0e-26;
 	Real lambda_cool = gamma_heat * (1.0e7 * std::exp(-114800. / (T + 1000.)) + 14. * std::sqrt(T) * std::exp(-92. / T));
-	Real rho_over_mh = rho / quokka::hydrogen_mass_cgs;
+	Real rho_over_mh = rho / C::m_u;
```
BenWibking: this should be $m_H$.

src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp
```diff
@@ -25,7 +25,7 @@ struct ShockProblem {
 constexpr double a_rad = 7.5646e-15;		  // erg cm^-3 K^-4
 constexpr double c = 2.99792458e10;		  // cm s^-1
 constexpr double k_B = C::k_B;			  // erg K^-1
-constexpr double m_H = quokka::hydrogen_mass_cgs; // mass of hydrogen atom [g]
+constexpr double m_H = C::m_u; // atomic mass unit
```
BenWibking: For clarity, this should probably be removed, and replaced with `C::m_u` directly wherever it's used.

psharda:
@BenWibking comments resolved. But, the radhydroshockCGS test will still fail.

src/ShockCloud/cloud.cpp
```diff
@@ -42,6 +42,7 @@ template <> struct quokka::EOS_Traits<ShockCloud> {
 	static constexpr double gamma = 5. / 3.; // default value
 	static constexpr double mean_molecular_weight = NAN;
 	static constexpr double boltzmann_constant = quokka::boltzmann_constant_cgs;
+	static constexpr double hydrogen_mass_code_units = quokka::hydrogen_mass_cgs;
```
BenWibking: Convert to amu.

src/SphericalCollapse/spherical_collapse.cpp
```diff
@@ -29,7 +29,8 @@ struct CollapseProblem {
 template <> struct quokka::EOS_Traits<CollapseProblem> {
 	static constexpr double gamma = 5. / 3.;
 	static constexpr double mean_molecular_weight = NAN;
-	static constexpr double boltzmann_constant = quokka::boltzmann_constant_cgs;
+	static constexpr double boltzmann_constant = C::k_B;
+	static constexpr double hydrogen_mass_code_units = C::m_u;
```
BenWibking: Rename.

src/CloudyCooling.hpp
```diff
@@ -81,7 +81,7 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto cloudy_cooling_function(Real const
 {
 	// interpolate cooling rates from Cloudy tables
 	const Real rhoH = rho * cloudy_H_mass_fraction; // mass density of H species
-	const Real nH = rhoH / quokka::hydrogen_mass_cgs;
+	const Real nH = rhoH / (C::m_p + C::m_e);
```
BenWibking, line 84: Can you define a variable `quokka::m_h` for this quantity ($m_p + m_e$)? I think that would make this more readable.

BenWibking (dismissed):
Minor changes.

For the RadhydroShockCGS problem, as long as the output looks ok, I think it's ok to increase the tolerance slightly so it passes.

src/CloudyCooling.hpp
psharda: Nooo lol
I will do that in a separate PR if you really want it. Because we should then do it everywhere.

psharda:
[radshock_cgs_temperature.pdf](https://github.com/quokka-astro/quokka/files/12148434/radshock_cgs_temperature.pdf)
This is how the plot looks for the radhydroshockCGS test. Looks slightly worse than before? The relative error norm is 0.0136 while the tolerance is 0.005

src/CloudyCooling.hpp
BenWibking: How many times is it used? I don't see how it's that big of a change...

BenWibking:
> [radshock_cgs_temperature.pdf](https://github.com/quokka-astro/quokka/files/12148434/radshock_cgs_temperature.pdf) This is how the plot looks for the radhydroshockCGS test. Looks slightly worse than before? The relative error norm is 0.0136 while the tolerance is 0.005

Well, that's annoying. It would be good to figure out why the error is now 3x worse than before...

Can you try changing the heat capacity `c_v` to use `m_H` instead of `C::m_u`? My guess is that the change to the heat capacity is causing the biggest difference.

psharda:
> > [radshock_cgs_temperature.pdf](https://github.com/quokka-astro/quokka/files/12148434/radshock_cgs_temperature.pdf) This is how the plot looks for the radhydroshockCGS test. Looks slightly worse than before? The relative error norm is 0.0136 while the tolerance is 0.005
> 
> Well, that's annoying. It would be good to figure out why the error is now 3x worse than before...
> 
> Can you try changing the heat capacity `c_v` to use `m_H` instead of `C::m_u`? My guess is that the change to the heat capacity is causing the biggest difference.

That reduces the tolerance to 0.0074.

BenWibking:
> > > [radshock_cgs_temperature.pdf](https://github.com/quokka-astro/quokka/files/12148434/radshock_cgs_temperature.pdf) This is how the plot looks for the radhydroshockCGS test. Looks slightly worse than before? The relative error norm is 0.0136 while the tolerance is 0.005
> > 
> > 
> > Well, that's annoying. It would be good to figure out why the error is now 3x worse than before...
> > Can you try changing the heat capacity `c_v` to use `m_H` instead of `C::m_u`? My guess is that the change to the heat capacity is causing the biggest difference.
> 
> That reduces the tolerance to 0.0074.

Ok, that's good. Can you also try reverting `mean_molecular_weight` to `m_H`, but keeping `mass_code_units` the same, and see if it passes?

psharda:
Yes, then it will pass. So I guess we leave the mean molecular weight for this test in terms of m_H.

BenWibking:
> Yes, then it will pass. So I guess we leave the mean molecular weight for this test in terms of m_H.

That sounds good to me. If the GPU tests pass, I'll approve and merge.

psharda:
I still need to modify the other two EOS functions in EOS.hpp to use Microphysics. I will do that now.

psharda:
This PR in microphysics will be helpful to get dedT from Microphysics in Quokka's EOS.hpp: https://github.com/AMReX-Astro/Microphysics/pull/1298

BenWibking:
RadPulse and RadhydroShock are failing. My guess is that $de/dT$ might be wrong for problems that use dimensionless units.

psharda:
Fixed.

src/CloudyCooling.hpp
psharda: We have also used it in a few other test problems, and I want to make sure we can keep track of where we use m_H versus amu. It is now much cleaner to do any changes needed for it in a separate PR.

BenWibking: Ok, that's fine.

BenWibking:
Can you add a new *.rst file in `docs/` that explains what `mass_code_units` is?

Since the rest of us were confused about this on the call, I think new users will be very confused without some documentation.

psharda:
We don't need `mass_code_units` anymore (I thought that was clear from the discussion?). Anyway I will push a commit soon.

psharda:
this is ready @BenWibking

BenWibking:
I'll take a look this afternoon.

src/Cooling/test_cooling.cpp
```diff
@@ -30,13 +30,14 @@ using amrex::Real;
...
+constexpr double m_H = C::m_u;
 constexpr double seconds_in_year = 3.154e7;
 
 template <> struct quokka::EOS_Traits<CoolingTest> {
 	static constexpr double gamma = 5. / 3.; // default value
-	static constexpr double mean_molecular_weight = quokka::hydrogen_mass_cgs;
-	static constexpr double boltzmann_constant = quokka::boltzmann_constant_cgs;
+	static constexpr double mean_molecular_weight = C::m_u;
+	static constexpr double boltzmann_constant = C::k_B;
+	static constexpr double mass_code_units = C::m_u;
```
BenWibking: We don't need this anymore, right?

BenWibking:
Looks good, except for the last remaining `mass_code_units` that can be deleted.

Automerge enabled.

BenWibking:
This does seem a bit slower than usual.

```
/usr/bin/cmake -E chdir . ctest -E MatterEnergyExchange* -T Test --output-on-failure
   Site: avatargpue.anu.edu.au
   Build name: Linux-g++
Create new tag: 20230725-1425 - Experimental
Test project /avatar/bwibking/azure-pipelines-agent/_work/4/s/build
      Start  1: HydroShocktubeCMA
 1/28 Test  #1: HydroShocktubeCMA ................   Passed  915.15 sec
      Start  2: PrimordialChem
 2/28 Test  #2: PrimordialChem ...................   Passed    2.53 sec
      Start  3: ScalarAdvection
 3/28 Test  #3: ScalarAdvection ..................   Passed    8.05 sec
      Start  4: Advection2D
 4/28 Test  #4: Advection2D ......................   Passed   46.84 sec
      Start  5: ScalarAdvectionSemiEllipse
 5/28 Test  #5: ScalarAdvectionSemiEllipse .......   Passed    7.94 sec
      Start  6: HydroBlast3D
 6/28 Test  #6: HydroBlast3D .....................   Passed  736.15 sec
      Start  7: HydroContact
 7/28 Test  #7: HydroContact .....................   Passed    2.99 sec
      Start  8: HydroLeblanc
 8/28 Test  #8: HydroLeblanc .....................   Passed   59.05 sec
      Start  9: HydroShocktube
 9/28 Test  #9: HydroShocktube ...................   Passed   86.79 sec
      Start 10: ShuOsher
10/28 Test #10: ShuOsher .........................   Passed   30.77 sec
      Start 11: HydroSlowMovingShock
11/28 Test #11: HydroSlowMovingShock .............   Passed    8.35 sec
      Start 12: HydroVacuum
12/28 Test #12: HydroVacuum ......................   Passed    2.52 sec
      Start 13: HydroWave
13/28 Test #13: HydroWave ........................   Passed    3.26 sec
      Start 14: HydroQuirk
14/28 Test #14: HydroQuirk .......................   Passed    3.49 sec
      Start 15: HydroHighMach
15/28 Test #15: HydroHighMach ....................   Passed    2.60 sec
      Start 16: RadForce
16/28 Test #16: RadForce .........................   Passed   57.97 sec
      Start 17: MarshakWave
17/28 Test #17: MarshakWave ......................   Passed    7.43 sec
      Start 18: MarshakWaveCGS
18/28 Test #18: MarshakWaveCGS ...................   Passed    3.74 sec
      Start 19: RadiationPulse
19/28 Test #19: RadiationPulse ...................   Passed   38.29 sec
      Start 20: RadiationStreaming
20/28 Test #20: RadiationStreaming ...............   Passed    4.80 sec
      Start 21: SuOlsonTest
21/28 Test #21: SuOlsonTest ......................   Passed   12.78 sec
      Start 22: RadiationPressureTube
22/28 Test #22: RadiationPressureTube ............   Passed   11.76 sec
      Start 23: RadhydroShock
23/28 Test #23: RadhydroShock ....................   Passed  119.09 sec
      Start 24: RadhydroShockCGS
24/28 Test #24: RadhydroShockCGS .................   Passed  128.29 sec
      Start 25: ODEIntegration
25/28 Test #25: ODEIntegration ...................   Passed    2.40 sec
      Start 26: PassiveScalar
26/28 Test #26: PassiveScalar ....................   Passed   47.20 sec
      Start 27: FCQuantities
27/28 Test #27: FCQuantities .....................   Passed    2.46 sec
      Start 28: SphericalCollapse
28/28 Test #28: SphericalCollapse ................   Passed  199.84 sec

100% tests passed, 0 tests failed out of 28

Total Test time (real) = 2553.28 sec
```

psharda:
> This does seem a bit slower than usual.
> 
> ```
> /usr/bin/cmake -E chdir . ctest -E MatterEnergyExchange* -T Test --output-on-failure
>    Site: avatargpue.anu.edu.au
>    Build name: Linux-g++
> Create new tag: 20230725-1425 - Experimental
> Test project /avatar/bwibking/azure-pipelines-agent/_work/4/s/build
>       Start  1: HydroShocktubeCMA
>  1/28 Test  #1: HydroShocktubeCMA ................   Passed  915.15 sec
>       Start  2: PrimordialChem
>  2/28 Test  #2: PrimordialChem ...................   Passed    2.53 sec
>       Start  3: ScalarAdvection
>  3/28 Test  #3: ScalarAdvection ..................   Passed    8.05 sec
>       Start  4: Advection2D
>  4/28 Test  #4: Advection2D ......................   Passed   46.84 sec
>       Start  5: ScalarAdvectionSemiEllipse
>  5/28 Test  #5: ScalarAdvectionSemiEllipse .......   Passed    7.94 sec
>       Start  6: HydroBlast3D
>  6/28 Test  #6: HydroBlast3D .....................   Passed  736.15 sec
>       Start  7: HydroContact
>  7/28 Test  #7: HydroContact .....................   Passed    2.99 sec
>       Start  8: HydroLeblanc
>  8/28 Test  #8: HydroLeblanc .....................   Passed   59.05 sec
>       Start  9: HydroShocktube
>  9/28 Test  #9: HydroShocktube ...................   Passed   86.79 sec
>       Start 10: ShuOsher
> 10/28 Test #10: ShuOsher .........................   Passed   30.77 sec
>       Start 11: HydroSlowMovingShock
> 11/28 Test #11: HydroSlowMovingShock .............   Passed    8.35 sec
>       Start 12: HydroVacuum
> 12/28 Test #12: HydroVacuum ......................   Passed    2.52 sec
>       Start 13: HydroWave
> 13/28 Test #13: HydroWave ........................   Passed    3.26 sec
>       Start 14: HydroQuirk
> 14/28 Test #14: HydroQuirk .......................   Passed    3.49 sec
>       Start 15: HydroHighMach
> 15/28 Test #15: HydroHighMach ....................   Passed    2.60 sec
>       Start 16: RadForce
> 16/28 Test #16: RadForce .........................   Passed   57.97 sec
>       Start 17: MarshakWave
> 17/28 Test #17: MarshakWave ......................   Passed    7.43 sec
>       Start 18: MarshakWaveCGS
> 18/28 Test #18: MarshakWaveCGS ...................   Passed    3.74 sec
>       Start 19: RadiationPulse
> 19/28 Test #19: RadiationPulse ...................   Passed   38.29 sec
>       Start 20: RadiationStreaming
> 20/28 Test #20: RadiationStreaming ...............   Passed    4.80 sec
>       Start 21: SuOlsonTest
> 21/28 Test #21: SuOlsonTest ......................   Passed   12.78 sec
>       Start 22: RadiationPressureTube
> 22/28 Test #22: RadiationPressureTube ............   Passed   11.76 sec
>       Start 23: RadhydroShock
> 23/28 Test #23: RadhydroShock ....................   Passed  119.09 sec
>       Start 24: RadhydroShockCGS
> 24/28 Test #24: RadhydroShockCGS .................   Passed  128.29 sec
>       Start 25: ODEIntegration
> 25/28 Test #25: ODEIntegration ...................   Passed    2.40 sec
>       Start 26: PassiveScalar
> 26/28 Test #26: PassiveScalar ....................   Passed   47.20 sec
>       Start 27: FCQuantities
> 27/28 Test #27: FCQuantities .....................   Passed    2.46 sec
>       Start 28: SphericalCollapse
> 28/28 Test #28: SphericalCollapse ................   Passed  199.84 sec
> 
> 100% tests passed, 0 tests failed out of 28
> 
> Total Test time (real) = 2553.28 sec
> ```

Yes, I found that too. I actually commented about it earlier but then edited my comment thinking it was probably a Gadi thingy, because on avatar it doesn't take any longer than before.

psharda:
Ready to be merged @BenWibking

BenWibking (approved):

PR #335, opened 2023-07-26
Compute pressure from microphysics eos

psharda:
This is my idea of how we want to compute pressure from microphysics. But, it might be better to move this function to `EOS.hpp` so all the EOS functions are together?

Anyway, first, need this to be approved: https://github.com/AMReX-Astro/Microphysics/pull/1301

### DIFF
--- a/src/hydro_system.hpp
+++ b/src/hydro_system.hpp
@@ -294,11 +294,13 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto HydroSystem<problem_t>::ComputePressure
 	const auto vz = pz / rho;
 	const auto kinetic_energy = 0.5 * rho * (vx * vx + vy * vy + vz * vz);
 	const auto thermal_energy = E - kinetic_energy;
-	amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(cons, i, j, k);
+	amrex::Real P = NAN;
+
 	if constexpr (is_eos_isothermal()) {
-		amrex::Real P = rho * cs_iso_ * cs_iso_;
+		P = rho * cs_iso_ * cs_iso_;
 	} else {
-		amrex::Real P = quokka::EOS<problem_t>::ComputePressure(rho, thermal_energy, massScalars);
+		amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(cons, i, j, k);
+		P = quokka::EOS<problem_t>::ComputePressure(rho, thermal_energy, massScalars);
 	}
 	return P;
 }
--- a/src/hydro_system.hpp
+++ b/src/hydro_system.hpp
@@ -213,7 +213,8 @@ template <typename problem_t> auto HydroSystem<problem_t>::maxSignalSpeedLocal(a
 					} else {
 						const auto Etot = cons[bx](i, j, k, HydroSystem<problem_t>::energy_index);
 						const auto Eint = Etot - kinetic_energy;
-						const auto P = ComputePressure(cons[bx], i, j, k); // const auto P =  Eint * (HydroSystem<problem_t>::gamma_ - 1.0);
+						const auto P =
+						    ComputePressure(cons[bx], i, j, k); // const auto P =  Eint * (HydroSystem<problem_t>::gamma_ - 1.0);
 						cs = std::sqrt(HydroSystem<problem_t>::gamma_ * P / rho);
 					}
 					return {cs + abs_vel};
--- a/src/hydro_system.hpp
+++ b/src/hydro_system.hpp
@@ -527,7 +527,8 @@ void HydroSystem<problem_t>::ComputeFlatteningCoefficients(amrex::MultiFab const
 
 		// Z is a measure of shock strength (Eq. 76 of Miller & Colella 2002)
 		amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(primVar, i, j, k);
-		const double K_S = std::pow(quokka::EOS<problem_t>::ComputeSoundSpeed(primVar(i, j, k, primDensity_index), P, massScalars), 2) * primVar(i, j, k, primDensity_index);
+		const double K_S = std::pow(quokka::EOS<problem_t>::ComputeSoundSpeed(primVar(i, j, k, primDensity_index), P, massScalars), 2) *
+				   primVar(i, j, k, primDensity_index);
 		const double Z = std::abs(Pplus1 - Pminus1) / K_S;
 
 		// check for converging flow along the normal direction DIR (Eq. 77)
--- a/src/hydro_system.hpp
+++ b/src/hydro_system.hpp
@@ -213,8 +213,7 @@ template <typename problem_t> auto HydroSystem<problem_t>::maxSignalSpeedLocal(a
 					} else {
 						const auto Etot = cons[bx](i, j, k, HydroSystem<problem_t>::energy_index);
 						const auto Eint = Etot - kinetic_energy;
-						const auto P =
-						    ComputePressure(cons[bx], i, j, k); // const auto P =  Eint * (HydroSystem<problem_t>::gamma_ - 1.0);
+						const auto P = ComputePressure(cons[bx], i, j, k);
 						cs = std::sqrt(HydroSystem<problem_t>::gamma_ * P / rho);
 					}
 					return {cs + abs_vel};
@@ -247,7 +246,7 @@ void HydroSystem<problem_t>::ComputeMaxSignalSpeed(amrex::Array4<const amrex::Re
 			AMREX_ASSERT(!std::isnan(E));
 			const auto kinetic_energy = 0.5 * rho * (vx * vx + vy * vy + vz * vz);
 			const auto thermal_energy = E - kinetic_energy;
-			const auto P = thermal_energy * (HydroSystem<problem_t>::gamma_ - 1.0);
+			const auto P = ComputePressure(cons, i, j, k);
 			cs = std::sqrt(HydroSystem<problem_t>::gamma_ * P / rho);
 		}
 		AMREX_ASSERT(cs > 0.);
@@ -275,7 +274,7 @@ template <typename problem_t> auto HydroSystem<problem_t>::CheckStatesValid(amre
 					const auto vz = pz / rho;
 					const auto kinetic_energy = 0.5 * rho * (vx * vx + vy * vy + vz * vz);
 					const auto thermal_energy = E - kinetic_energy;
-					const auto P = thermal_energy * (HydroSystem<problem_t>::gamma_ - 1.0);
+					const auto P = ComputePressure(cons[bx], i, j, k);
 
 					bool negativeDensity = (rho <= 0.);
 					bool negativePressure = (P <= 0.);
@@ -310,7 +309,6 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto HydroSystem<problem_t>::ComputePressure
 	const auto vz = pz / rho;
 	const auto kinetic_energy = 0.5 * rho * (vx * vx + vy * vy + vz * vz);
 	const auto thermal_energy = E - kinetic_energy;
-	// const auto P = thermal_energy * (HydroSystem<problem_t>::gamma_ - 1.0);
 
 	amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(cons, i, j, k);
 	amrex::Real P = quokka::EOS<problem_t>::ComputePressure(rho, thermal_energy, massScalars);
--- a/src/hydro_system.hpp
+++ b/src/hydro_system.hpp
@@ -493,20 +493,20 @@ void HydroSystem<problem_t>::ComputeFlatteningCoefficients(amrex::MultiFab const
 
 		if constexpr (reconstruct_eint) {
 			// compute (rho e) (gamma - 1)
-			amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(primVar, i + 2, j, k);
+			amrex::GpuArray<Real, nmscalars_> massScalars_plus2 = RadSystem<problem_t>::ComputeMassScalars(primVar, i + 2, j, k);
 			Pplus2 = quokka::EOS<problem_t>::ComputePressure(primVar(i + 2, j, k, primDensity_index),
-									 primVar(i + 2, j, k, primDensity_index) * Pplus2, massScalars);
-			massScalars = RadSystem<problem_t>::ComputeMassScalars(primVar, i + 1, j, k);
+									 primVar(i + 2, j, k, primDensity_index) * Pplus2, massScalars_plus2);
+			amrex::GpuArray<Real, nmscalars_> massScalars_plus1 = RadSystem<problem_t>::ComputeMassScalars(primVar, i + 1, j, k);
 			Pplus1 = quokka::EOS<problem_t>::ComputePressure(primVar(i + 1, j, k, primDensity_index),
-									 primVar(i + 1, j, k, primDensity_index) * Pplus1, massScalars);
-			massScalars = RadSystem<problem_t>::ComputeMassScalars(primVar, i, j, k);
+									 primVar(i + 1, j, k, primDensity_index) * Pplus1, massScalars_plus1);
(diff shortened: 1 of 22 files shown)


BenWibking:
I think it would make sense to move the new code to EOS.hpp, but make it a function that only needs scalar arguments (e.g., `amrex::Real rho, amrex::Real Eint`). Then the existing HydroSystem::ComputePressure just calls that function for a given cell.

psharda:
But, for primordial chem, we will need the mass scalars too

BenWibking:
Yeah, forgot about that for a sec :) The design you have now makes sense to me.

psharda:
Aha, encountered the first issue. Take the example of HydroHighMach test. For this test, the mean molecular weight is NAN. So, the gamma_law EOS in microphysics returns a NAN pressure, because it calculates the pressure using the temperature. The dependence on the mean molecular weight cancels out in the end (`state.mu / state.mu`), but because `state.mu = NAN`, `state.mu/state.mu = NAN`.

The fix I can think of is to modify microphysics to use the internal energy directly in place of temperature to calculate the pressure.

psharda:
Aah, can't do that because gamma_law uses pressure to update the internal energy.

BenWibking:
> Aha, encountered the first issue. Take the example of HydroHighMach test. For this test, the mean molecular weight is NAN. So, the gamma_law EOS in microphysics returns a NAN pressure, because it calculates the pressure using the temperature. The dependence on the mean molecular weight cancels out in the end (`state.mu / state.mu`), but because `state.mu = NAN`, `state.mu/state.mu = NAN`.
> 
> The fix I can think of is to modify microphysics to use the internal energy directly in place of temperature to calculate the pressure.

You could do that. Another option is to just set the mean molecular weight to some arbitrary but non-zero and non-NAN value.

psharda:
> > Aha, encountered the first issue. Take the example of HydroHighMach test. For this test, the mean molecular weight is NAN. So, the gamma_law EOS in microphysics returns a NAN pressure, because it calculates the pressure using the temperature. The dependence on the mean molecular weight cancels out in the end (`state.mu / state.mu`), but because `state.mu = NAN`, `state.mu/state.mu = NAN`.
> > The fix I can think of is to modify microphysics to use the internal energy directly in place of temperature to calculate the pressure.
> 
> You could do that. Another option is to just set the mean molecular weight to some arbitrary but non-zero and non-NAN value.

Oh, if that's possible, then that's the easiest thing to do without changing stuff in microphysics. Okay, I will then set the mean molecular weight to 1.0 for this test (and tests like these that might fail too)

BenWibking:
> > > Aha, encountered the first issue. Take the example of HydroHighMach test. For this test, the mean molecular weight is NAN. So, the gamma_law EOS in microphysics returns a NAN pressure, because it calculates the pressure using the temperature. The dependence on the mean molecular weight cancels out in the end (`state.mu / state.mu`), but because `state.mu = NAN`, `state.mu/state.mu = NAN`.
> > > The fix I can think of is to modify microphysics to use the internal energy directly in place of temperature to calculate the pressure.
> > 
> > 
> > You could do that. Another option is to just set the mean molecular weight to some arbitrary but non-zero and non-NAN value.
> 
> Oh, if that's possible, then that's the easiest thing to do without changing stuff in microphysics. Okay, I will then set the mean molecular weight to 1.0 for this test (and tests like these that might fail too)

Inelegant, but probably the easiest solution.

src/hydro_system.hpp
```diff
@@ -204,7 +213,8 @@ template <typename problem_t> auto HydroSystem<problem_t>::maxSignalSpeedLocal(a
 					} else {
 						const auto Etot = cons[bx](i, j, k, HydroSystem<problem_t>::energy_index);
 						const auto Eint = Etot - kinetic_energy;
-						const auto P = Eint * (HydroSystem<problem_t>::gamma_ - 1.0);
+						const auto P =
+						    ComputePressure(cons[bx], i, j, k); // const auto P =  Eint * (HydroSystem<problem_t>::gamma_ - 1.0);
```
psharda: @BenWibking my modification in this line returns a NAN for the pressure. It could be that the `cons` array the ComputePressure function is expecting is not the same as the `cons[bx]` array being used here.

psharda: Do you know why this would give a NAN for pressure

BenWibking: I don't see an obvious error. Might have to just step through it in a debugger.

psharda: Ugh, it was the mean molecular weight being NAN for the test where this error was coming from.

src/hydro_system.hpp
```diff
@@ -231,10 +231,8 @@ void HydroSystem<problem_t>::ComputeMaxSignalSpeed(amrex::Array4<const amrex::Re
 		if constexpr (is_eos_isothermal()) {
 			cs = cs_iso_;
 		} else {
-			const auto E = cons(i, j, k, energy_index); // *total* gas energy per unit volume
-			AMREX_ASSERT(!std::isnan(E));
-			const auto kinetic_energy = 0.5 * rho * (vx * vx + vy * vy + vz * vz);
 			const auto P = ComputePressure(cons, i, j, k);
+			AMREX_ASSERT(!std::isnan(P));
 			cs = std::sqrt(HydroSystem<problem_t>::gamma_ * P / rho);
```
BenWibking: You should call the sound speed function in the EOS here instead.

psharda: Do you mean we should create a new function to find the sound speed, in EOS.hpp?

BenWibking: Yes, we need to do that.

src/hydro_system.hpp
```diff
@@ -499,7 +526,9 @@ void HydroSystem<problem_t>::ComputeFlatteningCoefficients(amrex::MultiFab const
 		const double chi_min = std::max(0., std::min(1., (beta_max - beta) / (beta_max - beta_min)));
 
 		// Z is a measure of shock strength (Eq. 76 of Miller & Colella 2002)
-		const double K_S = gamma_ * P; // equal to \rho c_s^2
+		amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(primVar, i, j, k);
+		const double K_S = std::pow(quokka::EOS<problem_t>::ComputeSoundSpeed(primVar(i, j, k, primDensity_index), P, massScalars), 2) *
+				   primVar(i, j, k, primDensity_index);
```
psharda, line 536: @BenWibking in this line, should the second argument to `ComputeSoundSpeed` be just `P` or `P*primVar(i,j,k,primDensity_index)`? We use the latter a few lines above in the same function.

BenWibking: I should have rewritten this so it was less confusing. In the earlier code, `P` is actually specific internal energy $e$. Whenever `P` is actually pressure, it should not be multiplied by density.

psharda: Hmm, okay. So, in this function, we first use `P` in line 489: `amrex::Real P = primVar(i, j, k, pressure_index);`, and at this time, `P` is specific Eint. Now, if `reconstruct_eint` is true, `P` is the pressure, but if its not true, it is still the specific Eint (for gamma > 1)? So by the time we come to this line (529), we don't know a priori if `P` is specific Eint or pressure?

BenWibking: By the time we come to this line, `P` is always pressure.

psharda:
RadhydroShock fails but RadhydroShockCGS passes. Some units issue?

src/hydro_system.hpp
psharda: ah okay. so then what i have is correct.

src/hydro_system.hpp
```diff
@@ -826,8 +858,10 @@ void HydroSystem<problem_t>::ComputeFluxes(amrex::MultiFab &x1Flux_mf, amrex::Mu
 				// (pressure_index is actually eint)
 				const double eint_L = x1LeftState(i, j, k, pressure_index);
 				const double eint_R = x1RightState(i, j, k, pressure_index);
-				P_L = rho_L * eint_L * (gamma_ - 1.0);
-				P_R = rho_R * eint_R * (gamma_ - 1.0);
+				amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(x1LeftState, i, j, k);
+				P_L = quokka::EOS<problem_t>::ComputePressure(rho_L, eint_L * rho_L, massScalars);
+				massScalars = RadSystem<problem_t>::ComputeMassScalars(x1RightState, i, j, k);
+				P_R = quokka::EOS<problem_t>::ComputePressure(rho_R, eint_R * rho_R, massScalars);
```
psharda: @BenWibking the HydroContact test is failing because of this modification. The pressure returned by microphysics is initially exactly unity, but then as time goes on it starts slowly deviating away from being exactly 1. by the end of the simulation, it is quite far from 1 (e.g., pressures at the end step are 0.99000 or 1.002, etc.). This test has a zero tolerance because the solution should be exact. 

I wonder if it changing these lines but not updating the Riemann solver makes the input rho and eint deviate from their exact values, which causes microphysics to return a non-exact pressure value? This test passes if I keep all other modifications but revert back to the original code for here.

BenWibking: No, it should give exactly the same result here. This might be due to a subtle roundoff error.

BenWibking: Can you run the first timestep through the debugger and output `P_R - P_L`? It should be *exactly* zero. If it's some tiny (but non-zero) number, it's a numerical roundoff issue that should be fixed.

psharda: Yes, `P_R - P_L` is exactly 0 at the first timestep, but slowly starts deviating from 0 as time progresses.

psharda:
Okay, fixed the bug that was causing RadhydroShock to fail.

src/hydro_system.hpp
psharda: Okay, I have narrowed down the problem to Microphysics. If I use the _exact_ same equation in Microphysics as the one in Quokka, I get the exact solution right and the test passes.

To calculate pressure, Microphysics first finds the gas temperature (from eint), and uses the temperature to find pressure. There are a few cancellations that occur in this indirect way of calculating pressure from eint. My guess is that these cancellations are introducing a rounding off error. But, this error is not there in the beginning (at the first step), and starts to emerge over the course of the simulation.

psharda: FOUND IT! You will not believe what it was. https://github.com/AMReX-Astro/Microphysics/pull/1302#discussion_r1277577098

psharda:
@BenWibking some tests fail when `-DENABLE_ASAN=ON`, and the failure is caused by a division by zero error in `ComputePressure` in EOS.hpp, on the line where we do `state.e = Eint / rho`. This error does not appear when the address sanitizer is not enabled. How do I debug this error?

BenWibking:
> @BenWibking some tests fail when `-DENABLE_ASAN=ON`, and the failure is caused by a division by zero error in `ComputePressure` in EOS.hpp, on the line where we do `state.e = Eint / rho`. This error does not appear when the address sanitizer is not enabled. How do I debug this error?

You can add an `if` statement that checks whether `rho == 0`. If so, you can then trigger a breakpoint that you can then examine in gdb: https://stackoverflow.com/a/4326474.

psharda:
@BenWibking the remaining test (debug arm64 gcc) fails because `RadForce` fails. This test is likely failing because it uses isothermal gamma, and the function `ComputePressure` (which is what hydro_system.hpp uses everywhere) returns a NAN pressure for isothermal gamma.

What should be the pressure when gamma is isothermal? Its not zero right? Do you find it from the (finite, non-zero) isothermal sound speed, like here:  https://github.com/quokka-astro/quokka/blob/aa288b57d2dda961815af7f986ebbb8fc30b9234/src/hydro_system.hpp#L482

BenWibking:
> @BenWibking the remaining test (debug arm64 gcc) fails because `RadForce` fails. This test is likely failing because it uses isothermal gamma, and the function `ComputePressure` (which is what hydro_system.hpp uses everywhere) returns a NAN pressure for isothermal gamma.
> 
> What should be the pressure when gamma is isothermal? Its not zero right? Do you find it from the (finite, non-zero) isothermal sound speed, like here:
> 
> https://github.com/quokka-astro/quokka/blob/aa288b57d2dda961815af7f986ebbb8fc30b9234/src/hydro_system.hpp#L482

Yes, it's $\rho c_s^2$.

psharda:
all tests should pass now

BenWibking:
It looks like the SphericalCollapse test failed.

psharda:
all done @BenWibking

src/hydro_system.hpp
```diff
@@ -842,11 +879,13 @@ void HydroSystem<problem_t>::ComputeFluxes(amrex::MultiFab &x1Flux_mf, amrex::Mu
 				Eint_R = x1RightState(i, j, k, primEint_index);
 			}
 
-			cs_L = std::sqrt(gamma_ * P_L / rho_L);
-			cs_R = std::sqrt(gamma_ * P_R / rho_R);
+			amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(x1LeftState, i, j, k);
```
BenWibking: Can you rename this to `massScalars_L`?

src/hydro_system.hpp
```diff
@@ -842,11 +879,13 @@ void HydroSystem<problem_t>::ComputeFluxes(amrex::MultiFab &x1Flux_mf, amrex::Mu
...
+			cs_L = quokka::EOS<problem_t>::ComputeSoundSpeed(rho_L, P_L, massScalars);
+			E_L = quokka::EOS<problem_t>::ComputeEintFromPres(rho_L, P_L, massScalars) + ke_L;
 
-			E_L = P_L / (gamma_ - 1.0) + ke_L;
-			E_R = P_R / (gamma_ - 1.0) + ke_R;
+			massScalars = RadSystem<problem_t>::ComputeMassScalars(x1RightState, i, j, k);
```
BenWibking: Can you rename this to `massScalars_R`?

src/hydro_system.hpp
```diff
@@ -826,8 +861,10 @@ void HydroSystem<problem_t>::ComputeFluxes(amrex::MultiFab &x1Flux_mf, amrex::Mu
 				// (pressure_index is actually eint)
 				const double eint_L = x1LeftState(i, j, k, pressure_index);
 				const double eint_R = x1RightState(i, j, k, pressure_index);
-				P_L = rho_L * eint_L * (gamma_ - 1.0);
-				P_R = rho_R * eint_R * (gamma_ - 1.0);
+				amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(x1LeftState, i, j, k);
```
BenWibking: Rename to `massScalars_L`.

src/hydro_system.hpp
```diff
@@ -826,8 +861,10 @@ void HydroSystem<problem_t>::ComputeFluxes(amrex::MultiFab &x1Flux_mf, amrex::Mu
...
+				P_L = quokka::EOS<problem_t>::ComputePressure(rho_L, eint_L * rho_L, massScalars);
+				massScalars = RadSystem<problem_t>::ComputeMassScalars(x1RightState, i, j, k);
```
BenWibking: Rename to `massScalars_R`.

src/hydro_system.hpp
```diff
@@ -470,11 +493,20 @@ void HydroSystem<problem_t>::ComputeFlatteningCoefficients(amrex::MultiFab const
 
 		if constexpr (reconstruct_eint) {
 			// compute (rho e) (gamma - 1)
-			Pplus2 *= primVar(i + 2, j, k, primDensity_index) * (gamma_ - 1.0);
-			Pplus1 *= primVar(i + 1, j, k, primDensity_index) * (gamma_ - 1.0);
-			P *= primVar(i, j, k, primDensity_index) * (gamma_ - 1.0);
-			Pminus1 *= primVar(i - 1, j, k, primDensity_index) * (gamma_ - 1.0);
-			Pminus2 *= primVar(i - 2, j, k, primDensity_index) * (gamma_ - 1.0);
+			amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(primVar, i + 2, j, k);
```
BenWibking: Can you rename `massScalars` as appropriate for each index, e.g. `massScalars_plus2`, `massScalars_minus2`, etc.?

BenWibking (dismissed):
Minor code style changes.

BenWibking (approved):

PR #338, opened 2023-07-31
Implement general EOS equations in Riemann Solver

### LINKED ISSUE #337: compute Roe-average sound speed for a general EOS
To support a general EOS, this line in the Riemann solver must be changed:
https://github.com/quokka-astro/quokka/blob/aa288b57d2dda961815af7f986ebbb8fc30b9234/src/HLLC.hpp#L39

so that it matches Eq. 4.12 of Kershaw et al. (1998):

Screenshot 2023-07-30 at 4 01 38 PM

For an ideal gas, $C_{P}^{\star} = 1/(\gamma - 1)$ and $C_{\rho}^{\star} = 0$.

For a general EOS, the calculation of $C_{\rho}^{\star}$ and $C_{P}^{\star}$ is nontrivial.

An algorithm to calculate these quantities is given here: [Appendix_A.pdf](https://github.com/quokka-astro/quokka/files/12210059/Appendix_A.pdf). (In their notation, $I$ is the specific internal energy (i.e., $e_{int}$).)

psharda:
Resolves #337 and #271 and #278
Includes #339

### DIFF
--- a/src/EOS.hpp
+++ b/src/EOS.hpp
@@ -232,8 +232,8 @@ EOS<problem_t>::ComputeEintTempDerivative(const amrex::Real rho, const amrex::Re
 }
 
 template <typename problem_t>
-AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto
-EOS<problem_t>::ComputeOtherDerivatives(const amrex::Real rho, const amrex::Real P, const std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars)
+AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeOtherDerivatives(const amrex::Real rho, const amrex::Real P,
+										      const std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars)
 {
 	// compute derivative of specific internal energy w/r/t density, given density and pressure
 	amrex::Real deint_dRho = NAN;
@@ -277,7 +277,6 @@ EOS<problem_t>::ComputeOtherDerivatives(const amrex::Real rho, const amrex::Real
 #endif
 	return std::make_tuple(deint_dRho, deint_dP, dRho_dP);
 }
-
 
 template <typename problem_t>
 AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputePressure(amrex::Real rho, amrex::Real Eint,
--- a/src/EOS.hpp
+++ b/src/EOS.hpp
@@ -9,6 +9,7 @@
 
 #include <cmath>
 #include <optional>
+#include <tuple>
 
 #include "AMReX_Array.H"
 #include "AMReX_GpuQualifiers.H"
@@ -56,16 +57,7 @@ template <typename problem_t> class EOS
 	    -> amrex::Real;
 
 	[[nodiscard]] AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE static auto
-	ComputeeintDensDerivative(amrex::Real rho, amrex::Real P, const std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars = {})
-	    -> amrex::Real;
-
-	[[nodiscard]] AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE static auto
-	ComputeeintPresDerivative(amrex::Real rho, amrex::Real P, const std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars = {})
-	    -> amrex::Real;
-
-	[[nodiscard]] AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE static auto
-	ComputeDensPresDerivative(amrex::Real rho, amrex::Real P, const std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars = {})
-	    -> amrex::Real;
+	ComputeOtherDerivatives(amrex::Real rho, amrex::Real P, const std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars = {});
 
 	[[nodiscard]] AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE static auto
 	ComputePressure(amrex::Real rho, amrex::Real Eint, const std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars = {}) -> amrex::Real;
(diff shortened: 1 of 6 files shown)


src/HLLC.hpp
```diff
@@ -36,7 +36,10 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto HLLC(quokka::HydroState<N_scalars> cons
 	double cs_tilde = NAN;
 	if (gamma != 1.0) {
 		// TODO(ben): implement Roe average for general EOS
-		cs_tilde = std::sqrt((gamma - 1.) * (H_tilde - 0.5 * vsq_tilde));
+		const double C_star_P = 1.0 / (gamma - 1.0);
```
psharda: @BenWibking - the `HydroState` struct that is used in HLLC to define `sL` and `sR` does not store massScalars. It only stores the total scalars. Do we need to add massScalars to this struct? Because we need to access massScalars in HLLC and we can't use our `ComputeMassScalars` function from `RadSystem` here.

BenWibking: Yeah, I think it makes sense to add massScalars to the HydroState.

psharda:
@BenWibking I implemented equation A.5a of Kershaw+1998 to find $`C^{*}_{\rho}`$ for arbitrary EOS. I have kept $`C^{*}_{P}`$ the same as before (so, it is only applicable for ideal gas EOS). Now, for gamma_law, `dedr` = 0 in microphysics. So, equation A.5a of Kershaw reduces to: $`C^{*}_{\rho} = \frac{1}{2} \left(I^{-} + I^{+} \right)`$. However, since both $`I^{-}`$ and $`I^{+}`$ are > 0 for several Hydro test problems, I don't obtain $`C^{*}_{\rho} = 0`$ as we'd expect for ideal gas. Why?

All the tests (except for HydroSMS) pass anyway, which I don't understand. Is the issue here that equation A5.a is an approximate equation, or that I changed $`C^{*}_{\rho}`$ but not $`C^{*}_P`$?

psharda:
Ugh, why are my $$ not working above to display inline equations

BenWibking:
> All the tests (except for HydroSMS) pass anyway, which I don't understand. Is the issue here that equation A5.a is an approximate equation, or that I changed $C_{\rho}$ but not $C_P$?

The issue is the A.5a is approximate.

For reasons I don't fully understand (but have seen in experimenting with different ways to calculate the wavespeeds for ideal gases), the HydroSMS test is very sensitive to the wavespeed estimates.  So I don't think this is anything to worry about.

BenWibking:
Hmm. It does look bad though:
hydro_sms_1 0000

It doesn't get much better unless I add a huge amount of artificial viscosity, which is not something I'd want to do in a real simulation.

Let's hope it's less oscillatory once you use the numerical approximation for $C_P^{\star}$.

psharda:
[hydro_sms_1.0000.pdf](https://github.com/quokka-astro/quokka/files/12217373/hydro_sms_1.0000.pdf)
The oscillations seem less, but the spike is larger.

src/EOS.hpp
```diff
@@ -43,16 +43,33 @@ template <typename problem_t> class EOS
...
+	    -> amrex::Real;
+
+	[[nodiscard]] AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE static auto
+	ComputeeintPresDerivative(amrex::Real rho, amrex::Real P, const std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars = {})
+	    -> amrex::Real;
+
+	[[nodiscard]] AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE static auto
+	ComputeDensPresDerivative(amrex::Real rho, amrex::Real P, const std::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars = {})
+	    -> amrex::Real;
+
```
psharda, line 61: @BenWibking The three derivatives here - dedr, dedp and drdp, are only used for the Riemann solver, and at the same place in the solver. These three functions make some tests quite slow (e.g., the test HydroShocktubeCMA takes 100 seconds instead of the usual 35, and HydroLeBlanc takes 190 seconds instead of 75). Shall I club these three into one function, such that we only call Microphysics once, and fill in all the three derivatives?

BenWibking: If that makes it faster, sure. Is it the debug build that's slow, or all of the builds?

psharda: The numbers above are for the release build

BenWibking: Oh, yikes. Yes, definitely need to make sure we don't slow down that much.

psharda:
This is how it was before:

image

src/EOS.hpp
psharda: Okay, I clubbed these functions into one and now we are back to the same speed as before.

BenWibking:
The implementation looks good to me code-wise, but I don't want to merge until we have a more satisfactory solution for the slow-moving shock problem (or a consensus that we can't do much about it).

I think it would be ideal to also have a shock test problem that uses a non-ideal gas equation of state, such as one of the tests from https://ui.adsabs.harvard.edu/abs/2015ApJS..216...31Z/abstract.

BenWibking:
There are also some test problems here that might be useful: https://ui.adsabs.harvard.edu/abs/2020ApJS..248....7C/abstract

psharda:
Found a way to pass the test! Kershaw already mentions it but I overlooked it.

BenWibking:
> Found a way to pass the test! Kershaw already mentions it but I overlooked it.

I am intrigued... how did you do it?

psharda:
> > Found a way to pass the test! Kershaw already mentions it but I overlooked it.
> 
> I am intrigued... how did you do it?

@BenWibking Kershaw says: "If $` \Delta (\rho) \neq 0 `$ we can insert (A.5b) and (A.5c) into (A.1) and then solve for $` C^{*}_{\rho} `$; otherwise we use (A.5a)." That's what I did - I first find $` C^{*}_{P} `$ from their equation A5.b, and assuming $` C^{*}_{f} = 0`$ (which is true except for primordial chem), I use equation A.1 to find $` C^{*}_{\rho} `$ if $` \Delta (\rho) \neq 0 `$ and equation A.5a otherwise. 

This ensures the SMS test passes (because A1 is an exact equation), but it causes three other tests to fail (ShuOsher, ShocktubeCMA and RadhydroShock). To me, it seems that the reason these tests fail is that $` C^{*}_{\rho} `$ calculated from equation A1 is < 0 at some interfaces after a few timesteps, whereas the approximation (equation A5.a) yields a positive value always and thats why these tests passed earlier. I'm gonna look into it more tomorrow to see if I can come up with something that can pass all the four tests.

@markkrumholz do you have any suggestions?

BenWibking:
This might help: https://github.com/quokka-astro/quokka/pull/339

psharda:
So should I merge your branch into mine?

BenWibking:
> So should I merge your branch into mine?

Sure. Just note that I've hard-coded the value of $\mathcal{G}$, which you'll have to add to the EOS and then call it to obtain the value.

psharda:
> #339

@BenWibking when I calculate the $`\gamma`$ needed for $`\mathcal{G}`$, which state should I use - the left or the right? Because I have a $`\partial P / \partial \rho`$ for both the left and right states.

BenWibking:
> > #339
> 
> @BenWibking when I calculate the γ needed for G, which state should I use - the left or the right? Because I have a ∂P/∂ρ for both the left and right states.

That's a good question... I think the right thing to do is to compute it for both states, and then use the left state $\mathcal{G}_L$ for $S_L$ and the right state $\mathcal{G}_R$ for $S_R$.

BenWibking:
Also, are you computing $\partial P / \partial \rho$ or $(\partial P / \partial \rho)_s$ (i.e., at a constant entropy)?

psharda:
> Also, are you computing ∂P/∂ρ or (∂P/∂ρ)s (i.e., at a constant entropy)?

The former. I sort of ignored the $`s`$ haha. Microphysics only has the former, not the latter.

BenWibking:
> > Also, are you computing ∂P/∂ρ or (∂P/∂ρ)s (i.e., at a constant entropy)?
> 
> The former. I sort of ignored the s haha. Microphysics only has the former, not the latter.

It is there, but just under a different name :)
The square of the adiabatic sound speed $c_s^2 = (\partial P / \partial \rho)_s$.

psharda:
@BenWibking does Microphysics also have $`\left({\frac{\partial^2 P}{\partial^2 \rho}}\right)_{s}`$

BenWibking:
> @BenWibking does Microphysics also have (∂2P∂2ρ)s

Not sure. You'll have to ask Mike.

src/HLLC.hpp
```diff
@@ -34,17 +35,52 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto HLLC(quokka::HydroState<N_scalars> cons
...
+		S_L = std::min(sL.u - (sL.cs + s_NL), u_tilde - (cs_tilde + s_NL));
+		S_R = std::max(sR.u + (sR.cs + s_NR), u_tilde + (cs_tilde + s_NR));
+
 	} else {
 		cs_tilde = 0.5 * (sL.cs + sR.cs);
-	}
+		const double G_gamma_L = 1.0;
+		const double G_gamma_R = 1.0;
+
+		const double G_L = 0.5 * (G_gamma_L + 1.);
```
BenWibking, line 72: This drops the third term in 2.24:

257545030-2685e971-3233-414c-b442-f0af99f94f69

While it is zero for an ideal gas, it's not zero in general.

psharda: Yeah yeah, implementing in a step by step way. Its there now.

psharda:
hmm, I pushed some more commits but they haven't appeared here.

BenWibking:
Wow, it looks basically perfect: 
hydro_sms_1 0000

psharda:
> Wow, it looks basically perfect: hydro_sms_1 0000

Woah, didn't expect that. It's even better than what we had before this PR.

psharda:
all tests will pass now. any more comments @BenWibking ?

BenWibking (approved):

BenWibking:
As long as the GPU tests pass, it looks good to me.

PR #349, opened 2023-08-14
[WIP] require CMake 3.20+

BenWibking:
This enforces a minimum version requirement of CMake 3.20.

This allows us to remove some workarounds that were only needed in older CMake versions. CMake 3.20 was released in 2021, so it should be widely available for all users.

This causes a build error on GPU due to a change in how CMake treats `CMAKE_CUDA_ARCHITECTURES` that we currently don't have a fix for.


psharda:
There's some bug that doesn't let the tests compile on GPUs

BenWibking:
It looks like the issue is that `CUDA_ARCHITECTURES` is handled differently in CMake 3.20 (https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html).

It's not immediately obvious to me how to fix it. I will ask the AMReX developers if they have a fix.

BenWibking:
Since this doesn't work, I'm closing it for now. Hopefully CMake will work in the future.

PR #352, opened 2023-08-18
add Navier-Stokes Characteristic Boundary Conditions

BenWibking:
Adds Navier-Stokes Characteristic Boundary Conditions (NSCBC) from [Motheau, Almgren & Bell, AIAA, Vol. 55, No. 10, 2017](http://dx.doi.org/10.2514/1.J055885):

* The relevant boundary conditions are for subsonic inflow and subsonic outflow. The inflow boundary conditions are only partially implemented (for the x-direction lower boundary, for the exclusive purpose of running one of the test problems from the paper). The outflow BC is implemented for all faces.

* We do not include the special treatment of the corner boundaries that is done in the paper. (It may be necessary to implement this in the future if we want to have adjacent outflow-outflow boundaries.)

* Two test problems are included:
   * a 1D channel-flow test, where the left boundary is subsonic inflow and the right boundary is subsonic outflow. This is quantitatively checked against the exact solution.
   * a 2D vortex outflow test, where a 2D vortex is moving with a uniform velocity until it reaches the right subsonic outflow boundary. This seems impossible to check quantitatively, so I manually examined the output snapshots in VisIt to see that it flows outward without significant reflections or disturbances back into the box, matching row 2 of Figure 4 in the paper. I also ran with extrapolating boundaries and saw much poorer behavior, in agreement with row 1 of Figure 4. I also ran this problem with outflow boundaries on each lower/upper x/y boundary and the vortex flows outward without significant disturbances for all 4 cases.

Additionally, this adds exhaustive debugging output that can be enabled at runtime for all temporary/work MultiFabs used during the hydro update (including all ghost zones). This should only be enabled for debugging purposes, as it outputs *every* MultiFab used during *every* timestep.


BenWibking:
@markkrumholz This is a bit of a hefty PR, so please let me know if I should split it into smaller pieces and/or anything else that would be helpful in reviewing.

markkrumholz:
OK, will have a look, but it may not be until tomorrow or Friday due to other commitments.

src/NSCBC_inflow.hpp
markkrumholz, line 1: Typo in comment on line 32 "as shock" instead of "a shock"

src/NSCBC_outflow.hpp
markkrumholz, line 1: In lines 212 and 238, it looks like you have hardcoded in where the passive scalars start in the Q array. Should this be changed to the index defined in the hydro system description? Otherwise it looks like it will only work for 3d.

markkrumholz (commented):
I have left a couple comments on the PR, one just fixing a typo in a comment, one more substantive.

src/NSCBC_inflow.hpp
BenWibking: Fixed.

src/NSCBC_outflow.hpp
BenWibking: As a matter of code style, it would be better to use the index variables, but I don't think it matters in terms of correctness.

There are always 3 momentum components in the conserved variables and 3 velocity components in the primitive variables, regardless of the dimensionality of the code, so it should always work.

markkrumholz (approved):

src/NSCBC_outflow.hpp
markkrumholz: OK, thanks for the explanation. That's fine.

markkrumholz (approved):

PR #363, opened 2023-09-07
fill ghost cells before output

BenWibking:
This fills the ghost cells on all levels so that they are consistent before writing a plotfile. Since this modifies the state MultiFabs, we have to un-`const` the plotfile I/O functions.


BenWibking:
Oops, I didn't check if there were any face-centered variables before trying to fill `state_new_fc`. Fixing...

markkrumholz (approved):

PR #366, opened 2023-09-07
add history file output

BenWibking:
Add a "history file" output, inspired by the same feature in Athena.

This writes a text file named "history.txt" in the simulation working directory with scalar-valued statistics that can be computed by the problem generator in the `RadhydroSimulation<problem_t>::ComputeStatistics()` function. **The user is responsible for performing any needed MPI reductions inside this function before returning the computed statistics.**

### DIFF
--- a/src/RadhydroSimulation.hpp
+++ b/src/RadhydroSimulation.hpp
@@ -191,6 +191,9 @@ template <typename problem_t> class RadhydroSimulation : public AMRSimulation<pr
 	// compute derived variables
 	void ComputeDerivedVar(int lev, std::string const &dname, amrex::MultiFab &mf, int ncomp) const override;
 
+	// compute statistics
+  	auto ComputeStatistics() -> std::map<std::string, amrex::Real> override;
+
 	// fix-up states
 	void FixupState(int level) override;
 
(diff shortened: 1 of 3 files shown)


markkrumholz (commented):
It looks like you have left it up the the user to reduce statistics across MPI ranks, i.e., this has to be handled in their implementation of ComputeStatistics. It seems like this might be an invitation to user confusion. I understand the reason for the decision, since it allows maximum flexibility -- the user may want to reduce in different ways, e.g., computing extrema versus means -- but it seems like it would be good if the comment here included an explicit warning that the user is responsible for implementing MPI reduction in their ComputeStatistics implementation.

src/RadhydroSimulation.hpp
```diff
@@ -496,6 +499,14 @@ void RadhydroSimulation<problem_t>::ComputeDerivedVar(int lev, std::string const
 	// compute derived variables and save in 'mf' -- user should implement
 }
 
+template <typename problem_t>
+auto RadhydroSimulation<problem_t>::ComputeStatistics() -> std::map<std::string, amrex::Real> 
+{
+	// compute statistics and return a std::map<std::string, amrex::Real> -- user should implement
+	// IMPORTANT: the user is responsible for performing any necessary MPI reductions before returning
```
BenWibking, line 506: @markkrumholz I've added a comment about this here.

src/simulation.hpp
```diff
@@ -1741,6 +1765,51 @@ template <typename problem_t> void AMRSimulation<problem_t>::WritePlotFile() con
 #endif
 }
 
+template <typename problem_t>
+void AMRSimulation<problem_t>::WriteStatisticsFile() {
+  // append to statistics file
+  static bool isHeaderWritten = false;
+
+  // compute statistics
+  // IMPORTANT: the user is responsible for performing any necessary MPI reductions inside ComputeStatistics
```
BenWibking, line 1787: @markkrumholz I've also added a comment in the source here.

markkrumholz (approved):

markkrumholz (approved):

PR #367, opened 2023-09-07
add axis-aligned projection outputs

BenWibking:
This outputs 2D axis-aligned projections as AMReX plotfiles prefixed with `proj`.

The problem generator must call `computePlaneProjection(F const &user_f, const int dir)` where `user_f` is a lambda function that returns the value to project and `dir` is the axis along which the projection is taken.

### DIFF
--- a/src/AdvectionSimulation.hpp
+++ b/src/AdvectionSimulation.hpp
@@ -81,6 +81,8 @@ template <typename problem_t> class AdvectionSimulation : public AMRSimulation<p
 
 	// compute derived variables
 	void ComputeDerivedVar(int lev, std::string const &dname, amrex::MultiFab &mf, int ncomp) const override;
+	// compute projected vars
+  	[[nodiscard]] virtual auto ComputeProjections(int dir) const -> std::unordered_map<std::string, amrex::BaseFab<amrex::Real>> override;
 
 	void FixupState(int lev) override;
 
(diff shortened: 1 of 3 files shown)


markkrumholz (commented):
It looks like this code only products projection plots on the coarsest level. I assume that this is the intended option, but it does somewhat limit the utility for a problem that uses any significant degree of AMR. It might make more sense to project to the finest level, or give the user an option to do so. Fine if you don't want to tackle that in this PR and save it for a future enhancement, in which case I will approve now.

BenWibking:
> It looks like this code only products projection plots on the coarsest level. I assume that this is the intended option, but it does somewhat limit the utility for a problem that uses any significant degree of AMR. It might make more sense to project to the finest level, or give the user an option to do so. Fine if you don't want to tackle that in this PR and save it for a future enhancement, in which case I will approve now.

Yes, that is a significant limitation of the current PR. It's good enough for my cloud simulations, though, so I'd prefer to leave that to a future PR.

markkrumholz (dismissed):

BenWibking:
@markkrumholz This needs to be re-approved because I had to do a manual merge commit.

markkrumholz (dismissed):

BenWibking:
@markkrumholz There was another merge issue. Can you approve again? Hopefully it works this time.

markkrumholz (approved):

PR #369, opened 2023-09-08
Remove is_chemistry_enabled

psharda:
We don't need the Physics_Trait `is_chemistry_enabled` anymore.


BenWibking:
Looks fine to me. Should probably mention in this PR what the correct way to enable chemistry is instead.

psharda:
One can enable chemistry using `enableChemistry` in `RadhydroSimulation.hpp`, following the example of primordial chemistry. Also need to compile Chemistry.hpp with the problem in question.

BenWibking (approved):

PR #376, opened 2023-09-12
ensure first-order flux correction is consistent at box boundaries

### LINKED ISSUE #374: ensure consistent first-order flux correction at box boundaries
If a bad cell is adjacent to a box boundary, then the cell on the adjacent box on the other side of the boundary needs to have its flux updated. Currently, this does not happen, which is inconsistent.

BenWibking:
This adds 1 ghost cell to the `redoFlag` iMultiFab and communicates its ghost cells before doing flux correction. Then on each MPI rank, it loops over the entire `redoFlag` (including the ghost cells) and replaces the fluxes of adjacent valid cells.

Fixes https://github.com/quokka-astro/quokka/issues/374.


BenWibking:
@markkrumholz I think this will fix it.

markkrumholz (approved):

PR #378, opened 2023-09-13
Enforce checks and limits on mass scalars

psharda:
This PR will:

1.) Ensure we return an invalid State via `isStateValid` if the partial densities of any mass scalars are < 0.

2.) Ensure we enforce a partial density floor for mass scalars while doing hydro. We already do this while doing chemistry (via the `small_x` param in Microphysics)

### DIFF
--- a/src/hydro_system.hpp
+++ b/src/hydro_system.hpp
@@ -29,6 +29,9 @@
 #include "radiation_system.hpp"
 #include "valarray.hpp"
 
+// Microphysics headers
+#include "extern_parameters.H"
+
 // this struct is specialized by the user application code
 //
 template <typename problem_t> struct HydroSystem_Traits {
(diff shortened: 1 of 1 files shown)


psharda:
@BenWibking , regarding point 2 above, this line seems to enforce positivity for all scalars when it enforces the density floor: https://github.com/quokka-astro/quokka/blob/d17b5e4b742e3b9b1401c0418edaf4c625503ea4/src/hydro_system.hpp#L711

BenWibking:
Yes but it only triggers when the total density is negative, not when only the mass scalars are negative.

src/hydro_system.hpp
```diff
@@ -422,6 +425,18 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto HydroSystem<problem_t>::isStateValid(am
 	const amrex::Real rho = cons(i, j, k, density_index);
 	bool isDensityPositive = (rho > 0.);
 
+    bool isMassScalarPositive = true;
+	if (nmscalars_ > 0) {
```
BenWibking: Can this be made `if constexpr` instead of just `if`?

psharda: I dont know...

BenWibking: It should be possible if `nmscalars_` is a `constexpr` variable.

BenWibking (approved):

PR #380, opened 2023-09-13
add LLF fallback solver

### LINKED ISSUE #379: fallback to LLF solver
We should add the option to use the Local Lax-Friedrichs (LLF) Riemann solver when doing first-order flux correction.

When doing first-order flux correction, we revert to donor cell fluxes but currently use the same Riemann solver (HLLC). This may be insufficient for flows very close to vacuum, since the positivity-preserving timestep can be extremely small compared to the CFL timestep.

BenWibking:
Uses the Local-Lax Friedrichs (Rusanov) solver as a "fallback" Riemann solver.

Closes https://github.com/quokka-astro/quokka/issues/379.

### DIFF
--- a/src/LLF.hpp
+++ b/src/LLF.hpp
@@ -40,7 +40,7 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto LLF(quokka::HydroState<N_scalars, N_msc
 	const quokka::valarray<double, fluxdim> F_R = sR.u * U_R + sR.P * D_R;
 
 	// open the Riemann fan
-	quokka::valarray<double, fluxdim> F = 0.5 * (F_L + F_R) - 0.5 * Sp * (F_R - F_L);
+	quokka::valarray<double, fluxdim> F = 0.5 * (F_L + F_R) - 0.5 * Sp * (U_R - U_L);
 	return F;
 }
 } // namespace quokka::Riemann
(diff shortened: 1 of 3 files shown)


markkrumholz (commented):
The change to how Riemann solvers are referenced is going to cause conflicts with the MHD branch, since the same changes will need to applied to the HLLD solver in that branch. I'm not sure where @AstroKriel's MHD branch is with respect to the repo, but we might want to have a look just to make sure that there won't be any pain points when we need to integrate these changes with the MHD HLLD solver.

BenWibking:
> The change to how Riemann solvers are referenced is going to cause conflicts with the MHD branch, since the same changes will need to applied to the HLLD solver in that branch. I'm not sure where @AstroKriel's MHD branch is with respect to the repo, but we might want to have a look just to make sure that there won't be any pain points when we need to integrate these changes with the MHD HLLD solver.

The trivial solution is to just add another conditional case to the `if constexpr` statement in `HydroSystem::ComputeFluxes`. I don't think it should be more involved than that.

markkrumholz:
> > The change to how Riemann solvers are referenced is going to cause conflicts with the MHD branch, since the same changes will need to applied to the HLLD solver in that branch. I'm not sure where @AstroKriel's MHD branch is with respect to the repo, but we might want to have a look just to make sure that there won't be any pain points when we need to integrate these changes with the MHD HLLD solver.
> 
> The trivial solution is to just add another conditional case to the `if constexpr` statement in `HydroSystem::ComputeFluxes`. I don't think it should be more involved than that.

OK. I guess that should be done in a separate PR, which will have to happen before @AstroKriel's code can be merged into the main branch.

markkrumholz (approved):

PR #384, opened 2023-09-15
use minmod limiting for PLM

### LINKED ISSUE #377: add minmod limiter option for PLM reconstruction
The minmod limiter is much more diffusive than the current MC limiter, and may be useful for very unstable problems.

Also, minmod ensures the sign of the jump between left- and right- states at an interface is the same as the sign of the jump between the cell averages, which is not true of any other TVD limiter.

BenWibking:
Uses minmod limiting for the slopes when doing piecewise-linear reconstruction (PLM). This is more diffusive and more stable.

Closes https://github.com/quokka-astro/quokka/issues/377.


markkrumholz (approved):

BenWibking:
We can't merge this, since avatargpu is still not working for some reason.

PR #388, opened 2023-09-18
Multigroup radiation transport implementation

chongchonghe:
This PR implements multigroup radiation transport in Quokka. The key changes are:

- Radiation is now treated using N photon energy groups, with group boundaries and opacities defined for each group. 
- The radiation moment equations are solved separately for each group. Matter-radiation coupling is handled through a Newton-Raphson iteration that solves a set of 1+N equations with a sparse 1+N by 1+N Jacobian matrix. 
- Radiation energy exchange between gas and photons is apportioned among groups to conserve total energy. 
- Unit tests for the 1D RadForce and RadTube problems have been updated for the multigroup case. The RadTube test now verifies the separate evolution of two photon groups and compare it with analytic solution.
- ComputeRosselandOpacity is replaced with ComputeFluxMeanOpacity which returns the flux-mean opacity for each photon group. The energy-mean opacity is assumed to equal the flux-mean opacity. This assumption will be released via a future PR. 
- Infrastructure changes include adding the number of groups and group boundaries/units to traits classes, adding the number of variables for each fluid element, and refactoring some radiation-related data structures.

Additional quantitative test problems are planned for future PRs, such as a multigroup version of the SuOlson picket fence problem. This PR focuses on the core multigroup implementation and unit testing for 1D cases. The user should refer to the RadTube or RadForce example for usage of multigroup RT. For single-group (gray approximation) case, the only necessary change is to add `static constexpr int nGroups = 1;` to Physics_Traits. Feedback is welcome to further improve and test this new capability.


src/radiation_system.hpp
```diff
@@ -28,10 +28,15 @@
 #include "hyperbolic_system.hpp"
 #include "simulation.hpp"
 #include "valarray.hpp"
+#include "planck_integral.hpp"
 
 // physical constants in CGS units
 static constexpr double c_light_cgs_ = 2.99792458e10;	      // cgs
 static constexpr double radiation_constant_cgs_ = 7.5646e-15; // cgs
+static constexpr double pi = 3.14159265358979323846;
+static constexpr double hplanck = 6.62606957e-27;  // erg s
```
psharda: We don't need to define these again in quokka, we can just call the ones already defined in Microphysics. For example, k_B can be called as `C::k_B`

BenWibking: Yes, good catch. We should change all of these to just use the values from Microphysics. They are all in the `C` namespace (C is for "constants").

BenWibking:
@chongchonghe I can take a look at this in detail sometime later this week. In the meantime, can you fix the clang-tidy warnings? You can also see them locally in VSCode if you use the clangd extension (https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd)

See also: https://github.com/quokka-astro/quokka/wiki/How-to-use-clang%E2%80%90tidy

src/radiation_system.hpp
BenWibking: See this file for the available constants: https://github.com/AMReX-Astro/Microphysics/blob/main/constants/fundamental_constants.H

src/RadForce/test_radiation_force.cpp
```diff
@@ -56,6 +56,11 @@ template <> struct RadSystem_Traits<TubeProblem> {
 	static constexpr double radiation_constant = radiation_constant_cgs_;
 	static constexpr double Erad_floor = 0.;
 	static constexpr bool compute_v_over_c_terms = true;
+	static constexpr double energy_unit = ev2erg / k_B;
+	static constexpr double lower_rad_energy_bound = 0.0;
+	static constexpr double upper_rad_energy_bound = std::numeric_limits<double>::max();
+  static constexpr int nGroups = 2;
+  static constexpr std::array<double, nGroups - 1> radBoundaries {13.6};  // eV, size = nGroups - 1
```
BenWibking: I agree with Mark about changing this to `nGroups + 1` and including the lower energy bound and upper energy bound in this array.

src/RadForce/test_radiation_force.cpp
```diff
@@ -67,14 +72,35 @@ template <> struct Physics_Traits<TubeProblem> {
 	static constexpr bool is_radiation_enabled = true;
 	// face-centred
 	static constexpr bool is_mhd_enabled = false;
+  // CCH: a hack to get nvarTotal_cc_radhydro right in physics_info.hpp
+  static constexpr int nGroups = RadSystem_Traits<TubeProblem>::nGroups;
```
BenWibking: I think it might be necessary to just define this in `Physics_Traits` and reference it within `RadSystem_Traits`, rather than defining it in both traits classes.

src/radiation_system.hpp
```diff
@@ -133,18 +160,68 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
 
 	AMREX_GPU_HOST_DEVICE static auto ComputeEddingtonFactor(double f) -> double;
 	AMREX_GPU_HOST_DEVICE static auto ComputePlanckOpacity(double rho, double Tgas) -> double;
+	AMREX_GPU_HOST_DEVICE static auto ComputePlanckOpacityForGroup(double rho, double Tgas) -> quokka::valarray<double, nGroups_>;
```
BenWibking: I don't think it makes sense to have both `ComputePlanckOpacity` and `ComputePlanckOpacityForGroup`. I think it should be named `ComputePlanckOpacity` and it should just return an array of the opacities for all of the groups.

src/radiation_system.hpp
```diff
@@ -133,18 +160,68 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
...
+	AMREX_GPU_HOST_DEVICE static auto ComputeFluxMeanOpacityForGroup(double rho, double Tgas) -> quokka::valarray<double, nGroups_>;
```
BenWibking: Same comment as above.

src/radiation_system.hpp
```diff
@@ -133,18 +160,68 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
...
 	AMREX_GPU_HOST_DEVICE static auto ComputePlanckOpacityTempDerivative(double rho, double Tgas) -> double;
+	AMREX_GPU_HOST_DEVICE static auto ComputeOpacityTempDerivativeAtBins(double rho, double Tgas)->quokka::valarray<double, nGroups_>;
```
BenWibking: Same comment as above.

src/radiation_system.hpp
```diff
@@ -133,18 +160,68 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
...
 	AMREX_GPU_HOST_DEVICE static auto ComputeRosselandOpacity(double rho, double Tgas) -> double;
 	AMREX_GPU_HOST_DEVICE static auto ComputeEintFromEgas(double density, double X1GasMom, double X2GasMom, double X3GasMom, double Etot) -> double;
 	AMREX_GPU_HOST_DEVICE static auto ComputeEgasFromEint(double density, double X1GasMom, double X2GasMom, double X3GasMom, double Eint) -> double;
 
-	template <FluxDir DIR>
-	AMREX_GPU_DEVICE static auto ComputeCellOpticalDepth(const quokka::Array4View<const amrex::Real, DIR> &consVar,
-							     amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dx, int i, int j, int k) -> double;
+  template <FluxDir DIR>
+  AMREX_GPU_DEVICE static auto ComputeCellOpticalDepth(const quokka::Array4View<const amrex::Real, DIR> &consVar,
```
BenWibking: This needs to be a function of the energy group.

src/radiation_system.hpp
```diff
@@ -133,18 +160,68 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
...
+						       amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dx, int i, int j, int k) -> double;
 
-	AMREX_GPU_DEVICE static auto isStateValid(std::array<amrex::Real, nvarHyperbolic_> &cons) -> bool;
+  AMREX_GPU_DEVICE static auto isStateValid(std::array<amrex::Real, nvarHyperbolic_> &cons) -> bool;
 };
 
+// CCH: compute radiation energy fractions for each bin of a Planck function, given the temperature.
+template <typename problem_t>
+void RadSystem<problem_t>::ComputePlanckEnergyFractions(quokka::valarray<amrex::Real, nGroups_> &radEnergyFractions, amrex::Real const temperature)
```
BenWibking: Is this needed for anything other than setting the initial conditions in LTE?

src/valarray.hpp
```diff
@@ -59,6 +59,7 @@ template <typename T, int d> class valarray
 };
 } // namespace quokka
 
+// array + array
```
BenWibking: `valarray` is intended for representing mathematical vectors, with addition and multiplication by a scalar. I don't think it makes sense to use this object to add other operations. In general, it may be better to use a `std::array`.

BenWibking (commented):
I think the general approach to the code modifications is fine.

I think the additional variables you added to the traits classes can be simplified, so that RadSystem_Traits only needs one additional variable for the energy group boundaries and one additional variable to specify its units, and Physics_Traits only needs one additional variable for the number of energy groups.

I think you may be stretching `quokka::valarray` beyond its intended use case, which was originally for mathematical vectors with addition (and subtraction) of two vectors and multiplication by a scalar (and multiplication by the reciprocal of a scalar). In some cases, `std::array` may be a better choice. If that doesn't work, you may want to create a new array class for your purposes.

I don't think there should be multiple versions of the opacity functions. They should all return opacities for all of the energy groups. It doesn't make sense to keep the old opacity functions that return a scalar.

For multigroup test problems, I would suggest the multigroup radiative shock from Skinner et al. (https://arxiv.org/abs/1806.07390), the problem from Shestakov and Bolstad (https://ui.adsabs.harvard.edu/abs/2005JQSRT..91..133S/abstract), and the radiating sphere test from Graziani (2008) (https://link.springer.com/chapter/10.1007/978-3-540-77362-7_6).

@markkrumholz There is the more general question of whether to solve the comoving or mixed-frame equations. I think for multigroup the preferred option is actually to solve the comoving frame equations, because the group boundaries will otherwise change with the velocity boost. This is necessary to have a feasible number of energy groups where a small number of specific lines are of interest, e.g. Ly-alpha. But this can be implemented later.

src/radiation_system.hpp
chongchonghe: Yes, it's needed for calculating the source terms (Eq. 28 and 33 of the Quokka paper). Now, B is replace with B_i = B * fraction_i.

src/valarray.hpp
chongchonghe: I'm trying to extend it to the capacity of std::valarray. Is there anything wrong with this? I know the C++ community doesn't use valarray much (e.g. https://twitter.com/blelbach/status/839743137761259521?lang=en ), but I don't really understand why. I find it very convenient. Is it a speed issue?

BenWibking: I agree that it is very convenient. I'm not sure why the C++ community doesn't like it very much.

If you want to extend it, I am okay with it. Could you put the changes you want to make to quokka::valarray in a separate PR so I can review them independently of the multigroup code? That PR can be reviewed and merged before tackling the this PR with all of the multigroup stuff.

src/RadForce/test_radiation_force.cpp
```diff
@@ -67,14 +72,35 @@ template <> struct Physics_Traits<TubeProblem> {
...
 };
 
 template <> AMREX_GPU_HOST_DEVICE auto RadSystem<TubeProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> double
 {
 	return 0.; // no heating/cooling
 }
 
-template <> AMREX_GPU_HOST_DEVICE auto RadSystem<TubeProblem>::ComputeRosselandOpacity(const double /*rho*/, const double /*Tgas*/) -> double { return kappa0; }
+template <> AMREX_GPU_HOST_DEVICE auto RadSystem<TubeProblem>::ComputePlanckOpacityForGroup(const double /*rho*/, const double /*Tgas*/) -> quokka::valarray<double, nGroups_>
+{
```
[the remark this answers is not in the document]
BenWibking: This is why this is getting a warning: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-member-init.html

src/RadForce/test_radiation_force.cpp
```diff
@@ -67,14 +72,35 @@
...
+	quokka::valarray<double, nGroups_> kappaPVec;
+	fillin(kappaPVec, 0.);
+	return kappaPVec;
+}
+
+template <> AMREX_GPU_HOST_DEVICE auto RadSystem<TubeProblem>::ComputeFluxMeanOpacityForGroup(const double /*rho*/, const double Tgas) -> quokka::valarray<double, nGroups_>
+{
```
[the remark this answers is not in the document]
BenWibking: Same issue: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-member-init.html

BenWibking:
All of the "uninitialized record type" warnings are due to this: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-member-init.html. Can you fix those?

chongchonghe:
Sorry. Still some work to do before ready for merging. Need to fix 2D and 3D tests. Also need to remove all the clang-tidy warnings.

BenWibking:
It looks like the HIP build fails and the warnings CI fails. Can you take a look at those?

BenWibking:
The CUDA build produces these warnings:

```
../src/radiation_system.hpp(912): warning #780-D: reference is to structured binding "i"
(673): here -- under old for-init scoping rules it would have been variable "i"
(758): here

Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"

../src/radiation_system.hpp(913): warning #780-D: reference is to structured binding "i"
(673): here -- under old for-init scoping rules it would have been variable "i"
(758): here

../src/radiation_system.hpp(914): warning #780-D: reference is to structured binding "i"
(673): here -- under old for-init scoping rules it would have been variable "i"
(758): here

../src/radiation_system.hpp(915): warning #780-D: reference is to structured binding "i"
(673): here -- under old for-init scoping rules it would have been variable "i"
(758): here

../src/radiation_system.hpp(920): warning #780-D: reference is to structured binding "i"
(673): here -- under old for-init scoping rules it would have been variable "i"
(758): here

../src/radiation_system.hpp(921): warning #780-D: reference is to structured binding "i"
(673): here -- under old for-init scoping rules it would have been variable "i"
(758): here

../src/radiation_system.hpp(922): warning #780-D: reference is to structured binding "i"
(673): here -- under old for-init scoping rules it would have been variable "i"
(758): here

../src/radiation_system.hpp(923): warning #780-D: reference is to structured binding "i"
(673): here -- under old for-init scoping rules it would have been variable "i"
(758): here
```

Is this what you intended to do in this code?

BenWibking:
The CUDA build failed. Can you take a look?

BenWibking:
I'll try to review this in detail over the weekend.

Aside from the code itself, I am hesitant to approve this without quantative tests, although they could go in future PRs. What is the plan for quantitative test problems that check the group-by-group solution?

src/RadForce/test_radiation_force.cpp
```diff
@@ -66,14 +59,42 @@ template <> struct Physics_Traits<TubeProblem> {
...
+	return kappaPVec;
 }
 
-template <> AMREX_GPU_HOST_DEVICE auto RadSystem<TubeProblem>::ComputeRosselandOpacity(const double /*rho*/, const double /*Tgas*/) -> double { return kappa0; }
+template <>
+AMREX_GPU_HOST_DEVICE auto RadSystem<TubeProblem>::ComputeFluxMeanOpacity(const double /*rho*/, const double /*Tgas*/)
+    -> quokka::valarray<double, Physics_Traits<TubeProblem>::nGroups>
+{
+	quokka::valarray<double, Physics_Traits<TubeProblem>::nGroups> kappaFVec{};
+	// kappaFVec.fillin(kappa0);
```
BenWibking: Can you remove commented-out code? Unless there's a very good reason to keep it, it just makes it less readable to future developers/users. We've tried to avoid keeping commented-out code anywhere in Quokka (although there are a few places where it still needs to be removed).

src/RadForce/test_radiation_force.cpp
```diff
@@ -132,6 +153,13 @@ template <> void RadhydroSimulation<TubeProblem>::setInitialConditionsOnGrid(quo
 	auto const &Mach_ptr = Mach_arr_g.dataPtr();
 	int x_size = static_cast<int>(x_arr_g.size());
 
+	// CCH: calculate radEnergyFractions
+	quokka::valarray<amrex::Real, Physics_Traits<TubeProblem>::nGroups> radEnergyFractions{};
+	// radEnergyFractions.fillin(1.0 / Physics_Traits<TubeProblem>::nGroups);
```
BenWibking: Can you remove the commented-out code?

src/RadForce/test_radiation_force.cpp
```diff
@@ -132,6 +153,13 @@ template <> void RadhydroSimulation<TubeProblem>::setInitialConditionsOnGrid(quo
 	auto const &Mach_ptr = Mach_arr_g.dataPtr();
 	int x_size = static_cast<int>(x_arr_g.size());
 
+	// CCH: calculate radEnergyFractions
```
BenWibking: As a matter of style, we haven't added initials indicating the developer who has written the code, since it's always available with `git blame`.

src/RadForce/test_radiation_force.cpp
```diff
@@ -183,6 +215,15 @@ AMRSimulation<TubeProblem>::setCustomBoundaryConditions(const amrex::IntVect &iv
 	amrex::Real rho = NAN;
 	amrex::Real vel = NAN;
 
+	// CCH: calculate radEnergyFractions
+	// amrex::Real const temperature = 0.0;
+	// RadSystem<TubeProblem>::ComputeRadEnergyFractions(radEnergyFractions, temperature);
```
BenWibking: Remove the commented-out code.

src/RadForce/test_radiation_force.cpp
```diff
@@ -183,6 +215,15 @@ AMRSimulation<TubeProblem>::setCustomBoundaryConditions(const amrex::IntVect &iv
...
+	quokka::valarray<amrex::Real, Physics_Traits<TubeProblem>::nGroups> radEnergyFractions{};
+	// radEnergyFractions.fillin(1.0 / Physics_Traits<TubeProblem>::nGroups);
```
BenWibking: Remove the commented-out code.

src/RadForce/test_radiation_force.cpp
```diff
@@ -183,6 +215,15 @@ AMRSimulation<TubeProblem>::setCustomBoundaryConditions(const amrex::IntVect &iv
 	amrex::Real rho = NAN;
 	amrex::Real vel = NAN;
 
+	// CCH: calculate radEnergyFractions
```
BenWibking: This is not strictly necessary, since future users can always find that you wrote this line of code with `git blame`.

src/RadMarshak/test_radiation_marshak.cpp
```diff
@@ -52,13 +52,24 @@ template <> struct Physics_Traits<SuOlsonProblem> {
...
 	static constexpr bool is_mhd_enabled = false;
+	static constexpr int nGroups = 1; // number of radiation groups
 };
 
-template <> AMREX_GPU_HOST_DEVICE auto RadSystem<SuOlsonProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> double { return kappa; }
+template <>
+AMREX_GPU_HOST_DEVICE auto RadSystem<SuOlsonProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> quokka::valarray<double, nGroups_>
+{
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// kappaPVec.fillin(kappa);
```
BenWibking: Remove commented-out code.

src/RadMarshakAsymptotic/test_radiation_marshak_asymptotic.cpp
```diff
@@ -47,25 +47,40 @@ template <> struct Physics_Traits<SuOlsonProblemCgs> {
...
 
-template <> AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<SuOlsonProblemCgs>::ComputePlanckOpacity(const double rho, const double Tgas) -> double
+template <>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<SuOlsonProblemCgs>::ComputePlanckOpacity(const double rho, const double Tgas)
+    -> quokka::valarray<double, nGroups_>
 {
 	auto sigma = kappa * std::pow(Tgas / T_hohlraum, -3); // cm^-1
-	return (sigma / rho);				      // cm^2 g^-1
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// kappaPVec.fillin(sigma / rho);      // cm^2 g^-1
```
BenWibking: Remove commented-out code.

src/RadMarshakAsymptotic/test_radiation_marshak_asymptotic.cpp
```diff
@@ -47,25 +47,40 @@ template <> struct Physics_Traits<SuOlsonProblemCgs> {
...
 
 template <>
-AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<SuOlsonProblemCgs>::ComputePlanckOpacityTempDerivative(const double rho, const double Tgas) -> double
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<SuOlsonProblemCgs>::ComputePlanckOpacityTempDerivative(const double rho, const double Tgas)
+    -> quokka::valarray<double, nGroups_>
 {
+	quokka::valarray<double, nGroups_> opacity_deriv{};
 	auto sigma_dT = (-3.0 * kappa / Tgas) * std::pow(Tgas / T_hohlraum, -3); // cm^-1
-	return (sigma_dT / rho);
+	// opacity_deriv.fillin(sigma_dT / rho);
```
BenWibking: Remove commented-out code.

src/RadMarshakCGS/test_radiation_marshak_cgs.cpp
```diff
@@ -54,16 +54,25 @@ template <> struct Physics_Traits<SuOlsonProblemCgs> {
...
+	static constexpr int nGroups = 1; // number of radiation groups
 };
 
-template <> AMREX_GPU_HOST_DEVICE auto RadSystem<SuOlsonProblemCgs>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> double
+template <>
+AMREX_GPU_HOST_DEVICE auto RadSystem<SuOlsonProblemCgs>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> quokka::valarray<double, nGroups_>
 {
-	return kappa;
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// kappaPVec.fillin(kappa);
```
BenWibking: Remove commented-out code.

src/RadMatterCoupling/test_radiation_matter_coupling.cpp
```diff
@@ -52,13 +52,24 @@ template <> struct Physics_Traits<CouplingProblem> {
...
 	static constexpr bool is_mhd_enabled = false;
+	static constexpr int nGroups = 1; // number of radiation groups
 };
 
-template <> AMREX_GPU_HOST_DEVICE auto RadSystem<CouplingProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> double { return 1.0; }
+template <>
+AMREX_GPU_HOST_DEVICE auto RadSystem<CouplingProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> quokka::valarray<double, nGroups_>
+{
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// kappaPVec.fillin(1.0);
```
BenWibking: Remove commented-out code.

src/RadMatterCouplingRSLA/test_radiation_matter_coupling_rsla.cpp
```diff
@@ -54,13 +54,24 @@ template <> struct Physics_Traits<CouplingProblem> {
...
 	static constexpr bool is_mhd_enabled = false;
+	static constexpr int nGroups = 1; // number of radiation groups
 };
 
-template <> AMREX_GPU_HOST_DEVICE auto RadSystem<CouplingProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> double { return 1.0; }
+template <>
+AMREX_GPU_HOST_DEVICE auto RadSystem<CouplingProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> quokka::valarray<double, nGroups_>
+{
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// kappaPVec.fillin(1.0);
```
BenWibking: Remove commented-out code.

src/RadStreaming/test_radiation_streaming.cpp
```diff
@@ -44,16 +37,33 @@ template <> struct Physics_Traits<StreamingProblem> {
...
+	static constexpr double Erad_floor = initial_Erad;
+	static constexpr bool compute_v_over_c_terms = false;
+};
+
+template <>
+AMREX_GPU_HOST_DEVICE auto RadSystem<StreamingProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> quokka::valarray<double, nGroups_>
 {
-	return kappa0;
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// kappaPVec.fillin(kappa0);
```
BenWibking: Remove commented-out code.

src/RadStreaming/test_radiation_streaming.cpp
```diff
@@ -64,12 +74,22 @@ template <> void RadhydroSimulation<StreamingProblem>::setInitialConditionsOnGri
 	const auto Erad0 = initial_Erad;
 	const auto Egas0 = initial_Egas;
 
+	// CCH: calculate radEnergyFractions
+	quokka::valarray<amrex::Real, Physics_Traits<StreamingProblem>::nGroups> radEnergyFractions{};
+	// radEnergyFractions.fillin(1.0 / Physics_Traits<StreamingProblem>::nGroups);
```
BenWibking: Remove commented-out code.

src/RadStreaming/test_radiation_streaming.cpp
```diff
@@ -102,23 +122,36 @@ AMRSimulation<StreamingProblem>::setCustomBoundaryConditions(const amrex::IntVec
 	amrex::GpuArray<int, 3> lo = box.loVect3d();
 	amrex::GpuArray<int, 3> hi = box.hiVect3d();
 
+	// CCH: calculate radEnergyFractions
+	quokka::valarray<amrex::Real, Physics_Traits<StreamingProblem>::nGroups> radEnergyFractions{};
+	// radEnergyFractions.fillin(1.0 / Physics_Traits<StreamingProblem>::nGroups);
```
BenWibking: Remove commented-out code.

src/RadSuOlson/test_radiation_SuOlson.cpp
BenWibking, line 1: This test problem is designed for gray radiation transport. Does it make sense to make this a 2-group problem?

src/RadTube/test_radiation_tube.cpp
```diff
@@ -58,11 +52,36 @@ template <> struct Physics_Traits<TubeProblem> {
...
+	static constexpr double energy_unit = C::k_B;
+	static constexpr amrex::GpuArray<double, Physics_Traits<TubeProblem>::nGroups + 1> radBoundaries{0., T0, 3.3 * T0, inf}; // Kelvin
+};
+
+template <>
+AMREX_GPU_HOST_DEVICE auto RadSystem<TubeProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> quokka::valarray<double, nGroups_>
+{
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// amrex::Real const kappa0_g = kappa0;
+	// kappaPVec.fillin(kappa0_g);
```
BenWibking: Remove commented-out code.

src/RadhydroShell/test_radhydro_shell.cpp
```diff
@@ -123,15 +124,21 @@ void RadSystem<ShellProblem>::SetRadEnergySource(array_t &radEnergy, const amrex
 	});
 }
 
-template <> AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<ShellProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> double
+template <>
+AMREX_GPU_HOST_DEVICE auto RadSystem<ShellProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> quokka::valarray<double, nGroups_>
 {
-	return kappa0;
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// kappaPVec.fillin(kappa0);
```
BenWibking: Remove commented-out code.

src/RadhydroShock/test_radhydro_shock.cpp
```diff
@@ -75,16 +75,25 @@ template <> struct Physics_Traits<ShockProblem> {
...
 };
 
-template <> AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<ShockProblem>::ComputePlanckOpacity(const double rho, const double /*Tgas*/) -> double
+template <>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<ShockProblem>::ComputePlanckOpacity(const double rho, const double /*Tgas*/)
+    -> quokka::valarray<double, nGroups_>
 {
-	return (kappa / rho);
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// kappaPVec.fillin(kappa / rho);
```
BenWibking: Remove commented-out code.

src/RadhydroShockCGS/test_radhydro_shock_cgs.cpp
```diff
@@ -76,16 +76,25 @@ template <> struct Physics_Traits<ShockProblem> {
...
 };
 
-template <> AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<ShockProblem>::ComputePlanckOpacity(const double rho, const double /*Tgas*/) -> double
+template <>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<ShockProblem>::ComputePlanckOpacity(const double rho, const double /*Tgas*/)
+    -> quokka::valarray<double, nGroups_>
 {
-	return (kappa / rho);
+	quokka::valarray<double, nGroups_> kappaPVec{};
+	// kappaPVec.fillin(kappa / rho);
```
BenWibking: Remove commented-out code.

src/RadhydroSimulation.hpp
```diff
@@ -297,7 +297,14 @@ template <typename problem_t> void RadhydroSimulation<problem_t>::defineComponen
 	}
 	// add radiation state variables
 	if constexpr (Physics_Traits<problem_t>::is_radiation_enabled) {
-		std::vector<std::string> radNames = {"radEnergy", "x-RadFlux", "y-RadFlux", "z-RadFlux"};
+		// std::vector<std::string> radNames = {"radEnergy", "x-RadFlux", "y-RadFlux", "z-RadFlux"};
```
BenWibking: Remove commented-out code.

src/RadhydroSimulation.hpp
```diff
@@ -543,9 +550,17 @@ template <typename problem_t> void RadhydroSimulation<problem_t>::computeAfterEv
 	amrex::Real Etot0 = NAN;
 	amrex::Real Etot = NAN;
 	if constexpr (Physics_Traits<problem_t>::is_radiation_enabled) {
-		amrex::Real const Erad0 = initSumCons[RadSystem<problem_t>::radEnergy_index];
+		// amrex::Real const Erad0 = initSumCons[RadSystem<problem_t>::radEnergy_index];
```
BenWibking: Remove commented-out code.

src/RadhydroSimulation.hpp
```diff
@@ -543,9 +550,17 @@ template <typename problem_t> void RadhydroSimulation<problem_t>::computeAfterEv
...
+		amrex::Real Erad0 = 0.;
+		for (int g = 0; g < Physics_Traits<problem_t>::nGroups; ++g) {
+			Erad0 += initSumCons[RadSystem<problem_t>::radEnergy_index + Physics_NumVars::numRadVars * g];
+		}
 		Etot0 = Egas0 + (RadSystem<problem_t>::c_light_ / RadSystem<problem_t>::c_hat_) * Erad0;
-		amrex::Real const Erad = state_new_cc_[0].sum(RadSystem<problem_t>::radEnergy_index) * vol;
+		// amrex::Real const Erad = state_new_cc_[0].sum(RadSystem<problem_t>::radEnergy_index) * vol;
```
BenWibking: Remove commented-out code.

tests/Backtrace.0
BenWibking, line 1: This file should be removed from the PR.

src/planck_integral.hpp
```diff
@@ -0,0 +1,241 @@
...
+#include "AMReX_BLassert.H"
+#include "AMReX_Extension.H"
+#include "AMReX_GpuQualifiers.H"
+#include "AMReX_REAL.H"
+
+#include "valarray.hpp"
+
+using Real = amrex::Real;
+
+static constexpr double PI = 3.14159265358979323846;
```
BenWibking: Although it is not part of the standard, can you use `M_PI` instead (https://www.gnu.org/software/libc/manual/html_node/Mathematical-Constants.html)?

BenWibking: Once we switch to C++20, we can use `std::numbers::pi`: https://en.cppreference.com/w/cpp/numeric/constants

src/radiation_system.hpp
```diff
@@ -26,12 +27,16 @@
...
+#include "planck_integral.hpp"
 #include "simulation.hpp"
 #include "valarray.hpp"
 
 // physical constants in CGS units
-static constexpr double c_light_cgs_ = 2.99792458e10;	      // cgs
-static constexpr double radiation_constant_cgs_ = 7.5646e-15; // cgs
+static constexpr double c_light_cgs_ = C::c_light;	    // cgs
+static constexpr double radiation_constant_cgs_ = C::a_rad; // cgs
+static constexpr double pi = 3.14159265358979323846;
```
BenWibking: Can you use `M_PI` instead?

src/radiation_system.hpp
```diff
@@ -132,19 +150,76 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
...
 };
 
+// CCH: compute radiation energy fractions for each photon group from a Planck function, given nGroups, radBoundaries, and temperature
+template <typename problem_t>
+AMREX_GPU_HOST_DEVICE auto RadSystem<problem_t>::ComputePlanckEnergyFractions(amrex::GpuArray<double, nGroups_ + 1> const &boundaries, amrex::Real temperature)
+    -> quokka::valarray<amrex::Real, nGroups_>
+{
+	quokka::valarray<amrex::Real, nGroups_> radEnergyFractions{};
+	if constexpr (nGroups_ == 1) {
+		// CCH: for future reference, if we want to use a single group as a delta function, say at 13.6 eV, rethink this.
```
BenWibking: As a matter of code style, you could write this as `// TODO(cch): for future reference...`

src/radiation_system.hpp
```diff
@@ -132,19 +150,76 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
...
+		for (int g = 0; g < nGroups_; ++g) {
+			amrex::Real y = integrate_planck_from_0_to_x(boundaries[g + 1] * energy_unit_over_kT);
+			radEnergyFractions[g] = y - previous;
+			previous = y;
+		}
+		// A hack to avoid zeros in radEnergyFractions
+		// TODO(CCH): find a better way to do this
+		for (int g = 0; g < nGroups_; ++g) {
+			if (radEnergyFractions[g] < 1e-14) {
+				radEnergyFractions[g] = 1e-14;
```
BenWibking: Can you change this to be a variable instead of a hard-coded value of `1e-14`? A default value for this variable of 1e-14 seems fine to me, but it should not be hard-coded if we can avoid it.

src/radiation_system.hpp
```diff
@@ -611,264 +691,284 @@ void RadSystem<problem_t>::ComputeFluxes(array_t &x1Flux_in, array_t &x1FluxDiff
...
-	return 0.0;
+	quokka::valarray<double, nGroups_> kappaFVec{};
+	for (int g = 0; g < nGroups_; ++g) {
+		kappaFVec[g] = NAN;
+	}
+	return kappaFVec;
 }
 
-template <typename problem_t> AMREX_GPU_HOST_DEVICE auto RadSystem<problem_t>::ComputeRosselandOpacity(const double /*rho*/, const double /*Tgas*/) -> double
+// TODO(CCH): Derivative of Planck/energy-mean opacity with respect to gas temperature. In the future, they should be defined separately.
```
BenWibking: I don't understand what you mean here. Can you explain this comment?

src/radiation_system.hpp
```diff
@@ -917,189 +1022,244 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
+		quokka::valarray<double, nGroups_> kappaEVec{};
+		quokka::valarray<double, nGroups_> kappaFVec{};
 
-		if constexpr (gamma_ != 1.0) {
+		for (int g = 0; g < nGroups_; ++g) {
+			EradVec_guess[g] = NAN;
+		}
+
+		// if constexpr (gamma_ != 1.0) {
+		if (gamma_ != 1.0) {
```
BenWibking: Does this not work with `constexpr` anymore?

src/radiation_system.hpp
```diff
@@ -917,189 +1022,244 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
 				AMREX_ASSERT(T_gas >= 0.);
 
 				// compute opacity, emissivity
-				kappa = RadSystem<problem_t>::ComputePlanckOpacity(rho, T_gas);
-				AMREX_ASSERT(kappa >= 0.);
 				fourPiB = chat * a_rad * std::pow(T_gas, 4);
+				auto B = fourPiB / (4.0 * pi);
+				kappaPVec = ComputePlanckOpacity(rho, T_gas);
+				kappaFVec = ComputeFluxMeanOpacity(rho, T_gas);
+				kappaEVec = kappaFVec; // TODO(CCH): later, define ComputeEnergyMeanOpacity
```
BenWibking: Can you explain what you mean here?

src/radiation_system.hpp
```diff
@@ -917,189 +1022,244 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
-				dFR_dErad = 1.0 + dt * ((rho * kappa) * chat);
-				eta = -dFR_dEgas / dFG_dEgas;
-				eta = (eta > 0.0) ? eta : 0.0;
-
-				deltaErad = -(F_R + eta * F_G) / (dFR_dErad + eta * dFG_dErad);
-				deltaEgas = -(F_G + dFG_dErad * deltaErad) / dFG_dEgas;
+				// prepare to compute Jacobian elements
+				const double c_v = quokka::EOS<problem_t>::ComputeEintTempDerivative(rho, T_gas, massScalars);
+				dRtot_dErad = -dt * rho * kappaEVec * chat;
+				// TODO(CCH): fix this: dB_dTgas * radEnergyFractions is inaccurate. Define ComputePlanckEnergyFractionsTempDerivative
```
BenWibking: Can you explain this?

src/radiation_system.hpp
```diff
@@ -917,189 +1022,244 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
+
+				// compute energyLossFractions
+				quokka::valarray<double, nGroups_> energyLossFractions{};
+				for (int g = 0; g < nGroups_; ++g) {
+					Frad_t1[0] = consNew(i, j, k, x1RadFlux_index + numRadVars_ * g);
+					Frad_t1[1] = consNew(i, j, k, x2RadFlux_index + numRadVars_ * g);
+					Frad_t1[2] = consNew(i, j, k, x3RadFlux_index + numRadVars_ * g);
+					energyLossFractions[g] = kappaFVec[g] * (x1GasMom1 * Frad_t1[0] + x2GasMom1 * Frad_t1[1] + x3GasMom1 * Frad_t1[2]);
+				}
+				// TODO(CCH): tricky here. Have to avoid negavie radiation energy.
```
BenWibking: Can you explain the negative energy problem you encountered?

src/radiation_system.hpp
```diff
@@ -917,189 +1022,244 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
+			consNew(i, j, k, gasInternalEnergy_index) += Egas_guess - Egas0; // must compute difference
 		} else {
-			amrex::ignore_unused(Erad_guess);
+			amrex::ignore_unused(EradVec_guess);
 			amrex::ignore_unused(Egas_guess);
 		} // endif gamma != 1.0
 	});
 }
 
+// CCH: this function ComputeSourceTermsExplicit is never used.
```
BenWibking, line 1260: Yes, we can get rid of it in another PR.

src/radiation_system.hpp
```diff
@@ -917,189 +1022,244 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
-				// do not subtract radiation work in new radiation energy
-				dErad_work = 0.;
+				// else, do not subtract radiation work in new radiation energy
+				for (int g = 0; g < nGroups_; ++g) {
+					consNew(i, j, k, radEnergy_index + numRadVars_ * g) = EradVec_guess[g];
+				}
 			}
 
-			// compute difference between new and old internal energy
-			amrex::Real const dEint = Egas_guess - Egas0;
```
BenWibking, line 1090: Is there a reason to get rid of this variable?

src/radiation_system.hpp
```diff
@@ -917,189 +1022,244 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
+				const double c_v = quokka::EOS<problem_t>::ComputeEintTempDerivative(rho, T_gas, massScalars);
+				dRtot_dErad = -dt * rho * kappaEVec * chat;
+				// TODO(CCH): fix this: dB_dTgas * radEnergyFractions is inaccurate. Define ComputePlanckEnergyFractionsTempDerivative
+				dRvec_dEgas = dt * rho / c_v *
+					      (4 * pi * kappaPVec * dB_dTgas * radEnergyFractions + dkappaP_dTgas * fourPiB * radEnergyFractions -
+					       chat * dkappaE_dTgas * EradVec_guess);
+				dRtot_dEgas = sum(dRvec_dEgas);
+				// kappaP = RadSystem<problem_t>::ComputePlanckOpacity(rho, T_gas);
+				// dRtot_dEgas = dt * rho / c_v * (4 * pi * kappaP * dB_dTgas + dkappa_dTgas * fourPiB - chat * sum(dkappaVec_dTgas *
+				// EradVec_guess));
```
BenWibking: Can you remove the commented-out code?

src/radiation_system.hpp
```diff
@@ -917,189 +1022,244 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
 
 				// compute residuals
-				rhs = dt * (rho * kappa) * (fourPiB - chat * Erad_guess);
-				F_G = (Egas_guess - Egas0) + ((c / chat) * rhs);
-				F_R = (Erad_guess - Erad0) - (rhs + Src);
+				quokka::valarray<amrex::Real, nGroups_> radEnergyFractions{};
+				// TODO(CCH): fix this: if constexpr (nGroups_ > 1)
+				// if constexpr (nGroups_ > 1) {
+				// ../src/radiation_system.hpp(1089): error: An extended __device__ lambda cannot first-capture variable in constexpr-if context
+				// detected during: ...
```
BenWibking: This is an annoying limitation of the CUDA compiler. You can use this variable before the `ParallelFor` and then it will work.

BenWibking (changes requested):
There are a few style changes. In particular, we have tried to avoid commented-out code in Quokka in general. Unless there is a very good reason to keep it, it is usually available in the git history, or it never worked and was never part of the code, in which case it just confuses new developers/users.  We have also avoiding adding initials or developer names in comments, since this information is always available via `git blame`, although it's up to you if you want to keep that.

The main thing I am concerned about are the partially-explained comments in the radiation-matter update code that indicate TODO items. Can you explain those further? I think it's important to know if there are issues with the current implementation, or if you are just keeping track of where future improvements could be made.

src/RadForce/test_radiation_force.cpp
chongchonghe: I removed it and all other similar things.

src/RadForce/test_radiation_force.cpp
chongchonghe: Done

src/radiation_system.hpp
chongchonghe: Never mind this. It shouldn't be a TODO and I removed it.

src/radiation_system.hpp
chongchonghe: Not really. I like to keep the code concise and avoid unnecessary variables, but I also appreciate extra variables to better explain the physics. I put dEint back.

src/radiation_system.hpp
chongchonghe: Done.

src/radiation_system.hpp
chongchonghe: Good. Actually I realised this before but forgot to put the constexpr back.

chongchonghe:
> I'll try to review this in detail over the weekend.
> 
> Aside from the code itself, I am hesitant to approve this without quantative tests, although they could go in future PRs. What is the plan for quantitative test problems that check the group-by-group solution?

Let's discuss this on our next Quokka meeting.

BenWibking:
@chongchonghe @markkrumholz For mixed-frame, do the group boundaries need to be Doppler-shifted before doing the radiation-matter energy exchange? I think they do, but maybe I'm wrong.

BenWibking:
> > I'll try to review this in detail over the weekend.
> > Aside from the code itself, I am hesitant to approve this without quantative tests, although they could go in future PRs. What is the plan for quantitative test problems that check the group-by-group solution?
> 
> Let's discuss this on our next Quokka meeting.

For reference, see section 4.1: https://arxiv.org/abs/1207.3845.

markkrumholz:
Just a few quick comments on the physics here. First, we do not need to Doppler shift the group boundaries. In the mixed-frame formulation, all frequencies are written on the lab frame, and I do think we want to keep using mixed-frame, for reasons I’ll get to in a moment. 

Second, there are terms that involve shifts between energy bins due to work being done on the photons by matter, which we have not yet implemented. These are small for our problems, but should be included. I am unsure if it is with going the CASTRO route of treating them as an advection, or whether we should group them with the source terms. CASTRO does the former, but they’re doing problems with v/c ~ 0.1, whereas we are at ~10^-3. 

Third, I think we want to stick with mixed-frame because we are not doing line transfer. The only sharp features we care about in our opacities are ionization edges, and the mixed-frame formulation does fail near them. However, this is a tiny part of our frequency bin, so the error is small. Quantitatively, if for example we have a bin that goes from E_0 = 13.6 to E_1 = 20 eV, so Delta E = 6.4 eV then the error we make by applying mixed-frame is of order (v/c) E_0 / Delta E ~ 10^-3. This is smaller than the typical conservation error you would get in a comoving frame formulation.

BenWibking:
> Third, I think we want to stick with mixed-frame because we are not doing line transfer. The only sharp features we care about in our opacities are ionization edges, and the mixed-frame formulation does fail near them. However, this is a tiny part of our frequency bin, so the error is small. Quantitatively, if for example we have a bin that goes from E_0 = 13.6 to E_1 = 20 eV, so Delta E = 6.4 eV then the error we make by applying mixed-frame is of order (v/c) E_0 / Delta E ~ 10^-3. This is smaller than the typical conservation error you would get in a comoving frame formulation.

I would like to do line transfer in the future. However, maybe we can just add a runtime check that the bins are sufficiently coarse near ionization edges in the current implementation.

BenWibking:
@chongchonghe Can you take a look at the remaining unresolved comments?

src/RadForce/test_radiation_force.cpp
chongchonghe: All resolved.

src/RadSuOlson/test_radiation_SuOlson.cpp
chongchonghe: You are right. I changed it back to 1 group. In a seperate PR, I'll add the two-gropu SuOlson with Picket-Fence test following Sec 4.1.2 of the CASTRO paper.

src/RadForce/test_radiation_force.cpp
chongchonghe: Resolved this and all similar comments.

src/radiation_system.hpp
chongchonghe: To calculate d B_i / d Tgas accurately, I need to consider that both B_i and radFractions_i is a function of Tgas. Currently I am using an approximation that d B_i / d Tgas = radFractions_i * d B / d Tgas, which is exact in 1-group case and mostly accurate in multigropu case. The error caused by this is below 0.2% in the RadTube test. I'll do it rigorously in the next PR.

src/radiation_system.hpp
chongchonghe: There are kappa_P, kappa_F, and kappa_E. Currently I'm assuming kappa_E = kappa_F. In the future we want to release this assumption.

src/radiation_system.hpp
chongchonghe: `constexpr` works now. I've added it back.

src/radiation_system.hpp
chongchonghe: Similar to my other comments about d B_i / d T_gas, this is something I plan to do next. Mark and I discussed about this, and our plan is to define a class to allow the user to specify d kappa_P / d Tgas. We will also provide a convenient function to compute this from the Integral of Planck function if the user provide a tabulated kappa(nu). I removed this TODO and kept the other one about d B_i / d T_gas. 

I removed all TODOs except a few of them which I find it necessary to keep.

src/radiation_system.hpp
chongchonghe: Done.

src/radiation_system.hpp
chongchonghe: Done.

chongchonghe:
> There are a few style changes. In particular, we have tried to avoid commented-out code in Quokka in general. Unless there is a very good reason to keep it, it is usually available in the git history, or it never worked and was never part of the code, in which case it just confuses new developers/users. We have also avoiding adding initials or developer names in comments, since this information is always available via `git blame`, although it's up to you if you want to keep that.
> 
> The main thing I am concerned about are the partially-explained comments in the radiation-matter update code that indicate TODO items. Can you explain those further? I think it's important to know if there are issues with the current implementation, or if you are just keeping track of where future improvements could be made.

I removed all unnecessary TODOs and kept only a few necessary ones. They are not issues. They are planned work that will release some assumptions/approximates.

chongchonghe:
> I'll try to review this in detail over the weekend.
> 
> Aside from the code itself, I am hesitant to approve this without quantative tests, although they could go in future PRs. What is the plan for quantitative test problems that check the group-by-group solution?

There is one quantitative test in the current branch, the RadTube test. I set the boundary between two photon groups at near the temperature of the tube. In the new figure, radiation_pressure_tube_energy_density.pdf, you'll see that the exact solution for Erad_1 and Erad_2 are on top of the numerical curve. The proportions of the two groups changes with the temperature which changes from left to right. That's why you see it changed from group1-dominating to group2-dominating from left to right. The exact solution is calculated from the initial/equilibrium solution of Erad and Tgas and stored in the cpp file of the problem. The proportions are calculated using a Python script. 

In a seperate PR, I'll add the multigroup SuOlson picket-fence test (Su & Olson 1999, Case C, see also http://arxiv.org/abs/1207.3845). I may also add a few other tests, but they are not planed.

BenWibking:
The GPU build failed due to the first-capture issue: https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=6772&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=d735d76d-5905-5d5d-cb1c-7198326fbfa2

BenWibking:
Could you add a somewhat more detailed PR description? Since we don't have real documentation, our best existing documentation has been PR descriptions. About 1 paragraph in length would be good, along with a description of any test problems you've added, and any limitations or assumptions that potential users of the multigroup code should be aware of.

Once this is done and the CUDA build works, I'm happy to approve and merge.

extern/Microphysics
BenWibking, line 1: It looks like your submodules are out of date. Can you update your submodules?

chongchonghe: I don't see any outdated submodules. I believe they are up to date.

BenWibking:
Can you update the PR description?

src/RadPulse/test_radiation_pulse.cpp
```diff
@@ -63,22 +65,33 @@ auto compute_exact_Trad(const double x, const double t) -> double
...
+    -> quokka::valarray<double, nGroups_>
 {
+	quokka::valarray<double, nGroups_> opacity_deriv{};
+	double opacity_deriv_scalar = 0.;
 	if (Tgas > 1.0) {
-		return (kappa0 / rho) * (3.0 / T0) * std::pow(Tgas / T0, 2);
+		opacity_deriv_scalar = (kappa0 / rho) * (3.0 / T0) * std::pow(Tgas / T0, 2);
 	}
-	return 0.;
+	opacity_deriv.fillin(opacity_deriv_scalar);
```
BenWibking, line 93: Does `fillin` work on GPU?

src/radiation_system.hpp
```diff
@@ -611,264 +689,280 @@ void RadSystem<problem_t>::ComputeFluxes(array_t &x1Flux_in, array_t &x1FluxDiff
...
 }
 
-template <typename problem_t> AMREX_GPU_HOST_DEVICE auto RadSystem<problem_t>::ComputeRosselandOpacity(const double /*rho*/, const double /*Tgas*/) -> double
+template <typename problem_t>
+AMREX_GPU_HOST_DEVICE auto RadSystem<problem_t>::ComputePlanckOpacityTempDerivative(const double /* rho */, const double /* Tgas */)
+    -> quokka::valarray<double, nGroups_>
 {
-	return NAN;
+	quokka::valarray<double, nGroups_> kappa{};
+	kappa.fillin(0.0);
```
BenWibking, line 964: Does `fillin` work on GPU?

src/RadPulse/test_radiation_pulse.cpp
chongchonghe: It works on variables that are defined locally or hand-written number like 0.0. It doesn't work on globally defined constexpr like kappa0.

chongchonghe:
> Can you update the PR description?

I'll do it this noon when I get to work. What is exactly a PR description? Is it the first comments I made in the PR? I just edit that?

BenWibking:
> > Can you update the PR description?
> 
> I'll do it this noon when I get to work. What is exactly a PR description? Is it the first comments I made in the PR? I just edit that?

Yes, you can edit it.

chongchonghe:
> > > Can you update the PR description?
> > 
> > 
> > I'll do it this noon when I get to work. What is exactly a PR description? Is it the first comments I made in the PR? I just edit that?
> 
> Yes, you can edit it.

Done.

BenWibking (approved):
Looks good to me!

PR #391, opened 2023-09-20
Valarray extended

chongchonghe:


.gitignore
BenWibking, line 1: Can you remove the changes to this file? We could change this, but I don't think it makes sense to do so in this PR.

src/valarray.hpp
```diff
@@ -153,4 +191,101 @@ template <typename T, int d> AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto min(q
...
+// // do slicing with valarray, e.g. array_positive_terms = array[array > 0.]. Not working
+// template <typename T, int d>
+// AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto operator[](quokka::valarray<T, d> const &v, quokka::valarray<bool, d> const &mask) -> quokka::valarray<T, d>
+// {
+//   quokka::valarray<T, d> sliced;
+//   for (size_t i = 0; i < v.size(); ++i) {
+//     sliced[i] = mask[i] ? v[i] : 0.;
+//   }
+//   return sliced;
+// }
```
BenWibking: Can you remove the commented-out code? This can be kept on a separate branch until it's working and you want to merge it. In general, we want to avoid commented-out code in the main branch (there are some places where it exists and should be removed, but we haven't gotten around to it).

src/valarray.hpp
```diff
@@ -153,4 +191,101 @@ template <typename T, int d> AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto min(q
...
+  quokka::valarray<bool, d> comp;
+  for (size_t i = 0; i < a.size(); ++i) {
+    comp[i] = a[i] < scalar;
+  }
+  return comp;
+}
+
+// define fillin(valarray) for quokka::valarray
+template <typename T, int d> 
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void fillin(quokka::valarray<T, d> &v, T const &scalar)
```
BenWibking: Could you make this a static member function of `quokka::valarray`?

BenWibking (changes requested):

chongchonghe:
All modifications are done. Do I need to create a new PR? I don't see the option to. I suppose this PR is automatically updated.

src/valarray.hpp
BenWibking: You marked it `static`, but it is not a member function of `valarray`. Can you change this to `void valarray::fillin(...)`?

BenWibking:
> All modifications are done. Do I need to create a new PR? I don't see the option to. I suppose this PR is automatically updated.

It updates automatically.

See my new comment about `fillin`.

src/valarray.hpp
chongchonghe: Oh, that's what you mean. I was wondering about that since fillin apparently change the content of the array and shouldn't be 'static'.

chongchonghe: Not quite sure how. I tried `AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void valarray::fillin...` and `AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void quokka::valarray::fillin` but it didnt' work. It complained valarray is not defined, or quokka::valarray used without template arguments. I renamed fillin to valarray_fillin for now to avoid conflicting names. Let me know if you have more suggestions.

BenWibking:
Inside `namespace quokka {`, add:
```
// define fillin(valarray) for quokka::valarray
template <typename T, int d> 
AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void valarray<T, d>::fillin(quokka::valarray<T, d> &v, T const &scalar)
{
  for (size_t i = 0; i < v.size(); ++i) {
    v[i] = scalar;
  }
}
```

src/valarray.hpp
BenWibking: You have to define it within the `quokka` namespace. Member functions must also specify the template arguments like this `void valarray<T,d>::fillin(...)`. See the example code below.

BenWibking: There are some more examples here: https://learn.microsoft.com/en-us/cpp/cpp/class-templates?view=msvc-170#member-functions-of-class-templates and here: https://isocpp.org/wiki/faq/templates#class-templates

chongchonghe: Thanks. I used a different approach. Now fillin is a build in function of valarray. Use it like array.fill(value). Same thing with hasnan. Fully tested and working properly. How do you like it?

BenWibking (approved):

PR #392, opened 2023-09-20
update the flowchart on the website

chongchonghe:
1. a prettier look
2. add routines of the radiation module


docs/flowchart-v2.jpg
BenWibking, line 1: Is it possible to put an SVG file in the documentation, rather than a JPEG?

chongchonghe: I made this with Keynote. Not quite straight forward to convert it to SVG. I attached a PDF which is much better quality and you can select and copy texts from it (I have to zip it otherwise it will be ignored by git).

BenWibking: Ah, ok. I thought you might have used Inkscape or something like that. You should be able to add a PDF without zipping it: `git add --force figure.pdf`.

BenWibking (approved):

PR #395, opened 2023-09-22
fix src/CMakeLists for macOS with Clang

chongchonghe:
It's called Clang instead of AppleClang. I have to change this in order to make the compiler skip stdc++fs on my Mac with apple silicon


BenWibking:
That is an interesting problem. It works on my Apple Silicon laptop without this change.

The Apple-provided Clang is called AppleClang: https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html

Do you have another Clang compiler installed on your system?

psharda:
We should also make sure this doesn't break CMake on gadi and setonix in addition to avatar (which will be checked by the tests anyway)

BenWibking:
As you can see from the `CMake (macOS)` build, this breaks things that were working previously on macOS: https://github.com/quokka-astro/quokka/actions/runs/6271473074/job/17031116749?pr=395.

This appears to be an issue with how you've configured your computer.

chongchonghe:
> As you can see from the `CMake (macOS)` build, this breaks things that were working previously on macOS: https://github.com/quokka-astro/quokka/actions/runs/6271473074/job/17031116749?pr=395.
> 
> This appears to be an issue with how you've configured your computer.

OK, if this change works on my computer but not on the test platform, then it is indeed due to my configuration. However, we want to be as inclusive as possible, right? So, I changed it back to AppleClang but added another criterial, AND (NOT CMAKE_SYSTEM_NAME MATCHES "Darwin"). This should work on all macOS platforms, and it won't fail the linux test.

BenWibking:
> OK, if this change works on my computer but not on the test platform, then it is indeed due to my configuration. However, we want to be as inclusive as possible, right? So, I changed it back to AppleClang but added another criterial, AND (NOT CMAKE_SYSTEM_NAME MATCHES "Darwin"). This should work on all macOS platforms, and it won't fail the linux test.

I'm not sure if this will work if someone is compiling with gcc 8 on macOS. Can you test that and see if it works?

src/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ endif()
 find_package(HDF5 REQUIRED)
 
 # std::filesystem (required for GCC < 9)
-if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+if ((NOT CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang") AND (NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
```
BenWibking: Unfortunately, this won't work. On systems that use the GNU C++ standard library version < 9 but where the user is building with Clang, the code will not be able to use `std::filesystem` without linking with `stdc++fs`. This will affect all users of RHEL 8, which is currently the Linux version used on Gadi and many other clusters, including the Avatar GPU node.

chongchonghe:
The new criterial is to not include stdc++fs if the system is macOS AND the compiler is either AppleClang or Clang. Can you think of any case this would fail? This would work with gcc on macOS as well.

BenWibking:
> The new criterial is to not include stdc++fs if the system is macOS AND the compiler is either AppleClang or Clang. Can you think of any case this would fail? This would work with gcc on macOS as well.

I think that will work. I can't think of a case offhand. Let's make sure all the tests pass and then I will approve.

BenWibking (approved):

PR #398, opened 2023-09-24
Update README

psharda:
Updated README because `AMREX_GPUS_PER_NODE` is not required anymore, and is not a valid keyword now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ By default, Quokka compiles itself to run only on CPUs. Quokka can run on either
 ### NVIDIA GPUs
 If you want to run on NVIDIA GPUs, re-build Quokka as shown below. (*CUDA >= 11.7 is required. Quokka is only supported on Volta V100 GPUs or newer models. Your MPI library **must** support CUDA-aware MPI.*)
 ```
-cmake .. -DCMAKE_BUILD_TYPE=Release -DAMReX_GPU_BACKEND=CUDA -DAMREX_GPUS_PER_NODE=N -G Ninja
+cmake .. -DCMAKE_BUILD_TYPE=Release -DAMReX_GPU_BACKEND=CUDA -DAMReX_SPACEDIM=3 -G Ninja
 ninja -j6
 ```
 where $N$ is the number of GPUs available per compute node.
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ If you want to run on NVIDIA GPUs, re-build Quokka as shown below. (*CUDA >= 11.
 cmake .. -DCMAKE_BUILD_TYPE=Release -DAMReX_GPU_BACKEND=CUDA -DAMReX_SPACEDIM=3 -G Ninja
 ninja -j6
 ```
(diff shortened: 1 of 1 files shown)


psharda:
@BenWibking we don't really need to run any tests on this PR, except for the Build and Deploy Docs one maybe? But this test is skipped by default.

README.md
```diff
@@ -104,7 +104,7 @@ By default, Quokka compiles itself to run only on CPUs. Quokka can run on either
 ### NVIDIA GPUs
 If you want to run on NVIDIA GPUs, re-build Quokka as shown below. (*CUDA >= 11.7 is required. Quokka is only supported on Volta V100 GPUs or newer models. Your MPI library **must** support CUDA-aware MPI.*)
 ```
-cmake .. -DCMAKE_BUILD_TYPE=Release -DAMReX_GPU_BACKEND=CUDA -DAMREX_GPUS_PER_NODE=N -G Ninja
+cmake .. -DCMAKE_BUILD_TYPE=Release -DAMReX_GPU_BACKEND=CUDA -DAMReX_SPACEDIM=3 -G Ninja
 ninja -j6
 ```
 where $N$ is the number of GPUs available per compute node.
```
BenWibking: We can get rid of this phrase.

BenWibking: We can also get rid of these sentences: "It is necessary to use `-DAMREX_GPUS_PER_NODE` to specify the number of GPUs per compute node. Without this, performance will be very poor."

BenWibking (dismissed):

psharda:
Oops, maybe I deleted more than what you said?

BenWibking:
> Oops, maybe I deleted more than what you said?

The rest of the paragraph about the GPU bindings is still true and important, so we should keep that part. Just the part about manually setting the number of GPUs per node is no longer true.

README.md
```diff
@@ -104,12 +104,9 @@ By default, Quokka compiles itself to run only on CPUs. Quokka can run on either
 ### NVIDIA GPUs
 If you want to run on NVIDIA GPUs, re-build Quokka as shown below. (*CUDA >= 11.7 is required. Quokka is only supported on Volta V100 GPUs or newer models. Your MPI library **must** support CUDA-aware MPI.*)
 ```
-cmake .. -DCMAKE_BUILD_TYPE=Release -DAMReX_GPU_BACKEND=CUDA -DAMREX_GPUS_PER_NODE=N -G Ninja
+cmake .. -DCMAKE_BUILD_TYPE=Release -DAMReX_GPU_BACKEND=CUDA -DAMReX_SPACEDIM=3 -G Ninja
 ninja -j6
 ```
-where $N$ is the number of GPUs available per compute node.
-
-**It is necessary to use `-DAMREX_GPUS_PER_NODE` to specify the number of GPUs per compute node. Without this, performance will be very poor. All GPUs on a node must be visible from each MPI rank on the node for efficient GPU-aware MPI communication to take place via CUDA IPC.** When using the SLURM job scheduler, this means that `--gpu-bind` should be set to `none`.
```
BenWibking, line 112: We should keep this text:
```
All GPUs on a node must be visible from each MPI rank on the node for efficient GPU-aware MPI communication to take place via CUDA IPC.** When using the SLURM job scheduler, this means that `--gpu-bind` should be set to `none`.
```

BenWibking (approved):

Issue #393, opened 2023-09-21
Add a CPU job script for Setonix


psharda:
Will add this script below in a PR (opening this issue now to remember later):

```
#!/bin/bash --login

# SLURM directives
#
# Here we specify to SLURM we want an OpenMP job with 32 threads
# a wall-clock time limit of one hour.
#
# Replace [your-project] with the appropriate project name
# following --account (e.g., --account=project123).

#SBATCH --account=pawsey0807
#SBATCH --partition=work
#SBATCH --nodes=2
#SBATCH --ntasks-per-node=128
#SBATCH --cpus-per-task=1
#SBATCH --mem=58880M      #Indicate the amount of memory per node when asking for share resources
#SBATCH --time=01:00:00

# ---
# Load here the needed modules

# ---
# OpenMP settings
export OMP_NUM_THREADS=$SLURM_CPUS_PER_TASK   #To define the number of threads, in this case will be 32
export OMP_PLACES=cores     #To bind threads to cores
export OMP_PROC_BIND=close  #To bind (fix) threads (allocating them as close as possible). This option works together with the "places" indicated above, then: allocates threads in closest cores.

# ---
# Temporal workaround for avoiding Slingshot issues on shared nodes:
export FI_CXI_DEFAULT_VNI=$(od -vAn -N4 -tu < /dev/urandom)
# Run the desired code:
srun -N $SLURM_JOB_NUM_NODES -n $SLURM_NTASKS -c $OMP_NUM_THREADS -m block:block:block ./src/PopIII/popiii PopIII.in
```

psharda:
Should also add a setonix.profile for Setonix CPUs:

```
#!/bin/bash

source /opt/cray/pe/cpe/23.03/restore_lmod_system_defaults.sh

module load cmake/3.24.3
module load cray-mpich
module load cce/15.0.1
module load cray-hdf5
module load cray-python/3.9.13.1

# GPU-aware MPI
export MPICH_GPU_SUPPORT_ENABLED=0

# allow CMake to find Ascent
export Ascent_DIR=/software/projects/pawsey0807/bwibking/ascent_06082023/install/ascent-develop/lib/cmake/ascent/

# compiler environment hints
export CC=$(which cc)
export CXX=$(which CC)
export FC=$(which ftn)
export CFLAGS="-I${ROCM_PATH}/include"
export CXXFLAGS="-I${ROCM_PATH}/include"

```

psharda:
Closes by #399

PR #399, opened 2023-09-24
Add sample Setonix scripts

psharda:
This PR adds 1.) a sample Setonix CPU script, and 2.) separates setonix.profiles for CPUs and GPUs


psharda:
Closes #393

scripts/setonix-cpu.profile
```diff
@@ -0,0 +1,23 @@
...
+module load cray-mpich
+module load cce/15.0.1
+module load cray-hdf5
+module load cray-python/3.9.13.1
+
+# GPU-aware MPI
+export MPICH_GPU_SUPPORT_ENABLED=0
+
+# allow CMake to find Ascent
+export Ascent_DIR=/software/projects/pawsey0807/bwibking/ascent_06082023/install/ascent-develop/lib/cmake/ascent/
```
BenWibking: This Ascent build requires GPU support, so we should get rid of it for the CPU build.

scripts/cpu-setonix.sh
```diff
@@ -0,0 +1,32 @@
...
+#SBATCH --ntasks-per-node=128
+#SBATCH --cpus-per-task=1
+#SBATCH --mem=58880M      #Indicate the amount of memory per node when asking for share resources
+#SBATCH --time=01:00:00
+
+# ---
+# Load here the needed modules
+
+# ---
+# OpenMP settings
```
BenWibking: We don't use OpenMP, so we should get rid of all of the OMP stuff to avoid confusion.

scripts/cpu-setonix.sh
```diff
@@ -0,0 +1,32 @@
...
+export OMP_NUM_THREADS=$SLURM_CPUS_PER_TASK   #To define the number of threads, in this case will be 32
+export OMP_PLACES=cores     #To bind threads to cores
+export OMP_PROC_BIND=close  #To bind (fix) threads (allocating them as close as possible). This option works together with the "places" indicated above, then: allocates threads in closest cores.
+
+# ---
+# Temporal workaround for avoiding Slingshot issues on shared nodes:
+export FI_CXI_DEFAULT_VNI=$(od -vAn -N4 -tu < /dev/urandom)
```
BenWibking, line 19: We should always request a full node, so this should not be necessary.

scripts/cpu-setonix.sh
```diff
@@ -0,0 +1,32 @@
...
+# Run the desired code:
+srun -N $SLURM_JOB_NUM_NODES -n $SLURM_NTASKS -c $OMP_NUM_THREADS -m block:block:block ./src/PopIII/popiii ../tests/PopIII.in
```
BenWibking: I don't think any of the extra srun options are actually necessary. Specifying them using the `SBATCH` commands above should be sufficient.

scripts/cpu-setonix.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# Replace [your-project] with the appropriate project name
+# following --account (e.g., --account=project123).
+
+#SBATCH --account=pawsey0807
+#SBATCH --partition=work
+#SBATCH --nodes=2
+#SBATCH --ntasks-per-node=128
+#SBATCH --cpus-per-task=1
+#SBATCH --mem=58880M      #Indicate the amount of memory per node when asking for share resources
```
BenWibking: We should just request exclusive use of the node to avoid needing to do this.

BenWibking (dismissed):

scripts/cpu-setonix.sh
psharda: sure?

scripts/cpu-setonix.sh
psharda: how do you do that

scripts/cpu-setonix.sh
BenWibking: You could test it to be sure. But it probably won't hurt to leave it.

scripts/cpu-setonix.sh
BenWibking: SLURM cheat sheet: https://slurm.schedmd.com/pdfs/summary.pdf

BenWibking: `--exclusive`

scripts/setonix-cpu.profile
```diff
@@ -0,0 +1,20 @@
...
+module load cray-python/3.9.13.1
+
+# GPU-aware MPI
+export MPICH_GPU_SUPPORT_ENABLED=0
+
+# compiler environment hints
+export CC=$(which cc)
+export CXX=$(which CC)
+export FC=$(which ftn)
+export CFLAGS="-I${ROCM_PATH}/include"
```
psharda: Is this correct? Should we include `ROCM_PATH` here

scripts/cpu-setonix.sh
```diff
@@ -10,10 +10,7 @@
 
 #SBATCH --account=pawsey0807
 #SBATCH --partition=work
-#SBATCH --nodes=2
-#SBATCH --ntasks-per-node=128
-#SBATCH --cpus-per-task=1
```
BenWibking: I think it's still necessary to specify the three lines above.

scripts/setonix-cpu.profile
BenWibking: Not needed for CPU.

scripts/cpu-setonix.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+# Here we specify to SLURM we want an OpenMP job with 32 threads
+# a wall-clock time limit of one hour.
+#
+# Replace [your-project] with the appropriate project name
+# following --account (e.g., --account=project123).
+
+#SBATCH --account=pawsey0807
+#SBATCH --partition=work
+#SBATCH --exclusive=psharda
```
BenWibking: If someone else tries to use it, I don't think it will work unless they change this to their username. But it should also work without, i.e. just `--exclusive`.

scripts/cpu-setonix.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash --login
+
+# SLURM directives
+#
+# Here we specify to SLURM we want an OpenMP job with 32 threads
```
BenWibking: For the sake of avoiding confusion, we should probably get rid of the reference to "an OpenMP job with 32 threads", since Quokka doesn't use OpenMP.

BenWibking (approved):

PR #405, opened 2023-09-27
Renormalize mass scalars post enforcing floor

psharda:
If mass scalars are negative, enforce a floor value. Then renormalize to ensure mass conservation to machine precision. Without this renormalization, mass fractions are not conserved to machine precision in cells where chemistry is not carried out (e.g., in the low density background in a PopIII simulation).


psharda:
Also ensured proper indentation.

BenWibking:
I re-enabled the autoformatting bot. I thought it was buggy, but it seems to work now. If you push again, it should auto-format.

BenWibking:
Can you ask Mark to review this? I am stretched pretty thin at the moment.

markkrumholz:
Will try to get to it, but the NCMAS proposal is due Tuesday and I’m currently only working half time due to school holiday child care. It may not be until after NCMAS goes in. --Mark KrumholzTyped on a tiny keyboardOn Sep 28, 2023, at 6:27 AM, Ben Wibking ***@***.***> wrote:﻿
Can you ask Mark to review this? I am stretched pretty thin at the moment.

markkrumholz (dismissed):

src/hydro_system.hpp
```diff
@@ -744,12 +744,20 @@ void HydroSystem<problem_t>::EnforceLimits(amrex::Real const densityFloor, amrex
...
+					state[bx](i, j, k, scalar0_index + idx) = small_x * rho;
+				}
+				// get sum to renormalize
+				sp_sum += state[bx](i, j, k, scalar0_index + idx);
+			}
 
+			sp_sum /= rho; //get mass fractions
+			for (int idx = 0; idx < nmscalars_; ++idx) {
+				// renormalize			
+				state[bx](i, j, k, scalar0_index + idx) /= sp_sum;
```
BenWibking: The test is failing due to dividing by zero. My guess is that `sp_sum` is zero for some problem where the mass scalars are initialized to zero. You could add a conditional to only renormalize when `sp_sum` is nonzero. Otherwise `rho` is zero for some reason and the division-by-zero is comging from line 756, but I'm less sure how that would happen.

BenWibking (dismissed):

BenWibking:
Looks good, let's make sure tests pass.

BenWibking (approved):

PR #406, opened 2023-09-28
Save gravitational potential gpot as a derived var in pltfiles

### LINKED ISSUE #357: Save gravitational potential in plotfiles
We should have the ability to calculate gravitational potential and store it as a derived field in quokka snapshots.

psharda:
Will fix #357 (when it actually works).

Testing it out on the SphericalCollapse test. The code compiles but gives a segfault after writing chk00000

I'm sure there are bugs in what I did.


psharda:
ugh, the pre-commit bot is making more changes than I made. Likely because its finding other indentation issues that weren't caught when it was turned off

BenWibking:
> ugh, the pre-commit bot is making more changes than I made. Likely because its finding other indentation issues that weren't caught when it was turned off

Ahh, sorry. Can you create a PR that changes like 1 line and nothing else? Then it should fix the spacing in that PR rather than this one, and we can merge that PR in order to fix all the formatting.

psharda:
#407

> > ugh, the pre-commit bot is making more changes than I made. Likely because its finding other indentation issues that weren't caught when it was turned off
> 
> Ahh, sorry. Can you create a PR that changes like 1 line and nothing else? Then it should fix the spacing in that PR rather than this one, and we can merge that PR in order to fix all the formatting.

#407

src/simulation.hpp
```diff
@@ -901,14 +900,18 @@ template <typename problem_t> void AMRSimulation<problem_t>::ellipticSolveAllLev
 
 		// solve Poisson equation with open b.c. using the method of James (1977)
 		amrex::Vector<amrex::MultiFab> phi(finest_level + 1);
+		// define another fab to store phi
+		amrex::Vector<amrex::MultiFab> phi_data(finest_level + 1);
```
BenWibking: If you are defining `phi_data` as part of the AMRSimulation class, you don't want to re-define it here as a local variable, because then the local variable will take precedence over the class variable, and the class variable won't be updated (or even initialized!).

src/SphericalCollapse/spherical_collapse.cpp
```diff
@@ -104,6 +104,22 @@ template <> void RadhydroSimulation<CollapseProblem>::ErrorEst(int lev, amrex::T
...
+template <> void RadhydroSimulation<CollapseProblem>::ComputeDerivedVar(int lev, std::string const &dname, amrex::MultiFab &mf, const int ncomp_cc_in) const
+{
+	// compute derived variables and save in 'mf'
+	if (dname == "gpot") {
+		const int ncomp = ncomp_cc_in;
+		auto const &state = state_new_cc_[lev].const_arrays();
+		auto const &phi_data_ref = phi_data[lev].const_arrays();
+		auto output = mf.arrays();
+
+		// const amrex::Vector<amrex::MultiFab>& phi_data_ref = this->phi_data;
```
BenWibking: I think this comment can be deleted.

src/simulation.hpp
BenWibking: I think this will fix the segfault.

src/simulation.hpp
```diff
@@ -901,14 +900,18 @@ template <typename problem_t> void AMRSimulation<problem_t>::ellipticSolveAllLev
...
 		amrex::Vector<amrex::MultiFab> rhs(finest_level + 1);
 		const int nghost = 1;
 		const int ncomp = 1;
 		amrex::Real rhs_min = std::numeric_limits<amrex::Real>::max();
 		for (int lev = 0; lev <= finest_level; ++lev) {
 			phi[lev].define(grids[lev], dmap[lev], ncomp, nghost);
```
BenWibking: You could just get rid of `phi` and use `phi_data` instead. If the potential is kept in memory anyway, there's no point in creating a separate variable and copying it.

BenWibking (dismissed):
I think I found the issue -- you have a local variable that shadows the class member variable, so you're not updating the variable that you think you are. If you remove the local variable and just update the class member `phi_new` directly, then it should work.

psharda:
@BenWibking but can you look at this again? I still get a segfault when accessing `gpot`

src/simulation.hpp
```diff
@@ -912,7 +913,7 @@ template <typename problem_t> void AMRSimulation<problem_t>::ellipticSolveAllLev
 		}
 
 		// solve Poisson equation with open b.c. using the method of James (1977)
-		amrex::Vector<amrex::MultiFab> phi(finest_level + 1);
+		phi.resize(finest_level + 1);
```
BenWibking, line 919: I think the problem is that the first plotfile is output before the first timestep, so `phi` is not yet initialized the first time you access it. This code (plus the `phi[lev].define` and `phi[lev].setVal(0)` code below) needs to be called after the initial conditions are set, but before the first output is done.

BenWibking: It should go here: https://github.com/quokka-astro/quokka/pull/406/files#diff-1643e903f8d7916aa711335f79a095b83a7aa844db8dd37ba137dbc62a5eafbeR579

BenWibking: You want to initialize the grids and do the Poisson solve, but don't actually update the hydro momentum and energy with the $\rho g$ term.

psharda:
thanks!. this seems to work, but we should check the output file and see if `gpot` makes sense.

src/simulation.hpp
```diff
@@ -576,6 +577,14 @@ template <typename problem_t> void AMRSimulation<problem_t>::setInitialCondition
 		ReadCheckpointFile();
 	}
 
+	phi.resize(finest_level + 1);
+	const int nghost = 1;
+	const int ncomp = 1;
+	for (int lev = 0; lev <= finest_level; ++lev) {
+		phi[lev].define(grids[lev], dmap[lev], ncomp, nghost);
+		phi[lev].setVal(0); // set initial guess to zero
```
BenWibking: `phi` will be zero in the `plt00000` output. Is that what you want?

psharda: Umm I dont care about what it is in `plt00000`, but maybe other applications will? But how would we calculate it at t=0?

BenWibking: You could split `ellipticSolveAllLevels` into two functions -- one that calculates `phi` by calling the `OpenBC` object, and another that adds the gravitational acceleration to the gas. Then you could call the first function here.

psharda: So we need all of this at t=0:

```
                        phi[lev].define(grids[lev], dmap[lev], ncomp, nghost);
                        rhs[lev].define(grids[lev], dmap[lev], ncomp, nghost);
                        phi[lev].setVal(0); // set initial guess to zero
                        rhs[lev].setVal(0);
                        fillPoissonRhsAtLevel(rhs[lev], lev);
                        rhs_min = std::min(rhs_min, rhs[lev].min(0));
                }

                amrex::Real abstol = abstolPoisson_ * rhs_min;
                poissonSolver.solve(amrex::GetVecOfPtrs(phi), amrex::GetVecOfConstPtrs(rhs), reltolPoisson_, abstol);
                if (verbose) {
                        amrex::Print() << "\n";
                }
```
?

BenWibking: Yes, that would fix it so the phi output is valid in the first plotfile.

psharda:
Hmm, so this works on gadi on CPUs, but all the tests here fail.

src/simulation.hpp
```diff
@@ -576,6 +578,31 @@ template <typename problem_t> void AMRSimulation<problem_t>::setInitialCondition
 		ReadCheckpointFile();
 	}
 
+	// set up elliptic solve object
+	amrex::OpenBCSolver poissonSolver(Geom(0, finest_level), boxArray(0, finest_level), DistributionMap(0, finest_level));
```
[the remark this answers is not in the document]
psharda: @BenWibking does this warning make sense

psharda:
I think the tests failed because OpenBC solver is only used for 3D (which is why it ran on Gadi - because I compiled it in 3D). I have wrapped the t=0 `gpot` calculation in an `ifdef` in the initial conditions, just like that in `ellipticSolveAllLevels`

psharda:
Hmm, so SphericalCollapse test (where we use and store gpot) passes but other 3D tests fail.

src/simulation.hpp
```diff
@@ -165,6 +165,8 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 	amrex::Real reltolPoisson_ = 1.0e-5;	    // default
 	amrex::Real abstolPoisson_ = 1.0e-5;	    // default (scaled by minimum RHS value)
 	int doPoissonSolve_ = 0;		    // 1 == self-gravity enabled, 0 == disabled
+	amrex::Vector<amrex::MultiFab> phi;
+	amrex::Vector<amrex::MultiFab> rhs;
```
BenWibking: This is not needed and will cause excessive memory usage. Can you move rhs back to a local variable?

src/simulation.hpp
```diff
@@ -576,6 +578,33 @@ template <typename problem_t> void AMRSimulation<problem_t>::setInitialCondition
 		ReadCheckpointFile();
 	}
 
+#if AMREX_SPACEDIM == 3
+	// set up elliptic solve object
+	amrex::OpenBCSolver poissonSolver(Geom(0, finest_level), boxArray(0, finest_level), DistributionMap(0, finest_level));
```
BenWibking: We should avoid code duplication by putting this code in its own function, and calling it from here and from `ellipticSolveAllLevels`.

src/simulation.hpp
```diff
@@ -576,6 +578,33 @@ template <typename problem_t> void AMRSimulation<problem_t>::setInitialCondition
 		ReadCheckpointFile();
 	}
 
+#if AMREX_SPACEDIM == 3
```
BenWibking: You need to add a conditional here so that it only does this if gravity is enabled. Otherwise, it doesn't make sense. I think that's what is causing the test failures.

src/simulation.hpp
psharda: what do you want to call this function?

psharda: @BenWibking

src/simulation.hpp
```diff
@@ -576,6 +579,8 @@ template <typename problem_t> void AMRSimulation<problem_t>::setInitialCondition
 		ReadCheckpointFile();
 	}
 
+	calculateGpotAllLevels(dt_[0]);
```
psharda: @BenWibking what should be the argument passed to this function when it's called in the initial conditions?

BenWibking: I don't think `dt` is actually used by the Poisson solver, so you can just get rid of it entirely.

psharda: Ah yes I just realised that a minute ago!

psharda:
Nah, this is giving segfault while doing poisson solve during the initial conditions setup

BenWibking:
I took a quick look and everything seems reasonable. Once all the tests pass, let me know and I'll submit a review.

psharda:
@BenWibking all tests passed

BenWibking (approved):

PR #417, opened 2023-10-03
Update GPU CI timeout

### LINKED ISSUE #415: GPU CI fails (need to increase GPU CI timeout to 90 minutes)
**Describe the bug**
The GPU tests time out and fail because they hit the maximum walltime of 60 minutes.

**To Reproduce**
1. Run the GPU test suite on avatargpu
3. After 60 minutes, see timeout error

**Additional context**
Our test suite has grown increasingly lengthy. For some test problems, we might be able to reduce the number of timesteps we take by increasing the CFL number or simulating for less time. A few problems take longer to run on the GPU than on the CPU due to kernel launch overheads for essentially 1D/single-zone problems.

The easiest fix is to just increase the maximum time we are willing to wait for the test suite to complete on the GPU node.

BenWibking:
Set timeout to 90 minutes (increased from the default value of 60 minutes).

Fixes https://github.com/quokka-astro/quokka/issues/415.

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -12,31 +12,30 @@ pr:
     include:
     - development
 
-pool: avatar
-
-timeoutInMinutes: 90
-
-steps:
-
-- task: CMake@1
-  displayName: 'Configure CMake'
-  inputs:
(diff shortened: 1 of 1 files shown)


BenWibking:
@psharda Can you approve?

psharda (approved):

PR #421, opened 2023-10-09
Add RadhydroShockMultigroup test

chongchonghe:
Extending the original single-group radiation shock test to multigroup gray-opacity radiation shock, following Sec 9.5 of Skinner et al. 2019. We use 8 radiation groups logarithmically spaced over [1e15, 1e19] Hz. The run time increases from 22.0 s in the single-group case to 122 s in the 8-group case. 

We can discuss whether we want to keep both RadhyroShock and RadhyroShockMultigroup, or we want to keep the latter only and rename it to the former.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -147,6 +147,7 @@ add_subdirectory(RadTube)
 add_subdirectory(RadhydroShell)
 add_subdirectory(RadhydroShock)
 add_subdirectory(RadhydroShockCGS)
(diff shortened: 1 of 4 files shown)


BenWibking:
It looks like the tests have failed for unrelated reasons. The `CMake (macOS)` action fails due to something that broke with the Python dependencies.

The `debug arm64 gcc` action fails due to taking longer than 60 minutes to run. We can increase the timeout for this.

PR #425, opened 2023-10-13
workaround macOS GitHub actions bug

### LINKED ISSUE #424: macOS CI fails due to misconfigued GitHub actions runners
**Describe the bug**
This is the bug: https://github.com/actions/runner-images/issues/11545

**To Reproduce**
1. Wait for an update to Python to be pushed to Homebrew.
2. Run a macOS GitHub action.
3. See error, the action will fail.

**Additional context**
This is because the GitHub runner images put extra things into /usr/local that the system install does not, which conflicts with files that Homebrew then adds to /usr/local. Because the runner images modify /usr/local and also include Homebrew installed to /usr/local, this is a misconfiguration issue that is entirely GitHub's fault.

A temporary workaround is to ignore Homebrew upgrade failures:
```
      - run: |
          brew update
          brew upgrade || true
```
and then manually force Homebrew to overwrite the extraneous files in /usr/local (but this must be specified for each package that has a conflict, and it must include the version number, which will change in the future):
```
brew link --overwrite python@3.11
```

BenWibking:
Fixes https://github.com/quokka-astro/quokka/issues/424.

### DIFF
--- a/.github/workflows/cmake-macos.yml
+++ b/.github/workflows/cmake-macos.yml
@@ -31,7 +31,9 @@ jobs:
       run: cmake -E make_directory ${{runner.workspace}}/build
 
     - name: Install Homebrew dependencies
-      run: brew install python3 openmpi hdf5
+      run: |
+        brew install python3 openmpi hdf5 || true
+        brew link --overwrite python@3.11
 
     - name: Install pip dependencies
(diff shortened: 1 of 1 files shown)


BenWibking:
@psharda this fixes the CI issue.

psharda (approved):

PR #430, opened 2023-10-22
fix out-of-bounds bug in NSCBC outflow

### LINKED ISSUE #375: out-of-bounds access in NSCBC transverse derivatives
When computing the transverse derivatives for the NSCBC physical boundaries, it is naively necessary to access cells in other FABs. However, the corner cells that require these accesses do not actually need to be filled, so we should just avoid filling them.

Diagram:
IMG-7802

BenWibking:
Fixes https://github.com/quokka-astro/quokka/issues/375.


BenWibking:
@aditivijayan @markkrumholz This should make the outflow NSCBC usable.

markkrumholz (approved):

PR #435, opened 2023-10-27
add Summit (OLCF) config

BenWibking:
This adds a build configuration with the appropriate modules for running on [Summit](https://docs.olcf.ornl.gov/systems/summit_user_guide.html) at Oak Ridge National Laboratory.

Scripts to run scaling tests are included. Note that GPU-aware MPI does not work on Summit with the CUDA version that we need for Quokka.

Original scaling plot _(see updated scaling plot in comment below)_:
chart

### DIFF
--- a/cmake/summit.cmake
+++ b/cmake/summit.cmake
@@ -1,0 +1,16 @@
+# host configuration
+# run with `cmake -C host_config.cmake ..` from inside build directory
+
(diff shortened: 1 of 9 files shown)


markkrumholz (approved):

BenWibking:
Updated scaling plot (without`-gpu` flag, not needed since we disable GPU-aware MPI):
updated_summit_scaling_plot

PR #445, opened 2023-11-11
Advecting pulse test

chongchonghe:
Advecting pulse problem (Krumholz 2007). We use exactly the same initial conditions parameters. However, this test failed: not producing the desired result. The radiation diffuses too fast.

The diffusion coefficient is D ~ c / 3 kappa, and the width of the Gaussian pulse should be w^2 ~ 4/3 * c * t / kappa = 120 cm at t = 4.8e5. However, in this simulation, radiation diffuses to ~300 cm with chat = 0.1 c, and this width depends on the value of chat.

I can't spot any bug in the code. Is this due to the operator splitter problem we were talking about? Is this the same reason why the Asymptotic Marshak wave test failed?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -147,6 +147,7 @@ add_subdirectory(RadTube)
 add_subdirectory(RadhydroShell)
 add_subdirectory(RadhydroShock)
 add_subdirectory(RadhydroShockCGS)
+add_subdirectory(RadhydroPulse)
 
 add_subdirectory(ODEIntegration)
 add_subdirectory(Cooling)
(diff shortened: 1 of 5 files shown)


chongchonghe:
I don't need to create a PR for this... anyway. 

Run the test with `ninja -j4 test_radhydro_pulse && ninja src/RadhydroPulse/test`

BenWibking:
Yes, this problem won't give the right answer without an asymptotic-preserving time integrator.

It would be good to include in the test suite, but we need the IMEX PD-ARS integrator first.

chongchonghe:
Replaced by [advecting_pulse_test_new](https://github.com/quokka-astro/quokka/tree/advecting_pulse_test_new)

BenWibking:
I'm going to go ahead and close this. It will still be accessible from the list of closed PRs, and this branch remains in the repo.

PR #454, opened 2023-11-29
Fix the convergence problem of the Newton-Raphson

### LINKED ISSUE #451: On the convergence criteria of Newton-Raphson iteration
I found the following scenario where the convergence criteria for the  Newton-Raphson iteration, i.e. | F_G / E_tot | < epsilon = 1e-10, could fail. 

Imagine a test problem with matter and radiation in equilibrium with temperature T. Let's calculate how the radiation-matter exchange would work. We have $R = c \Delta t ~ \rho \kappa ~ (4 \pi B / c - E_r) = \frac{c \Delta t}{l} (a_r T^4 - E_r) $, where $l$ is the mean free path. In the diffusion regime with very high optical depth, $c \Delta t / l \gg 1$, let's say, it's 1e5. In the first iteration, the guessed value is the same as the original value, so $F_G = R$. Assume $E_r >> E_g$, then $E_{\rm tot} \approx E_r$. Then, $F_G / E_{\rm tot} = R / E_r = 10^5 (a_r T^4 / E_r - 1)$. Ideally, the part in the parathesis should be 0 since we assumed equilibrium in the beginning, but due to machine accuracy/roundoff error, it might be $\sim 10^{-14}$, then we get $F_G / E_{\rm tot} \approx 10^{-9}$. Therefore, the iteration won't break if we set epsilon to 1e-10. What I observed is that the iteration would bounce back and forth between 1e-9 and -1e-9 and never converge. 

Simply increasing epsilon won't solve the problem because the optical depth could go higher. 

@BenWibking @markkrumholz Is this a known issue? I have some idea but haven't tested it yet. I may update on this later after hearing your thoughts.

chongchonghe:
### Description

Fix the convergence problem of the Newton-Raphson iteration at high opatical depth by adding relative and absolute checks of Erad. This solves the problem of the Newton-Raphson iteration fails to converge when `c dt kappa rho >~ 10^4`. Now, all tests pass even with `resid_tol = 1e-13` while originally the RadTube test will fail with this resid_tol. 

### Related issues

Fix #451 

### Checklist

### DIFF
--- a/src/RadPulse/test_radiation_pulse.cpp
+++ b/src/RadPulse/test_radiation_pulse.cpp
@@ -23,7 +23,7 @@ constexpr double rho0 = 1.0;	  // g cm^-3 (matter density)
 constexpr double a_rad = 4.0e-10; // radiation constant == 4sigma_SB/c (dimensionless)
 constexpr double c = 1.0e8;	  // speed of light (dimensionless)
 constexpr double chat = 1.0e7;
-constexpr double erad_floor = a_rad * (1.0e-8);
+constexpr double erad_floor = a_rad * (1.0e-10);
 
 // constexpr double Lx = 1.0; // dimensionless length
 constexpr double initial_time = 1.0e-8;
--- a/src/radiation_system.hpp
+++ b/src/radiation_system.hpp
@@ -1091,7 +1091,7 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 			quokka::valarray<double, nGroups_> deltaErad;
 			quokka::valarray<double, nGroups_> F_R;
 
-			const double resid_tol = 1.0e-10; // 1.0e-15;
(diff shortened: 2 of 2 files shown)


chongchonghe:
That's strange. All tests passed on my Mac Studio, but the RadiationPulse test failed on the automatic tests.

BenWibking:
> That's strange. All tests passed on my Mac Studio, but the RadiationPulse test failed on the automatic tests.

Maybe different compiler flags? It fails on both macOS (AppleClang) and Linux CI (gcc).

chongchonghe:
> > That's strange. All tests passed on my Mac Studio, but the RadiationPulse test failed on the automatic tests.
> 
> Maybe different compiler flags? It fails on both macOS (AppleClang) and Linux CI (gcc).

I think I find the problem, although I don't know why it passed on one computer but not on others. When using the absolute check, we need to set Erad_floor carefully (usually 1e-10 * a_rad * T0^4).

BenWibking (approved):

PR #476, opened 2023-12-23
Fix RadShadow test problem

BenWibking:
### Description
Updates `RadShadow` to use `Physics_Traits`. This test problem had not been updated to use `Physics_Traits`, so the variable indexing was wrong and caused it to crash.

Now it crashes at runtime due to the Newton-Raphson iteration failing to converge about ~40% of the way through the simulation.

### Related issues
N/A

### Checklist


BenWibking:
It crashes with the error:
```
Coarse STEP 1373 at t = 2.059424724e-11 (41.18849448%) starts ...
Creating new distribution map on level: 1
Creating new distribution map on level: 2
[Level 0 step 1373] ADVANCE with time = 2.059424724e-11 dt = 1.501038428e-14
        Radiation substeps: 1   dt: 1.501038428e-14
[Level 0 step 1373] Advanced 22400 cells
[Level 1 step 2745] ADVANCE with time = 2.059424724e-11 dt = 7.505192142e-15
        Radiation substeps: 1   dt: 7.505192142e-15
[Level 1 step 2745] Advanced 12928 cells
[Level 2 step 5489] ADVANCE with time = 2.059424724e-11 dt = 3.752596071e-15
        Radiation substeps: 1   dt: 3.752596071e-15
[Level 2 step 5489] Advanced 34176 cells
[Level 2 step 5490] ADVANCE with time = 2.059799983e-11 dt = 3.752596071e-15
        Radiation substeps: 1   dt: 3.752596071e-15
[Level 2 step 5490] Advanced 34176 cells
[Level 1 step 2746] ADVANCE with time = 2.060175243e-11 dt = 7.505192142e-15
        Radiation substeps: 1   dt: 7.505192142e-15
[Level 1 step 2746] Advanced 12928 cells
[Level 2 step 5491] ADVANCE with time = 2.060175243e-11 dt = 3.752596071e-15
        Radiation substeps: 1   dt: 3.752596071e-15
[Level 2 step 5491] Advanced 34176 cells
[Level 2 step 5492] ADVANCE with time = 2.060550503e-11 dt = 3.752596071e-15
        Radiation substeps: 1   dt: 3.752596071e-15
  what():  Assertion `n < maxIter' failed, file "/home/benwibking/quokka/src/radiation_system.hpp", line 1201, Msg: "Newton-Raphson iteration failed to converge!"
```

We probably need to implement linesearch with backtracking in order to fix this.

cc @chongchonghe

chongchonghe:
working on it...

chongchonghe:
Line-search backtracking didn't fix this problem. The problem is that the rate of converging is too slow for certain problems and it failed to converge after 400 (or even higher) iterations. It seems to be converging at the right direction There are two ways to fix this:

1. Remove the `Assertion 'n < maxIter'` line, or simply put a line in the for loop to let it break after 100 iterations. This way the test will run with success. I believe the result should be correct although I'm not sure since there isn't an error check in the code. In the variable-kappa advecting pulse test this trick worked. The reason this worked is that in certain steps it would take a huge number of Newton iterations to balance the gas and radiation temperature, but we can just let it go after, say 100, iterations, resulting in an unbalanced state. In the next few time steps, they will catch up and eventually come to equilibrium. 
2. Use 2nd-order Newton method. This should make it converge much faster. 

I may create a PR to implement method 2. For now, method 1 works just fine.

BenWibking:
> Line-search backtracking didn't fix this problem. The problem is that the rate of converging is too slow for certain problems and it failed to converge after 400 (or even higher) iterations. It seems to be converging at the right direction There are two ways to fix this:
> 
> 1. Remove the `Assertion 'n < maxIter'` line, or simply put a line in the for loop to let it break after 100 iterations. This way the test will run with success. I believe the result should be correct although I'm not sure since there isn't an error check in the code. In the variable-kappa advecting pulse test this trick worked. The reason this worked is that in certain steps it would take a huge number of Newton iterations to balance the gas and radiation temperature, but we can just let it go after, say 100, iterations, resulting in an unbalanced state. In the next few time steps, they will catch up and eventually come to equilibrium.
> 2. Use 2nd-order Newton method. This should make it converge much faster.
> 
> I may create a PR to implement method 2. For now, method 1 works just fine.

I don't think this is a good idea. It should reach the convergence tolerance or fail. Ignoring it might work sometimes, but in a simulation without a known solution, we have no idea if we are getting the right answer.

I don't know what you mean by this. Since we are using the exact analytic derivatives, the existing Newton-Raphson method should converge asymptotically at second order (https://en.wikipedia.org/wiki/Newton%27s_method#Proof_of_quadratic_convergence_for_Newton's_iterative_method).

For hydro, if the solution isn't good enough, it goes back and re-does the step with a halved timestep. Something like that might work here.

Also, is it possible there is a numerical precision issue?

chongchonghe:
I meant higher-order Householder's methods where a higher order of derivative is used. 

Numerical precision issue is a good point. I'll take a look at that.

BenWibking:
The recent fix to the Newton solver should fix this problem. Need to check.

chongchonghe:
The only thing I did to the Newton iteration in my recent update is I added an extra break condition: |d Egas / Egas_guess| < 1e-7. This makes the iteration to break when the relative change in Egas or Erad is small but the residual F_G isn't (probably because both F_G and E_tot are small). 

All tests are passed.  You can change it from Draft to PR and I'll approve and merge.

chongchonghe (approved):

PR #486, opened 2024-01-09
Multigroup advecting pulse test

chongchonghe:
### Description
Define the multigroup, variable opacity advecting pulse test. See Sec. 4.1.6 of the CASTRO III paper (Zhang et al. 2013). The code perfectly reproduces Figure 6 of the paper. We use the original Newton-Raphson method on the (Egas, Erad) basis.

### Related issues
N/A

### Checklist


chongchonghe:
@markkrumholz @BenWibking 
I identified the origin of the problem of the Newton-Rasphon iteration failing to converge. It's probably due to the nature of the problem, not the failure of the code. Here's why. For any given parameters and kappa function at a given cell, I can plot the values of F_G, F_R, and F = sqrt(F_G^2 + F_R^2) as a function of E_g and E_r (for grey radiation). From this plot, we can visualize how the Newton-Rasphon iteration would approach the solution.

For example, consider the following simple test: E_g = E_r = 1, kappa=1, and all constants are 1. The expected solution (equilibrium state) is (1, 1), which is also what the plot suggests. 

F_G_F_R-unit

Here is what it looks like for the advecting pulse test. This is unexpected and there are a few problems: 1. There are infinite equilibrium states (roots of F), so it's mathematically impossible to find a unique solution. If this situation is physically feasible, we have to figure out how to deal with it.  2. Even though there are infinite solutions located on a curve, I would expect the algorithm to find one of them very quickly, given that F is a smooth function of E_g and E_r. This suggests there is something wrong with my code or the algorithm. 

F_G_F_R-pulse

_F, F_G, and F_R of one of the cells in the advecting pulse test. The 'initial state' is the initial state of the Newton iteration, and the 'final' state is where the code gets to after 100 iterations. I get a similar plot when kappa is set to a constant 180._

chongchonghe:
I confirm that this behaviour occurs whenever the opacity is high, e.g. with the 'everything is unity' initial condition (Egas = Erad = rho = c = c_v = 1), if I set kappa = 1e3 and dt = 0.1, the solution of the Newton iteration is not unique. With kappa = 1, the solution is unique (Fig. 1 from my previous comment). However, when I set kappa = 1e3 and dt = 1e-4, the F plot is a nice oval and the solution is unique. Bottom line is, with the current method, we need extremely small time step when the opacity is high.

F_G_F_R-unit-kappa1e3-dt1e-1
_All parameters are 1 except kappa = 1e3 and dt = 0.1_

F_G_F_R-unit-kappa1e3-dt1e-4
_All parameters are 1 except kappa = 1e3 and dt = 1e-4_

BenWibking:
These plots show that there is some structure that is making it difficult to converge, but it does not conclusively demonstrate that the root is non-unique. For instance, it could be that there is one zero, but a large band of near-zero values of the residual.

Can you try re-plotting the above plots with a log-scaled colorbar?

chongchonghe:
> These plots show that there is some structure that is making it difficult to converge, but it does not conclusively demonstrate that the root is non-unique. For instance, it could be that there is one zero, but a large band of near-zero values of the residual.
> 
> Can you try re-plotting the above plots with a log-scaled colorbar?

Fig 1 is a remake of the first figure from my previous comment. Even with a very narrow x and y range (+- 0.005) and 1000x1000 pixel sampling, there is a line of pixels with near zero values. Note that the vmin is set to be free so it reflects the minimum value of F, which is 0.00107. I tend to believe there is a unique solution (without proving it), but it lies in the middle of a long stripe of near-zero values. See Fig 2. This, plus the fact that the gradient descent direction is perpendicular to this stripe, makes it hard to converge to a unique answer. 

a2
Fig. 1. _All parameters are 1 except kappa = 1e3 and dt = 0.1. Log scale._

a3
Fig. 2

chongchonghe:
Here is a simple script to make these plots: 
[plot_F_residual.txt](https://github.com/quokka-astro/quokka/files/13876374/plot_F_residual.txt)

BenWibking:
It would help to see the individual iterations plotted. You could do this by running on CPU and saving the iterations to a `std::vector`.

BenWibking:
Backtracking linesearch with Newton's method is supposed to work for these kind of banana-shaped functions. See the plot of the iterates for the Rosenbrock function here: https://www.brnt.eu/phd/node10.html#fig:newton-method

If it doesn't work, then there is probably either a numerical precision issue or a bug. This should manifest as very strange behavior in the iterations.

markkrumholz:
Just coming online in morning here and seeing this discussion. It seems to me like @BenWibking's idea would work, but we might be able to do something easier by exploiting our knowledge of the physical nature of the problem, rather than resorting to more general methods that work on arbitrary banana-shaped functions. 

My diagnosis is that the phenomenon @chongchonghe is finding occurs when R (equation 28 of the Quokka paper) has a very large coefficient dt * kappa, such that R >> E_g^(n+1) - E_g^(n) and also R >> E_r^(n+1) - E_r^(n). The narrow canyon in the plots appears because the large coefficient guarantees that, unless you choose E_g^(n+1) and E_r^(n+1) to lie almost exactly along the line where R = 0, you get huge residuals. If you continue with the analogy of a canyon, Newton's method has trouble because the solution landscape looks like a curving, narrowly-walled canyon, and Netwon's method winds up taking tiny steps going back and forth across the short dimension of the canyon, making very little progress following its much more gradually varying curve toward the solution.

One way of dealing with this, as @BenWibking says, is to use a method that can deal with functions with this sort of shape. But another is that we can exploit our knowledge of this geometry of the problem to make a change of variables from one where the canyon is curved (which is the problem here) to one where is straight and along one of the coordinate axes; Newton's method should do just find in this case. One way to do this would be to make of change of variables to use E_g^(n+1) and R^(n+1) as opposed to E_g^(n+1) and E_r^(n+1) as iteration variables. That is, analytically rewrite the equations to solve for Delta E_g and Delta R. I bet this would converge fine for this problem.

BenWibking:
That could work. However, I'm concerned that we might end up with weirdly-shaped residual surfaces for the general multigroup case, especially with variable opacities.

markkrumholz:
I guess I'm a bit less concerned, because the situation where this seems to be occurring is one that is physically extremely simple and easy-to-understand: the opacity times time step is huge, so the solution is that the system is in perfect equilibrium, with radiation-gas exchange term vanishing. So you could in this situation the right answer just by setting R = 0 manually and then solving for the remaining variables subject to the initial conditions. Thus it seems to me like my strategy is entirely analogous to the departure coefficient formulation in non-LTE line transfer: when you expect the solution to be very close to equilibrium, you rewrite the problem so that, instead of using the population of each level as your variable, you use the ratio of the population to the LTE value as your variable. This had the advantage of normalising all the variables to be at similar order of magnitude, and making the problem orthogonal such that levels that are in equilibrium can just go to it immediately, and you can focus all your computational effort on the dimensions that are non-equilibrium. It seems like using R as a variable when the problem is very close to equilibrium is in the same spirit here.

chongchonghe:
I tried the line search method and it didn't work. It's easy to understand why it failed in the extremely high opacity case, because it does not prevent it from oscillating back and forth across the short dimension of the canyon. In some cases, when the trial arrives at the center of the canyon at the short dimension (but not at exactly the root), the slope might be zero and the line search failed because even a tiny step (1e-14) won't meet the Wolfe condition. The "orthogonal" steps method looks interesting and I might try that if necessary.

I like the idea of solving it from the physics point of view. I will look into it and see how it works.

BenWibking:
> I tried the line search method and it didn't work. It's easy to understand why it failed in the extremely high opacity case, because it does not prevent it from oscillating back and forth across the short dimension of the canyon. In some cases, when the trial arrives at the center of the canyon at the short dimension (but not at exactly the root), the slope might be zero and the line search failed because even a tiny step (1e-14) won't meet the Wolfe condition. The "orthogonal" steps method looks interesting and I might try that if necessary.
> 
> I like the idea of solving it from the physics point of view. I will look into it and see how it works.

I suspect there's a numerical precision issue that prevents this from working. In any case, Mark's suggestion should also have the benefit of rescaling the variables to avoid steps that are too close to machine precision.

chongchonghe:
The issue was resolved and now the multigroup advecting radiation pulse test (Sec. 4.1.6 of CASTRO III paper) is fully passed. The mistake comes from a stupid mistake I made on calculating d_kappa_d_T. It's off by 10^7 and the calculated Newton step is 10^7 times smaller than it should be, causing the iteration to never converge.

BenWibking:
@chongchonghe Can you take a look at the CI failures?

BenWibking:
> The issue was resolved and now the multigroup advecting radiation pulse test (Sec. 4.1.6 of CASTRO III paper) is fully passed. The mistake comes from a stupid mistake I made on calculating d_kappa_d_T. It's off by 10^7 and the calculated Newton step is 10^7 times smaller than it should be, causing the iteration to never converge.

I made similar mistakes several times when writing the original version of the radiation code.

BenWibking:
It looks like there are still test failures for the debug build.

chongchonghe:
> It looks like there are still test failures for the debug build.

The multigroup test takes too long to run with debugger enabled. I have to turn it off.

BenWibking:
I would prefer to have this test on for at least some of the builds. You could disable it from running for a given build the way the matter-energy exchange test is disabled for the GPU runs:
https://github.com/quokka-astro/quokka/blob/b4222b7e92d41231fc115e0298c75faf7d6300ca/.ci/azure-pipelines-amdgpu.yml#L33

BenWibking:
You could alternatively label the test using ctest labels: https://cmake.org/cmake/help/latest/manual/ctest.1.html#label-matching

And then you can do `ctest -LE test_label_to_exclude` to avoid running all the tests with label `test_label_to_exclude`.

BenWibking (approved):

PR #502, opened 2024-01-18
add CIC particles

### LINKED ISSUE #491: add cloud-in-cell particles (without AMR subcycling)
**Describe the proposal**
With a global simulation timestep, It should be possible to implement particles that interact gravitationally without needing virtual or ghost particles.

It should be possible to do the cloud-in-cell (CIC) deposition of the particles self-consistently by depositing the particles onto the AMR levels, then averaging down the rhs of all of the AMR levels before solving the Poisson equation.

The only possible complication I can imagine is due to particles that lie across a coarse-fine boundary, where the deposited particle has a larger size on the coarse side of the boundary (and a smaller size on the fine side). I imagine most codes just ignore this, but this seems wrong.

The alternative is to implement gravitationally-interacting (i.e., cloud-in-cell) particles with ghost particles and virtual particles (https://github.com/quokka-astro/quokka/issues/332). This is necessary in order to i) handle the interactions that can occur at coarse-fine boundaries during a subcycle and ii) to deposit particles that live at finer levels. With a global timestep, problem i does not happen and problem ii is handled by averaging down the rhs of the Poisson equation after particle deposition.

ORION avoids this problem by replicating all particles on every MPI rank. This will not scale to a large number of particles.

**Additional context**
Since our gravity solver implementation also currently requires a global timestep, we don't regress any physics capabilities compared to what we have right now.

*AMReX cloud-in-cell functions:*
* `amrex::ParticleToMesh`: https://github.com/AMReX-Codes/amrex/blob/b96b731d4a6fa496f80e54c2c0476a50b9dd4e86/Tests/Particles/ParticleMeshMultiLevel/main.cpp#L91
* `amrex::MeshToParticle`: https://github.com/AMReX-Codes/amrex/blob/b96b731d4a6fa496f80e54c2c0476a50b9dd4e86/Tests/Particles/ParticleMesh/main.cpp#L96

The [Nyx paper](https://arxiv.org/abs/1301.4498) describes the interaction of the leapfrog particle integrator with the hydro update.

BenWibking:
### Description
This adds cloud-in-cell particles that interact with the gravitational field and are integrated in time using a kick-drift-kick leapfrog assuming a global simulation timestep (i.e., no AMR subcycling).

### Related issues
Closes https://github.com/quokka-astro/quokka/issues/491.

### Checklist


BenWibking:
Spherical collapse test with particles:

https://github.com/quokka-astro/quokka/assets/1781919/a63dc69b-a38e-455c-ae60-9176acedc7bc

BenWibking:
@psharda can you test this out with your simulation?

BenWibking:
@markkrumholz the basic particle infrastructure is in place. Suggestions for test problems are welcome.

psharda:
> @psharda can you test this out with your simulation?

Test it how?  Like you did for spherical collapse?

BenWibking:
> > @psharda can you test this out with your simulation?
> 
> Test it how?

Can you merge it into your PrimordialChem problem and add some particles to the initial conditions, and see if it crashes? :)

src/SphericalCollapse/spherical_collapse.cpp
```diff
@@ -85,6 +85,18 @@ template <> void RadhydroSimulation<CollapseProblem>::setInitialConditionsOnGrid
 	});
 }
 
+template <> void RadhydroSimulation<CollapseProblem>::createInitialParticles()
+{
+	// add particles at random positions in the box
+	bool generate_on_root_rank = true;
+	int iseed = 42;
+	int num_particles = 1000;
+	double particle_mass = 0.5 / static_cast<double>(num_particles);
```
psharda: what should be the particle mass for a particular problem?

psharda: like how would you set it for the pop iii test

src/SphericalCollapse/spherical_collapse.cpp
```diff
@@ -114,8 +126,7 @@ template <> void RadhydroSimulation<CollapseProblem>::ComputeDerivedVar(int lev,
 		auto const &phi_data_ref = phi[lev].const_arrays();
 		auto output = mf.arrays();
 
-		amrex::ParallelFor(
-		    mf, [=] AMREX_GPU_DEVICE(int bx, int i, int j, int k) noexcept { output[bx](i, j, k, ncomp) = phi_data_ref[bx](i, j, k, ncomp); });
+		amrex::ParallelFor(mf, [=] AMREX_GPU_DEVICE(int bx, int i, int j, int k) noexcept { output[bx](i, j, k, ncomp) = phi_data_ref[bx](i, j, k); });
 	}
```
psharda, line 129: what changed here? just some indentation?

src/SphericalCollapse/spherical_collapse.cpp
BenWibking: I tried to figure out what total mass I wanted in particles vs the gas (about equal), then divide by the number of particles. This will be different for a primordial gas cloud. There, it might be appropriate to choose the total gas mass relative to total particle+gas mass based on the universal baryon fraction $\Omega_b / \Omega_m$.

src/SphericalCollapse/spherical_collapse.cpp
BenWibking: You referenced a non-existent component of `phi[lev]`. Since `phi[lev]` only has one component, you can omit the component index in the array and just give it the $(i,j,k)$ cell indices.

psharda:
Does this already work with AMR?

BenWibking:
> Does this already work with AMR?

I tested with max_level = 1, we'll see if it works with more levels, or if it will crash in an exciting way :)

BenWibking:
> > Does this already work with AMR?
> 
> I tested with max_level = 1, we'll see if it works with more levels, or if it will crash in an exciting way :)

Hmm, it's crashed in an exciting way when max_level = 2. Maybe don't use AMR until I figure out what I missed...

BenWibking:
I think I was too enthusiastic since it worked in the very simple case. I think there's a missing factor in converting from mass to density and/or from gravitational potential to acceleration...

I'll debug it and then let you know when it's fully working with AMR for the SphericalCollapse problem.

BenWibking:
Inscrutable CUDA error:
```
/tmp/bwibking/_work/1/s/extern/amrex/Src/Base/AMReX_MFParallelForG.H(87): error: A type local to a function ("lambda [](int, int, int, int, int)->void") cannot be used in the template argument of the enclosing parent function (and any parent classes) of an extended __device__ or __host__ __device__ lambda
          detected during:
            instantiation of "std::enable_if_t<amrex::IsFabArray<MF, void>::value, void> amrex::experimental::detail::ParallelFor<MT,MF,F>(const MF &, const amrex::IntVect &, int, const amrex::IntVect &, __nv_bool, F &&) [with MT=256, MF=amrex::MultiFab, F=lambda [](int, int, int, int, int)->void]" 
(153): here
            instantiation of "std::enable_if_t<amrex::IsFabArray<MF, void>::value, void> amrex::experimental::detail::ParallelFor(const MF &, const amrex::IntVect &, int, const amrex::IntVect &, __nv_bool, F &&) [with MF=amrex::MultiFab, F=lambda [](int, int, int, int, int)->void]" 
/tmp/bwibking/_work/1/s/extern/amrex/Src/Base/AMReX_MFParallelFor.H(149): here
            instantiation of "std::enable_if_t<amrex::IsFabArray<MF, void>::value, void> amrex::experimental::ParallelFor(const MF &, const amrex::IntVect &, int, F &&) [with MF=amrex::MultiFab, F=lambda [](int, int, int, int, int)->void]" 
/tmp/bwibking/_work/1/s/src/simulation.hpp(1109): here
            instantiation of "void AMRSimulation<problem_t>::kickParticlesAllLevels(amrex::Real) [with problem_t=KelvinHelmholzProblem]"
```

BenWibking:
### Reproducer for OpenBC grid nesting problem
```console
git clone git@github.com:quokka-astro/quokka.git
cd quokka
gh pr checkout 502
git submodule update --recursive --init
mkdir build; cd build
cmake .. -DAMReX_SPACEDIM=3
cmake --build . --target spherical_collapse
cd ../tests
../build/src/SphericalCollapse/spherical_collapse SphericalCollapse_reproducer.in
```

psharda:
@BenWibking so this is now done?

BenWibking:
No AMR yet.

psharda:
Oh

BenWibking:
@psharda You can test this with your problem now. AMR should work, and I've tested that it does work for the SphericalCollapse problem.

BenWibking:
AMR SphericalCollapse with gravitating particles:

https://github.com/quokka-astro/quokka/assets/1781919/dea36b7e-62ca-4b12-9bd7-a281bd2fcc69

src/simulation.hpp
```diff
@@ -1021,6 +1079,131 @@ template <typename problem_t> void AMRSimulation<problem_t>::ellipticSolveAllLev
...
+			if (lev == 0) {
+				accel[lev].FillBoundary(geom[lev].periodicity());
+				fineBoundaryFunctor(accel[lev], 0, accel[lev].nComp(), accel[lev].nGrowVect(), 0., 0);
+			} else {
+				amrex::PhysBCFunct<amrex::GpuBndryFuncFab<setFunctorParticleAccel>> coarseBoundaryFunctor(geom[lev - 1], accelBC,
+															  boundaryFunctor);
+				amrex::Vector<amrex::MultiFab *> fineData{&accel[lev]};
+				amrex::Vector<amrex::MultiFab *> coarseData{&accel[lev - 1]};
+				// N.B.: all multifabs are at the same time, so we ignore the time arguments
+				amrex::FillPatchTwoLevels(accel[lev], 0., coarseData, {0.}, fineData, {0.}, 0, 0, AMREX_SPACEDIM, geom[lev - 1], geom[lev],
```
BenWibking: This can be simplified by instead using `amrex::InterpFromCoarseLevel`: https://github.com/AMReX-Codes/amrex/blob/022f97ea9ebbc28fca25f38f3eda26ce68b6230b/Src/AmrCore/AMReX_FillPatchUtil_I.H#L934.

src/simulation.hpp
```diff
@@ -981,7 +1021,20 @@ template <typename problem_t> void AMRSimulation<problem_t>::calculateGpotAllLev
 			rhs[lev].define(grids[lev], dmap[lev], ncomp, nghost);
 			phi[lev].setVal(0); // set initial guess to zero
 			rhs[lev].setVal(0);
+		}
+
+#ifdef AMREX_PARTICLES
+		if (do_cic_particles != 0) {
+			// deposit particles using amrex::ParticleToMesh
+			amrex::ParticleToMesh(*CICParticles, amrex::GetVecOfPtrs(rhs), 0, finest_level,
+					      quokka::CICDeposition{Gconst_, quokka::ParticleMassIdx, 0, 1});
```
BenWibking, line 1030: Unclear whether `amrex::ParticleToMesh` or `CICParticles->AssignDensity` is the appropriate function to use here.

src/simulation.hpp
```diff
@@ -1021,6 +1079,131 @@ template <typename problem_t> void AMRSimulation<problem_t>::ellipticSolveAllLev
...
+				}
+				if (n == 2) {
+					accel_arr[bx](i, j, k, n) = -0.5 * dx_inv[2] * (phi_arr[bx](i, j, k + 1) - phi_arr[bx](i, j, k - 1));
+				}
+			});
+			amrex::Gpu::streamSynchronizeAll();
+
+			// fill ghost cells for accel[lev]
+			amrex::GpuBndryFuncFab<setFunctorParticleAccel> boundaryFunctor(setFunctorParticleAccel{});
+			amrex::PhysBCFunct<amrex::GpuBndryFuncFab<setFunctorParticleAccel>> fineBoundaryFunctor(geom[lev], accelBC, boundaryFunctor);
```
BenWibking: This can probably be replaced with `PhysBCFunctNoOp`, e.g.: https://github.com/AMReX-Codes/amrex/blob/022f97ea9ebbc28fca25f38f3eda26ce68b6230b/Src/AmrCore/AMReX_AmrParticles.H#L218.

BenWibking: Actually, it can't.

src/simulation.hpp
BenWibking: Done.

BenWibking:
With 1e6 particles:

https://github.com/quokka-astro/quokka/assets/1781919/0235b85c-5711-4f51-807d-8ca405a63e6c

BenWibking:
Something strange is happening with SphericalCollapse on AMDGPU: https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=8261&view=logs&j=0410d313-00ef-550d-37c8-facbf7a1cb10&t=62d0f0fc-52cd-5dfc-847e-751a25f0b5e6. It times out, but it also should not take as many timesteps as it is, suggesting a miscompile or similar issue.

src/simulation.hpp
BenWibking: `AssignDensity` only does CIC deposition, whereas `ParticleToMesh` is more general. Either should work.

markkrumholz (commented):
I have looked through the code, and in general it looks fine. However, I would like to suggest two additional changes, which could either go into this PR or be done in separate PRs:

1. Right now your naming scheme is that just particles are tracer particles, while cic particles are different. In the future we plan to add even more particles types (sink and stellar feedback particles), so it would be good to rename the variables for tracer particles everywhere from do_particles to do_tracer_particles, and similarly for all other variable / input file option names.
2. I would suggest adding an additional test, which tests if your kick-drift and CIC schemes work properly: initialize two particles separated by ~10 cells, and with initial masses and velocities chosen to produce an orbit with a particular eccentricity and angular momentum. Make the gas density very low and gas temperature very high, such that the Bondi radius is larger than the domain size, so the gas should just sit there, and exerts negligible gravitational drag. Then run the problem for a few orbits, and check the orbital elements. If everything is working correctly, the semi-major axis and eccentricity of the orbit (or the energy and angular momentum) should be preserved very well; at most you should see some precession of the orbit. This is a much stronger test than spherical collapse, because it tests that your time stepping is properly symplectic.

BenWibking:
> 1. Right now your naming scheme is that just particles are tracer particles, while cic particles are different. In the future we plan to add even more particles types (sink and stellar feedback particles), so it would be good to rename the variables for tracer particles everywhere from do_particles to do_tracer_particles, and similarly for all other variable / input file option names.

I've renamed the "particles" output group to "tracer_particles".

For future reference, this name has to be manually specified in the Paraview GUI to view the particles (https://amrex-codes.github.io/amrex/docs_html/Visualization.html#visualizing-particle-data).

The particles can also been examined in yt with:
```
In [1]: import yt

In [2]: from yt.frontends.boxlib.data_structures import AMReXDataset

In [3]: ds = AMReXDataset("plt02758")
yt : [INFO     ] 2024-01-22 21:14:58,704 Parameters: current_time              = 0.05
yt : [INFO     ] 2024-01-22 21:14:58,704 Parameters: domain_dimensions         = [64 64 64]
yt : [INFO     ] 2024-01-22 21:14:58,704 Parameters: domain_left_edge          = [-0.6 -0.6 -0.6]
yt : [INFO     ] 2024-01-22 21:14:58,704 Parameters: domain_right_edge         = [0.6 0.6 0.6]

In [4]: ds.particle_types
Out[4]: ('CIC_particles',)
```

BenWibking:
Re comment 2, I'll replicate the `BinaryOrbitCIC` problem from the ORION2 repo.

BenWibking:
~~Still waiting on a workaround for a bug in the yt AMReX particles reader to check the orbital elements.~~

In the meantime, here's a movie of the BinaryOrbitCIC test:

https://github.com/quokka-astro/quokka/assets/1781919/a067c29a-5130-47b1-a363-5a0f9c0a1a7d

BenWibking:
For future reference, this is the fractional variation in the distance between the particles using the same test problem run with ORION2 (left: error as fraction of initial orbit, right: error as fraction of cell width):

image

BenWibking:
Quokka test run (at twice the spatial resolution of the ORION2 test):
orbit

BenWibking:
@markkrumholz How does this result look to you?

markkrumholz:
> @markkrumholz How does this result look to you?

What is this in units of cell spacings? In general it looks ok, because it's oscillating and there is no secular drift. However, I just want to verify that the error is of order 10% of a cell size, which is about the level I expect due to the CIC approximation of gravity.

BenWibking:
> > @markkrumholz How does this result look to you?
> 
> What is this in units of cell spacings? In general it looks ok, because it's oscillating and there is no secular drift. However, I just want to verify that the error is of order 10% of a cell size, which is about the level I expect due to the CIC approximation of gravity.

Here it is in units of cell spacings:

orbit

markkrumholz (approved):

PR #510, opened 2024-01-28
add HWASAN CI run

### LINKED ISSUE #509: add HWASAN CI run
**Describe the proposal**
We should add a CI run using Azure Pipelines that runs on ARM64 with HWASAN enabled. This will help detect additional memory errors.

We currently have ASAN running in CI, but it can only detect out-of-bounds memory accesses immediately after/before an allocated array/variable. HWASAN uses hardware features in ARM64 CPUs to detect out of bounds accesses at arbitrarily large jumps from the actual memory address. (It can also detect use-after-free arbitrarily long after the first free of a memory allocation.) The downside is that it does so probabilistically, with a ~0.3% chance of missing a memory access error on any given access.

**Additional context**
HWASAN (Hardware-assisted AddressSanitizer) is explained here: https://clang.llvm.org/docs/HardwareAssistedAddressSanitizerDesign.html

BenWibking:
### Description
This adds a CI run on Oracle Cloud Infrastructure (with Ampere aarch64 processors) with HWASAN enabled.

Processor info:
```
Architecture:           aarch64
  CPU op-mode(s):       32-bit, 64-bit
  Byte Order:           Little Endian
CPU(s):                 4
  On-line CPU(s) list:  0-3
Vendor ID:              ARM
  Model name:           Neoverse-N1
    Model:              1
    Thread(s) per core: 1
    Core(s) per socket: 4
    Socket(s):          1
    Stepping:           r3p1
    BogoMIPS:           50.00
    Flags:              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
```

### Related issues
Closes https://github.com/quokka-astro/quokka/issues/509.

### Checklist


BenWibking:
Currently fails with a linking error:
```
/usr/bin/ld: cannot find -lhwasan
```

src/CMakeLists.txt
```diff
@@ -60,6 +60,13 @@ if(ENABLE_ASAN)
   add_link_options(-fsanitize=address -fsanitize=undefined)
 endif(ENABLE_ASAN)
 
+if(ENABLE_HWASAN)
+  # enable HWAddressSanitizer for debugging
+  message(STATUS "Compiling with HWAddressSanitizer *enabled*")
+  add_compile_options(-fsanitize=hwaddress)
+  add_link_options(-fsanitize=hwaddress)
+endif(ENABLE_ASAN)
```
BenWibking: fix argument mismatch

BenWibking:
It failed because Oracle Linux does not package libhwasan for gcc with appstream. Maybe Clang will work.

BenWibking:
Now it fails with:
```
/opt/rh/gcc-toolset-13/root/usr/lib/gcc/aarch64-redhat-linux/13/../../../../bin/ld: ../../amrex/Src/libamrex_1d.a(AMReX.cpp.o): relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol `stderr@@GLIBC_2.17' which may bind externally can not be used when making a shared object; recompile with -fPIC
```

BenWibking:
Now it fails at runtime with an exception thrown for all tests:
```
 1/31 Test  #1: ScalarAdvection ..................***Exception: Illegal  0.11 sec
```
etc.

BenWibking:
The error is:
```
Illegal instruction (core dumped)
```
No core dump is produced. This looks like a compiler bug :(

BenWibking:
This looks like the same issue: https://github.com/google/sanitizers/issues/1241

.ci/azure-pipelines-aarch64-hwasan.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+- job: BuildAndTest
+  timeoutInMinutes: 180 # how long to run the job before automatically cancelling
+  pool: oracle-cloud
+  steps:
+  - task: CMake@1
+    displayName: 'Configure CMake'
+    inputs:
+      cmakeArgs: '.. -DCMAKE_BUILD_TYPE=Release -DENABLE_HWASAN=ON -DCMAKE_CXX_FLAGS="-fPIC" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DAMReX_SPACEDIM=1'
```
BenWibking: It may help to add `-fuse-ld=lld` to the linker flags.

BenWibking:
@psharda can you review?

psharda:
Just so I understand, you say HWASAN can miss 0.3% memory errors. Are these 0.3% memory errors also anyway missed by ASAN? Or, are these errors ASAN cannot even detect. That is, is HWASAN = 0.3*(ASAN + more stuff) or HWASAN = ASAN + 0.3*more_stuff, error wise?

BenWibking:
It's the former. There are more errors it can detect, but since there are only a finite number of memory tags, it will always miss a small fraction of the of them. However, the tags are assigned at random in a nonrepeatable way, so any given bug could be detected by repeating the run a large number of times.

psharda:
> It's the former. There are more errors it can detect, but since there are only a finite number of memory tags, it will always miss a small fraction of the of them. However, the tags are assigned at random in a nonrepeatable way, so any given bug could be detected by repeating the run a large number of times.

Okay, so this means something that is not new could give an error in a future PR that did not give an error in the past PR?

BenWibking:
> > It's the former. There are more errors it can detect, but since there are only a finite number of memory tags, it will always miss a small fraction of the of them. However, the tags are assigned at random in a nonrepeatable way, so any given bug could be detected by repeating the run a large number of times.
> 
> 
> 
> Okay, so this means something that is not new could give an error in a future PR that did not give an error in the past PR?

That is possible. Since the tags are randomized on each run, it is exceedingly unlikely that a given bug would be missed more than once. In any case, it will print where the bug is, so it is straightforward to check whether it's in new or old code.

psharda (approved):

PR #512, opened 2024-01-30
Make ComputeRadPressure return F and S

chongchonghe:
### Description
Make ComputeRadPressure in radiation_system.hpp return F and S instead of modifying them at place.

### Checklist

### DIFF
--- a/src/radiation_system.hpp
+++ b/src/radiation_system.hpp
@@ -1246,18 +1246,18 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 				auto fz = Fz / (c_light_ * erad);
 
 				std::array<quokka::valarray<double, 4>, 3> P{};
-        {
-          auto [F, S] = ComputeRadPressure<FluxDir::X1>(erad, Fx, Fy, Fz, fx, fy, fz);
-          P[0] = F;
-        }
-        {
-          auto [F, S] = ComputeRadPressure<FluxDir::X2>(erad, Fx, Fy, Fz, fx, fy, fz);
-          P[1] = F;
-        }
-        {
-          auto [F, S] = ComputeRadPressure<FluxDir::X3>(erad, Fx, Fy, Fz, fx, fy, fz);
-          P[2] = F;
-        }
+				{
+					auto [F, S] = ComputeRadPressure<FluxDir::X1>(erad, Fx, Fy, Fz, fx, fy, fz);
+					P[0] = F;
+				}
+				{
+					auto [F, S] = ComputeRadPressure<FluxDir::X2>(erad, Fx, Fy, Fz, fx, fy, fz);
+					P[1] = F;
+				}
+				{
+					auto [F, S] = ComputeRadPressure<FluxDir::X3>(erad, Fx, Fy, Fz, fx, fy, fz);
+					P[2] = F;
+				}
 
 				// loop over spatial dimensions
 				for (int n = 0; n < 3; ++n) {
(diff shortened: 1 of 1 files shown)


src/radiation_system.hpp
```diff
@@ -786,9 +789,9 @@ AMREX_GPU_DEVICE void RadSystem<problem_t>::ComputeRadPressure(TARRAY &F, double
 	AMREX_ASSERT(Pny != NAN);
 	AMREX_ASSERT(Pnz != NAN);
 
-	F = {Fn, Pnx, Pny, Pnz};
+	const double S = std::max(0.1, std::sqrt(Tnormal));
 
-	S = std::max(0.1, std::sqrt(Tnormal));
+	return {Fn, Pnx, Pny, Pnz, S};
```
BenWibking: I think this makes the code less readable, since the meaning of each component of the array depends on arbitrary array indices. I think it should instead return a struct, where each component is a named variable. This should not be any less efficient than returning an array.

BenWibking (changes requested):

src/radiation_system.hpp
chongchonghe: Indeed. I have changed it so it returns a struct of valarray and double.

src/radiation_system.hpp
```diff
@@ -63,6 +63,12 @@ template <typename problem_t> struct RadSystem_Traits {
 	static constexpr amrex::GpuArray<double, Physics_Traits<problem_t>::nGroups + 1> radBoundaries = {0., inf};
 };
 
+// A struct to hold the results of the ComputeRadPressure function.
+struct RadPressureResult {
+	quokka::valarray<double, 4> F;
+	double S;
```
BenWibking: can you add a comment noting this is the wavespeed for the radiation system?

src/radiation_system.hpp
```diff
@@ -63,6 +63,12 @@ template <typename problem_t> struct RadSystem_Traits {
 	static constexpr amrex::GpuArray<double, Physics_Traits<problem_t>::nGroups + 1> radBoundaries = {0., inf};
 };
 
+// A struct to hold the results of the ComputeRadPressure function.
+struct RadPressureResult {
+	quokka::valarray<double, 4> F;
```
BenWibking: Can you add a comment that these are components of the radiation pressure tensor?

src/radiation_system.hpp
```diff
@@ -1241,12 +1247,24 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
 
 				std::array<std::array<double, numRadVars_>, 3> P{};
-				double S = NAN;
-				// loop DIR over {FluxDir::X1, FluxDir::X2, FluxDir::X3}
-				// ComputeRadPressure will modify P[DIR] and S
-				ComputeRadPressure<FluxDir::X1>(P[0], S, erad, Fx, Fy, Fz, fx, fy, fz);
-				ComputeRadPressure<FluxDir::X2>(P[1], S, erad, Fx, Fy, Fz, fx, fy, fz);
-				ComputeRadPressure<FluxDir::X3>(P[2], S, erad, Fx, Fy, Fz, fx, fy, fz);
+				{
+					auto [F, S] = ComputeRadPressure<FluxDir::X1>(erad, Fx, Fy, Fz, fx, fy, fz);
```
BenWibking, line 1248: does it work if you assign it directly to `P[0]`? something like `auto [P[0], S] = ...`?

BenWibking (changes requested):

src/radiation_system.hpp
chongchonghe: It does not work. Structured bindings do not allow array elements. However, `P[0] = F` works and is simpler.

BenWibking (approved):

PR #535, opened 2024-02-13
Newton iteration on new basis (Egas, R / tau0)

chongchonghe:
### Description
Use a new basis in the Newton-Raphson iteration for the radiation-matter exchange. Instead of (Egas, Erad), we use the new basis (Egas, D), where D = R / tau0 and tau0 = c dt rho kappa^(t). This makes the iteration converge SO much faster. The code is also much simpler. The basis (Egas, R) may work equally well, but the 1/tau0 factor scales the second coordinate to the same scale as the first coordinate and should speed up the convergence rate and improve stability. 

The Jacobian of $F$ with respect to (Egas, R) is

```math
\begin{aligned}
    \frac{\partial F_G}{\partial E_g} &= 1 \\
    \frac{\partial F_G}{\partial R_i} &= \frac{c}{\hat{c}} \\
    \frac{\partial F_{R,i}}{\partial E_g} &= \frac{\partial E_{r,i}}{\partial E_g} |_{R_i = {\rm const.}} = \frac{1}{\hat{c}} \frac{1}{C_v} \frac{\partial}{\partial T} \left( \frac{\kappa_{P,i}}{\kappa_{E,i}} 4 \pi B_i \right) \\
    \frac{\partial F_{R,i}}{\partial R_i} &= \frac{\partial E_{r,i}}{\partial R_i}|_{T = {\rm const.}} - 1 = - \frac{1}{\hat{c} \Delta t \rho \kappa_{E,i}} - 1
\end{aligned}
```

For (Egas, D), simply multiple the second and fourth lines by $\tau_i^{(t)}$. 

If we assume $\kappa_{P, i} / \kappa_{E, i} = $ const., we can take this term out of the derivative, and we don't need to compute the derivative of opacity with respect to temperature at all. 

### Related issues
Should solve most 'Newton iteration fail to converge' error, even when error_tol = 1e-13. No extra break condition is required other than the one used in the original version of QUOKKA.

### Checklist

### DIFF
--- a/src/radiation_system.hpp
+++ b/src/radiation_system.hpp
@@ -1096,7 +1096,7 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 		quokka::valarray<double, nGroups_> fourPiB{};
 		quokka::valarray<double, nGroups_> EradVec_guess{};
 		quokka::valarray<double, nGroups_> kappaPVec{};
-    quokka::valarray<double, nGroups_> kappaEVec{};
+		quokka::valarray<double, nGroups_> kappaEVec{};
 		quokka::valarray<double, nGroups_> kappaFVec{};
 		quokka::valarray<double, nGroups_> tau0{}; // optical depth across c * dt at old state
 		quokka::valarray<double, nGroups_> tau{};  // optical depth across c * dt at new state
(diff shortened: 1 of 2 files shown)


BenWibking:
This looks very nice. I will try to review it in detail later today.

src/radiation_system.hpp
```diff
@@ -1121,20 +1122,23 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 
 			// BEGIN NEWTON-RAPHSON LOOP
 			Egas_guess = Egas0;
-			EradVec_guess = Erad0Vec;
+			T_gas = quokka::EOS<problem_t>::ComputeTgasFromEint(rho, Egas_guess, massScalars);
+			AMREX_ASSERT(T_gas >= 0.);
+			fourPiB = chat * ComputeThermalRadiation(T_gas, radBoundaries_g_copy);
+			kappaPVec = ComputePlanckOpacity(rho, T_gas);
+			AMREX_ASSERT(!kappaPVec.hasnan());
+			// note that I assume kappa_P = kappa_E here
```
BenWibking: I don't think we should assume $\kappa_P = \kappa_E$ for multigroup. This PR seems to require this to be the case.

BenWibking (changes requested):
I don't think assuming $\kappa_P = \kappa_E$ makes sense now that we have multigroup...

Can you write (in LaTeX) the Jacobian of the system in the new variables for reference?

src/radiation_system.hpp
markkrumholz: The algorithm that @chongchonghe and I designed should need to assume $\kappa_P = \kappa_E$, so I'm not sure why that comment is there. However, you still do need to compute $\kappa_P$ because that is the correct opacity to use to compute the matter emissivity. That is, the absorption term looks like $\kappa_E E$ for each group, but the emission term still looks like $\kappa_P(T_\mathrm{gas}) B(T_\mathrm{gas})$.

chongchonghe: See the updated description. Assuming this is for convenience, although we don't have to do that. It's safe to do so as of now because we always assume kappa_E = kappa_P in our code. In the future, we may want to do interpolation on the radiation energy and compute kappa_E based on the local radiation energy profile.

chongchonghe: Note that this assumption only affects the convergence rate of the Newton-Raphson iteration. It does not affect the result at all, even if kappa_P and kappa_E are not equally and their ratio depends on T. As a matter of fact, in all the tests where kappa depends on T (RadMarshakAsymptotic, RadhydroPulse, RadhydroPulseMG), you can set ComputePlanckOpacityTempDerivative to return 0.0 and the code with the old base variables would pass the tests without any problem.

BenWibking: Ok, I think it makes sense now. I was also confused as to why the derivative of opacity w/r/t temperature dropped out.

I would strongly prefer that this is also documented in comments in the code, since we don't yet have a new code paper for this algorithm. Can you add your PR description to the source code? Or at least the comment about why the opacity derivatives drop out, and the LaTeX version of the Jacobian?

BenWibking: Also, I think the current assumption that $\kappa_E = \kappa_P$ needs to be prominently highlighted in a comment at the start of this function.

chongchonghe: OK. I added back kappa_E and now kappa_E and kappa_P are allowed to differ. The only place where I assume kappa_E = kappa_P is at the calculation of the Jacobian. And again, this does not affect the converged result. 

I've also added sufficient comments to the code and made everything clear.

BenWibking:
I think the GPU machines need to be restarted. I'll do that and then the tests should run there.

BenWibking:
Ok, just waiting on the GPU tests to pass in order to approve.

markkrumholz:
Looks like this has passed the GPU tests now. @BenWibking, can you resolve your change request so this can be merged?

BenWibking (approved):

PR #536, opened 2024-02-13
Use fully implicit scheme for flux update

chongchonghe:
### Description
Before this PR, the gas momentum is delayed when updating radiation flux using a backward Euler scheme. Now, a set of 3N + 3 linear equations is solved for the updated v and F_{r, i}. This is fully tested on all test problems with success. However, this implementation brings in a lot more computation but does not significantly change the result of any of the current tests, therefore I keep it disabled by default. I have enabled it in the shock_multigroup test.

By default, we update the flux using

```math
\mathbf{F}_{r, i}^{(t+1)} = \frac{\mathbf{F}_{r, i}^{(t)} + (\kappa_P^{(t+1)} (4 \pi B)^{(t+1)} \mathbf{v}^{(t)} + c \kappa_F^{(t+1)} \mathbf{v}^{(t)}:\mathbf{P}_r^{(t)} ) \rho \hat{c} \Delta t / c}{1+\rho \kappa_{F, i}^{(t+1)} \hat{c} \Delta t}
```

Optionally, by adding the following line in RadSystem_Traits, 
```
  static bool flux_full_implicit_update; // NOLINT
```
one switch on the fully implicit scheme where the following set of $3N+3$ linear equations is solved for $\boldsymbol{v}^{(t+1)}, \boldsymbol{F}_{r, i}^{(t+1)}$. 

```math
    \frac{\boldsymbol{F}_{r,i}^{(t+1)} - \boldsymbol{F}_{r,i}^{(t)}}{\Delta t} = -\hat{c} \rho \kappa_{F,i}^{(t+1)} \boldsymbol{F}_{r,i}^{(t+1)} + \rho \kappa_{P,i}^{(t+1)}\left(4 \pi B_i^{(t+1)} \right) \frac{\hat{c}}{c} \boldsymbol{v}^{(t+1)} + \hat{c} \rho \kappa_{F,i}^{(t+1)} \boldsymbol{v}^{(t+1)} :\boldsymbol{T}_{i}^{(t)} E_{r,i}^{(t+1)},
```

```math
   \frac{\rho \boldsymbol{v}^{(t+1)} - \rho \boldsymbol{v}^{(t)}}{\Delta t} = -\frac{1}{\hat{c}c} \left[ -\hat{c} \rho \sum_i \kappa_{F,i}^{(t+1)} \boldsymbol{F}_{r,i}^{(t+1)} + \rho \frac{\hat{c}}{c} \boldsymbol{v}^{(t+1)} \sum_i \kappa_{P,i}^{(t+1)} 4 \pi B_i^{(t+1)} + \hat{c} \rho \sum_i \kappa_{F,i}^{(t+1)} \boldsymbol{v}^{(t+1)} :\boldsymbol{T}_{i}^{(t)} E_{r,i}^{(t+1)} \right],
```

Note that the only term that is delayed in this set of equations is the Eddington tensor $T_i$, but the error this brings is very minimal because $T_i = 1/3 ~I$ is constant in the diffusion limit, which is the only regime where the implicit scheme matters. 

The matrix inversion is done in $O(3N)$ steps using a dedicated Gaussian elimination method. The extra memory cost is $3^2 + 3^2 = 18$ double precision floats. 

### Related issues
I am yet to find a test problem where this change will make a difference. It should be significant when there is a stiff change in both gas momentum and radiation flux. This is probably seen in radiation-driven shocks.

### Checklist

### DIFF
--- a/src/RadhydroShockMultigroup/test_radhydro_shock_multigroup.cpp
+++ b/src/RadhydroShockMultigroup/test_radhydro_shock_multigroup.cpp
@@ -66,8 +66,8 @@ template <> struct RadSystem_Traits<ShockProblem> {
 	static constexpr double radiation_constant = a_rad;
 	static constexpr double Erad_floor = Erad_floor_;
 	static constexpr bool compute_v_over_c_terms = true;
-  // Note that flux_full_implicit_update should not be defined as constexpr.
-  static bool flux_full_implicit_update; // NOLINT
+	// Note that flux_full_implicit_update should not be defined as constexpr.
+	static bool flux_full_implicit_update;		  // NOLINT
 	static constexpr double energy_unit = C::hplanck; // set boundary unit to Hz
 	static constexpr amrex::GpuArray<double, Physics_Traits<ShockProblem>::nGroups + 1> radBoundaries{
 	    1.00000000e+15, 3.16227766e+15, 1.00000000e+16, 3.16227766e+16, 1.00000000e+17, 3.16227766e+17, 1.00000000e+18, 3.16227766e+18, 1.00000000e+19};
--- a/src/RadhydroShockMultigroup/test_radhydro_shock_multigroup.cpp
+++ b/src/RadhydroShockMultigroup/test_radhydro_shock_multigroup.cpp
@@ -66,6 +66,8 @@ template <> struct RadSystem_Traits<ShockProblem> {
 	static constexpr double radiation_constant = a_rad;
 	static constexpr double Erad_floor = Erad_floor_;
 	static constexpr bool compute_v_over_c_terms = true;
+  // Note that flux_full_implicit_update should not be defined as constexpr.
+  static bool flux_full_implicit_update; // NOLINT
 	static constexpr double energy_unit = C::hplanck; // set boundary unit to Hz
 	static constexpr amrex::GpuArray<double, Physics_Traits<ShockProblem>::nGroups + 1> radBoundaries{
 	    1.00000000e+15, 3.16227766e+15, 1.00000000e+16, 3.16227766e+16, 1.00000000e+17, 3.16227766e+17, 1.00000000e+18, 3.16227766e+18, 1.00000000e+19};
--- a/src/radiation_system.hpp
+++ b/src/radiation_system.hpp
@@ -69,20 +69,16 @@ struct RadPressureResult {
 };
 
 // Define a helper type trait to check for member existence
-template <typename T>
-struct has_member_flux_full_implicit_update {
-private:
-    template <typename C>
-    static constexpr auto check(C*) -> typename std::is_same<decltype(C::flux_full_implicit_update), bool>::type;
+template <typename T> struct has_member_flux_full_implicit_update {
+      private:
+	template <typename C> static constexpr auto check(C *) -> typename std::is_same<decltype(C::flux_full_implicit_update), bool>::type;
 
-    template <typename>
-    static constexpr auto check(...) -> std::false_type;
+	template <typename> static constexpr auto check(...) -> std::false_type;
 
-public:
-    using type = decltype(check<T>(nullptr));
-    static constexpr bool value = type::value;
+      public:
+	using type = decltype(check<T>(nullptr));
+	static constexpr bool value = type::value;
 };
-
 
 /// Class for the radiation moment equations
 ///
@@ -127,10 +123,10 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
 
 	static constexpr bool compute_v_over_c_terms_ = RadSystem_Traits<problem_t>::compute_v_over_c_terms;
 
-  // Set flux_full_implicit_update_ to true if it is defined, otherwise default to false.
-  static constexpr bool flux_full_implicit_update_ = []() constexpr {
-    return has_member_flux_full_implicit_update<RadSystem_Traits<problem_t>>::value;
-  }();
+	// Set flux_full_implicit_update_ to true if it is defined, otherwise default to false.
+	static constexpr bool flux_full_implicit_update_ = []() constexpr {
+		return has_member_flux_full_implicit_update<RadSystem_Traits<problem_t>>::value;
+	}();
 
 	static constexpr int nGroups_ = Physics_Traits<problem_t>::nGroups;
 	static constexpr amrex::GpuArray<double, nGroups_ + 1> radBoundaries_ = []() constexpr {
@@ -204,7 +200,8 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
 							 const double &y0, const quokka::valarray<double, nGroups_> &yi, double &x0,
 							 quokka::valarray<double, nGroups_> &xi);
 
-  AMREX_GPU_HOST_DEVICE static auto Solve3x3matrix(double C00, double C01, double C02, double C10, double C11, double C12, double C20, double C21, double C22, double Y0, double Y1, double Y2) -> std::tuple<double, double, double>;
+	AMREX_GPU_HOST_DEVICE static auto Solve3x3matrix(double C00, double C01, double C02, double C10, double C11, double C12, double C20, double C21,
+							 double C22, double Y0, double Y1, double Y2) -> std::tuple<double, double, double>;
 
 	AMREX_GPU_HOST_DEVICE static auto ComputePlanckEnergyFractions(amrex::GpuArray<double, nGroups_ + 1> const &boundaries, amrex::Real temperature)
 	    -> quokka::valarray<amrex::Real, nGroups_>;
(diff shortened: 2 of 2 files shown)


BenWibking:
Could you write out the set of equations that you are solving in this PR? I've lost track all of the terms that have been changed recently.

Can you also explicitly specify which terms were previously treated explicitly and are now fully implicit?

In the original code, I think the v/c terms were calculated based on the old state before the matter-energy exchange. Are these now updated implicitly along with the radiation flux?

BenWibking:
This is definitely mathematically appealing. However, I am not sure whether the v/c terms are stiff in the nonrelativistic regime. I think in ORION they are always treated explicitly (i.e., using v from time $t$) -- @markkrumholz ?

markkrumholz:
ORION2 does indeed always treat these terms explicitly. However, ORION2 also does not work in the dynamic diffusion regime, which is where these terms become important, so that's not very meaningful.

Those terms are never going to be stiff in the sense that they evolve on a timescale faster than the hydro timestep -- these terms can be understood as representing advection of radiation by gas, so they always operate on the advective time scale. Thus l suspect that you never get into a situation where you need to do backwards Euler on these terms for stability. However, it is definitely more consistent to difference them in time the same way as the other radiation terms. Whether that actually improves results in practice is another question. If there are any situations where it does, they will be in the dynamic diffusion regime, because that's the only place where those terms are at the same order as the radiation-matter exchange terms.

BenWibking:
> ORION2 does indeed always treat these terms explicitly. However, ORION2 also does not work in the dynamic diffusion regime, which is where these terms become important, so that's not very meaningful.

I am a bit puzzled by this. Why doesn't it work in the dynamic diffusion regime? FLD should generically work in the dynamic diffusion regime, right?

markkrumholz:
> > ORION2 does indeed always treat these terms explicitly. However, ORION2 also does not work in the dynamic diffusion regime, which is where these terms become important, so that's not very meaningful.
> 
> I am a bit puzzled by this. Why doesn't it work in the dynamic diffusion regime? FLD should generically work in the dynamic diffusion regime, right?

The problem isn't the FLD part. The problem is that treating the v/c terms using a simple operator split forward-Euler resulted in an instability in ORION.

BenWibking:
@chongchonghe Ok, I am fine with adding this to the code.

However, I am not certain that the existing radiative shock problem is in the dynamic diffusion regime. At least, the side of the shock with the temperature spike is optically thin. The other side of the shock might only be in the static diffusion regime -- can you check?

chongchonghe:
> @chongchonghe Ok, I am fine with adding this to the code.
> 
> However, I am not certain that the existing radiative shock problem is in the dynamic diffusion regime. At least, the side of the shock with the temperature spike is optically thin. The other side of the shock might only be in the static diffusion regime -- can you check?

I'll try the advecting pulse test in the dynamic diffusion regime. However, such a test ($\tau = 2e4, \beta = 3e-3, \beta \tau = 60$ only works well with very high resolution (512 grids) and takes a long time to run. Therefore, it is not suitable for a regular test. I will test it on my computer. 

Besides, I still need to update the equations as shown in the description to add the (kappa_P - kappa_F) term. Fortunately, it's not too much work.

@markkrumholz 
I'm not sure why the advecting pulse test in the dynamic diffusion regime works well with nx = 512 while it fails (with 30% relative error) with 64 grids. It would be nice to have a regular test for the dynamic diffusion regime.

BenWibking:
We should probably create a short version of the test suite and a "long" version of the test suite, so we can still include it in the "long" version. We could run the longer version nightly or weekly, rather than on each PR.

However, I agree something strange is going on if it requires that many cells to resolve. This seems like an order reduction phenomenon, or maybe just a bug.

chongchonghe:
Compare the plots below for the dynamic diffusion test ($\beta \tau = 60$) at different resolutions. The instability issue in the non-advecting pulse is a big problem. We should continue our discussion on this issue. 

The asymmetry in the advecting pulse at low resolution is also suspicious. This could come from a bug in the code, or fundamental flaw in our method. The beta in this run is 3e-3, so beta^2 = 1e-5, not a negligible number. So, the radiation four-force accurate to the first order of beta might not be able to capture this. This also explains why it works well only at high resolution. If we really care about dynamic diffusion limit, we probably need to include the O(beta^2) terms. It's pretty straight forward to add those terms in the semi-implicit formula, although it's not feasible anymore to solve the fully implicit equations. 

64 grids,
CleanShot 2024-02-18 at 11 22 04

512 grids,
CleanShot 2024-02-18 at 11 17 40

markkrumholz:
The asymmetry could well be an issue of missing the next-order terms. The way the ordering works in RHD is that you can't just expand in powers of $\beta$, because in the diffusion limit, $\tau \gg 1$, there are actually two small parameters, $\beta$ and $\tau^{-1}$. Dynamic diffusion is defined by $\beta\tau \gg 1$, which means that the term ordering is $\beta\tau \gg 1 \gg \beta^2 \tau \gg \beta \ldots$. We have get the $\beta\tau$ and $\beta$ terms, but perhaps not the $\beta^2\tau$ ones, which we might need to re-add. Not sure that will fix the instability issue, but it may well be necessary to avoid the asymmetry.

I should add: working out this ordering is the explicit purpose of my 2007 paper, so that it the place to start looking.

markkrumholz:
Thinking about this some more, I think the solution to the instability issue may be in the CASTRO paper. My recollection is that the fundamental source of the instability in dynamic diffusion is that the radiation transport step is basically irrelevant, because the optical depth is so high that the amount of energy transferred between cells is tiny compared to the amount transferred by advection / work. The reason the instability arises is that the advection and work terms are being operator-split from the hydro advection step. So this would be analogous to a scheme where we do a Godunov update for hydrodynamics, but we don't update the energy during that update, and instead we then do a separate, operator-split update to the gas energy after the fact, and with a different time centering than was used during the hydro update. That's effectively what we're doing here. My recollection is that the CASTO paper (not paper III, which describes multigroup, but paper I or II) describes a way around this by moving the work terms into the hydro update step.

BenWibking:
> Thinking about this some more, I think the solution to the instability issue may be in the CASTRO paper. My recollection is that the fundamental source of the instability in dynamic diffusion is that the radiation transport step is basically irrelevant, because the optical depth is so high that the amount of energy transferred between cells is tiny compared to the amount transferred by advection / work. The reason the instability arises is that the advection and work terms are being operator-split from the hydro advection step. So this would be analogous to a scheme where we do a Godunov update for hydrodynamics, but we don't update the energy during that update, and instead we then do a separate, operator-split update to the gas energy after the fact, and with a different time centering than was used during the hydro update. That's effectively what we're doing here. My recollection is that the CASTO paper (not paper III, which describes multigroup, but paper I or II) describes a way around this by moving the work terms into the hydro update step.

If this is the problem, then it should be possible to reduce the splitting error by several orders of magnitude by using an integration scheme to couple hydro and radiation that is very similar to the IMEX scheme we use to couple the hyperbolic radiation system and radiation four-force terms. Such a scheme is demonstrated here: https://arxiv.org/abs/2305.05452.

BenWibking:
Although here the instability only appears when $v=0$, so I think that might not be the solution. I think it is caused by of the odd-even decoupling of the stencil that appears when writing a diffusion equation in first order form with an upwind discretization.

I think it might be fixed by using the $1/\tau$ correction in the Riemann solver (rather than the $1/\tau^{-2}$ correction) and then disabling this correction at odd-indexed interfaces.

Alternatively, we could disable PPM reconstruction at interfaces where the mean free path is unresolved in the adjacent cells and simply average the values from the adjacent cells.

markkrumholz:
Definitely worth trying this first, before we go mucking with the more complex fix I suggested.

BenWibking:
I thought about this a bit more, inspired by the paper I linked (https://arxiv.org/abs/2305.05452).

It may be significantly more stable to do the $O(v/c)$ terms explicitly if they are done in an operator-split step *before* the radiation update (see section 3.1 of the paper and the "radiation material motion correction"). I think this makes sense, because since they only advect the radiation on the hydro timescale. Then the radiation emission/absorption can be done afterward, which proceeds on a much shorter timescale.

In general, ordering by increasing stiffness is the most accurate operator splitting (https://doi.org/10.1006/jcph.2000.6495).

chongchonghe:
> I thought about this a bit more, inspired by the paper I linked (https://arxiv.org/abs/2305.05452).
> 
> It may be significantly more stable to do the O(v/c) terms explicitly if they are done in an operator-split step _before_ the radiation update (see section 3.1 of the paper and the "radiation material motion correction"). I think this makes sense, because since they only advect the radiation on the hydro timescale. Then the radiation emission/absorption can be done afterward, which proceeds on a much shorter timescale.
> 
> In general, ordering by increasing stiffness is the most accurate operator splitting (https://doi.org/10.1006/jcph.2000.6495).

Did you mean "It may be significantly more stable to do the $O(v/c)$ terms **implicitly** if they are done ... *before* the radiation update"? If done explicitly, it's the same idea as Mark's and it will be done before the radiation update anyway.

chongchonghe:
> Thinking about this some more, I think the solution to the instability issue may be in the CASTRO paper. My recollection is that the fundamental source of the instability in dynamic diffusion is that the radiation transport step is basically irrelevant, because the optical depth is so high that the amount of energy transferred between cells is tiny compared to the amount transferred by advection / work. The reason the instability arises is that the advection and work terms are being operator-split from the hydro advection step. So this would be analogous to a scheme where we do a Godunov update for hydrodynamics, but we don't update the energy during that update, and instead we then do a separate, operator-split update to the gas energy after the fact, and with a different time centering than was used during the hydro update. That's effectively what we're doing here. My recollection is that the CASTO paper (not paper III, which describes multigroup, but paper I or II) describes a way around this by moving the work terms into the hydro update step.

I recall in CASTRO paper II they use FLD approximation, so there is no flux update like we have. I'll take a look at the paper I. 

I think it's not just the instability that we need to fix. The code fails to produce the right result in the dynamic diffusion limit unless with super high resolution. However, the instability only occurs when $\beta = 0$.

I thought the whole point of doing implicit update of the flux is to maintain the correct asymptotic behavior in the diffusion regime, as Skinner+2019 suggested. I can also (kind of) see this as I write down the proof of the asymptotic preserving property of the IMEX scheme. I'll take a closer look at it later.

chongchonghe:
With the new odd-even instability fix, the noise in velocity is much smaller, although it's still there. The noise in the temperature curve disappeared completely, which is good. However, the change of the temperature curve is tiny, suggesting that the odd-even instability issue is not the primary cause of the error. We do need the $\beta^2 \tau$ terms. 

CleanShot 2024-02-20 at 16 03 42

chongchonghe:
> The asymmetry could well be an issue of missing the next-order terms. The way the ordering works in RHD is that you can't just expand in powers of β, because in the diffusion limit, τ≫1, there are actually two small parameters, β and τ−1. Dynamic diffusion is defined by βτ≫1, which means that the term ordering is βτ≫1≫β2τ≫β…. We have get the βτ and β terms, but perhaps not the β2τ ones, which we might need to re-add. Not sure that will fix the instability issue, but it may well be necessary to avoid the asymmetry.
> 
> I should add: working out this ordering is the explicit purpose of my 2007 paper, so that it the place to start looking.

If I understand the Table 1 in your 2007 paper correctly, the (v/c)^2 terms in Eq. 25 is of order $\beta^2$ in static diffusion limit and of order $\beta^3 \tau = 6 \times 10^{-3}$ in dynamic diffusion limit, which is quite significant. I'll add that term in.

chongchonghe:
OK. I have shown that G (up to $O(\beta^2 \tau)$) in dynamic diffusion limit does not vanish if you plug in Eq. 31 and 32 of your 2007 paper into Eq. 25. What you get is $\beta^2 \tau F$​. Here's the math:

Consider $E = E_0, F = 0$ in comoving frame of uniform media moving with velocity $v$.  In lab frame, 

$$
E = (1 + \frac{4}{3} \beta^2) E_0
$$

$$
F = \frac{4}{3} v E_0
$$

From Eq. 25, assuming kappa_R = kappa_P,

$$
-\frac{c^2 G}{\rho \kappa} = - c F + v c (\frac{4\pi B}{c} - E) + v c (E + P) - \frac{1}{2} \beta^2 c F
$$

Assuming $\frac{4\pi B}{c} - E = \frac{4}{3} \beta^2 E$ as suggested in the appendix, we get

$$
\begin{align}
{\rm LHS} &= -\frac{4}{3} vcE_0 + \frac{4}{3} vc \beta^2 E + \frac{4}{3} v c E - \frac{1}{6} \beta^2 v c E_0 \\
&= -\frac{4}{3} vcE_0 + \frac{4}{3} vc E (1 + \beta^2) - \frac{1}{6} \beta^2 v c E_0 \\
&\approx -\frac{4}{3} vcE_0 + \frac{4}{3} vc E_0 (1 + \beta^2) (1 + \frac{4}{3} \beta^2) - \frac{1}{6} \beta^2 v c E_0 \\
&\approx -\frac{4}{3} vcE_0 + \frac{4}{3} vc E_0 (1 + \frac{7}{3} \beta^2) - \frac{1}{6} \beta^2 v c E_0 \\
&\approx 3 \beta^2 c v E_0
\end{align}
$$

By seting $\tau = \rho \kappa c \Delta t$ we get

$$
-c^2 G \Delta t \approx 3 \beta^2 \tau v E_0 \approx 3 \beta^2 \tau F
$$

Notice that $\beta^2 \tau$​ is the next term after 1 in your term ordering, so it's quite significant. Actually, $\beta^2 \tau \approx 0.2$​​ in this test. 

If instead $\tau = \rho \kappa v \Delta t$, as the flux propogation is dominated by advection, it becomes $3 \beta \tau F$. Even bigger. 

If we assume $\frac{4\pi B}{c} - E = 0$ instead, the result is $1.6 \beta^2 \tau F$. 

@markkrumholz 
Am I missing something? In the Eq. (A1) of your paper, you have $\gamma^2$ in the energy expression. I don't know where that comes from.

chongchonghe:
This implementation does not bring any foreseeable benefits, so I am gonna close this PR.

Issue #447, opened 2023-11-12
[HIP] memory errors affecting Microphysics codes on AMD GPU


BenWibking:
**Describe the bug**
AMD's AddressSanitizer reports a memory corruption bug in `amrex::InitRandom()`. It's currently unclear if this is a real bug or a false positive.

**To Reproduce**
1. Compile the HydroBlast3D problem with the `moth-sanitizer.profile` settings on Moth (https://github.com/quokka-astro/quokka/pull/446).
2. Run the problem.
3. See error from ASAN:
```console
bwibking@moth:~/quokka/build> ./src/HydroBlast3D/test_hydro3d_blast ../tests/benchmark_unigrid_256.in
Initializing AMReX (23.10-23-g601cc4ee80e0)...
MPI initialized with 1 MPI processes
MPI initialized with thread support level 0
Initializing HIP...
HIP initialized with 1 device.
=================================================================
==1379366==ERROR: AddressSanitizer: global-buffer-overflow on address 0x00000326cfe8 at pc 0x7fc7796a5ea7 bp 0x7ffd90522b00 sp 0x7f
fd905222c0
READ of size 32 at 0x00000326cfe8 thread T0
    #0 0x7fc7796a5ea6 in __interceptor_memcpy (/opt/rocm-5.7.0/llvm/lib/clang/17.0.0/lib/linux/libclang_rt.asan-x86_64.so+0xa5ea6)
(BuildId: e2f6676d7d0ade0de2c4ac32fa5856892b18b70a)
    #1 0x7fc7781440a9  (/opt/rocm-5.7.0/lib/libamdhip64.so.5+0x3440a9) (BuildId: 7342fbe1c361ada40d7aa3c1da36c32f3fbe143d)
    #2 0x7fc7781462f6  (/opt/rocm-5.7.0/lib/libamdhip64.so.5+0x3462f6) (BuildId: 7342fbe1c361ada40d7aa3c1da36c32f3fbe143d)
    #3 0x7fc7781465a6  (/opt/rocm-5.7.0/lib/libamdhip64.so.5+0x3465a6) (BuildId: 7342fbe1c361ada40d7aa3c1da36c32f3fbe143d)
    #4 0x7fc778112434  (/opt/rocm-5.7.0/lib/libamdhip64.so.5+0x312434) (BuildId: 7342fbe1c361ada40d7aa3c1da36c32f3fbe143d)
[...]

Issue #394, opened 2023-09-21
AMD GPUs produce incorrect results


psharda:
For the exact same setup, a Pop III simulation on Setonix GPUs show a very different evolution as compared to CPUs. The initial density projections on both are identical:

![plt00000_Projection_z_gasDensity_gasDensity](https://github.com/quokka-astro/quokka/assets/34922596/7f65a7c7-ea83-4a0f-acd7-4eb7c995d4f4)

However, after a few 100 steps, GPU projections look weird whereas CPU projections look more reasonable. GPU projection:

![plt00900_Projection_z_gasDensity_gasDensity](https://github.com/quokka-astro/quokka/assets/34922596/5ce599dc-c4f9-4d86-93d7-a9b50a72fc5c)

CPU projection:

![plt00900_Projection_z_gasDensity_gasDensity](https://github.com/quokka-astro/quokka/assets/34922596/85530bac-4560-4edb-9e6e-de1c8ea23970)

GPU run is much faster, but it needs flux correction in several cells in each timestep almost from t = 0, whereas CPU run doesn't need any flux corrections. At some point, the GPU run aborts because of rho = 0 or NAN during regridding.

There seems to be some issue with Setonix GPUs.

BenWibking:
cc @markkrumholz @aditivijayan

markkrumholz:
Can you try this on Gadi GPU and see what happens?

markkrumholz:
Also to test: run one time step only on both Setonix CPU and GPU. The result should be bit-wise identical. Are they, and, if not, are they different in all cells, or just some?

psharda:
@markkrumholz - the last I tried on Gadi GPUs, they gave a memory error. Lately, the runs on gpuvolta are being put on hold, or do not start for hours. Is there an express gpu queue for testing and debugging on gadi?

BenWibking:
There is no express GPU queue. Can you run on the RSAA GPU nodes at NCI?

psharda:
After one timestep, the min/max ratio of densities of the CPU to the GPU runs is:

```
In [17]: np.min(dens_gpu / dens_cpu), np.max(dens_gpu / dens_cpu)
Out[17]:
(unyt_quantity(0.00999067, '(dimensionless)'),
 unyt_quantity(100.00105867, '(dimensionless)'))
```

[...]

PR #539, opened 2024-02-16
Increase max timesteps to 300 for the PopIII Test

psharda:
### Description
This PR increases the maximum steps to run the PopIII test problem to 300, to compare the evolution on AMD and NVIDIA GPUs (which starts to differ around step 58).

### Related issues
Will further help debug #394 #447 

### Checklist
- [x ] I have added a link to any related issues see (see above).
- [x ] I have tested this PR on my local computer and all tests pass.
- [x ] I have manually triggered the GPU tests with the magic comment `/azp run`.
- [x ] I have requested a reviewer for this PR.

### DIFF
--- a/tests/PopIII.in
+++ b/tests/PopIII.in
@@ -22,7 +22,7 @@ amr.grid_eff        = 0.7   # default
 
 hydro.reconstruction_order = 3  # PPM
 cfl = 0.15
-max_timesteps = 10
+max_timesteps = 300
 stop_time = 1e16
 
 do_reflux = 1


psharda:
Ah, of course the ROCm test does not pass. Is it still worth accepting this PR? I created it so that the AMD experts do not have to manually change the max number of timesteps when they try our test. Thoughts - @BenWibking @markkrumholz ?

BenWibking:
> Ah, of course the ROCm test does not pass. Is it still worth accepting this PR? I created it so that the AMD experts do not have to manually change the max number of timesteps when they try our test. Thoughts - @BenWibking @markkrumholz ?

They really want a standalone reproducer with just the one function from chemistry.hpp that they can compile (without the rest of Quokka), so I think that should just go in its own GitHub repository.

BenWibking:
@psharda It turns out that this magic set of compiler flags fixes the problem for Castro:
```
CXXFLAGS="-mllvm -amdgpu-function-calls=true"
```
We should check whether it fixes it for us too.

psharda:
I inserted this line in `quokka/CMakeLists.txt`:

```
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mllvm -amdgpu-function-calls=true")
```

but during compilation I get this warning:

```
clang: warning: argument unused during compilation: '-mllvm -amdgpu-function-calls=true' [-Wunused-command-line-argument]
```

I get the same warning if I instead do `set(CMAKE_CXX_FLAGS "-mllvm -amdgpu-function-calls=true")`

The CXXFLAGS being set during compilation are:

```
--    C++ flags                = -O3 -DNDEBUG -mllvm -amdgpu-function-calls=true --offload-arch=gfx90a -m64 -munsafe-fp-atomics -std=c++17 -fgpu-rdc -x hip

```
I guess the rest of these flags come from amrex?

As an alternative, I replaced the export CXXFLAGS line in `quokka/scripts/moth.profile` with `export CXXFLAGS="-mllvm -amdgpu-function-calls=true" ` but I get the same compilation warning as above.

However, in all the three instances, despite the compilation warning, if I run the PopIII test, it does NOT fail with a memory access fault, but because of FOFC and subsequent NaNs in hydro which makes hydro crash.

Shall I disable the other CXXFLAGS that amrex sets, and try with just the ones you suggested above?

Perhaps unrelated. but anytime I build on moth, I get this message:

```
CMake Warning at extern/amrex/Tools/CMake/AMReXParallelBackends.cmake:257 (message):
  You are using the legacy wrapper 'hipcc' as the HIP compiler.

  Use AMD's 'clang++'/'amdclang++' compiler, or on Cray the CC compiler
  wrapper instead.
```

BenWibking:
All of these warnings are irrelevant. The flags work fine on files compiled for GPU. CMake does not have a way to specify separate flags for host and device code when compiling with HIP. The warning about hipcc was added to AMReX a few years ago and does not apply anymore.

Good news! It solved the memory access error. If you still want to debug this, I think the next step would be to figure out why the hydro is crashing. I think the possibilites are either i) the chemistry network is still giving a different answer on AMD (turning hydro off and `fcompare`ing the outputs), or ii) the hydro update is giving a different answer on AMD (by turning chemistry off and `fcompare`ing the outputs).

BenWibking:
@psharda The AMD folks are still interested in this example code so they can improve/fix their compiler.

Since we now have a workaround (that at least appears to work so far), this is probably less important for us, but if you are interested, I wanted to let you know.

(Of course, this all depends on verifying that we get roundoff-identical results on AMDGPU, NVIDIA, and CPU when using the current Quokka+Microphysics.)

BenWibking:
@psharda Is this PR needed anymore, or can it be closed?

BenWibking:
Ok, I ran this with ROCm 6.3.1 and it works without any burn failures!!
Full log: [popiii_log.txt](https://github.com/user-attachments/files/18350514/popiii_log.txt)

PR #540, opened 2024-02-19
[WIP] characteristic reconstruction

BenWibking:
### Description
This does PPM reconstruction for hydro in the characteristic variables defined by the eigensystem of $[\rho^{-1}, u, P, e_{int}]^T$. (Following CW84, we use the specific volume $V \equiv 1/\rho$ instead of density $\rho$.)

This reduces the post-shock oscillations for several test problems. Balsara claims that characteristic reconstruction has improved entropy stability compared to primitive reconstruction [1].

However, the HydroHighMach test does not appear to converge to the correct answer with characteristic reconstruction. (This may be due to the way it falls back to primitive variable reconstruction if characteristic reconstruction produces a negative density or negative pressure state.)

*More testing is needed to determine whether this remains stable in isobaric thermally unstable flows (otherwise, it is necessary to use the temperature eigensystem).*

[1] https://ui.adsabs.harvard.edu/abs/2017LRCA....3....2B/abstract

### Related issues
N/A

### Checklist


BenWibking:
It turns out this is not a good idea for our problems. The gradients are too steep for characteristic reconstruction to work, so it goes negative. This cannot be fixed by falling back to primitive reconstruction cell-by-cell.

Issue #541, opened 2024-02-20
Odd-even instability error in diffusion limit


chongchonghe:
**Describe the bug**
The advecting radiation pulse test has what seems to be an odd-even instability issue when the pulse is not advecting (v = 0) and it goes away when v > 0. 

Method 1: 

`epsilon = {S_corr * S_corr, S_corr, S_corr, S_corr}` for even indices and `epsilon = {S_corr, 1.0, 1.0, 1.0}` for odd indices.

![radhydro_pulse_velocity original](https://github.com/quokka-astro/quokka/assets/24463821/71bfc158-e7bc-4e65-9c78-3139444059ef)

Method 2:

`epsilon = {S_corr, 1.0, 1.0, 1.0}` for even indices and `epsilon = {1.0, 1.0, 1.0, 1.0}` for odd indices.

![radhydro_pulse_velocity](https://github.com/quokka-astro/quokka/assets/24463821/05e9b8a5-3811-4f90-84d8-0bcdb11876b4)

Method 2 seems to work well for the MarshakAsymptotic problem as well: compare method 1 and method 2:

![CleanShot 2024-02-20 at 13 18 26](https://github.com/quokka-astro/quokka/assets/24463821/6ace4aa9-703b-49f9-a80a-741afe72becf)

**To Reproduce**
1. Compile this problem 'test_radhydro_pulse'
2. Run this problem 'test_radhydro_pulse'
3. See the error in the velocity curve.

BenWibking:
Can you fill out the sections requested in the description for this issue?

BenWibking:
My proposed solution is to modify the reconstruction so that Equation 27 of [1] is obtained in the diffusion limit. This will require interpolating between the standard reconstruction and Equation 27 as a function of $\tau_j$ and $\tau_{j+1}$.

[1] https://ui.adsabs.harvard.edu/abs/2001JQSRT..69..475L/abstract

chongchonghe:
@BenWibking @markkrumholz 
See PR #544 . What Ben suggested is also used in the THC_M1 code (Radice2022). So, I followed their approach and also tried a few variations. However, none of them worked. I'll have to other methods, like switching the order of flux and energy update.

BenWibking:
No, I don't think it is the case that my suggestion above is the same method as that used by THC_M1. (It is a little unclear since the notation in that paper is not fully defined in the text.)

[...]

PR #544, opened 2024-02-28
Implemented Radice2022's approach to fix odd-even instability 

chongchonghe:
Following Section 3.1 of https://ui.adsabs.harvard.edu/abs/2022MNRAS.512.1499R 

However, this does not solve the problem. The instability persists. I also tried making the following changes, but none of them worked
1. Set epsilon to {1.0, 1.0, 1.0, 1.0} at odd indices instead of at local zigzags.
2. Start with `epsilon = {S_corr, 1.0, 1.0, 1.0}`.
3. Remove flux_limiter (set it to 1).
4. Check the local minima and maxima of gas velocity instead of radiation quantities.

### Description
Please write a short paragraph describing what changes this pull request makes to the code. It should provide sufficient context so that a reviewer can efficiently browse the source code changes.

### Related issues

### Checklist

### DIFF
--- a/src/radiation_system.hpp
+++ b/src/radiation_system.hpp
@@ -435,21 +435,21 @@ void RadSystem<problem_t>::ReconstructStatesPLM(arrayconst_t &q_in, array_t &lef
 template <typename problem_t>
 template <FluxDir DIR>
 void RadSystem<problem_t>::ReconstructStatesAVE(arrayconst_t &q_in, array_t &leftState_in, array_t &rightState_in, amrex::Box const &indexRange,
-						     const int nvars)
+						const int nvars)
 {
 	// construct ArrayViews for permuted indices
 	quokka::Array4View<amrex::Real const, DIR> q(q_in);
 	quokka::Array4View<amrex::Real, DIR> leftState(leftState_in);
 	quokka::Array4View<amrex::Real, DIR> rightState(rightState_in);
 
-	amrex::ParallelFor(indexRange, nvars, [=] AMREX_GPU_DEVICE(int i_in, int j_in, int k_in, int n) noexcept { 
-    // permute array indices according to dir 
-    auto [i, j, k] = quokka::reorderMultiIndex<DIR>(i_in, j_in, k_in);
+	amrex::ParallelFor(indexRange, nvars, [=] AMREX_GPU_DEVICE(int i_in, int j_in, int k_in, int n) noexcept {
+		// permute array indices according to dir
+		auto [i, j, k] = quokka::reorderMultiIndex<DIR>(i_in, j_in, k_in);
 
-    const auto ave = 0.5 * (q(i - 1, j, k, n) + q(i, j, k, n));
+		const auto ave = 0.5 * (q(i - 1, j, k, n) + q(i, j, k, n));
 		leftState(i, j, k, n) = ave;
 		rightState(i, j, k, n) = ave;
-  });
+	});
 }
 
 template <typename problem_t>
@@ -906,9 +906,9 @@ void RadSystem<problem_t>::ComputeFluxes(array_t &x1Flux_in, array_t &x1FluxDiff
 			// const double tau_corr = 1.0;  // optically thin, no correction
 			// const double tau_corr = 0.0;  // optically thick
 
-      // low optical depth
+			// low optical depth
 			double erad_L_RECON = x1LeftState(i, j, k, primRadEnergy_index + numRadVars_ * g);
-			double erad_R_RECON  = x1RightState(i, j, k, primRadEnergy_index + numRadVars_ * g);
+			double erad_R_RECON = x1RightState(i, j, k, primRadEnergy_index + numRadVars_ * g);
 
 			double fx_L_RECON = x1LeftState(i, j, k, x1ReducedFlux_index + numRadVars_ * g);
 			double fx_R_RECON = x1RightState(i, j, k, x1ReducedFlux_index + numRadVars_ * g);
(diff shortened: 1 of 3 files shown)


chongchonghe:
The old 1/tau correction:

radhydro_pulse_velocity
radhydro_pulse_temperature


The new method:

radhydro_pulse_velocity
radhydro_pulse_temperature

BenWibking:
This is useful to have for reference. However, since this makes things worse, I don't think this should be merged.

chongchonghe:
> This is useful to have for reference. However, since this makes things worse, I don't think this should be merged.

No worry. I'm creating this PR to show you what I have done in my attempt to fix this instability issue. This is not meant to be merged yet.

chongchonghe:
@BenWibking 
Check if this is what you have in mind. It didn't fix the issue. In fact, it makes it even worse (see figure below). 

radhydro_pulse_velocity-AVE

src/radiation_system.hpp
```diff
@@ -426,6 +432,26 @@ void RadSystem<problem_t>::ReconstructStatesPLM(arrayconst_t &q_in, array_t &lef
...
+	// construct ArrayViews for permuted indices
+	quokka::Array4View<amrex::Real const, DIR> q(q_in);
+	quokka::Array4View<amrex::Real, DIR> leftState(leftState_in);
+	quokka::Array4View<amrex::Real, DIR> rightState(rightState_in);
+
+	amrex::ParallelFor(indexRange, nvars, [=] AMREX_GPU_DEVICE(int i_in, int j_in, int k_in, int n) noexcept {
+		// permute array indices according to dir
+		auto [i, j, k] = quokka::reorderMultiIndex<DIR>(i_in, j_in, k_in);
+
+		const auto ave = 0.5 * (q(i - 1, j, k, n) + q(i, j, k, n));
```
BenWibking, line 449: My suggestion was to modify the existing reconstruction so that it interpolates between PPM and this expression as a function of optical depth.

BenWibking:
> @BenWibking Check if this is what you have in mind. It didn't fix the issue. In fact, it makes it even worse (see figure below).

I'm not sure whether what you did corresponds to what I had in mind. It looks like the new reconstruction is not actually used in the code, or you made changes that are not reflected in the latest commit on this branch. Can point me to the exact commit that you tested this with?

src/radiation_system.hpp
chongchonghe: I did interpolation between PPM and AVERAGE in Lines 909 - 946.

chongchonghe:
> > @BenWibking Check if this is what you have in mind. It didn't fix the issue. In fact, it makes it even worse (see figure below).
> 
> I'm not sure whether what you did corresponds to what I had in mind. It looks like the new reconstruction is not actually used in the code, or you made changes that are not reflected in the latest commit on this branch. Can point me to the exact commit that you tested this with?

I meant your comment -- "My suggestion above was to modify the reconstruction, not the Riemann solver." 

I did interpolation between PPM and AVERAGE in Lines 909 - 946 of [src/radiation_system.hpp](https://github.com/quokka-astro/quokka/pull/544/files#diff-74b2cb839d76f48507c09be9164633e006916bfdb5536497976ec64526b8a8a7) . I pass two extra variables to ComputeFluxes: x1LeftStateAVE_in and x1RightStateAVE_in, and the interpolation is done inside ComputeFluxes.

BenWibking:
Thanks, I was confused by the fact that you did the interpolation inside the Riemann solver. I understand now.

Can you try only doing the interpolation for the radiation energy density, and using the original PPM reconstruction for the flux variables?

chongchonghe:
> Thanks, I was confused by the fact that you did the interpolation inside the Riemann solver. I understand now.
> 
> Can you try only doing the interpolation for the radiation energy density, and using the original PPM reconstruction for the flux variables?

Unfortunately, that didn't work either.

BenWibking:
Ok, one last idea: can you try enabling interpolation on Erad only for the even indices?

chongchonghe:
I have a few observations:

1. The reconstruction matters a lot in this test. Changing `radiationReconstructionOrder_` to 2 or 1 completely destroys the non-advecting pulse run -- the temperature is almost flat, but the advecting pulse has the correct output. 
2. The odd-even correction won't make a big difference. I've tried turning on/off the correction for all cells. The difference in the outcome is very minimal. 
3. Both the radiation flux and gas momentum are oscillating. That means the issue must come from the work term ($v \cdot F$).

chongchonghe:
> Ok, one last idea: can you try enabling interpolation on Erad only for the even indices?

That failed as well...

BenWibking:
> I have a few observations:
> 
> 1. The reconstruction matters a lot in this test. Changing `radiationReconstructionOrder_` to 2 or 1 completely destroys the non-advecting pulse run -- the temperature is almost flat, but the advecting pulse has the correct output.

This is interesting. What this means for the oscillations seems unclear.

> 2. The odd-even correction won't make a big difference. I've tried turning on/off the correction for all cells. The difference in the outcome is very minimal.

I am a bit confused by this statement. It seemed to make a significant difference for "Method 2" in https://github.com/quokka-astro/quokka/issues/541. Is that not actually the case?

> 3. Both the radiation flux and gas momentum are oscillating. That means the issue must come from the work term (v⋅F).

I don't see why this is the case. If the radiation flux is oscillating, I would expect the work term $\kappa F/c$ to impart oscillations to the gas momentum, regardless of the underlying cause of the oscillations.

BenWibking:
Since this doesn't help, I would be in favor of closing this PR and instead implementing "Method 2" from https://github.com/quokka-astro/quokka/issues/541, which you have shown *does* mitigate the odd-even decoupling.

I think the residual oscillations may be small enough to just live with (as long as they don't continue to grow with time).

PR #551, opened 2024-03-06
Defined advecting pulse test in dynamic diffusion regime

chongchonghe:
1. Defined advecting pulse test in the dynamic diffusion regime. 
2. Use a new odd-even correction: S_corr = (1/tau, 1, 1, 1) at high tau and (1, 1, 1, 1) otherwise. 
3. Increased error_tol in some tests as we use a new odd-even correction

### Related issues
None.

### Checklist

### DIFF
--- a/src/RadMarshakAsymptotic/test_radiation_marshak_asymptotic.cpp
+++ b/src/RadMarshakAsymptotic/test_radiation_marshak_asymptotic.cpp
@@ -326,7 +326,7 @@ auto problem_main() -> int
 		sol_norm += std::abs(Tmat_exact[i]);
 	}
 
-	const double error_tol = 0.05; // 5 per cent
+	const double error_tol = 0.09;
 	const double rel_error = err_norm / sol_norm;
 	amrex::Print() << "Relative L1 error norm = " << rel_error << std::endl;
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -172,6 +172,7 @@ add_subdirectory(RadhydroShockCGS)
 add_subdirectory(RadhydroShockMultigroup)
 add_subdirectory(RadhydroUniformAdvecting)
 add_subdirectory(RadhydroPulse)
+add_subdirectory(RadhydroPulseDyn)
 add_subdirectory(RadhydroPulseGrey)
 add_subdirectory(RadhydroPulseMG)
 
--- a/src/RadMarshakAsymptotic/test_radiation_marshak_asymptotic.hpp
+++ b/src/RadMarshakAsymptotic/test_radiation_marshak_asymptotic.hpp
@@ -5,7 +5,7 @@
 // Copyright 2020 Benjamin Wibking.
 // Released under the MIT license. See LICENSE file included in the GitHub repo.
 //==============================================================================
-/// \file test_radiation_marshak.hpp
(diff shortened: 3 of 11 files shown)


markkrumholz (approved):

BenWibking:
The GPU tests still need to run. Please don't merge yet.

src/RadMarshakAsymptotic/test_radiation_marshak_asymptotic.cpp
```diff
@@ -326,7 +326,7 @@ auto problem_main() -> int
 		sol_norm += std::abs(Tmat_exact[i]);
 	}
 
-	const double error_tol = 0.05; // 5 per cent
+	const double error_tol = 0.09;
```
BenWibking, line 329: This seems a bit large. Can you post a plot of the solution?

chongchonghe: Here is the new plot:
marshak_wave_asymptotic_gastemperature

compare to the old plot:
marshak_wave_asymptotic_gastemperature-cfl0 9

The new plot is smoother thanks to the new odd-even correction, but there is one single dot at x=0.14 that is ahead of the analytic solution by (maybe) 0.005, but the relative error is huge because of the sharp drop at the wave front. I think we can live with this.

markkrumholz: In cases like this is may just be that using |sim - exact| / exact as your definition of the error is not a good choice of how to define the error metric, since formally the exact solution is zero -- the only reason that the error didn't come out as infinite is that you have had to impose some finite temperature even in the upstream reason for numerical reasons. A better metric for error for this sort of problem might be to define the position of the front somehow (for example the first cell for which T > 1.1 * T_init) and check that the front position is accurate to a cell.

chongchonghe: We compare the total L1 norm of the error with the total L1 norm of the exact solution, sum(|sim - exact|) / sum(|exact|), so there is no problem of relative error being infinite. 

Indeed, I agree. We could add an extra metric to check the position of the wavefront.

chongchonghe:
The GPU tests are staying in the queue forever ... I thought we get our own AMD GPU for testing, so I'm expecting the GPU amdclang test to start very fast. Are we not using it, or the machine is malfunctioning? @BenWibking

BenWibking:
You can check what's running on this page: https://dev.azure.com/quokka-code/quokka-CI/_build

It looks like the answer is "nothing", which happens when the machine is rebooted and Chris doesn't tell me. I have to manually re-start the CI agent. I will do that now.

BenWibking:
`moth` is running tests now. `avatargpue` is currently offline (presumably having its hard drive replaced).

markkrumholz:
> `moth` is running tests now. `avatargpue` is currently offline (presumably having its hard drive replaced).

Yes -- unfortunately the drive did not arrive as scheduled yesterday, so hopefully it got here today.

chongchonghe:
@BenWibking What's the reason for the nvcc failure and is it working now?

BenWibking:
The reason is that the drivers weren't re-installed after the machine died. They are now and it works.

PR #552, opened 2024-03-07
New source terms that include the work term

chongchonghe:
### Description

A new scheme for radiation-matter coupling that includes the work term in the source terms and also includes up to $\beta^2$ terms in the flux update. 

- A new parameter `beta_order` is introduced in RadSystem_Traits. It's used to control the order of the $\beta$ terms to include in the radiation four-force. Available options are 1, 2, and 3. 
- The user can set `include_work_term_in_source = false` in radiation_system.hpp to restore the old scheme. I didn't add this to RadSystem_Traits, so you have to do it in the source file. 

The radiation-matter coupling is done in three steps:

##### Step 1: Energy exchange

The motivation to include the work term in the source terms is as fellows. The energy equations excluding the transport part are:

$$
\begin{gather}
\frac{\partial}{\partial t} E = c G^0 \\
\frac{\partial}{\partial t} E_r = -\hat{c} G^0
\end{gather}
$$

where E is the gas **total** energy. The fully relativistic expression of the time-like component of the four-force is (there are extra terms if $\kappa_F \ne \kappa_E$): 

$$
-cG^0 = \hat{c} \gamma (\kappa_P B - \kappa_E E_r) + \hat{c} c^{-2} \gamma_3 \kappa_F v^i F^i
$$

where I denote $4 \pi B/c$ as $B$. In the scheme presented in the Quokka paper, only the $B-E$ term is included in the source function, and it's okay to assume the $E$ in the above RHD equation is the internal energy. However, this simplification breaks down in the advecting diffusive uniform medium test. Imagine a test problem where the radiation (with uniform energy density $E_0$) is in thermal equilibrium with a uniform medium and you boost it to a moving frame with velocity $v$. In the lab frame, $B = E_0$ but $E_r > E_0$ due to Lorentz transformation. Therefore, in this scheme the source term is non-zero and the radiation is out of equilibrium with matter. The matter temperature will increase in the energy exchange step. In the next steps, the gas internal energy is not touched, so it's off from the correct equilibrium forever.

This issue goes away if you include the work ($v^i F^i$) in the source term. We can show that the source $G^0$ in the lab frame vanishes precisely in the aforementioned test problem. 

##### Step 2: Flux update

The momentum and flux equations (excluding the transport part) are 

$$
\begin{gather}
\frac{\partial}{\partial t} \rho v^i = G^i, \\
\frac{\partial}{\partial t} F^i = - \hat{c} c G^i
\end{gather}
$$

and 

$$
\begin{align}
-c \hat{c} G^i = &- \hat{c} \kappa_F \gamma ~F^i + \hat{c} \kappa_P B \gamma ~v^i + \hat{c} \kappa_F \gamma ~v^j P^{ji} \\
& + \hat{c} (\kappa_F - \kappa_E)(E \gamma^3 - 2 \frac{v^j F^j}{c^2} \gamma^3) ~v^i 
\end{align}
$$

We use the backward Euler scheme for flux update:

$$
\frac{F^i - F_0^i}{\Delta t} = - \hat{c} c G^i
$$

where the subscript $0$ indicates the old state and the variables without subscript $0$ are in the new state. In $G^i$, the following two terms are lagged: the velocity $v^i$ and Edington tensor $T^{ij}$. We take a valid assumption that the Eddington tensor is unchanged, because the regime where the flux could have significant variable, which is the diffusion regime, is when the Eddington tensor is nearly constant at 1/3 I. The lagged velocity can cause huge error when the radiation force is stronge. We eliminate this lag in velocity via iterations in the next step. 

Assuming $F^i$ is the only implicit quantity in the equation, it's straight forward to solve $F^i$. In the case of $\kappa_F - \kappa_E = 0$​, the solution is

$$
F_i = \frac{F_0^i + \hat{c} \Delta t (\kappa_P B \gamma v^i + \kappa_F \gamma v^j P^{ji})}{1 + \hat{c} \Delta t \kappa_F \gamma}
$$

In the case of  $\kappa_F - \kappa_E \ne 0$, $G^i$ would contain all three components of $F$, therefore we need to solve for $F^i$​ by inverting a 3x3 matrix. This is done by the hand-crafted `Solve3x3matrix` function that avoids dividing any of the off-diagonal elements which could be zero. 

The value of $\gamma$ is controled by the `beta_order` parameter. $\gamma = 1$ when `beta_order = 0` or `beta_order = 1`, and $\gamma = 1 + \beta^2/2$ when `beta_order = 2`. When `beta_order >= 3`, we use the fully relativistic formula $\gamma = 1/\sqrt{1 - \beta^2}$. Furthermore, the $\beta^2$ term in $G^i$ is ignored when `beta_order < 2`. 

##### Step 3: Perform iteration to ensure the gas velocity is up to date 

To complete radiation-matter coupling, we iteration step 1 and step 2 to make sure the gas velocity is up to date and the energy is conserved.

1. Use the new source term, $S = \Delta t \hat{c} \gamma (\kappa_{P} B - \kappa_E E) + \Delta t \hat{c} c^{-2} \gamma \kappa_F v_i F_i$, to update the gas internal energy $E_g$ and radiation energy $E_r$. In this step, the total energy is conserved. We record the (lagged) work term as $W^\star \sim v_i F_i$. 
2. Update flux using backward Euler scheme.
3. Update gas momentum following $(\rho v_i)^{(t+1)} = (\rho v_i)^{\star} - \Delta F_i / (c \hat{c})$. 
4. $E_g^{(t+1)} \leftarrow E_g^{(t+1)} - \Delta K$ where $\Delta K$ is the change in kinetic energy. In this and previous step, the gas total energy is conserved. 
5. Calculate the updated work term: $W \sim v_i F_i$. 
6. Check if $W$ converges to $W^\star$ with the criteria listed below. If yes, break; otherwise, restart from step 1 with $W^\star \leftarrow W$. The convergence criteria I use are: $(c/\hat{c}) |W - W^{\star}|/E_{\rm tot} < 10^{-13}$, or $|W - W^{\star}|/W^{\star} < 10^{-13}$, or  $|W - W^{\star}|/S < 10^{-13}$. 

This scheme passed all the tests, and at most 5 iterations is used. The results from all the tests are nearly identical to the previous results.

### Related issues

I adjusted the parameters in the `RadhydroUniformAdvecting` test problem so that it's in the dynamic diffusion regime. The old scheme failed at this test but the new scheme passed.

### Alternative Approaches

We can also set the old scheme to default and optionally turn on this new scheme. But since this new scheme passes tests that the old scheme failed, and it is ONLY more expensive in the diffusion regime where it gives more accurate result, we should probably keep it as default.

We can remove `compute_v_over_c_terms` in RadSystem_Traits and replace it with `beta_order = 0`. 

### Checklist


markkrumholz (commented):
Looks good. Please just have a look at the codeQL warnings, and one those are resolved I am happy to approve.

BenWibking:
It's a little strange to me that we use an expansion in $v/c$ for other terms, but here use the full Lorentz factor.

It might be ok, but I am concerned about the possbility of a subtle inconsistency to not expand everything to the same order in $v/c$. Is there a reason to prefer it this way in a non-relativistic code?

markkrumholz:
> It's a little strange to me that we use an expansion in v/c for other terms, but here use the full Lorentz factor.
> 
> It might be ok, but I am concerned about the possbility of a subtle inconsistency to not expand everything to the same order in v/c. Is there a reason to prefer it this way in a non-relativistic code?

Well this isn't the default option -- the default is to use an expansion in v/c. I think the option to use the full Lorentz factor is mostly a debugging / testing capability. I can't see any good reason why one would use it in practice.

BenWibking:
> > It's a little strange to me that we use an expansion in v/c for other terms, but here use the full Lorentz factor.
> > It might be ok, but I am concerned about the possbility of a subtle inconsistency to not expand everything to the same order in v/c. Is there a reason to prefer it this way in a non-relativistic code?
> 
> Well this isn't the default option -- the default is to use an expansion in v/c. I think the option to use the full Lorentz factor is mostly a debugging / testing capability. I can't see any good reason why one would use it in practice.

Ok, I see that now. Maybe the use of the full Lorentz factor could be controlled by a different parameter, since it's not actually $O(\beta^3)$? I think that it is very confusing to enable it by setting `beta_order=3`.

chongchonghe:
> > > It's a little strange to me that we use an expansion in v/c for other terms, but here use the full Lorentz factor.
> > > It might be ok, but I am concerned about the possbility of a subtle inconsistency to not expand everything to the same order in v/c. Is there a reason to prefer it this way in a non-relativistic code?
> > 
> > 
> > Well this isn't the default option -- the default is to use an expansion in v/c. I think the option to use the full Lorentz factor is mostly a debugging / testing capability. I can't see any good reason why one would use it in practice.
> 
> Ok, I see that now. Maybe the use of the full Lorentz factor could be controlled by a different parameter, since it's not actually O(β3)? I think that it is very confusing to enable it by setting `beta_order=3`.

That's true. Maybe use `beta_order=-1` for Lorentz factor?

markkrumholz:
> > > > It's a little strange to me that we use an expansion in v/c for other terms, but here use the full Lorentz factor.
> > > > It might be ok, but I am concerned about the possbility of a subtle inconsistency to not expand everything to the same order in v/c. Is there a reason to prefer it this way in a non-relativistic code?
> > > 
> > > 
> > > Well this isn't the default option -- the default is to use an expansion in v/c. I think the option to use the full Lorentz factor is mostly a debugging / testing capability. I can't see any good reason why one would use it in practice.
> > 
> > 
> > Ok, I see that now. Maybe the use of the full Lorentz factor could be controlled by a different parameter, since it's not actually O(β3)? I think that it is very confusing to enable it by setting `beta_order=3`.
> 
> That's true. Maybe use `beta_order=-1` for Lorentz factor?

That would be fine with me. I personally don't find `beta_order=3` confusing, but having `-1` be the flag for using the full Lorentz factor would perhaps be clearer.

BenWibking:
@chongchonghe Can you add `beta_order` to the list of runtime parameters in the docs: https://quokka-astro.github.io/quokka/parameters.html?

chongchonghe:
> @chongchonghe Can you add `beta_order` to the list of runtime parameters in the docs: https://quokka-astro.github.io/quokka/parameters.html?

Oh, but `beta_order` is not a runtime parameter. You can't specify it in the .in file. What's more, I define it as a constexpr and it has consequences at compile time.

BenWibking:
> > @chongchonghe Can you add `beta_order` to the list of runtime parameters in the docs: https://quokka-astro.github.io/quokka/parameters.html?
> 
> Oh, but `beta_order` is not a runtime parameter. You can't specify it in the .in file. What's more, I define it as a constexpr and it has consequences at compile time.

Ah, ok, nevermind then. We should at some point add a documentation page for compile-time parameters, but no need to do that in this PR.

chongchonghe:
> Looks good. Please just have a look at the codeQL warnings, and one those are resolved I am happy to approve.

What codeQL warnings? I don't see any.

markkrumholz:
> > Looks good. Please just have a look at the codeQL warnings, and one those are resolved I am happy to approve.
> 
> What codeQL warnings? I don't see any.

If you look earlier in the discussion, you will see comments on radsystem.hpp line 1134 from CodeQL. They are provided by "github-advanced-security bot".

chongchonghe:
> > > Looks good. Please just have a look at the codeQL warnings, and one those are resolved I am happy to approve.
> > 
> > 
> > What codeQL warnings? I don't see any.
> 
> If you look earlier in the discussion, you will see comments on radsystem.hpp line 1134 from CodeQL. They are provided by "github-advanced-security bot".

OK, I saw that. I believe they are false positives, so I ignored them.

markkrumholz:
> > > > Looks good. Please just have a look at the codeQL warnings, and one those are resolved I am happy to approve.
> > > 
> > > 
> > > What codeQL warnings? I don't see any.
> > 
> > 
> > If you look earlier in the discussion, you will see comments on radsystem.hpp line 1134 from CodeQL. They are provided by "github-advanced-security bot".
> 
> OK, I saw that. I believe they are false positives, so I ignored them.

OK, that's fine. I am happy to proceed.

chongchonghe:
OK, all CPU checks have passed. I'll wait till my other PR is merged into dev branch and I'll merge the dev branch into this branch, then I'll trigger GPU tests.

markkrumholz (dismissed):

chongchonghe:
@markkrumholz Can you approve this PR again? It merged in recent updates and need another approve.

markkrumholz (approved):

PR #570, opened 2024-03-18
Refactor Reconstruction Functions

AstroKriel:
### Description
I have relocated the reconstruction functions to all sit within the HyperbolicSystem class, and tried to minimise code duplication.

### Checklist
- [ ] (Not Applicable) I have added a link to any related issues see (see above).
- [ ] (Not Applicable) I have added tests for any new physics that this PR adds to the code.


markkrumholz:
@AstroKriel: were you attempting to ask me to review this? You assigned it to me, which is not the same thing as requesting my review.

AstroKriel:
@markkrumholz, ah yes, I see! Fixed it now.

markkrumholz (commented):
Generally looks fine, but just two comments. 

(1) Though it's not part of the automated test suite, can you confirm that the 3d problems still pass with these changes? Just want to make sure that we're still reconstructing the number of cells we need to be reconstructing in 3d.

(2) It would be good to add a comment to the header in hyperbolic system explaining the relationship between the index ranges used in the multiFab and Array4 versions of the reconstruct functions. That is, the comment should say that calling the multiFab versions with nghost = ng is equivalent to looping over the Fabs and then calling the Array4 versions with cellRange set equal to fabBox.lo + ng to fabBox.hi - ng, or something like that. This will be useful for people who need to interface with these versions in the future.

src/hyperbolic_system.hpp
```diff
@@ -34,13 +34,27 @@ namespace quokka
...
       public:
+	template <SlopeLimiter limiter> static auto SlopeFunc(amrex::Real x, amrex::Real y) -> amrex::Real
+	{
+		if (limiter == SlopeLimiter::minmod) {
+			return minmod(x, y);
+		}
+		if (limiter == SlopeLimiter::MC) {
+			return MC(x, y);
+		}
+		throw std::invalid_argument("Invalid slope limiter specified.");
```
BenWibking: We should **never** throw exceptions. They do not work on GPU. Since this is a template argument, you can use `static_assert` here.

src/hyperbolic_system.hpp
```diff
@@ -34,13 +34,27 @@ namespace quokka
...
 using array_t = amrex::Array4<amrex::Real> const;
 using arrayconst_t = amrex::Array4<const amrex::Real> const;
 
 /// Class for a hyperbolic system of conservation laws
 template <typename problem_t> class HyperbolicSystem
 {
       public:
+	template <SlopeLimiter limiter> static auto SlopeFunc(amrex::Real x, amrex::Real y) -> amrex::Real
+	{
+		if (limiter == SlopeLimiter::minmod) {
```
BenWibking: This should be `if constexpr`.

src/hyperbolic_system.hpp
```diff
@@ -34,13 +34,27 @@ namespace quokka
...
+			return minmod(x, y);
+		}
+		if (limiter == SlopeLimiter::MC) {
```
BenWibking: This should be `if constexpr`

BenWibking (changes requested):
You should be aware that C++ exceptions do not work on GPUs. They simply do nothing when executed on GPUs. For that reason, they should *never* be used. If they are throwing an error based on something that can be computed at compile time, use `static_assert` instead.

BenWibking:
> (1) Though it's not part of the automated test suite, can you confirm that the 3d problems still pass with these changes? Just want to make sure that we're still reconstructing the number of cells we need to be reconstructing in 3d.
> 

@markkrumholz The GPU tests run 3D problems, so that should be covered. However, the GPU tests currently fail.

src/hyperbolic_system.hpp
AstroKriel: Done.

src/hyperbolic_system.hpp
AstroKriel: Done.

src/hyperbolic_system.hpp
AstroKriel: Done.

AstroKriel:
I'm digging into what's happening with the 3D tests. Looks like its failing to compile on the GPU.

src/hyperbolic_system.hpp
```diff
@@ -208,12 +297,7 @@ template <FluxDir DIR>
 void HyperbolicSystem<problem_t>::ReconstructStatesPPM(amrex::MultiFab const &q_mf, amrex::MultiFab &leftState_mf, amrex::MultiFab &rightState_mf,
 						       const int nghost, const int nvars, const int iReadFrom, const int iWriteFrom)
 {
-	BL_PROFILE("HyperbolicSystem::ReconstructStatesPPM()");
-
-	// By convention, the interfaces are defined on the left edge of each
-	// zone, i.e. xleft_(i) is the "left"-side of the interface at the left
-	// edge of zone i, and xright_(i) is the "right"-side of the interface
-	// at the *left* edge of zone i.
+	BL_PROFILE("HyperbolicSystem::ReconstructStatesPPM(MultiFabs)");
```
BenWibking: You should be able to fix the clang-tidy warning by changing this to
`const BL_PROFILE("HyperbolicSystem::ReconstructStatesPPM(MultiFabs)");`

src/hyperbolic_system.hpp
```diff
@@ -227,91 +311,125 @@ void HyperbolicSystem<problem_t>::ReconstructStatesPPM(amrex::MultiFab const &q_
...
+	});
+}
+
+template <typename problem_t>
+template <FluxDir DIR>
+AMREX_GPU_HOST_DEVICE void HyperbolicSystem<problem_t>::ReconstructStatesPPM(arrayconst_t &q_in, array_t &leftState_in, array_t &rightState_in,
+									     amrex::Box const &cellRange, amrex::Box const & /*interfaceRange*/,
+									     const int nvars, const int iReadFrom, const int iWriteFrom)
+{
+	BL_PROFILE("HyperbolicSystem::ReconstructStatesPPM(Arrays)");
```
BenWibking: You can fix the warning by changing this line to:
`const BL_PROFILE("HyperbolicSystem::ReconstructStatesPPM(Arrays)");`

src/hyperbolic_system.hpp
AstroKriel: Done.

src/hyperbolic_system.hpp
AstroKriel: Done.

BenWibking:
It looks good to me, but I think Chongchong's PR introduced a conflict in `radiation_system.hpp`. Hopefully it is straightforward to fix.

BenWibking:
@AstroKriel Can you fix the merge conflict? That's the only thing holding this up.

AstroKriel:
> You should be aware that C++ exceptions do not work on GPUs. They simply do nothing when executed on GPUs. For that reason, they should _never_ be used. If they are throwing an error based on something that can be computed at compile time, use `static_assert` instead.

I have fixed this, could you resolve this request, please?

BenWibking (approved):

Issue #566, opened 2024-03-17
refuse to build when Cray compiler is detected for CPU-only builds


BenWibking:
**Describe the proposal**
We refuse to build when the Cray C/C++ Compiler (cce) is detected *and* we are doing a CPU-only build. It miscompiles the code and produces wrong output.

However, when doing a GPU build, it works fine.

**Additional context**
@chongchonghe found that compiling using `cce/15.0.1` causes the test suite to fail on CPU, but works on GPU.

psharda:
So this is an issue if you compile using `setonix-cpu.profile` and `setonix-gpu.profile`?

BenWibking:
> So this is an issue if you compile using `setonix-cpu.profile` and `setonix-gpu.profile`?

Yes.

PR #571, opened 2024-03-18
Load rocm for setonix CPUs

### LINKED ISSUE #563: setonix-cpu profile needs rocm
**Describe the bug**
We need `module load rocm/5.4.3` in `setonix-cpu.profile` otherwise setonix cannot find HIP. 

**To Reproduce**
```
source scripts/setonix-cpu.profile
mkdir build
cd build
cmake -C ../cmake/setonix.cmake .. -DCMAKE_BUILD_TYPE=Release -DQUOKKA_PYTHON=OFF -DAMReX_SPACEDIM=3 -DCMAKE_C_COMPILER=cc -DCMAKE_CXX_COMPILER=CC
```

**Additional context**
Maybe something changed in setonix which is why loading rocm for CPU only build on setonix was not needed.
Thanks to @aditivijayan for figuring it out.

psharda:
### Description
We need to load rocm while compiling on setonix CPUs.

### Related issues
Fixes #563 

### Checklist

### DIFF
--- a/scripts/setonix-cpu.profile
+++ b/scripts/setonix-cpu.profile
@@ -14,6 +14,6 @@ export MPICH_GPU_SUPPORT_ENABLED=0
 
 # compiler environment hints
 export CC=$(which cc)
-export CXX="$(which CC) -fno-cray"
+export CXX="$(which CC)"
 export FC=$(which ftn)
-
+export CXXFLAGS="$CXXFLAGS -fno-cray"
--- a/scripts/setonix-cpu.profile
+++ b/scripts/setonix-cpu.profile
@@ -7,6 +7,7 @@ module load cray-mpich
 module load cce/15.0.1
 module load cray-hdf5
 module load cray-python/3.9.13.1
+module load rocm/5.4.3
 
 # GPU-aware MPI
 export MPICH_GPU_SUPPORT_ENABLED=0
--- a/scripts/setonix-cpu.profile
+++ b/scripts/setonix-cpu.profile
@@ -14,6 +14,6 @@ export MPICH_GPU_SUPPORT_ENABLED=0
 
 # compiler environment hints
 export CC=$(which cc)
-export CXX=$(which CC)
+export CXX="$(which CC) -fno-cray"
 export FC=$(which ftn)
 
--- a/scripts/setonix-gpu.profile
+++ b/scripts/setonix-gpu.profile
@@ -24,5 +24,5 @@ export CC=$(which cc)
 export CXX=$(which CC)
 export FC=$(which ftn)
 export CFLAGS="-I${ROCM_PATH}/include"
-export CXXFLAGS="-I${ROCM_PATH}/include"
+export CXXFLAGS="-I${ROCM_PATH}/include -fno-cray"
 


BenWibking:
For production runs on CPUs, it might be better to avoid the Cray compiler altogether and use gcc instead.

scripts/setonix-cpu.profile
```diff
@@ -7,12 +7,13 @@ module load cray-mpich
...
 module load cray-python/3.9.13.1
+module load rocm/5.4.3
 
 # GPU-aware MPI
 export MPICH_GPU_SUPPORT_ENABLED=0
 
 # compiler environment hints
 export CC=$(which cc)
-export CXX=$(which CC)
+export CXX="$(which CC) -fno-cray"
```
psharda: @BenWibking will this fix #566 for CPUs? Will it work even if the compiler is not cray?

BenWibking: It's probably a better idea to detect the cray compiler based on the value of `CMAKE_CXX_COMPILER_ID`, e.g.
https://github.com/quokka-astro/quokka/blob/a2acbb741a282be88b48ad52279196c3d421aaeb/src/CMakeLists.txt#L79
except the appropriate compiler ID is "CrayClang" (https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html).

psharda: oops, put it in the wrong place

BenWibking (dismissed):

psharda:
Ugh, using `fno-cray` seems to give a linker error while compiling on Setonix CPUs.

BenWibking:
@chongchonghe got it to work on Setonix -- maybe @chongchonghe can help?

chongchonghe:
On top of the setonix-gpu.profile, I have to add `-ffp=0` to both `CFLAGS=` and `CXXFLAGS=` and the code compiles and produces correct results for some pure hydro simulations. The radiation runs failed.

BenWibking:
> On top of the setonix-gpu.profile, I have to add `-ffp=0` to both `CFLAGS=` and `CXXFLAGS=` and the code compiles and produces correct results for some pure hydro simulations. The radiation runs failed.

But it works if you add `-fno-cray`, right?

chongchonghe:
> > On top of the setonix-gpu.profile, I have to add `-ffp=0` to both `CFLAGS=` and `CXXFLAGS=` and the code compiles and produces correct results for some pure hydro simulations. The radiation runs failed.
> 
> But it works if you add `-fno-cray`, right?

If I add `-fno-cray`, the job failed with segfault when running with MPI.

psharda:
So then `fno-cray` is not going to work for all the test problems? @BenWibking

BenWibking:
> So then `fno-cray` is not going to work for all the test problems? @BenWibking

It should be possible to avoid the Cray compiler entirely on Setonix. Then no special options should be necessary.

This should be possible by loading `PrgEnv-gnu` and then setting the compiler to `hipcc`:
```
module load PrgEnv-gnu
export CC=$(which hipcc)
export CXX=$(which hipcc)
```

psharda:
> > So then `fno-cray` is not going to work for all the test problems? @BenWibking
> 
> It should be possible to avoid the Cray compiler entirely on Setonix. Then no special options should be necessary.
> 
> This should be possible by loading `PrgEnv-gnu` and then setting the compiler to `hipcc`:
> 
> ```
> module load PrgEnv-gnu
> export CC=$(which hipcc)
> export CXX=$(which hipcc)
> ```

This by itself does not seem to be sufficient. We still need HDF5 and ROCm for example. When I load HDF5 and ROCm modules and `make` a test problem, I get these warnings: 

```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LC_CTYPE = "UTF-8",
	LANG = (unset)
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```

psharda:
I still have some cray modules loaded. Is this an issue:

```
psharda@setonix-06:~> module list

Currently Loaded Modules:
  1) craype-x86-milan     4) perftools-base/23.03.0                  7) gcc/12.2.0        10) cray-mpich/8.1.25      13) pawsey         16) python/3.10.10
  2) libfabric/1.15.2.0   5) xpmem/2.5.2-2.4_3.47__gd0f7936.shasta   8) craype/2.7.20     11) cray-libsci/23.02.1.1  14) pawseytools
  3) craype-network-ofi   6) pawseyenv/2023.08                       9) cray-dsmml/0.2.2  12) PrgEnv-gnu/8.3.3       15) slurm/22.05.2
```

BenWibking:
I think this is a better solution: https://github.com/quokka-astro/quokka/issues/566.

BenWibking:
Since there is no activity on this for the past 2 months, I am going to close this issue for now. If it turns out to be needed, we can re-open it.

PR #572, opened 2024-03-18
Avoid imaginary roe average sound speed

### LINKED ISSUE #555: HLLC wavespeeds are not computed correctly if cs_tilde is imaginary
**Describe the bug**
For some left and right states, it is possible that the Roe-average sound speed is imaginary:
https://github.com/quokka-astro/quokka/blob/fcde1aa11205ed64365e42ce7e084a9964aa7e9f/src/HLLC.hpp#L58

Since `std::min` behaves differently for different compilers when one argument is NAN (in particular, gcc returns a the non-NAN argument while Clang returns NAN), we have problems when using Clang-based compilers to run problems that have interfaces where the Roe average is not defined:
https://github.com/quokka-astro/quokka/blob/fcde1aa11205ed64365e42ce7e084a9964aa7e9f/src/HLLC.hpp#L64

A reasonable solution would be to check whether $\tilde c_s^2$ is positive, and, if not, use the arithmetic average of the left and right sound speeds.

**To Reproduce**
1. Compile the PopIII problem with an Clang-based compiler.
2. Run this problem for one timestep.
3. Print out the HLLC wavespeeds for at each interface.
4. Observe NAN values.

**Additional context**
Since the AMD GPU compiler is based on Clang, it also has this problem, and is the reason the hydro update behaves differently on AMD GPUs.

Thanks to @psharda for discovering this after a truly exceptional debugging effort.

psharda:
### Description
There can be instances when the Roe-averaged sound speed (`cs_tilde` in `HLLC.hpp`) can be negative. In such cases, the current implementation will generate a NAN cs_tilde, which is handled differently by different compilers (e.g., GCC versus HIPCC). This was one of the causes of StarCluster and PopIII problems failing on HIPCC-based CPUs and AMD GPUs.

### Related issues
Fixes #555.

### Checklist

### DIFF
--- a/src/HLLC.hpp
+++ b/src/HLLC.hpp
@@ -54,8 +54,14 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto HLLC(quokka::HydroState<N_scalars, N_ms
 		// equation A.5b of Kershaw+1998
 		const double C_tilde_P = 0.5 * ((sL.Eint / sL.rho) * drdp_L + (sR.Eint / sR.rho) * drdp_R + sL.rho * dedp_L + sR.rho * dedp_R);
 
-		// equation 4.12 of Kershaw+1998
-		cs_tilde = std::sqrt((1.0 / C_tilde_P) * (H_tilde - 0.5 * vsq_tilde - C_tilde_rho));
+		// avoid NAN cs_tilde
+		if (H_tilde - 0.5*vsq_tilde - C_tilde_rho < 0) {
+			// fall back to the gamma=1 case
+			cs_tilde = 0.5 * (sL.cs + sR.cs);
+		} else {
+			// equation 4.12 of Kershaw+1998
+			cs_tilde = std::sqrt((1.0 / C_tilde_P) * (H_tilde - 0.5 * vsq_tilde - C_tilde_rho));
+		}
 
 		const double s_NL = 0.5 * G_L * std::max(dU, 0.); // second-order wavespeed correction
 		const double s_NR = 0.5 * G_R * std::max(dU, 0.);
--- a/src/HLLC.hpp
+++ b/src/HLLC.hpp
@@ -55,12 +55,13 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto HLLC(quokka::HydroState<N_scalars, N_ms
 		const double C_tilde_P = 0.5 * ((sL.Eint / sL.rho) * drdp_L + (sR.Eint / sR.rho) * drdp_R + sL.rho * dedp_L + sR.rho * dedp_R);
 
 		// avoid NAN cs_tilde
-		if (H_tilde - 0.5 * vsq_tilde - C_tilde_rho < 0) {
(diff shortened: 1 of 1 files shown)


src/HLLC.hpp
```diff
@@ -54,8 +54,14 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto HLLC(quokka::HydroState<N_scalars, N_ms
 		// equation A.5b of Kershaw+1998
 		const double C_tilde_P = 0.5 * ((sL.Eint / sL.rho) * drdp_L + (sR.Eint / sR.rho) * drdp_R + sL.rho * dedp_L + sR.rho * dedp_R);
 
-		// equation 4.12 of Kershaw+1998
-		cs_tilde = std::sqrt((1.0 / C_tilde_P) * (H_tilde - 0.5 * vsq_tilde - C_tilde_rho));
+		// avoid NAN cs_tilde
+		if (H_tilde - 0.5 * vsq_tilde - C_tilde_rho < 0) {
```
BenWibking: I think it would be better to check whether this is less than _or equal to_ 0. A zero value produces a non-NaN cs_tilde, but has other problems.

A minor optimization, but can you extract this to a separate variable to avoid computing it twice?

src/HLLC.hpp
```diff
@@ -54,8 +54,14 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto HLLC(quokka::HydroState<N_scalars, N_ms
...
+			// fall back to the gamma=1 case
```
BenWibking: I think this comment is a bit confusing. This is the same code used when $\gamma = 1$, but is not conceptually related to having an isothermal equation of state. It's just an arithmetic average of the left and right sound speeds.

src/HLLC.hpp
```diff
@@ -54,8 +54,15 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto HLLC(quokka::HydroState<N_scalars, N_ms
...
-		// equation 4.12 of Kershaw+1998
-		cs_tilde = std::sqrt((1.0 / C_tilde_P) * (H_tilde - 0.5 * vsq_tilde - C_tilde_rho));
+		// avoid NAN cs_tilde
+		const double cs_exp = H_tilde - 0.5 * vsq_tilde - C_tilde_rho;
+		if (cs_exp <= 0) {
+			// take the average of the left and right states
+			cs_tilde = 0.5 * (sL.cs + sR.cs);
+		} else {
+			// equation 4.12 of Kershaw+1998
+			cs_tilde = std::sqrt((1.0 / C_tilde_P) * cs_exp);
```
BenWibking: You could simplify this to `cs_tilde = std::sqrt(cs_exp / C_tilde_P)`.

BenWibking (changes requested):
Ok, looks good. I've noted one suggestion inline that will avoid an extra floating-point operation when the compiler interprets the parentheses strictly.

BenWibking (approved):

PR #573, opened 2024-03-18
Define K_S for isothermal gamma

### LINKED ISSUE #553: handle isothermal EOS entirely within EOS.hpp
**Describe the proposal**
Before general EOS support was added to the code, an isothermal EOS was handled as a special case whenever the pressure was needed in `HydroSystem`. Now that there is general EOS support, we should move all of the conditionals to handle `gamma = 1.0` into EOS.hpp, rather than scattered elsewhere in the code.

We could leave it as is, but this appears to cause problems with shock flattening for isothermal problems, since a special case to handle an isothermal EOS was never written for that.

**Additional context**
This appears to be the underlying cause (or at least one of the causes) of different results for the StarCluster problem when compiled with `hipcc` versus `gcc` on CPU (presumably since this causes a NAN to be produced in a way that is handled differently by different compilers).

cc @psharda

psharda:
### Description
Currently, `K_S` is undefined in the `computeFlatenningCoefficients` function for the case of isothermal gamma, because the function to compute sound speed within `EOS.hpp` that it calls only works if gamma != 1.

### Related issues
Fixes #553 for now. In the long run, this wont be needed because we will move all EOS calculations (including for gamma=1) to EOS.hpp. But for now, we should apply this patch so that the gamma=1 case is correctly handled while computing flattening coefficients. 

### Checklist

### DIFF
--- a/src/hydro_system.hpp
+++ b/src/hydro_system.hpp
@@ -600,8 +600,12 @@ void HydroSystem<problem_t>::ComputeFlatteningCoefficients(amrex::MultiFab const
 
 		// Z is a measure of shock strength (Eq. 76 of Miller & Colella 2002)
 		amrex::GpuArray<Real, nmscalars_> massScalars = RadSystem<problem_t>::ComputeMassScalars(primVar, i, j, k);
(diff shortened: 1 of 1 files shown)


BenWibking (approved):

BenWibking:
Feel free to merge after the GPU tests pass.

Issue #543, opened 2024-02-23
amrex::Abort does not unconditionally abort on GPU


BenWibking:
**Describe the bug**
We have been relying on `amrex::Abort` to unconditionally abort the code when called from GPU code. However, this only works when compiled in debug mode (or with additional compiler flags).

We need to rewrite the error handling code for iterative solves so that it correctly handles errors when `-DCMAKE_BUILD_TYPE=Release` is used (the default).

The Quokka documentation incorrectly describes `amrex::Abort` based on the old AMReX documentation (which was incorrect): https://quokka-astro.github.io/quokka/error_checking.html

**To Reproduce**
1. Compile the PopIII problem for GPU
2. It should crash with a VODE failure at timestep 58
3. Observe that it does not.

**Additional context**
The AMReX documentation was recently changed (October 2023) to reflect this new behavior, but this change was not communicated to the user community: https://github.com/AMReX-Codes/amrex/pull/3605.

**This affects both the Newton-Raphson solve used by the radiation code as well as the Microphysics chemistry integrator.**
cc @psharda @chongchonghe

chongchonghe:
This is important to know. I always reply on `AMREX_ASSERT`, `AMREX_ALWAYS_ASSERT`, or `AMREX_ALWAYS_ASSERT_WITH_MESSAGE`. I guess they all reply on amrex::Abort, right? Then, what is the solution? How do we "rewrite the error handling code" so that the code handles errors in release build? According to the documentation, we should set `USE_ASSERTION=TRUE` at compile time to enable `AMREX_ALWAYS_ASSERT` in GPU release build.

BenWibking:
> This is important to know. I always reply on `AMREX_ASSERT`, `AMREX_ALWAYS_ASSERT`, or `AMREX_ALWAYS_ASSERT_WITH_MESSAGE`. I guess they all reply on amrex::Abort, right?

Yes.

> Then, what is the solution? How do we "rewrite the error handling code" so that the code handles errors in release build?

You have to rewrite the function so it returns an error code that you then check. For an example, see the cooling code:
https://github.com/quokka-astro/quokka/blob/fcde1aa11205ed64365e42ce7e084a9964aa7e9f/src/CloudyCooling.hpp#L278
https://github.com/quokka-astro/quokka/blob/fcde1aa11205ed64365e42ce7e084a9964aa7e9f/src/CloudyCooling.hpp#L298

[...]

PR #575, opened 2024-03-18
Fix amrex::abort not aborting unconditionally on GPUs (for primordial chem and popiii problems)

psharda:
### Description
As described in #543 , `amrex::Abort()` does not unconditionally abort on GPUs. Following the workaround proposed in #543, I have implemented a way to circumvent this issue for problems using microphysics' `burn` (PrimordialChem and PopIIII).

### Related issues
#543 

### Checklist


src/Chemistry.hpp
```diff
@@ -88,14 +93,12 @@ template <typename problem_t> void computeChemistry(amrex::MultiFab &mf, const R
...
-
+			burnsteps(i, j, k) = chemstate.success;
+		
 			if (std::isnan(chemstate.xn[0]) || std::isnan(chemstate.rho)) {
 				amrex::Abort("Burner returned NAN");
 			}
 
-			if (!chemstate.success) {
-				amrex::Abort("VODE integration was unsuccessful!");
-			}
```
psharda, line 109: Now there is no point in having this loop here, right?

src/Chemistry.hpp
```diff
@@ -143,6 +146,11 @@ template <typename problem_t> void computeChemistry(amrex::MultiFab &mf, const R
 			}
 		});
 	}
+
+	int burnmin = burnstepsMF.min(0);
+	if (burnmin == 0) {
+		amrex::Abort("VODE integration was unsuccessful!");
+	}
```
psharda, line 168: I see a potential issue here. We do not do the burn in all the cells, only in cells that have rho > rho_threshold. So, what is the value of `burnstepsMF` for cells where a burn is not done? Is it 0? If yes, then this won't work here.

src/Chemistry.hpp
```diff
@@ -88,15 +93,12 @@ template <typename problem_t> void computeChemistry(amrex::MultiFab &mf, const R
 			// do it in .cpp so that it is not built at compile time for all tests
 			// which would otherwise slow down compilation due to the large RHS file
 			chemburner(chemstate, dt);
+			burnsteps(i, j, k) = chemstate.success;
 
 			if (std::isnan(chemstate.xn[0]) || std::isnan(chemstate.rho)) {
 				amrex::Abort("Burner returned NAN");
 			}
```
psharda, line 100: Don't think this loop is needed? But I must have encountered a case in the past where the burn seemed to be successful but the mass fraction or density returned were NAN, which is why I check for NAN here. But again, this won't work on GPUs all the time. Not sure if 1.) this is still needed and 2.) if yes, how to fix it other than defining new MFs in the way I do to store `chemstate.success`.

BenWibking: You could instead have a MultiFab that stores whether it's successful or not (either 0 or 1 or `true` or `false`). Then you can set `false` for any cell that returns NAN.

src/Chemistry.hpp
BenWibking: If you want to determine whether it succeeded based on the number of steps, then you should instead do `int burnmax = burnstepsMF.max(0);` and check whether `burnmax == max_steps_for_burn`.

Or you could create a boolean / integer multifab where it just stores true or false for success or failure.

BenWibking (changes requested):
Some changes are needed.

You should either check whether the maximum number of steps is equal to a preset maximum number of steps to see if it failed, *or* you should follow Castro and just propagate `true` or `false` values.

src/Chemistry.hpp
psharda: But we dont want to determine if the burn succeeded based on the number of steps. We want to determine if the burn succeeded based on the value of the boolean variable `chemstate.success`

psharda:
We dont use the max number of steps to evaluate if the burn succeeded, we use `chemstate.success`

src/Chemistry.hpp
BenWibking: Ok. In that case, I would recommend following the implementation used in Castro in `Castro::react_state`:
https://github.com/AMReX-Astro/Castro/blob/5728a640ea7e5e7c47532402ff754e6a1a030fe6/Source/reactions/Castro_react.cpp#L407

BenWibking:
> We dont use the max number of steps to evaluate if the burn succeeded, we use `chemstate.success`

That makes sense. I was thinking of what I did for the cooling integrator, where the max steps is used for this purpose. I would recommend carefully looking at what Castro does to handle this:
https://github.com/AMReX-Astro/Castro/blob/5728a640ea7e5e7c47532402ff754e6a1a030fe6/Source/reactions/Castro_react.cpp#L407

If it's not clear, I can explain it in more detail, or you can ask Mike.

src/Chemistry.hpp
```diff
@@ -94,8 +102,16 @@ template <typename problem_t> void computeChemistry(amrex::MultiFab &mf, const R
...
-				amrex::Abort("VODE integration was unsuccessful!");
+				burn_failed = 1;
+			}
+
+#if defined(AMREX_USE_GPU)
+			if (burn_failed) {
+				amrex::Gpu::Atomic::Add(p_num_failed, burn_failed);
 			}
+#else
+			num_failed += burn_failed;
```
psharda: I get an error on this line:

```
/Users/psharda/Documents/quokka/src/Chemistry.hpp:113:15: error: cannot assign to a variable captured by copy in a non-mutable lambda
                        num_failed += burn_failed;
                        ~~~~~~~~~~ ^
```

psharda:
Can you review my implementation

src/Chemistry.hpp
```diff
@@ -31,15 +31,21 @@ AMREX_GPU_DEVICE void chemburner(burn_t &chemstate, Real dt);
...
 	const BL_PROFILE("Chemistry::computeChemistry()");
 	for (amrex::MFIter iter(mf); iter.isValid(); ++iter) {
 		const amrex::Box &indexRange = iter.validbox();
 		auto const &state = mf.array(iter);
 
-		if (dt < 0) {
-			amrex::Abort("Cannot do chemistry with dt < 0!");
-		}
-
 		amrex::ParallelFor(indexRange, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
```
BenWibking, line 47: You can try changing this line to:
`amrex::ParallelFor(indexRange, [=, &num_failed] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {`

src/Chemistry.hpp
```diff
@@ -142,7 +158,19 @@ template <typename problem_t> void computeChemistry(amrex::MultiFab &mf, const R
...
+#endif
 	}
+
+#if defined(AMREX_USE_GPU)
+	num_failed = *(d_num_failed.copyToHost());
+#endif
+
+	burn_success = !num_failed;
+
+	amrex::ParallelDescriptor::ReduceIntMin(burn_success);
```
BenWibking, line 164: You have to do something here with the value of `burn_success` at the end of this function. The reasonable options are to either i) abort the code if burn_success is false, or ii) return the value of burn_success, and let the hydro update figure out what to do.

src/Chemistry.hpp
BenWibking: This is necessary because of https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture.

psharda:
HIP build fails, and gives the same error even though I am using `num_failed` in the ParallelFor loop.

Unrelated, but the HIP build does not set `amdgpu-function-calls=true`.

BenWibking:
> HIP build fails, and gives the same error even though I am using `num_failed` in the ParallelFor loop.
> 
> Unrelated, but the HIP build does not set `amdgpu-function-calls=true`.

Oh, I see. It turns out you can actually use `amrex::Gpu::Atomic::add` on CPU as well, so you don't need separate code for GPU vs CPU. You can just always use the GPU version.

psharda:
> > HIP build fails, and gives the same error even though I am using `num_failed` in the ParallelFor loop.
> > Unrelated, but the HIP build does not set `amdgpu-function-calls=true`.
> 
> Oh, I see. It turns out you can actually use `amrex::Gpu::Atomic::add` on CPU as well, so you don't need separate code for GPU vs CPU. You can just always use the GPU version.

But the quantity we add, `p_num_failed` is defined above only for GPU

BenWibking:
> > > HIP build fails, and gives the same error even though I am using `num_failed` in the ParallelFor loop.
> > > Unrelated, but the HIP build does not set `amdgpu-function-calls=true`.
> > 
> > 
> > Oh, I see. It turns out you can actually use `amrex::Gpu::Atomic::add` on CPU as well, so you don't need separate code for GPU vs CPU. You can just always use the GPU version.
> 
> But the quantity we add, `p_num_failed` is defined above only for GPU

You can remove the ifdefs so it's always defined.

psharda:
Still crashes during build with the same error :/

src/Chemistry.hpp
```diff
@@ -31,16 +31,20 @@ AMREX_GPU_DEVICE void chemburner(burn_t &chemstate, Real dt);
...
 	for (amrex::MFIter iter(mf); iter.isValid(); ++iter) {
 		const amrex::Box &indexRange = iter.validbox();
 		auto const &state = mf.array(iter);
 
-		if (dt < 0) {
-			amrex::Abort("Cannot do chemistry with dt < 0!");
-		}
-
-		amrex::ParallelFor(indexRange, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+		amrex::ParallelFor(indexRange, [=, &num_failed] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
```
BenWibking: You can get rid of `&num_failed`.

src/Chemistry.hpp
```diff
@@ -94,9 +100,14 @@ template <typename problem_t> void computeChemistry(amrex::MultiFab &mf, const R
...
 
 			if (!chemstate.success) {
-				amrex::Abort("VODE integration was unsuccessful!");
+				burn_failed = 1;
 			}
 
+			if (burn_failed) {
+				amrex::Gpu::Atomic::Add(p_num_failed, burn_failed);
+			}
+			num_failed += burn_failed;
```
BenWibking: You can delete this line.

BenWibking:
> Still crashes during build with the same error :/

The point is to avoid using `num_failed` at all inside the `ParallelFor`, so you should delete the capture argument `&num_failed` and remove the assignment to it.

psharda:
> > Still crashes during build with the same error :/
> 
> The point is to avoid using `num_failed` at all inside the `ParallelFor`, so you should delete the capture argument `&num_failed` and remove the assignment to it.




But do you keep these lines then, or remove `num_failed` completely:

https://github.com/psharda/quokka/blob/0c104a569b449cd4f19222c01f03df0caf6a4a50/src/Chemistry.hpp#L40

https://github.com/psharda/quokka/blob/0c104a569b449cd4f19222c01f03df0caf6a4a50/src/Chemistry.hpp#L163

https://github.com/psharda/quokka/blob/0c104a569b449cd4f19222c01f03df0caf6a4a50/src/Chemistry.hpp#L166

BenWibking:
You should probably keep line 163:
https://github.com/psharda/quokka/blob/0c104a569b449cd4f19222c01f03df0caf6a4a50/src/Chemistry.hpp#L163
but remove the ifdefs.

psharda:
> You should probably keep line 163: https://github.com/psharda/quokka/blob/0c104a569b449cd4f19222c01f03df0caf6a4a50/src/Chemistry.hpp#L163 but remove the ifdefs.

Yeah thats what I was thinking, otherwise there's no point of `num_failed` on CPUs 🗡️

psharda:
Okay, the tests should pass now. But 1.) we still have some other `amrex::Abort()` calls in the `ParallelFor` loop, and 2.) I should test that this actually works by starting from 'wrong' initial mass fractions that should make the burn unsuccessful, and we should see an abort on GPU.

src/Chemistry.hpp
```diff
@@ -142,6 +152,19 @@ template <typename problem_t> void computeChemistry(amrex::MultiFab &mf, const R
 				state(i, j, k, HydroSystem<problem_t>::scalar0_index + nn) = inmfracs[nn] * rho; // scale by rho to return partial densities
 			}
 		});
+
+#if defined(AMREX_USE_HIP)
+		amrex::Gpu::streamSynchronize(); // otherwise HIP may fail to allocate the necessary resources.
```
psharda, line 157: @BenWibking why should this be defined only if HIP is used, instead of being defined for all GPU? As in, shouldn't the correct ifdef be `if defined(AMREX_USE_GPU)`?

psharda:
> Unrelated, but the HIP build does not set `amdgpu-function-calls=true`.

How do we ensure this is not the case, in the workflow file for the HIP test? It's not setting it explicitly, but calling a cache.

src/Chemistry.hpp
BenWibking: It's to work around a bug on AMD.

psharda:
I can confirm that this works on both CPU and GPU, and aborts if the burn fails. We can merge it once the last remaining test passes.

BenWibking (approved):

PR #581, opened 2024-03-27
add histogram (PDF) output

BenWibking:
### Description
This adds histogram / probability density function (PDF) outputs (as fixed-width text files) at timestep intervals as the simulation runs.

Bins can be optionally log-spaced. Normalization of the output is left up to the user, with bin volumes calculated (for convenience) in both original variable coordinates and optionally-log-transformed coordinates.

*Example usage:*
```
quokka.hist_temp.type = DiagPDF                         # Diagnostic type
quokka.hist_temp.file = PDFTempDens                     # Output file prefix
quokka.hist_temp.int  = 10                              # Output cadence (in number of coarse steps)
quokka.hist_temp.weight_by = mass                       # (Optional) Weight by: mass, volume, cell_counts
quokka.hist_temp.var_names = temperature gasDensity     # Variable(s) of interest (compute a N-D histogram)

quokka.hist_temp.temperature.nBins = 20                 # temperature: Number of bins
quokka.hist_temp.temperature.log_spaced_bins = 1        # temperature: (Optional, default: 0) Use log-spaced bins
quokka.hist_temp.temperature.range = 1e3 1e7            # temperature: (Optional) Specify the min/max of the bins

quokka.hist_temp.gasDensity.nBins = 5                   # gasDensity: Number of bins
quokka.hist_temp.gasDensity.log_spaced_bins = 1         # gasDensity: (Optional, default: 0) Use log-spaced bins
quokka.hist_temp.gasDensity.range = 1e-29 1e-23         # gasDensity: (Optional) Specify the min/max of the bins
```

*Filters (based on any variables, not necessary those used for the histogram) can be optionally added:*
```
quokka.hist_temp.filters = dense                       # (Optional) List of filters
quokka.hist_temp.dense.field_name = gasDensity         # Filter field
quokka.hist_temp.dense.value_greater = 1e-25           # Filters: value_greater, value_less, value_inrange
```

*Output snippet:*
```
#                     time:   0.00000000000000000e+00
#                    cycle:                         0
#                variables:               temperature                gasDensity
#            is_log_spaced:                         1                         1
          temperature_idx           temperature_min           temperature_max           gasDensity_idx            gasDensity_min            gasDensity_max                  mass_sum
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         0   9.99999999999999943e-30   1.58489319246111091e-28   0.00000000000000000e+00
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         1   1.58489319246111091e-28   2.51188643150957179e-27   0.00000000000000000e+00
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         2   2.51188643150957179e-27   3.98107170553495291e-26   0.00000000000000000e+00
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         3   3.98107170553498563e-26   6.30957344480194281e-25   0.00000000000000000e+00
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         4   6.30957344480194281e-25   9.99999999999999960e-24   0.00000000000000000e+00
```

Based on the [PeleLMeX implementation](https://amrex-combustion.github.io/PeleLMeX/manual/html/LMeXControls.html#run-time-diagnostics) with extensions to N-D histograms, log binning, and mass weighting.

### Related issues
N/A

### Checklist


src/DiagPDF.cpp
```diff
@@ -0,0 +1,337 @@
...
+#include "AMReX_SPACE.H"
+
+#include "DiagPDF.H"
+
+void DiagPDF::init(const std::string &a_prefix, std::string_view a_diagName)
+{
+	DiagBase::init(a_prefix, a_diagName);
+
+	amrex::ParmParse const hist_pp(a_prefix);
+	hist_pp.query("normalized", m_normalized);
```
markkrumholz: What does normalized do? I can't see where in the value of m_normalized is ever referenced again, and from what I can follow you are normalizing regardless of the value of this variable.

src/DiagPDF.cpp
```diff
@@ -0,0 +1,337 @@
...
+	hist_pp.query("volume_weighted", m_volWeighted);
```
markkrumholz: What is the alternative to volume weighted? It seems like it is cell count-weighted, but it is not obvious to me that this is a useful thing to offer or be able to compute. Are there any real circumstances where you'd want cell count-weighted rather than volume-weighted? It seems like the obvious alternative is mass-weighted, and if you don't specify that I would naively expect that selecting volume-weighted to be false would result in mass-weighted, not cell-count weighted. I would suggest either making this the behavior, or, if you want to preserve cell count weighting as an option, changing this to an integer value that can be 0, 1, or 2, with each numerical value corresponding to one of the alternatives for weighting (cell count, volume, mass).

src/simulation.hpp
```diff
@@ -1802,29 +1802,44 @@ void AMRSimulation<problem_t>::FillPatchWithData(int lev, amrex::Real time, amre
...
+
+	amrex::GpuBndryFuncFab<setBoundaryFunctorFaceVar<problem_t>> boundaryFunctor_fc(setBoundaryFunctorFaceVar<problem_t>{});
+	amrex::PhysBCFunct<amrex::GpuBndryFuncFab<setBoundaryFunctorFaceVar<problem_t>>> finePhysicalBoundaryFunctor_fc(geom[lev], BCs, boundaryFunctor_fc);
 
 	if (lev == 0) { // NOTE: used by RemakeLevel
 		// copies interior zones, fills ghost zones
-		amrex::FillPatchSingleLevel(mf, time, fineData, fineTime, 0, icomp, ncomp, geom[lev], finePhysicalBoundaryFunctor, 0);
+		if (cen == quokka::centering::cc) {
+			amrex::FillPatchSingleLevel(mf, time, fineData, fineTime, 0, icomp, ncomp, geom[lev], finePhysicalBoundaryFunctor_cc, 0);
+		} else if (cen == quokka::centering::fc) {
```
markkrumholz: This seems to be in conflict with the PR I just reviewed about fixing face centered BC's. Can you resolve?

markkrumholz (commented):
See inline comments.

src/DiagPDF.cpp
BenWibking: That's a good point. I'll change it so it can be weighted by one of (cell count, volume, mass) with volume as the default.

src/DiagPDF.cpp
BenWibking: ~~Oops, that was a mistake. I will fix it.~~

Fixed.

src/simulation.hpp
BenWibking: Should be fixed now.

src/DiagPDF.cpp
```diff
@@ -0,0 +1,367 @@
...
+			amrex::Real binvol = 1;
+
+			// calculate bin edges, bin volume
+			for (int n = 0; n < nvars; ++n) {
+				int const i = idxVec[n];
+				amrex::Real const transformed_bin_left = transformed_lowBnd[n] + static_cast<amrex::Real>(i) * transformed_binWidth[n];
+				amrex::Real const transformed_bin_right = transformed_bin_left + transformed_binWidth[n];
+				amrex::Real bin_left{NAN};
+				amrex::Real bin_right{NAN};
+
```
markkrumholz, line 355: There is a subtle question of how to normalize when your bins are logarithmically spaced, and your choice here is probably not the one I would make. If you have some variable x by which you're binning, and your bins are spaced logarithmically, then you have a choice about whether to define the width of a bin whose edges are x_1 and x_2 as log(x_2) - log(x_1) = log(x_2 / x_1), or to define it as just x_2 - x_1. The first choice corresponds to computing the probability distribution dp / d log(x), while the second to computing dp/dx over non-uniform bins. Both are valid choices, but personally I think it would make more sense, if the user requests uniformly-spaced logarithmic bins, to compute dp / d log(x) in uniform bins rather than dp / dx in non-uniform bins. If you agree, then you should remove this if clause, and just take bin_left and bin_right to be equal to transformed_bin_left and transformed_bin_right regardless of whether the bins are logarithmic or not. If you do want to keep this code as is, you should just very clearly document what setting the bins to logarithmic means, i.e., it means that even though the bins are logarithmically spaced, the quantity returned is still dp / dx and not dp / d log(x).

BenWibking: The problem is that for multidimensional histograms, it's possible to have some bins that are log-spaced and others that are not, and we don't have a good way to indicate this in a CSV file.

In particular, there is no standard way to included parsable metadata in a csv file, so users are likely to ignore any comments at the top of the file that might attempt to explain this and instead make a severe units error when this scenario presents itself.

~~We could output YAML format files instead, and add metadata that specifies what the precise units are. This should solve the problem, since it will require writing significantly more code on the part of the user to parse the file, and they are more likely to have to carefully examine the format in order to get the histogram read back into memory. On the other hand, that is significantly more work.~~

BenWibking: Actually, we could output the probability mass function instead, and let the user convert it to a density function with whatever units are preferred, since all the information needed to do so in within each row.

I am leaning toward this solution. Does this seem workable to you?

BenWibking:
I've changed it so the normalization is left up to the user.

The output now includes: the bin volume calculated in the original coordinates, the bin volume calculated in the optionally-log-transformed coordinates, and the histogram sum of mass, volume, or cell counts (labeled as either `mass_sum`, `volume_sum`, or `cell_counts_sum`).

Example output (first 5 rows only):
```
              temperature_idx               temperature_min               temperature_max               gasDensity_idx                gasDensity_min                gasDensity_max                  mass_sum                bin_volume    transformed_bin_volume
0   1.00000000000000000e+03   1.58489319246111404e+03 0   9.99999999999999943e-30   1.58489319246111091e-28   0.00000000000000000e+00   8.68503919802354668e-26   2.40000000000000074e-01
0   1.00000000000000000e+03   1.58489319246111404e+03 1   1.58489319246111091e-28   2.51188643150957179e-27   0.00000000000000000e+00   1.37648595012054237e-24   2.40000000000000074e-01
0   1.00000000000000000e+03   1.58489319246111404e+03 2   2.51188643150957179e-27   3.98107170553495291e-26   0.00000000000000000e+00   2.18158321186441207e-23   2.40000000000000074e-01
0   1.00000000000000000e+03   1.58489319246111404e+03 3   3.98107170553498563e-26   6.30957344480194281e-25   0.00000000000000000e+00   3.45757638127138013e-22   2.40000000000000074e-01
0   1.00000000000000000e+03   1.58489319246111404e+03 4   6.30957344480194281e-25   9.99999999999999960e-24   0.00000000000000000e+00   5.47988926909133217e-21   2.40000000000000074e-01
```

src/DiagPDF.cpp
BenWibking: See comment below for my proposed solution.

markkrumholz:
Your proposed solution so just outputting the total cell count / volume / mass in each bin is fine, since then people can indeed normalize however they want, and there is no particular reason to do the normalization in real time as opposed to when plotting / post-processing. I actually think you don't even need to output the bin volume, since this is trivially calculable from the bin edges. So if you just write out the bin edges that should be sufficient.

BenWibking:
> Your proposed solution so just outputting the total cell count / volume / mass in each bin is fine, since then people can indeed normalize however they want, and there is no particular reason to do the normalization in real time as opposed to when plotting / post-processing. I actually think you don't even need to output the bin volume, since this is trivially calculable from the bin edges. So if you just write out the bin edges that should be sufficient.

Ok, done.

I've also added some metadata as a `#` comment that should be ignored by NumPy and other common csv readers, but easy to parse with a regular expression (or something equally simple):
```
#                     time:   0.00000000000000000e+00
#                    cycle:                         0
#                variables:               temperature                gasDensity
#            is_log_spaced:                         1                         1
          temperature_idx           temperature_min           temperature_max           gasDensity_idx            gasDensity_min            gasDensity_max                  mass_sum
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         0   9.99999999999999943e-30   1.58489319246111091e-28   0.00000000000000000e+00
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         1   1.58489319246111091e-28   2.51188643150957179e-27   0.00000000000000000e+00
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         2   2.51188643150957179e-27   3.98107170553495291e-26   0.00000000000000000e+00
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         3   3.98107170553498563e-26   6.30957344480194281e-25   0.00000000000000000e+00
                        0   1.00000000000000000e+03   1.58489319246111404e+03                         4   6.30957344480194281e-25   9.99999999999999960e-24   0.00000000000000000e+00
 ```

markkrumholz (approved):

PR #582, opened 2024-03-27
Remove wavespeed correction in radiation flux

chongchonghe:
### Description

The wavespeed adjustment, $S = c \sqrt{f/\tau}$, in the HLL flux of radiation energy (e.g. [Jiang+2013](https://ui.adsabs.harvard.edu/abs/2013ApJ...767..148J/abstract), [Skinner+2019](https://ui.adsabs.harvard.edu/abs/2019ApJS..241....7S/abstract)) is not necessary any more since we implemented the IMEX PD-ARS scheme. All previous tests passed after this change. Two of the tests get slightly worse results but still within the error tolerance. The fact that it passed all the tests seems to suggest that this correction at diffusion limit is no longer required. In fact, I have shown in the IMEX paper that, at diffusion limit, the wavespeed is roughly $c \sqrt{f/\tau}$. 

Our numerical scheme gives the following discretization:

$$
\frac{\partial}{\partial t} E = \frac{\cal C}{3 \cal L} \frac{\partial^2}{\partial x^2} \frac{E}{3}
$$

which is a diffusion equation with the correct diffusion coefficient. The diffusion speed in time interval $t$ is approximated 

$$
C_{\rm eff} \equiv \frac{\sqrt{D t}}{t} = \sqrt{\frac{\cal C}{3 {\cal L} t}}
$$

In dimensional terms,

$$
C_{\rm eff,\ with \ dim} = 
    \frac{c}{\sqrt{3}} \frac{1}{\sqrt{{\rm CFL} ~\tau_{\rm cell}}}
$$

The $1/\tau$ correction is a natural result of the IMEX scheme. 

A side note: it might be possible to remove the 0.1 c floor of the wavespeeds S_L and S_R while not causing any issues. Simply removing it causes S_L = S_R = 0 and F = NAN in the 3D RadForce test *only*. I'll take a second look at this in a future PR.

### Related issues
Odd-even instability #541 

### Checklist


chongchonghe:
Compare new results of the RadTube test with the old result (1/tau correction). Every other test passed and the results are virtually the same as before. 

New
radiation_pressure_tube

Old
radiation_pressure_tube

src/radiation_system.hpp
```diff
@@ -971,32 +971,15 @@ void RadSystem<problem_t>::ComputeFluxes(array_t &x1Flux_in, array_t &x1FluxDiff
...
-			// quokka::valarray<double, numRadVars_> epsilon = {S_corr, S_corr, S_corr, S_corr}; // Jiang et al. (2013)
-			// quokka::valarray<double, numRadVars_> epsilon = {S_corr * S_corr, S_corr, S_corr, S_corr}; // this code
-
-			// fix odd-even instability that appears in the asymptotic diffusion limit
-			// [for details, see section 3.1: https://ui.adsabs.harvard.edu/abs/2022MNRAS.512.1499R/abstract]
-			if ((i + j + k) % 2 == 1) {
-				// revert to more diffusive flux (has no effect in optically-thin limit)
-				epsilon = {1.0, 1.0, 1.0, 1.0};
-			}
+			// Adjusting wavespeeds is no longer necessary with the IMEX PD-ARS scheme.
```
BenWibking, line 977: Maybe add a link here to this PR for future reference?

BenWibking:
This is interesting! If all the tests pass, then I'm certainly in favor of this.

markkrumholz:
I'm working on the paper now, and I definitely think this should be included, and is in fact a feature we can highlight in the writeup. The corrections to the wavespeeds required in earlier explicit RHD methods always struck me as a bit ad hoc and not very well justified -- they sort of worked in limiting cases, but were of generally unknown accuracy in intermediate cases. Doing away with the need for them is a victory for code comprehensibility and cleanliness.

BenWibking:
> I'm working on the paper now, and I definitely think this should be included, and is in fact a feature we can highlight in the writeup. The corrections to the wavespeeds required in earlier explicit RHD methods always struck me as a bit ad hoc and not very well justified -- they sort of worked in limiting cases, but were of generally unknown accuracy in intermediate cases. Doing away with the need for them is a victory for code comprehensibility and cleanliness.

I agree that doing away with an ad-hoc fix would be a great improvement.

I don't think we really know the accuracy of the method in the intermediate regime, at least with respect to the transition between optically-thin and optically-thick regions. There are examples of transport discretizations that have the asymptotic diffusion limit but still perform terribly in unresolved layers between optically thin and optically-thick regions: https://doi.org/10.1016/0021-9991(89)90229-5. Some of the test problems in that paper might be worth implementing.

chongchonghe:
@BenWibking I noticed that you set the wavespeed estimates to $\max(0.1, \sqrt{T}) c$. What is the consideration for setting the $0.1 c$ floor? Is it from the appendix of Jiang+2013? I removed the floor in the most recent push and all tests still pass, just for testing.

BenWibking:
> @BenWibking I noticed that you set the wavespeed estimates to max(0.1,T)c. What is the consideration for setting the 0.1c floor? Is it from the appendix of Jiang+2013? I removed the floor in the most recent push and all tests still pass, just for testing.

I would not change this. It's there to prevent the wavespeed from becoming zero in pathological cases.

BenWibking:
> @BenWibking I noticed that you set the wavespeed estimates to max(0.1,T)c. What is the consideration for setting the 0.1c floor? Is it from the appendix of Jiang+2013? I removed the floor in the most recent push and all tests still pass, just for testing.

I think this was probably due to have a zero or negative $E_r$. Since you added a floor for the radiation energy, if the tests all pass without it, I'm fine with removing it.

src/RadTube/test_radiation_tube.cpp
```diff
@@ -384,12 +384,12 @@ auto problem_main() -> int
...
-		sol_norm += std::abs(E1_exact[i]);
-		err_norm += std::abs(Erad_arr_numerical_interp_at_group_2[i] - E2_exact[i]);
-		sol_norm += std::abs(E2_exact[i]);
-	}
+	// for (int i = 0; i < static_cast<int>(xs_exact.size()); ++i) {
+	// 	err_norm += std::abs(Erad_arr_numerical_interp_at_group_1[i] - E1_exact[i]);
+	// 	sol_norm += std::abs(E1_exact[i]);
+	// 	err_norm += std::abs(Erad_arr_numerical_interp_at_group_2[i] - E2_exact[i]);
+	// 	sol_norm += std::abs(E2_exact[i]);
+	// }
```
BenWibking, line 392: Is there any reason to keep this commented-out code?

BenWibking (dismissed):
Looks good to me. I would prefer to remove the commented-out code (see inline comment), but if it needs to stay for some reason, I'm fine with merging as-is.

chongchonghe:
I have been working on the paper and other PRs recently. 

We probably want to keep the code for calculating tau_cell and wavespeed adjustments because it helps reducing odd-even instability errors. I'll turn it off by default and the users can optionally turn it on if the instability is a big issue. 

Also, i believe the IMEX scheme will preserve asymptotic diffusion limit in dynamic diffusion regime as well. I'll test it right now. 

I will update this PR shortly.

src/RadTube/test_radiation_tube.cpp
chongchonghe: Yes, I'll keep them for future use. This check of E_r for individual bins is compromised because I precalculated E1 and E2 based on numerical results from one run. I will change it to use the analytical solution later.

chongchonghe:
This is strange. I didn't do anything to the RadForce test but it failed with "max retry" error on GPU. I'll debug it later.

chongchonghe:
Found it. Removing the 0.1 c floor of the wavespeed causes S_L  = S_R = 0 and F = NAN in the 3D RadForce test *only*. 

I'll bring back the 0.1 c floor for now. It might be possible to remove this floor without causing any issues. I'll take a second look at this in a future PR.

Should pass all the tests now and it'll be ready to merge.

BenWibking (approved):
Looks good to me!

chongchonghe:
Here's the status of the odd-even instability issue after this update. 

The instability issue is not as bad as I previously thought. 

The results from the RadhydroPulse test with 64 grids are virtually the same -- they have the same level of instability. However, the magnitude of the speed is more correct -- the peak should be at around 4. 

Left: the old method with $1/\tau$ correction; right: without correction.

CleanShot 2024-04-10 at 01 45 33@2x

With higher resolution (n=512), the instability is tiny but visible, and it's as strong in the new result as with the $1/\tau$ correction. With $1/\tau^2$ correction, there is no visible instability.

No correction:
radhydro_pulse_velocity-without

$1/\tau$:
radhydro_pulse_velocity-with

$1/\tau^2$:
radhydro_pulse_velocity-with-tau-squared

PR #584, opened 2024-03-28
fix face-centered BCs for refined levels

BenWibking:
### Description
Face-centered boundary conditions weren't using the correct physical boundary functor for refined levels.

Since we currently don't evolve any face-centered vars with AMR, this bug should not have affected anyone.

### Related issues
N/A

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -1819,7 +1819,7 @@ void AMRSimulation<problem_t>::FillPatchWithData(int lev, amrex::Real time, amre
 		amrex::PhysBCFunct<amrex::GpuBndryFuncFab<setBoundaryFunctor<problem_t>>> coarsePhysicalBoundaryFunctor_cc(geom[lev - 1], BCs,
 															   boundaryFunctor_cc);
 		amrex::PhysBCFunct<amrex::GpuBndryFuncFab<setBoundaryFunctorFaceVar<problem_t>>> coarsePhysicalBoundaryFunctor_fc(geom[lev - 1], BCs,
-															   boundaryFunctor_fc);
+																  boundaryFunctor_fc);
 
 		// copies interior zones, fills ghost zones with space-time interpolated
 		// data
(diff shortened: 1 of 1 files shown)


src/simulation.hpp
```diff
@@ -1802,29 +1802,44 @@ void AMRSimulation<problem_t>::FillPatchWithData(int lev, amrex::Real time, amre
...
-			// N.B.: this only works for cell-centered data
-			fillpatcher_[lev]->fill(mf, mf.nGrowVect(), time, coarseData, coarseTime, fineData, fineTime, 0, icomp, ncomp,
-						coarsePhysicalBoundaryFunctor, 0, finePhysicalBoundaryFunctor, 0, BCs, 0, pre_interp, post_interp);
+			if (cen == quokka::centering::cc) {
+				// N.B.: this only works for cell-centered data
+				fillpatcher_[lev]->fill(mf, mf.nGrowVect(), time, coarseData, coarseTime, fineData, fineTime, 0, icomp, ncomp,
+							coarsePhysicalBoundaryFunctor_cc, 0, finePhysicalBoundaryFunctor_cc, 0, BCs, 0, pre_interp,
+							post_interp);
+			} else {
+				amrex::Abort("FillPatchType::fillpatch_class cannot be used with non-cell-centered data!");
```
markkrumholz: Is there a problem here? I'm not sure under what circumstances fptype would be set to fillpatch_class, but if this is normally allowed but will cause failures for face-centered data, we should add some comments about the recommended alternative.

markkrumholz (commented):
See my inline comment / question.

src/simulation.hpp
BenWibking: I've added an inline comment that says to use `FillPatchType::fillpatch_function` instead.

Both perform the same operations, but, as an optimization, `fillpatch_class` avoids a communication step in certain circumstances. See https://github.com/AMReX-Codes/amrex/pull/2972 for the full explanation.

markkrumholz (approved):

markkrumholz (approved):

PR #588, opened 2024-03-31
update histogram plotting script

BenWibking:
### Description
Updates the histogram plotting script to read the variables, time, and timestep from the header.

### Related issues
N/A

### Checklist

### DIFF
--- a/scripts/plot_PDF_2D_outputs.py
+++ b/scripts/plot_PDF_2D_outputs.py
@@ -1,6 +1,5 @@
 import warnings
 import numpy as np
-import pandas as pd
 import matplotlib.pyplot as plt
 import argparse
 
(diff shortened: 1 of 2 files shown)


markkrumholz (commented):
Some suggestions:

1. Add a description to the ArgumentParser instantiation line, and a help keyword to the filenames argument, to explain what the script does and what the user is supposed to provide; that way doing "python plot_PDF_outputs.py --help" will work correctly.
2. Plotting using imshow seems appropriate for a 2D PDF, but not for a 1D PDF; that would be better shown as a plot or histogram. Do you want to add an if statement that handles the 1D case? If not, I'd recommend changing the file name to explicitly specify a 2D PDF.
3. You've used the pandas read_csv file, which in general is fine, but adds an extra dependency, since not everyone will have pandas installed. It seems like you could achieve the same effect using numpy.genfromtxt and thereby not pull in the extra dependency. It might be worth doing it that way.

BenWibking:
@markkrumholz I've updated to address your comments.

markkrumholz (approved):

PR #595, opened 2024-04-07
Fix "Newton-Rasphon iteration failed to converge" bug when tau = 0.

chongchonghe:
### Description

Since we base our Newton-Rasphon iteration on (E_gas, R), where R = tau (4 pi B / c - E), when tau = 0, E is not deducible from R. This causes division-by-zero error and E = NAN, which causes the iteration fail to converge. In this PR, I fixed this issue by enforcing d E_g = 0 when tau_g = 0. Here E_g is the radiation energy of the g'th photon group.

To validate this fix, I modified the ShockMultigroup test to include an extra redundant photon bin where kappa is set to 0.

### Checklist


BenWibking (changes requested):
Can you sync your submodules? Those shouldn't be changed in this PR.

chongchonghe:
@BenWibking How can I debug this error on CMake / build? `valarray.hpp:117:17: runtime error: division by zero` . I can't reproduce this error on my macOS or Gadi, even with debug.

BenWibking:
Did you configure with the same CMake options:
`cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=gcc-11 -DCMAKE_CXX_COMPILER=g++-11 -DENABLE_ASAN=ON`

chongchonghe:
Somehow the submodules are mismatched and I don't know how to fix them, so I created a new PR #603 to replace this one.

PR #597, opened 2024-04-07
add computeVolumeIntegral function

BenWibking:
### Description
Adds a function that can be used to a compute a volume integral of a user-defined function that takes an Array4 for a box of the `state_new_cc_` MultiFab as input.

### Related issues
N/A

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -307,6 +307,9 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 
 	template <typename ReduceOp, typename F> auto computePlaneProjection(F const &user_f, int dir) const -> amrex::BaseFab<amrex::Real>;
 
+	// compute volume integrals
+	template <typename F> auto computeVolumeIntegral(F const &user_f) -> amrex::Real;
+
 	// I/O functions
 	[[nodiscard]] auto PlotFileName(int lev) const -> std::string;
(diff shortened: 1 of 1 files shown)


markkrumholz (commented):
Doesn't necessarily have to be in this PR, but a very useful enhancement to this might be to let the user optionally provide a number of ghost zones and grow the state variable by that amount. The reason is that right now this will only work for volume integrals of quantities that don't include any derivatives, so for example you couldn't use this to compute the volume integral of vorticity or something like that. It's fine if you don't want to add that now and you want to leave that as an enhancement for a future PR. I just thought I'd mention it as an idea. If you do want to skip it for now, just let me know and I'll approve this as-is.

BenWibking:
> Doesn't necessarily have to be in this PR, but a very useful enhancement to this might be to let the user optionally provide a number of ghost zones and grow the state variable by that amount. The reason is that right now this will only work for volume integrals of quantities that don't include any derivatives, so for example you couldn't use this to compute the volume integral of vorticity or something like that. It's fine if you don't want to add that now and you want to leave that as an enhancement for a future PR. I just thought I'd mention it as an idea. If you do want to skip it for now, just let me know and I'll approve this as-is.

Let's do that in a future PR.

markkrumholz (approved):

PR #603, opened 2024-04-11
Fix Newton-Rasphon convergence bug when tau = 0

chongchonghe:
### Description

This is a replacement of #595 .

Since we base our Newton-Rasphon iteration on (E_gas, R), where R = tau (4 pi B / c - E), when tau = 0, E is not deducible from R. This causes division-by-zero error and E = NAN, which causes the iteration fail to converge. In this PR, I fixed this issue by enforcing d E_g = 0 when tau_g = 0. Here E_g is the radiation energy of the g'th photon group.

To validate this fix, I modified the ShockMultigroup test to include an extra redundant photon bin where kappa is set to 0.

### Checklist


markkrumholz (approved):

chongchonghe:
The avatargpu test failed 1 test with memory issue. This probably has nothing to do with the code.

PR #610, opened 2024-04-17
Use mass scalars while computing EOS quantities in NSCBC

psharda:
### Description
We currently do not pass mass scalars as an input to the `ComputeSoundSpeed` function in NSCBC. 

### Checklist

### DIFF
--- a/src/NSCBC_outflow.hpp
+++ b/src/NSCBC_outflow.hpp
@@ -38,7 +38,7 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto dQ_dx_outflow(quokka::valarray<amrex::R
 	static constexpr int nmscalars_ = Physics_Traits<problem_t>::numMassScalars;
 	amrex::GpuArray<Real, nmscalars_> massScalars;
 	for (int n = 0; n < nmscalars_; ++n) {
-		massScalars[n] = Q[scalar0_index + n];
+		massScalars[n] = Q[RadSystem<problem_t>::scalar0_index + n];
 	}
 
 	// normal derivatives
--- a/src/NSCBC_outflow.hpp
+++ b/src/NSCBC_outflow.hpp
@@ -35,6 +35,7 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto dQ_dx_outflow(quokka::valarray<amrex::R
 	const amrex::Real w = Q[3];
 	const amrex::Real P = Q[4];
 
+	static constexpr int nmscalars_ = Physics_Traits<problem_t>::numMassScalars;
 	amrex::GpuArray<Real, nmscalars_> massScalars;
 	for (int n = 0; n < nmscalars_; ++n) {
 		massScalars[n] = Q[scalar0_index + n];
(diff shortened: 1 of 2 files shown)


src/NSCBC_outflow.hpp
```diff
@@ -35,6 +35,12 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto dQ_dx_outflow(quokka::valarray<amrex::R
 	const amrex::Real w = Q[3];
 	const amrex::Real P = Q[4];
 
+	static constexpr int nmscalars_ = Physics_Traits<problem_t>::numMassScalars;
+	amrex::GpuArray<Real, nmscalars_> massScalars;
+	for (int n = 0; n < nmscalars_; ++n) {
+		massScalars[n] = Q[RadSystem<problem_t>::scalar0_index + n];
```
psharda: @BenWibking  Is it correct to access the indices of scalars this way?

If yes, we should also use this way to specify the indices of density, etc. above, which are currently hardcoded.

psharda:
I don't think any of the current test problems are using the NSCBC? If true, how can we check the changes actually work

BenWibking:
> I don't think any of the current test problems are using the NSCBC? If true, how can we check the changes actually work

Test problems are here: https://github.com/quokka-astro/quokka/tree/development/src/NSCBC

src/NSCBC_outflow.hpp
BenWibking: Here they follow the ordering defined by the `HydroSystem<problem_t>::ComputePrimVars` function:
https://github.com/quokka-astro/quokka/blob/9d44f9156beba9a5e4b91ce460b4323f2e1663c0/src/hydro_system.hpp#L322

This is not necessarily the same as the ordering used by the HydroSystem/RadSystem indices, although in the current implementation they are the same.

src/NSCBC_outflow.hpp
```diff
@@ -35,6 +35,13 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto dQ_dx_outflow(quokka::valarray<amrex::R
 	const amrex::Real w = Q[3];
 	const amrex::Real P = Q[4];
 
+	static constexpr int nmscalars_ = Physics_Traits<problem_t>::numMassScalars;
+	amrex::GpuArray<Real, nmscalars_> massScalars;
+	for (int n = 0; n < nmscalars_; ++n) {
+		// indexing here follows primVars
+		massScalars[n] = Q[HydroSystem<problem_t>::primScalar0_index + n];
```
psharda, line 42: Is this fine @BenWibking

BenWibking: Yes, I think so.

psharda:
@BenWibking can it be approved now?

BenWibking (approved):

psharda:
Hmm, @BenWibking , one of the radiation tests timed out during merge.

BenWibking:
Ah, that keeps happening. You can retry the merge until it works, or you can increase the timeout for that test. Or ask Chongchong to adjust the parameters so it doesn't take as long to run.

PR #626, opened 2024-05-02
Mean opacity for variable kappa based on piecewise-power-law approximation

chongchonghe:
### Description

This PR implements piecewise-power-law approximation for multigroup radiation transfer. We calculate Planck-mean, energy-mean, and flux-mean opacity given an arbitrary function kappa(nu, T, rho) based on piecewise-power-law fitting to kappa(nu), E(nu), and F(nu). The user can enable this routine by setting `OpacityModel` to `OpacityModel::piecewisePowerLaw`. See examples in test_radhydro_shock_multigroup.cpp and test_radhydro_pulse_MG_int.hpp. The original approach that the opacities are defined via `ComputePlanckOpacity`, `ComputeFluxMeanOpacity`, etc are retained by setting `OpacityModel` to `OpacityModel::user`, which is the default. 

We calculate the power-law slope of a radiation quantity via the following relations:

$$
\alpha_{i} = {\rm minmod}(s_{i-1/2}, s_{i+1/2})
$$

where

$$
s_{i+1/2} = \frac{\ln \frac{X_{i+1}}{\nu_{i+3/2} - \nu_{i+1/2}} - \ln \frac{X_i}{\nu_{i+1/2} - \nu_{i-1/2}}}{\ln \sqrt{\nu_{i+1/2} \nu_{i+3/2}} - \ln \sqrt{\nu_{i-1/2} \nu_{i+1/2}}}
$$

I will try to construct a better estimate of the power-law slope in a separate PR. The user can choose to overwrite this by defining your own `RadSystem<problem_t>::ComputeRadQuantityExponents`; see example in test_radhydro_pulse_MG_int.cpp. 

The power-law fitting of $\kappa(\nu)$ can be specified in two ways. The first method is to specify its power-law slope and lower-bound value for every photon bin, given $T$ and $\rho$. The second method is to specify a precise definition of $\kappa(\nu, T, \rho)$. Then, a power-law fitting to $\kappa(\nu)$ is done on the fly in the following way: 

$$
\begin{gather}
\kappa_{L}^i = \kappa(\nu_{i-1/2}, T, \rho) \\
\alpha^i_{\kappa} = \frac{\ln(\kappa(\nu_{i+1/2}, T, \rho) / \kappa(\nu_{i-1/2},T,\rho))}{\ln(\nu_{i+1/2} / \nu_{i-1/2})} \\
\kappa_{{\rm fit}}^i(\nu) = \kappa_{L}^i \left( \frac{\nu}{\nu_{i-1/2}} \right)^{\alpha_{\kappa}^i} 
\end{gather}
$$

This is definitely not the most accurate way to do this as it is not volume conservative, but it's simple and this is important since it's done for every cell. An alternative is to construct $\kappa_{L,i}$ and $\alpha_{\kappa,i}$ in the beginning if $\kappa$ does not depend on $T$ or $\rho$. 

### Related issues

N/A

### Checklist

### DIFF
--- a/src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
+++ b/src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
@@ -16,6 +16,8 @@ struct MGintProblem {
 struct GreyProblem { // Advecting grey pulse
 };
 
+AMREX_GPU_MANAGED double spec_power = -1.0;
+
 // set include_grey to false for the convenience of debugging multigroup interpolation
 static constexpr bool include_grey = true;
 static constexpr bool export_csv = true;
@@ -26,10 +28,10 @@ static constexpr bool export_csv = true;
 // constexpr amrex::GpuArray<double, n_groups_ + 1> rad_boundaries_{1e16, 1e18, 1e20};
 // constexpr int n_groups_ = 4;
 // constexpr amrex::GpuArray<double, n_groups_ + 1> rad_boundaries_{1e16, 1e17, 1e18, 1e19, 1e20};
-// constexpr int n_groups_ = 8;
-// constexpr amrex::GpuArray<double, n_groups_ + 1> rad_boundaries_{1e16, 3.16e16, 1e17, 3.16e17, 1e18, 3.16e18, 1e19, 3.16e19, 1e20};
-constexpr int n_groups_ = 16;
-constexpr amrex::GpuArray<double, n_groups_ + 1> rad_boundaries_{1.00000000e+16, 1.77827941e+16, 3.16227766e+16, 5.62341325e+16, 1.00000000e+17, 1.77827941e+17, 3.16227766e+17, 5.62341325e+17, 1.00000000e+18, 1.77827941e+18, 3.16227766e+18, 5.62341325e+18, 1.00000000e+19, 1.77827941e+19, 3.16227766e+19, 5.62341325e+19, 1.00000000e+20};
+constexpr int n_groups_ = 8;
+constexpr amrex::GpuArray<double, n_groups_ + 1> rad_boundaries_{1e16, 3.16e16, 1e17, 3.16e17, 1e18, 3.16e18, 1e19, 3.16e19, 1e20};
+// constexpr int n_groups_ = 16;
+// constexpr amrex::GpuArray<double, n_groups_ + 1> rad_boundaries_{1.00000000e+16, 1.77827941e+16, 3.16227766e+16, 5.62341325e+16, 1.00000000e+17, 1.77827941e+17, 3.16227766e+17, 5.62341325e+17, 1.00000000e+18, 1.77827941e+18, 3.16227766e+18, 5.62341325e+18, 1.00000000e+19, 1.77827941e+19, 3.16227766e+19, 5.62341325e+19, 1.00000000e+20};
 // constexpr int n_groups_ = 32;
 // constexpr amrex::GpuArray<double, n_groups_ + 1> rad_boundaries_{1.00000000e+16, 1.33352143e+16, 1.77827941e+16, 2.37137371e+16, 3.16227766e+16, 4.21696503e+16, 5.62341325e+16, 7.49894209e+16, 1.00000000e+17, 1.33352143e+17, 1.77827941e+17, 2.37137371e+17, 3.16227766e+17, 4.21696503e+17, 5.62341325e+17, 7.49894209e+17, 1.00000000e+18, 1.33352143e+18, 1.77827941e+18, 2.37137371e+18, 3.16227766e+18, 4.21696503e+18, 5.62341325e+18, 7.49894209e+18, 1.00000000e+19, 1.33352143e+19, 1.77827941e+19, 2.37137371e+19, 3.16227766e+19, 4.21696503e+19, 5.62341325e+19, 7.49894209e+19, 1.00000000e+20};
 // constexpr int n_groups_ = 64;
@@ -53,6 +55,7 @@ constexpr double k_B = C::k_B;
 
 // static diffusion: (for single group) tau = 2e3, beta = 3e-5, beta tau = 6e-2
 constexpr double v0_adv = 1.0e6;    // advecting pulse
+// constexpr double v0_adv = 0.0;
 constexpr double max_time = 4.8e-5; // max_time = 2.0 * width / v1;
 // constexpr double max_time = 4.8e-6;
 
@@ -118,6 +121,17 @@ template <> struct RadSystem_Traits<GreyProblem> {
 	static constexpr int opacity_model = 0;
 };
 
+template <>
+template <typename ArrayType>
+AMREX_GPU_HOST_DEVICE auto RadSystem<MGintProblem>::ComputeRadQuantityExponents(ArrayType const &/*quant*/, amrex::GpuArray<double, nGroups_ + 1> const &/*boundaries*/) -> amrex::GpuArray<double, nGroups_>
+{
+	amrex::GpuArray<double, nGroups_> exponents{};
+	for (int g = 0; g < nGroups_; ++g) {
+		exponents[g] = spec_power;
+	}
+	return exponents;
+}
+
 AMREX_GPU_HOST_DEVICE
 auto compute_initial_Tgas(const double x) -> double
 {
@@ -143,21 +157,58 @@ auto compute_kappa(const double nu, const double Tgas) -> double
 	return kappa0 * std::pow(T_, -0.5) * std::pow(nu_, -3) * (1.0 - std::exp(-coeff_ * nu_ / T_));
 }
 
+// template <>
+// AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<MGintProblem>::DefineOpacityExponentsAndLowerValues(const double rho, const double Tgas)
+//     -> amrex::GpuArray<amrex::GpuArray<double, nGroups_>, 2>
+// {
+// 	amrex::GpuArray<double, nGroups_> exponents{};
+// 	amrex::GpuArray<double, nGroups_> kappa_lower{};
+// 	for (int g = 0; g < nGroups_; ++g) {
+// 		auto kappa_up = compute_kappa(rad_boundaries_[g + 1], Tgas);
+// 		auto kappa_down = compute_kappa(rad_boundaries_[g], Tgas);
+// 		exponents[g] = std::log(kappa_up / kappa_down) / std::log(rad_boundaries_[g + 1] / rad_boundaries_[g]);
+// 		kappa_lower[g] = kappa_down / rho;
+// 		AMREX_ASSERT(!std::isnan(exponents[g]));
+// 		AMREX_ASSERT(kappa_lower[g] >= 0.);
+// 	}
+// 	amrex::GpuArray<amrex::GpuArray<double, nGroups_>, 2> const exponents_and_values{exponents, kappa_lower};
+// 	return exponents_and_values;
+// }
+
 template <>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto RadSystem<MGintProblem>::DefineOpacityExponentsAndLowerValues(const double rho, const double Tgas)
     -> amrex::GpuArray<amrex::GpuArray<double, nGroups_>, 2>
 {
 	amrex::GpuArray<double, nGroups_> exponents{};
+	// amrex::GpuArray<double, nGroups_> kappa_at_edges{};
 	amrex::GpuArray<double, nGroups_> kappa_lower{};
+	amrex::GpuArray<double, nGroups_> kappa_lower_over_rho{};
 	for (int g = 0; g < nGroups_; ++g) {
+		// kappa_at_edges[g] = compute_kappa(rad_boundaries_[g], Tgas);
 		auto kappa_up = compute_kappa(rad_boundaries_[g + 1], Tgas);
 		auto kappa_down = compute_kappa(rad_boundaries_[g], Tgas);
 		exponents[g] = std::log(kappa_up / kappa_down) / std::log(rad_boundaries_[g + 1] / rad_boundaries_[g]);
-		kappa_lower[g] = kappa_down / rho;
-		AMREX_ASSERT(!std::isnan(exponents[g]));
-		AMREX_ASSERT(kappa_lower[g] >= 0.);
+		kappa_lower_over_rho[g] = kappa_down / rho;
+		// double log_nu_up = std::log(rad_boundaries_[g + 1]);
+		// double log_nu_down = std::log(rad_boundaries_[g]);
+		// double delta_x = log_nu_up - log_nu_down;
+		// double delta_y = std::log(kappa_up) - std::log(kappa_down);
+		// // trapezoidal rule
+		// const int N = 10;
+		// double area = std::log(kappa_down);
+		// for (int i = 1; i < N; ++i) {
+		// 	double log_nu = log_nu_down + i * delta_x / N;
+		// 	double kappa = compute_kappa(std::exp(log_nu), Tgas);
+		// 	area += 2.0 * std::log(kappa);
+		// }
+		// area += std::log(kappa_up);
+		// area *= delta_x / (2.0 * N);
+		// kappa_lower[g] = std::exp((area - 0.5 * delta_x * delta_x * exponents[g]) / delta_x);
+		// kappa_lower_over_rho[g] = kappa_lower[g] / rho;
+		// AMREX_ASSERT(!std::isnan(exponents[g]));
+		// AMREX_ASSERT(kappa_lower[g] >= 0.);
 	}
-	amrex::GpuArray<amrex::GpuArray<double, nGroups_>, 2> const exponents_and_values{exponents, kappa_lower};
+	amrex::GpuArray<amrex::GpuArray<double, nGroups_>, 2> const exponents_and_values{exponents, kappa_lower_over_rho};
 	return exponents_and_values;
 }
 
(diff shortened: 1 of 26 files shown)


src/radiation_system.hpp
```diff
@@ -1119,11 +1284,24 @@
...
+					for (int g = 0; g < nGroups_; ++g) {
+						if (tau[g] > 0.0) {
+							F_D_abs_sum += std::abs(F_D[g]);
+							F_G += (c / chat) * Rvec[g];
+						}
+					}
+
+					// Add a breakpoint for debugging
+					if (n >= maxIter - 50) {
+						n = n;
```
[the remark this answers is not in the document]
chongchonghe: Resolved

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
markkrumholz, line 1: A note on naming: I think calling the frequency-integrated case grey is a bit confusing, because normally grey means an approximation where you ignore the frequency-dependence of the opacity. That's not what you're doing in the case you call grey; instead, you're using the exact result for the highly optically thick regime and evaluating the integrals analytically. I therefore think "grey" here should be replaced by "exact" or "analytic" or "frequency-integrated" or words to that effect.

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
```diff
@@ -0,0 +1,551 @@
...
+struct GreyProblem { // Advecting grey pulse
+};
+
+AMREX_GPU_MANAGED double spec_power = -1.0; // NOLINT
+static constexpr bool export_csv = true;
+
+// constexpr int n_groups_ = 2;
+// constexpr amrex::GpuArray<double, n_groups_ + 1> rad_boundaries_{1e16, 1e18, 1e20};
+constexpr int n_groups_ = 4;
+constexpr amrex::GpuArray<double, n_groups_ + 1> rad_boundaries_{1e16, 1e17, 1e18, 1e19, 1e20};
```
markkrumholz, line 26: A design choice: it looks here like you're requiring the groups to be set at compile time. Do we want to keep that, or have the number of groups / group boundaries be settable at run-time, via the problem file? I guess I would favor setting things at run-time in the same way that we set the number of cells in the spatial dimensions at run-time, unless there are strong code speed / efficiency reasons to set them at compile time.

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
```diff
@@ -0,0 +1,551 @@
...
+template <> struct RadSystem_Traits<MGintProblem> {
+	static constexpr double c_light = c;
+	static constexpr double c_hat = chat;
+	static constexpr double radiation_constant = a_rad;
+	static constexpr double Erad_floor = erad_floor;
+	static constexpr bool compute_v_over_c_terms = true;
+	static constexpr double energy_unit = h_planck;
+	static constexpr amrex::GpuArray<double, n_groups_ + 1> radBoundaries = rad_boundaries_;
+	static constexpr int beta_order = 1;
+	static constexpr int opacity_model = 1;
```
markkrumholz: Can we provide an enum here in place of an integer value for opacity model, in order to make the code more readable?

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
```diff
@@ -0,0 +1,551 @@
...
+#include "physics_info.hpp"
+#include "planck_integral.hpp"
+#include "radiation_system.hpp"
+
+struct MGintProblem {
+};		     // dummy type to allow compile-type polymorphism via template specialization
+struct GreyProblem { // Advecting grey pulse
+};
+
+AMREX_GPU_MANAGED double spec_power = -1.0; // NOLINT
```
markkrumholz, line 20: We discussed issues with hard-coding this value. Has there been progress on that?

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
```diff
@@ -0,0 +1,551 @@
...
+		const auto Egas = values.at(RadSystem<MGintProblem>::gasInternalEnergy_index)[i];
+		xs.at(i) = x - drift;
+		rhogas.at(index_) = rho_t;
+		Trad.at(index_) = Trad_t;
+		Tgas.at(index_) = quokka::EOS<MGintProblem>::ComputeTgasFromEint(rho_t, Egas);
+		Vgas.at(index_) = 1e-5 * (v_t - v0_adv);
+	}
+	// END OF PROBLEM 1
+
+	// Problem 3: grey radiation
```
markkrumholz: Shouldn't this be problem 2?

markkrumholz (commented):
See inline comments. I know we're still playing around with how to get the exponents for the radiation energy density and flux, so if you want as an intermediate step I can approve this version that has a fixed value for them. Just let me know.

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
chongchonghe: I thought about that and found there are difficulties in reading rad_boundaries from the parameter file. The main difficulty is that many functions take `GpuArray(double, nGroups_)` as an input, so if nGroups_ has to read from the parameter file, then you have to replace GpuArray with a dynamic memory array like Vector. A lot of things have to be rewritten. I'll try to make that work in a future PR.

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
chongchonghe: Done. I changed it to ExactProblem.

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
markkrumholz: OK, happy to leave this as set at compile-time for now.

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
chongchonghe: You can demand free slope to be fit on the fly by removing the ComputeRadQuantityExponents function in the cpp file. Here spec_power = -1 is set for testing purpose only. It gives a solution very close to the exact one, so I can compare it with the exact solution and set a reasonable error_tol. Currently, a free power-law slope from the minmod method would produce a result with a much larger error when nGroup <= 4, which is what I use for testing. In a future PR, I will try to work out a volume-averaged Piecewise power law method.

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
chongchonghe: Done.

chongchonghe:
> See inline comments. I know we're still playing around with how to get the exponents for the radiation energy density and flux, so if you want as an intermediate step I can approve this version that has a fixed value for them. Just let me know.

Yes, I plan to use a constant power law as an intermediate step and implement a better real-time fitting in a future PR.

BenWibking:
Is there any way this can be done with an approximate logarithm function? The issue is that transcendental functions on GPUs have to use special hardware units that are not able to process as many threads simultaneously as the normal arithmetic units can, so there can be a severe performance cost.

If this is not possible (and I think it might not be), can you benchmark the radiation solver performance before/after this PR on either AMD GPUs or NVIDIA GPUs (or both)?

chongchonghe:
> Is there any way this can be done with an approximate logarithm function? The issue is that transcendental functions on GPUs have to use special hardware units that are not able to process as many threads simultaneously as the normal arithmetic units can, so there can be a severe performance cost.
> 
> If this is not possible (and I think it might not be), can you benchmark the radiation solver performance before/after this PR on either AMD GPUs or NVIDIA GPUs (or both)?

Yes. $\log(x)$ can be approximated with 

$$
\log(x) = \lim_{a \rightarrow 0} \frac{x^a - 1}{a}
$$

By setting $a = 10^{-7}$, the relative error is below 4e-6 for x between 1e-30 and 1e30, according to Mathematica. @BenWibking Do you think this is worth doing? 

I was doing the opposite in ComputeGroupMeanOpacity. I need to calculate the above equation, and when a < 1e-8, I turned to using std::log(x). See here: 

https://github.com/quokka-astro/quokka/blob/926370c3b71c06dd4da7193ce868f27e2b578cf0/src/radiation_system.hpp#L967

BenWibking:
> > Is there any way this can be done with an approximate logarithm function? The issue is that transcendental functions on GPUs have to use special hardware units that are not able to process as many threads simultaneously as the normal arithmetic units can, so there can be a severe performance cost.
> > If this is not possible (and I think it might not be), can you benchmark the radiation solver performance before/after this PR on either AMD GPUs or NVIDIA GPUs (or both)?
> 
> Yes. log⁡(x) can be approximated with
> 
> log⁡(x)=lima→0xa−1a
> 
> By setting a=10−7, the relative error is below 4e-6 for x between 1e-30 and 1e30, according to Mathematica. @BenWibking Do you think this is worth doing?
> 
> I was doing the opposite in ComputeGroupMeanOpacity. I need to calculate the above equation, and when a < 1e-8, I turned to using std::log(x). See here:
> 
> https://github.com/quokka-astro/quokka/blob/926370c3b71c06dd4da7193ce868f27e2b578cf0/src/radiation_system.hpp#L967

Unfortunately, I think `std::pow` has the same issue, so I don't think that will help.

Any approximate form should either be a rational polynomial interpolant, or else use `std::ldexp` and `std::frexp`, which can be used to approximately compute 2^x and log base 2: https://en.cppreference.com/w/cpp/numeric/math/ldexp.

chongchonghe:
> > > Is there any way this can be done with an approximate logarithm function? The issue is that transcendental functions on GPUs have to use special hardware units that are not able to process as many threads simultaneously as the normal arithmetic units can, so there can be a severe performance cost.
> > > If this is not possible (and I think it might not be), can you benchmark the radiation solver performance before/after this PR on either AMD GPUs or NVIDIA GPUs (or both)?
> > 
> > 
> > Yes. log⁡(x) can be approximated with
> > log⁡(x)=lima→0xa−1a
> > By setting a=10−7, the relative error is below 4e-6 for x between 1e-30 and 1e30, according to Mathematica. @BenWibking Do you think this is worth doing?
> > I was doing the opposite in ComputeGroupMeanOpacity. I need to calculate the above equation, and when a < 1e-8, I turned to using std::log(x). See here:
> > https://github.com/quokka-astro/quokka/blob/926370c3b71c06dd4da7193ce868f27e2b578cf0/src/radiation_system.hpp#L967
> 
> Unfortunately, I think `std::pow` has the same issue, so I don't think that will help.
> 
> Any approximate form should either be a rational polynomial interpolant, or else use `std::ldexp` and `std::frexp`, which can be used to approximately compute 2^x and log base 2: https://en.cppreference.com/w/cpp/numeric/math/ldexp.

In that case, I'll leave that for future work. If you can approximate log_2(x), then you can approximate log(x) with log(x) = log_2(x) * log(2). 

For now, I'll do a quick benchmark as you suggested.

chongchonghe:
The 3D GPU test is taking very long. This PR is fine: pulse_MG_int takes 2.5 hours on gadi. However, the old pulse_MG test will take a projected time of 100 hours to finish. I suspect this is because of the low `resid_tol` value at 1e-13, while in this new PR I changed it to 1e-11. I reduced the resid_tol of the old pulse_MG test and I'll wait till it finishes and compare with the new test.

chongchonghe:
OK. I did a benchmark on gadi gpuvolta for the old pulse-MG test (bin-centered opacity) and the new pulse-MG-int (bin integrated opacity with piecewise power-law) test. The elapsed time is 45min vs 37min. The new test is faster, but I realized it is not a fair test. The old test runs two simulations, each with 4 photon groups. The new test runs two simulations, one with 4 photon groups and the other with 1 group (exact solution where kappa is integrated over the whole spectrum). The amount of computation is roughly proportional to the nGroups, so the elapsed time per group are: old: 45/8=5.625, new: 37/5=7.4. So, the difference is not large. I wouldn't worry too much about transcendental functions on GPU.

BenWibking:
> OK. I did a benchmark on gadi gpuvolta for the old pulse-MG test (bin-centered opacity) and the new pulse-MG-int (bin integrated opacity with piecewise power-law) test. The elapsed time is 45min vs 37min. The new test is faster, but I realized it is not a fair test. The old test runs two simulations, each with 4 photon groups. The new test runs two simulations, one with 4 photon groups and the other with 1 group (exact solution where kappa is integrated over the whole spectrum). The amount of computation is roughly proportional to the nGroups, so the elapsed time per group are: old: 45/8=5.625, new: 37/5=7.4. So, the difference is not large. I wouldn't worry too much about transcendental functions on GPU.

Can you run a test with a 128^3 grid (with `blocking_factor=128` and `max_grid=128`)? For a performance benchmark, it is not necessary to run the whole problem, only $\sim 10$ timesteps.

chongchonghe:
@BenWibking 
OK, here are the test results on gadi with 1 GPU. The parameters are 128^3 grids running for 10 time steps. Both the total run time and `μs/zone-update` are very similar, so the conclusion is there is no significant performance issue from this PR. 

Old, bin-centered approach, `opacity_model = OpacityModel::user`:

```
186 Performance figure-of-merit: 0.7211298889 μs/zone-update [1.386712734 Mupdates/s]
187 Zone-updates on level 0: 20971520
188
189 Relative L1 error norm = 1.646387658e-05
190 elapsed time: 1459.157303 seconds.
```

New, `opacity_model = OpacityModel::piecewisePowerLaw`

```
186 Performance figure-of-merit: 0.8290508373 μs/zone-update [1.206198649 Mupdates/s]
187 Zone-updates on level 0: 20971520
188
189 Relative L1 error norm = 1.60272655e-05
190 elapsed time: 1394.70028 seconds.
```

With 4 GPUs, the performance is similar, too. These takes (about 4x) shorter time to run, although the reported μs/zone-update is bigger.

Old,

```
116 Performance figure-of-merit: 17.29543015 μs/zone-update [0.05781874122 Mupdates/s]
117 Zone-updates on level 0: 20971520
```

new:

```
116 Performance figure-of-merit: 16.46246597 μs/zone-update [0.06074424098 Mupdates/s]
117 Zone-updates on level 0: 20971520
```

BenWibking:
Ok, that seems reasonable to me.

As a final check, can you see how many iterations per zone there are on average for your benchmark runs?

When I tested the radiation code in the methods paper using 128^3 grids, I got 40 million updates/sec/GPU. However, that problem was optically thin, so there number of iterations should have been close to 1.

chongchonghe:
> Ok, that seems reasonable to me.
> 
> As a final check, can you see how many iterations per zone there are on average for your benchmark runs?
> 
> When I tested the radiation code in the methods paper using 128^3 grids, I got 40 million updates/sec/GPU. However, that problem was optically thin, so there number of iterations should have been close to 1.

@BenWibking 
Do you mean the number of Newton-Raphson iterations? How would you check that? I don't see a global counter for it.

BenWibking:
> > Ok, that seems reasonable to me.
> > As a final check, can you see how many iterations per zone there are on average for your benchmark runs?
> > When I tested the radiation code in the methods paper using 128^3 grids, I got 40 million updates/sec/GPU. However, that problem was optically thin, so there number of iterations should have been close to 1.
> 
> @BenWibking Do you mean the number of Newton-Raphson iterations? How would you check that? I don't see a global counter for it.

Yes. There currently isn't one, but there is one for the number of cooling iterations, which you could very nearly copy and paste from: https://github.com/quokka-astro/quokka/blob/d7f4b43f4519a2c662de1d725db40833b52dde9c/src/TabulatedCooling.hpp#L309

chongchonghe:
> > > Ok, that seems reasonable to me.
> > > As a final check, can you see how many iterations per zone there are on average for your benchmark runs?
> > > When I tested the radiation code in the methods paper using 128^3 grids, I got 40 million updates/sec/GPU. However, that problem was optically thin, so there number of iterations should have been close to 1.
> > 
> > 
> > @BenWibking Do you mean the number of Newton-Raphson iterations? How would you check that? I don't see a global counter for it.
> 
> Yes. There currently isn't one, but there is one for the number of cooling iterations, which you could very nearly copy and paste from:
> 
> https://github.com/quokka-astro/quokka/blob/d7f4b43f4519a2c662de1d725db40833b52dde9c/src/TabulatedCooling.hpp#L309

OK, I'll try that later. Gadi is currently down.

chongchonghe:
I did the same test with the RadForce problem instead. This is in the optically-thin limit and the performance is 8.3 Mupdates/s. Note that this is done with 2 photon groups, and the computational complexity is supposed to scale linearly with the number of photon groups (with a ~2.5 jump from 1 group to 2 groups due to additional setup for multigroup, e.g. interpolating the Planck integral). This brings this test to roughly 20 Mupdates/s/GPU per group. This also applies to the RadhydroPulseMGint problem presented above which uses 4 groups. 

Old, 
```
 86 Performance figure-of-merit: 0.119798939 μs/zone-update [8.347319334 Mupdates/s]
 87 Zone-updates on level 0: 20971520
```

New, PPL 

```
 86 Performance figure-of-merit: 0.1201915345 μs/zone-update [8.320053521 Mupdates/s]
 87 Zone-updates on level 0: 20971520
```

BenWibking:
> I did the same test with the RadForce problem instead. This is in the optically-thin limit and the performance is 8.3 Mupdates/s. Note that this is done with 2 photon groups, and the computational complexity is supposed to scale linearly with the number of photon groups (with a ~2.5 jump from 1 group to 2 groups due to additional setup for multigroup, e.g. interpolating the Planck integral). This brings this test to roughly 20 Mupdates/s/GPU per group. This also applies to the RadhydroPulseMGint problem presented above which uses 4 groups.
> 
> Old,
> 
> ```
>  86 Performance figure-of-merit: 0.119798939 μs/zone-update [8.347319334 Mupdates/s]
>  87 Zone-updates on level 0: 20971520
> ```
> 
> New, PPL
> 
> ```
>  86 Performance figure-of-merit: 0.1201915345 μs/zone-update [8.320053521 Mupdates/s]
>  87 Zone-updates on level 0: 20971520
> ```

Ok, I am satisfied with this.

(Sometime in the future, we should figure out why it's a factor of ~2 slower than the version from the original paper.)

chongchonghe:
@markkrumholz Should we merge this PR? I think the only thing I haven't addressed is the Piecewise powerlaw reconstruction of the radiation quantities, which I haven't finished and I'll leave for the next PR.

markkrumholz:
> @markkrumholz Should we merge this PR? I think the only thing I haven't addressed is the Piecewise powerlaw reconstruction of the radiation quantities, which I haven't finished and I'll leave for the next PR.

Will look today and approve. It’s first thing in the morning here.

markkrumholz (approved):

PR #627, opened 2024-05-04
change initial condition of RadForce

chongchonghe:
### Description

The RadForce test can be improved to be more realistic and less `ad hoc`. Instead of initializing with the exact solution and letting the simulation maintain a static state, we start the simulation with a uniform density of rho0 and zero velocity. The left boundary condition is kept the same as before, but the right boundary is set to be foextrap. The end result is exactly the same as before: the radiation blows away the gas and the system comes to the static state given by the differential equation. 

### Checklist

### DIFF
--- a/src/RadForce/test_radiation_force.cpp
+++ b/src/RadForce/test_radiation_force.cpp
@@ -286,10 +286,8 @@ auto problem_main() -> int
 	std::vector<double> xs(nx);
 	std::vector<double> rho_arr(nx);
 	std::vector<double> rho_exact_arr(nx);
-	std::vector<double> rho_err(nx);
 	std::vector<double> vx_arr(nx);
 	std::vector<double> vx_exact_arr(nx);
-	std::vector<double> Frad_err(nx);
 
 	auto *const x_ptr = x_arr.data();
 	auto *const rho_ptr = rho_arr.data();
@@ -313,8 +311,6 @@ auto problem_main() -> int
 
 		vx_arr.at(i) = vx / a0;
 		vx_exact_arr.at(i) = vx_exact / a0;
-		Frad_err.at(i) = (Frad - Frad0) / Frad0;
-		rho_err.at(i) = (rho - rho_exact) / rho_exact;
 		rho_exact_arr.at(i) = rho_exact;
 		rho_arr.at(i) = rho;
 	}
--- a/src/RadForce/test_radiation_force.cpp
+++ b/src/RadForce/test_radiation_force.cpp
@@ -15,6 +15,7 @@
 #include "AMReX_REAL.H"
 
 #include "ArrayUtil.hpp"
+#include "NSCBC_outflow.hpp"
 #include "RadhydroSimulation.hpp"
 #include "fextract.hpp"
 #include "hydro_system.hpp"
@@ -22,7 +23,6 @@
 #include "physics_info.hpp"
 #include "radiation_system.hpp"
 #include "test_radiation_force.hpp"
-#include "NSCBC_outflow.hpp"
 #ifdef HAVE_PYTHON
 #include "matplotlibcpp.h"
 #endif
--- a/src/RadForce/test_radiation_force.cpp
+++ b/src/RadForce/test_radiation_force.cpp
@@ -286,10 +286,8 @@ auto problem_main() -> int
 	std::vector<double> xs(nx);
 	std::vector<double> rho_arr(nx);
 	std::vector<double> rho_exact_arr(nx);
-	std::vector<double> rho_err(nx);
 	std::vector<double> vx_arr(nx);
 	std::vector<double> vx_exact_arr(nx);
-	std::vector<double> Frad_err(nx);
 
 	auto *const x_ptr = x_arr.data();
 	auto *const rho_ptr = rho_arr.data();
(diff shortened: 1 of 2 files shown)


chongchonghe:
I made this change in order to do a scaling test for the RHD solver, but the result is surprising: the R1 norm of the error does not change with decreasing dt (see table below). I don't know why this happens. 

| dt    | nx   | L1 norm         | elasped time |
| ----- | ---- | --------------- | ------------ |
| 2e8   | 128  | 0.0004315252173 | 4.344377     |
| 1e8   | 128  | 0.0004296931455 | 6.812964     |
| 5e7   | 128  | 0.0004292276377 | 9.439731     |
| 2e8   | 64   | 0.0008842796415 | 2.600089     |
| 1e8   | 64   | 0.0008833405693 | 3.727579     |
| 2.5e7 | 64   | 0.0008808892701 | 12.810238    |

chongchonghe:
I just realized that this problem is in the diffusion limit, so it's not a good test for scaling of the time integration scheme. But in any case, this PR is worth merging.

BenWibking:
> I made this change in order to do a scaling test for the RHD solver, but the result is surprising: the R1 norm of the error does not change with decreasing dt (see table below). I don't know why this happens.
> 
> dt	nx	L1 norm	elasped time
> 2e8	128	0.0004315252173	4.344377
> 1e8	128	0.0004296931455	6.812964
> 5e7	128	0.0004292276377	9.439731
> 2e8	64	0.0008842796415	2.600089
> 1e8	64	0.0008833405693	3.727579
> 2.5e7	64	0.0008808892701	12.810238

This will happen if the error is dominated by the spatial discretization rather than the time discretization. In that case, you only get significantly smaller errors if you increase the spatial resolution.

BenWibking:
For a test problem, I don't think there's anything wrong with initializing to the exact solution. Is there any other benefit to changing this?

chongchonghe:
> For a test problem, I don't think there's anything wrong with initializing to the exact solution. Is there any other benefit to changing this?

I just think a test that is initialized with the exact solution is not super reliable. Support you accidentally changed some of the problem parameters and the simulation finishes at t = 0, or the system is frozen for some reason and never evolves, the test will still pass. This PR shows you can start from a realistic initial condition and end with the exact analytic solution, so why not apply it?

src/RadForce/test_radiation_force.cpp
```diff
@@ -258,7 +239,7 @@ auto problem_main() -> int
 	for (int n = 0; n < nvars; ++n) {
 		// for x-axis:
 		BCs_cc[n].setLo(0, amrex::BCType::ext_dir);
-		BCs_cc[n].setHi(0, amrex::BCType::ext_dir);
+		BCs_cc[n].setHi(0, amrex::BCType::foextrap); // extrapolate
```
BenWibking: Using `foextrap` makes the problem mathematically ill-posed. Can you try using `NSCBC::setOutflowBoundary` or `NSCBC::setOutflowBoundaryLowOrder`: https://github.com/quokka-astro/quokka/blob/4ab068ebda59e45ab21eccb05aae17985d570548/src/NSCBC_outflow.hpp#L358?

BenWibking: This has to called from within `setCustomBoundaryConditions` above.

chongchonghe:
@BenWibking Can you take a look at the `setOutflowBoundaryLowOrder` boundary condition in test_radiation_force.cpp? Why does it not work? The error message and where it comes from are copied below. 

```
Assertion !S_filled.contains_nan() failed,

	AMREX_ASSERT(!S_filled.contains_nan()); // check ghost zones (usually this is caused by
						// forgetting to fill some components when
						// using custom Dirichlet BCs, e.g., radiation
						// variables in a hydro-only problem)
```

BenWibking:
> @BenWibking Can you take a look at the `setOutflowBoundaryLowOrder` boundary condition in test_radiation_force.cpp? Why does it not work? The error message and where it comes from are copied below.
> 
> ```
> Assertion !S_filled.contains_nan() failed,
> 
> 	AMREX_ASSERT(!S_filled.contains_nan()); // check ghost zones (usually this is caused by
> 						// forgetting to fill some components when
> 						// using custom Dirichlet BCs, e.g., radiation
> 						// variables in a hydro-only problem)
> ```

Ah, I am almost certain the cause is that I wrote this function to only fill the hydro variables. You will have to fill the ghost cells for the radiation variables yourself.

However, I forgot that this was all designed for the full Euler equations, i.e. where there is also an energy equation, which is missing for an isothermal EOS. So your original approach is probably fine. The serious issues only arise when the pressure boundary condition is not specified at an outflow boundary, but here it is not an issue since the pressure is not an independent variable. Somehow I didn't connect the dots here, sorry.

chongchonghe:
It shouldn't matter anyway, because the gas is definitely outflowing at the upper boundary. In my implementation of `setOutflowBoundaryLowOrder`, I used something similar to foextrap to pressure (= rho_R * cs^2). 

So, the foextrap B.C. is okay? I'll change it back.

BenWibking (dismissed):

BenWibking:
@chongchonghe Can you take a look at the GPU test that failed?

BenWibking (approved):

PR #635, opened 2024-05-30
Fix gas mtm update bug when hydro is off

chongchonghe:
### Description
Due to a mistake at the end of `AddSourceTerms`, the gas momentum is updated to balance with radiation flux update even when hydro is turned off. This is unexpected behaviour. I fixed it and tested it with an extra velocity check in the RadMarshak test. 

### Checklist

### DIFF
--- a/src/RadMarshak/test_radiation_marshak.cpp
+++ b/src/RadMarshak/test_radiation_marshak.cpp
@@ -244,6 +244,7 @@ auto problem_main() -> int
 		std::vector<double> Tgas(nx);
 		std::vector<double> Erad(nx);
 		std::vector<double> Egas(nx);
+		std::vector<double> vgas(nx);
 
 		for (int i = 0; i < nx; ++i) {
 			const double x = position[i];
@@ -261,6 +262,8 @@ auto problem_main() -> int
 			const double Egas_t = (Etot_t - Ekin);
 			Egas.at(i) = Egas_t;
 			Tgas.at(i) = quokka::EOS<SuOlsonProblem>::ComputeTgasFromEint(rho, Egas_t);
+
+			vgas.at(i) = x1GasMom / rho;
 		}
 
 		// read in exact solution
@@ -315,6 +318,18 @@ auto problem_main() -> int
 		amrex::Print() << "Relative L1 error norm = " << rel_error << std::endl;
 
 		if ((rel_error > error_tol) || std::isnan(rel_error)) {
+			status = 1;
+		}
+
+		// check if velocity is strictly zero
+		const double error_v_tol = 1.0e-10;
+		double error_v = 0.0;
+		const double cs = std::sqrt(5. / 3. * T_initial); // sound speed
+		for (size_t i = 0; i < xs.size(); ++i) {
+			error_v += std::abs(vgas[i])/ cs;
+		}
+		amrex::Print() << "Sum of abs(v) / cs = " << error_v << std::endl;
+		if ((error_v > error_v_tol) || std::isnan(error_v)) {
 			status = 1;
 		}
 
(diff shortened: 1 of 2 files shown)


BenWibking:
This is not a bug -- this is actually the intended behavior. Why do you want to disable the momentum update?

chongchonghe:
> This is not a bug -- this is actually the intended behavior. Why do you want to disable the momentum update?

Oh, your original code works like this? I didn't notice that. The question is, in many of the tests in the literature as well as your Quokka paper, do you allow gas momentum update in radiation-only tests?

chongchonghe:
I just thought it makes sense to prevent gas from moving in a radiation-only run, because if you allow gas momentum to update, the gas would do work on radiation via the $v F$ term, but I don't think those terms are included in the analytical calculation in most of the tests (e.g. Marshak, streaming, Su-Olson). 

@markkrumholz What's your opinion?

BenWibking:
> I just thought it makes sense to prevent gas from moving in a radiation-only run, because if you allow gas momentum to update, the gas would do work on radiation via the vF term, but I don't think those terms are included in the analytical calculation in most of the tests (e.g. Marshak, streaming, Su-Olson).
> 
> @markkrumholz What's your opinion?

Those terms should be disabled by setting `beta_order = 0`. Is there a case that doesn't cover?
https://github.com/quokka-astro/quokka/blob/7a597a57579780a8e0e25ee0220296de479ff4ea/src/RadSuOlson/test_radiation_SuOlson.cpp#L52

markkrumholz:
Hmmm... this one is actually tricky and subtle. If you set `beta_order = 0` that disables all velocity-dependent terms for both radiation and gas exchange. While that doesn't matter for our existing no-hydro tests, because they all use a zero-velocity background, one can certainly envision tests that posit a background with a finite velocity and where these terms are non-zero. So that argues for explicitly disabling updates to the gas momentum for no-hydro tests, as @chongchonghe suggests. 

On the other hand, I can also envision that one might want to track those terms as part of the test. That is, while none of our current tests do this, I can also envision a situation where one of the quantities you might want to test for accuracy and convergence is the amount of momentum deposited in the matter by the radiation. You obviously can't do that if you disable matter-radiation exchange. Moreover, it is somewhat inconsistent to disable momentum exchange while still enabling energy exchange. That is unphysical -- we do it for tests because that allows us to obtain analytic solutions, and the actual physics error we make is small since v/c is small in these tests, but we might not want to code to default to behaving in an unphysical way, and when this is required for a test we might want to handle that by explicitly disabling physics by setting `beta_order = 0` for that test, as @BenWibking suggests.

So I think my bottom line is that for our current suite of tests it doesn't matter whether we explicitly disable updates to the radiation momentum when hydro is off or enforce that it be off test-by-test by setting `beta_order = 0`. There are good reasons to do it both way, and both options have the potential to cause confusion if we forget how we set up the default behavior. All things being equal I think I probably lean slightly toward the option of not disabling momentum updates and instead using the `beta_order = 0` option, just because that forces us to be explicit in the various tests about the physical assumptions that have been made in deriving the analytic solutions, i.e., we have to acknowledge that the analytic solution is itself an approximation, in exactly the same way that in some problems we explicitly override the calculation of the Eddington tensor at manually set it to (1/3) I, because that's the assumption that was used to derive the analytic solution.

BenWibking:
Since we don't have any tests that depend on velocity-dependent terms, I strongly prefer we keep things as-is and use `beta_order` to handle tests that are derived under the assumption of no velocity dependence.

If someone adds tests at some point in the future with velocity-dependent terms that assume static velocities, then I think it would make sense to re-visit this PR.

markkrumholz:
OK, I'm happy to go with the option that @BenWibking suggests of using `beta_order = 0` at least for now.

chongchonghe:
I was a little confused but now it's clear. I worried that the v*F term in the source term R would be an extra source of work, but now I realized the default behaviour is that the v*F term is not included in the source term when `beta_order=0`, so there is nothing to worry about. @BenWibking @markkrumholz

PR #670, opened 2024-07-07
Calculate the delta terms via linear interpolation in the MG scheme 

chongchonghe:
### Description

In this PR, I defined the following Opacity models. In all the models (except user), the integration terms and the delta terms (Equations 11 and 12 of the multigroup paper) are treated separately. The integration terms can be calculated with piecewise powerlaw (PPL) approximation with fixed slope of -1 or free slopes, and the delta terms can either be ignored or calculated via PPL with free slopes or linear fitting. 

On how to choose which model to use: In most cases when the number of photon groups is small (< ~8), we want to use `PPL_fixed_slope`, because it's hard to reconstruct the correct shapes of the spectrum with a small number of photon bins, and in this case we'd better use a fixed slope of -1, which is the 'average' slope of any spectrum. This is shown to give the best result in the case of variable opacities. If we want to resolve the correct spectrum with lots of photon bins, we should use `PPL_free_slope` or `PPL_free_slope_with_PPL_delta_terms` or `PPL_free_slope_with_delta_terms`. 

I have shown that it's the correct powerlaw slopes that is most important for resolving the correct spectrum, not the delta terms. This is obvious in the optically thick limit, where the two delta terms in Equation 12 cancel with each other and the 'transport' term reside in - int nu chi d P = - alpha_P chi_g P_g. In the single group case, this terms is chi * P, therefore more precisely the actual transport term is - alpha_P - 1. From this we can understand why the spectrum F_nu looks like what it is in Figure 1. Wherever the spectrum slope is bigger than -1, this transport term is positive and the spectrum is higher than what is expected from a un-redshifed blackbody (the solid black curve), and vice versa. We can perhaps understand this physically. When you have a more positive spectrum (slope > -1), there are more radiation coming from the higher-nu group than there are radiation leaking into lower-nu group. 

```
enum class OpacityModel {
  user = 0,	 // User-defined opacity for each group, given as a function of density and temperature. For single-group ONLY.
  PPL_fixed_slope, // int term: PPL with fixed slope of -1; delta term: 0. In the special case of constant opacity, this is equivalent to the user case
  PPL_fixed_slope_with_delta_terms,    // int term: PPL with fixed slope of -1; delta term: linear fitting
  PPL_free_slope,			     // int term: PPL with free slope; delta term: 0
  PPL_free_slope_with_PPL_delta_terms, // int term: PPL with free slope; delta term: PPL
  PPL_free_slope_with_delta_terms	     // int term: PPL with free slopes; delta term: linear fitting
};
```

### Checklist


markkrumholz:
What is the motivation for having the default be `PPL_fixed_slope` instead of `PPL_fixed_slope_with_delta_terms`? It seems like one should try to include these terms in most cases, even if they're not dominant. Is there a reason not to?

chongchonghe:
> What is the motivation for having the default be `PPL_fixed_slope` instead of `PPL_fixed_slope_with_delta_terms`? It seems like one should try to include these terms in most cases, even if they're not dominant. Is there a reason not to?

It's speed. With `PPL_fixed_slope`, kappa_g is constant over time and space, so the amount of calculation per group is the same as the single-group, constant kappa case. With the introduction of the delta terms, the slow down is about a factor of 2. Then, it's a compromise between speed and accuracy. I don't have strong opinions on using `PPL_fixed_slope` or `PPL_fixed_slope_with_delta_terms` as the default.

markkrumholz:
> > What is the motivation for having the default be `PPL_fixed_slope` instead of `PPL_fixed_slope_with_delta_terms`? It seems like one should try to include these terms in most cases, even if they're not dominant. Is there a reason not to?
> 
> It's speed. With `PPL_fixed_slope`, kappa_g is constant over time and space, so the amount of calculation per group is the same as the single-group, constant kappa case. With the introduction of the delta terms, the slow down is about a factor of 2. Then, it's a compromise between speed and accuracy. I don't have strong opinions on using `PPL_fixed_slope` or `PPL_fixed_slope_with_delta_terms` as the default.

Not quite sure I understand this. Is the extra cost due to the need to do more iterations in the implicit solve? The cost of evaluating the terms in the matrix? Is this something that is limited to the idealized case of an opacity that doesn't depend on gas properties, and where the speed advantage will vanish for a more realistic case where opacity does depend on gas properties?

chongchonghe:
The extra cost comes from doing linear reconstruction of the spectrum to calculate the delta terms in the flux update, particularly the pressure terms which is expensive to calculate, because you have to compute the all 9 components of the pressure tensor for all groups and store them in memory before doing a linear interpolation. 

This is not related to whether or not the opacity depends on material property, because if the opacity does dependend on material property, the PPL group-mean opacity has to be calculated in each cell in both models.

BenWibking:
This was the earlier, obsolete version of the multigroup interpolation code, right? Can this PR be closed?

PR #684, opened 2024-07-25
Clean up code for single-group radiation 

chongchonghe:
### Description

This PR updates the code structure for single-group radiation so that functions like `ComputePlanckOpacity` return a `double`. This change enforces the use of the correct opacity models (i.e., `piecewise_constant_opacity` or `PPL_opacity_fixed_slope_spectrum`, or `PPL_opacity_full_spectrum`) when dealing with multiple radiation groups. The previous code provides an option to naively extend the single-group formulation in the IMEX paper to the multigroup case, i.e. replacing $\kappa_P$ and $\kappa_F$ with $\kappa_g$. However, that method does not correctly describe multigroup radiation transport and should be removed, because it misses the important delta term in the flux equation. See Equation 16 of the multigroup paper and the discussion thereafter. 

In summary, here's how you should define gas opacity in a problem:
- Single-group opacity definition:
  - Define `ComputePlanckOpacity` and optionally `ComputeFluxMeanOpacity` and `ComputeEnergyMeanOpacity`. By default, the latter two are set equal to the first. 
- Multigroup opacity definition:
  - Choose an opacity model by setting `opacity_model` in `RadSystem_Traits` to one of the following: `piecewise_constant_opacity`, `PPL_opacity_fixed_slope_spectrum`, or `PPL_opacity_full_spectrum`
  - Define `DefineOpacityExponentsAndLowerValues`. See the example in RadhydroShockMultigroup. 

Bug fix:
- fixed bug related to multigroup opacity in `ComputeCellOpticalDepth`. This function is only used when you turn on wave speed correction by hand and is not turned on by default. We can't have a regression test for this because it is controlled by `use_wavespeed_correction` inside `radiation_system.hpp` and we don't have a way to turn it on for a specific problem. I tested it on my computer and it worked as expected. 

### Checklist


BenWibking:
We could createa new version of an existing test that uses the wavespeed correction in the Riemann solver to test this. I think that would help avoid this problem.

chongchonghe:
> We could createa new version of an existing test that uses the wavespeed correction in the Riemann solver to test this. I think that would help avoid this problem.

The issue is, you can't turn it on for any specific problem. It's controlled by `use_wavespeed_correction` which is not in any traits. You have to change `radiation_system.hpp` to turn it on. I can, of course, add another parameter in the Radiation Trait, but I think it adds another fuss to the user and it's not worth it since we will almost never use it.

BenWibking:
> > We could createa new version of an existing test that uses the wavespeed correction in the Riemann solver to test this. I think that would help avoid this problem.
> 
> The issue is, you can't turn it on for any specific problem. It's controlled by `use_wavespeed_correction` which is not in any traits. You have to change `radiation_system.hpp` to turn it on. I can, of course, add another parameter in the Radiation Trait, but I think it adds another fuss to the user and it's not worth it since we will almost never use it.

We could do this: https://github.com/AMReX-Codes/amrex/pull/2954

chongchonghe:
> > > We could createa new version of an existing test that uses the wavespeed correction in the Riemann solver to test this. I think that would help avoid this problem.
> > 
> > 
> > The issue is, you can't turn it on for any specific problem. It's controlled by `use_wavespeed_correction` which is not in any traits. You have to change `radiation_system.hpp` to turn it on. I can, of course, add another parameter in the Radiation Trait, but I think it adds another fuss to the user and it's not worth it since we will almost never use it.
> 
> We could do this: [AMReX-Codes/amrex#2954](https://github.com/AMReX-Codes/amrex/pull/2954)

That feature is interesting. Have you tried it? Is there any example in Quokka? I'm not sure how to set `runtime_option` in problem.cpp.

BenWibking:
> That feature is interesting. Have you tried it? Is there any example in Quokka? I'm not sure how to set `runtime_option` in problem.cpp.

No, it's not currently used in Quokka.

In the code below, you can set `runtime_option` just like any other variable in the code. You could initialize its value by querying the `ParmParse` database, for instance, so that it's set from a command-line parameter.

```
    int runtime_option = ...; // set this based on the value of any variable, read it from an input file, read it from the command-line, do whatever you want
    enum All_options : int { A0, A1, A2, A3};
    // Four ParallelFors will be generated.
    ParallelFor(TypeList<CompileTimeOptions<A0,A1,A2,A3>>{}, {runtime_option},
                box, [=] AMREX_GPU_DEVICE (int i, int j, int k, auto control)
    {
        ...
        if constexpr (control.value == A0) {
            ...
        } else if constexpr (control.value == A1) {
            ...
        } else if constexpr (control.value == A2) {
            ...
        else {
            ...
        }
        ...
    });
```

BenWibking:
Here's an example in WarpX:
https://github.com/ECP-WarpX/WarpX/blob/675369d9797aa2bcfda8b64c8e500fff6a2b9f04/Source/Particles/RigidInjectedParticleContainer.cpp#L424

chongchonghe:
@BenWibking Can you review this PR?

src/RadMarshakAsymptoticCorr/test_radiation_marshak_asymptotic_corr.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+	sim.radiationReconstructionOrder_ = 3; // PPM
+	sim.stopTime_ = max_time;
+	sim.initDt_ = initial_dt;
+	sim.maxDt_ = max_dt;
+	sim.radiationCflNumber_ = CFL_number;
+	sim.maxTimesteps_ = max_timesteps;
+	sim.plotfileInterval_ = -1;
+
+	sim.use_wavespeed_correction_ = true;
```
BenWibking: You could read this from the input file, e.g.:
```
amrex::ParmParse pp("marshak");
pp.query("use_wavespeed_correction", sim.use_wavespeed_correction_);
```
Then you could re-use the same problem file, and just create two tests (using different input files):
```
marshak.use_wavespeed_correction = 0
```
and
```
marshak.use_wavespeed_correction = 1
```

src/radiation_system.hpp
```diff
@@ -630,11 +637,21 @@ AMREX_GPU_DEVICE auto RadSystem<problem_t>::ComputeCellOpticalDepth(const quokka
 	} else if constexpr (DIR == FluxDir::X3) {
 		dl = dx[2];
 	}
-	quokka::valarray<double, nGroups_> const tau_L = dl * rho_L * RadSystem<problem_t>::ComputeFluxMeanOpacity(rho_L, Tgas_L);
-	quokka::valarray<double, nGroups_> const tau_R = dl * rho_R * RadSystem<problem_t>::ComputeFluxMeanOpacity(rho_R, Tgas_R);
 
-	return (tau_L * tau_R * 2.) / (tau_L + tau_R); // harmonic mean
-						       // return 0.5*(tau_L + tau_R); // arithmetic mean
+	quokka::valarray<double, nGroups_> optical_depths{};
+	if constexpr (nGroups_ == 1) {
```
BenWibking: It looks like this will only work for the gray case, and will not actually give the right answer for the multi-group case. Is that correct?

src/radiation_system.hpp
```diff
@@ -203,9 +210,9 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
 	AMREX_GPU_DEVICE static auto ComputeMassScalars(ArrayType const &arr, int i, int j, int k) -> amrex::GpuArray<Real, nmscalars_>;
 
 	AMREX_GPU_HOST_DEVICE static auto ComputeEddingtonFactor(double f) -> double;
-	AMREX_GPU_HOST_DEVICE static auto ComputePlanckOpacity(double rho, double Tgas) -> quokka::valarray<double, nGroups_>;
-	AMREX_GPU_HOST_DEVICE static auto ComputeFluxMeanOpacity(double rho, double Tgas) -> quokka::valarray<double, nGroups_>;
-	AMREX_GPU_HOST_DEVICE static auto ComputeEnergyMeanOpacity(double rho, double Tgas) -> quokka::valarray<double, nGroups_>;
+	AMREX_GPU_HOST_DEVICE static auto ComputePlanckOpacity(double rho, double Tgas) -> Real;
+	AMREX_GPU_HOST_DEVICE static auto ComputeFluxMeanOpacity(double rho, double Tgas) -> Real;
+	AMREX_GPU_HOST_DEVICE static auto ComputeEnergyMeanOpacity(double rho, double Tgas) -> Real;
```
BenWibking: These functions only work for the gray case, right? If so, I think there should be a comment in the source clearly specifying this.

src/RadhydroPulseMGconst/test_radhydro_pulse_MG_const_kappa.cpp
```diff
@@ -113,17 +113,9 @@ template <> struct RadSystem_Traits<SGProblem> {
...
-template <>
-AMREX_GPU_HOST_DEVICE auto RadSystem<SGProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> quokka::valarray<double, nGroups_>
-{
-	quokka::valarray<double, nGroups_> kappaPVec{};
-	for (int g = 0; g < nGroups_; ++g) {
-		kappaPVec[g] = kappa0;
-	}
-	return kappaPVec;
-}
+template <> AMREX_GPU_HOST_DEVICE auto RadSystem<SGProblem>::ComputePlanckOpacity(const double /*rho*/, const double /*Tgas*/) -> amrex::Real { return kappa0; }
```
BenWibking: Why is this function used for a multi-group test? I don't understand what's going on here.

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
```diff
@@ -203,21 +203,16 @@ RadSystem<MGProblem>::DefineOpacityExponentsAndLowerValues(amrex::GpuArray<doubl
...
 }
 
-template <> AMREX_GPU_HOST_DEVICE auto RadSystem<ExactProblem>::ComputePlanckOpacity(const double rho, const double Tgas) -> quokka::valarray<double, nGroups_>
+template <> AMREX_GPU_HOST_DEVICE auto RadSystem<ExactProblem>::ComputePlanckOpacity(const double rho, const double Tgas) -> amrex::Real
 {
 	const double sigma = scaleup * 3063.96 * std::pow(Tgas / T0, -3.5);
-	quokka::valarray<double, nGroups_> kappaPVec{};
-	kappaPVec.fillin(sigma / rho);
-	return kappaPVec;
+	return sigma / rho;
```
BenWibking: I also don't understand what's going on here. I thought this function was only for gray problems.

BenWibking (changes requested):
I am confused. Can you explain more explicitly what opacity functions should be used by gray problems and what opacity functions should be used (and how) by multi-group problems?

src/RadMarshakAsymptoticCorr/test_radiation_marshak_asymptotic_corr.cpp
chongchonghe: Done.

src/RadhydroPulseMGconst/test_radhydro_pulse_MG_const_kappa.cpp
chongchonghe: Because SGProblem is a single-group problem. This test has two problems: a single-group problem and a multigroup problem with constant opacity. This is to confirm that MG RHD gives the same result as the single-group RHD.

src/RadhydroPulseMGint/test_radhydro_pulse_MG_int.cpp
chongchonghe: Same here. There are two problems in this test. ExactProblem is a single-group test with analytical planck-mean and Rossland-mean opacity and is used and the reference result. MGProblem is the multigroup problem.

src/radiation_system.hpp
chongchonghe: You are right. I added comments to make this clear.

src/radiation_system.hpp
chongchonghe: That's not the case. It will work for both single-group and multigroup. That's what the if constexpr (nGroups_ == 1) is doing. Note that DefineOpacityExponentsAndLowerValues and ComputeBinCenterOpacity returns an valarray, so optical_depths is a valarray of length nGroups_.

chongchonghe: Although, I'm kind of cutting corner here by using ComputeBinCenterOpacity regardless of the OpacityModel the user chooses to use. But I think this won't matter since this wavespeed correction is only an approximation method.

chongchonghe:
> I am confused. Can you explain more explicitly what opacity functions should be used by gray problems and what opacity functions should be used (and how) by multi-group problems?

I updated the PR comments to better explain this.

src/problems/RadMarshakAsymptoticCorr/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_executable(test_radiation_marshak_asymptotic_corr ../RadMarshakAsymptotic/test_radiation_marshak_asymptotic.cpp ../../util/fextract.cpp ../../math/interpolate.cpp ${QuokkaObjSources})
+
+if(AMReX_GPU_BACKEND MATCHES "CUDA")
+    setup_target_for_cuda_compilation(test_radiation_marshak_asymptotic_corr)
+endif(AMReX_GPU_BACKEND MATCHES "CUDA")
+
+add_test(NAME MarshakWaveAsymptoticDiffusion COMMAND test_radiation_marshak_asymptotic_corr MarshakAsymptoticCorr.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
```
BenWibking: Is there a reason to build this twice? You could just call `add_test` with the `test_radiation_marshak_asymptotic` executable.

BenWibking (changes requested):

src/problems/RadMarshakAsymptoticCorr/CMakeLists.txt
BenWibking: You could even have both `add_test` statements in the same `CMakeLists.txt` file.

chongchonghe: Good point. Done!

BenWibking (approved):

PR #687, opened 2024-07-26
Add VSCode/Codespaces Dev Container settings

BenWibking:
### Description
This adds `.devcontainer/devcontainer.json` so that VSCode will automatically download and start a Docker container and load a development environment for Quokka inside this container.

For more information, see: https://code.visualstudio.com/docs/devcontainers/containers#_quick-start-open-a-git-repository-or-github-pr-in-an-isolated-container-volume.

This will also enable GitHub Codespaces: https://docs.github.com/en/codespaces/overview#what-is-a-codespace.

### Related issues
N/A

### Checklist


chongchonghe:
How do you test GitHub Codespaces? This PR looks harmless so I'll approve it. Before that I'll test Docker in VScode.

BenWibking:
I don't think it will work until it's merged to the development branch, since the Docker container is only tagged appropriately in that case.

However, you can modify this line so that it uses the `Dockerfile` directly to build a new container image instead of downloading the image: 
https://github.com/quokka-astro/quokka/blob/4c346f4122a8c7c88a48d785951201b0a4fad944/.devcontainer/devcontainer.json#L4
and replace it with (e.g., https://containers.dev/guide/dockerfile):
```
    "build": {
        // Path is relative to the devcontainer.json file. We want to get the Dockerfile in the root of the repo.
        "dockerfile": "../Dockerfile"
    }
```
Then it should work locally for testing.

BenWibking:
You can test the Codespace by clicking on `<> Code` and then clicking the  `Open in Codespace` button:
Screenshot 2024-07-26 at 10 42 52 PM

BenWibking:
The Codespace will currently fail for the same reason: the `ghcr.io/quokka-astro/quokka:development` image doesn't exist yet.

I can change this PR to use the Dockerfile for testing purposes. (The point of using the cloud image is just to avoid the overhead of having to re-build the image locally.)

BenWibking:
Ok, the Codespace now works for me. However, it doesn't fetch the submodules by default, which is confusing. So users still need to do `git submodule update --init` before things will work.

BenWibking:
Ok, I've simplified the `devcontainer.json` configuration.

The Codespace sort of works now, but it tries to configure CMake before it has checked out all of the submodules, which then causes it to suggest to the user to start the CMake debugger, which is extremely confusing and just wrong.

Minor annoyances:
* ~~It suggests installing the Docker extension, which does not make sense and is confusing to the user~~
* The virtual machine it runs on by default is too underpowered to compile Quokka. It takes a *long* time to compile. (This can be partially fixed by setting `hostRequirements` in `devcontainer.json`.)

@chongchonghe if you have any ideas of how to fix any of the above problems, let me know. They are all GitHub or VSCode bugs, in my opinion, but it would be good to work around them if we can.

chongchonghe:
> Ok, I've simplified the `devcontainer.json` configuration.
> 
> The Codespace sort of works now, but it tries to configure CMake before it has checked out all of the submodules, which then causes it to suggest to the user to start the CMake debugger, which is extremely confusing and just wrong.
> 
> Minor annoyances:
> 
> * It suggests installing the Docker extension, which does not make sense and is confusing to the user
> * The virtual machine it runs on by default is too underpowered to compile Quokka. It takes a _long_ time to compile.
> 
> @chongchonghe if you have any ideas of how to fix any of the above problems, let me know. They are all GitHub or VSCode bugs, in my opinion, but it would be good to work around them if we can.

OK, I'll take a look.

chongchonghe:
> Ok, I've simplified the `devcontainer.json` configuration.
> 
> The Codespace sort of works now, but it tries to configure CMake before it has checked out all of the submodules, which then causes it to suggest to the user to start the CMake debugger, which is extremely confusing and just wrong.
> 
> Minor annoyances:
> 
> * It suggests installing the Docker extension, which does not make sense and is confusing to the user
> * The virtual machine it runs on by default is too underpowered to compile Quokka. It takes a _long_ time to compile. (This can be partially fixed by setting `hostRequirements` in `devcontainer.json`.)
> 
> @chongchonghe if you have any ideas of how to fix any of the above problems, let me know. They are all GitHub or VSCode bugs, in my opinion, but it would be good to work around them if we can.

OK, I tested Docker in vscode and it worked great.

Problem 1 is easy. Simply add Docker to extensions devcontainer.json. It's useful to install it anyway, since if you are going to use Docker, you probably have installed Docker extension in your VScode anyway. Speaking of extensions, I prefer to install `ms-vscode.cpptools` and turn it on, and set `"C_Cpp.intelliSenseEngine": "disabled"` to avoid conflict with clangd. I use MS cpptools for debugging because it's very nice. But of course the user can always install the extension themself. 

Regarding Problem 2, I don't have this issue on my Mac Studio, which is what I use most of the time, so it's not a concern to me. On my Mac Studio, compilation in the container is only slightly slower than local build, so it's totally fine. I tried it in github codespace, and it's indeed noticeably slower. It uses AMD EPYC 7763 64-Core Processor, and you can request at most 16 cores, so it's understandable that single-core performance is very limited. I wouldn't worry about it since using github codespace is a very rare occasion(at least for me; I only use it when I don't have access to my laptop and I have to use an iPad).

chongchonghe:
@BenWibking Do you really want Docker to be default when you first open Quokka in vscode? Not everyone is using Docker. Is there a way to disable Docker by default and manually enable it?

BenWibking:
> @BenWibking Do you really want Docker to be default when you first open Quokka in vscode? Not everyone is using Docker. Is there a way to disable Docker by default and manually enable it?

I have received so many queries about how to install dependencies needed for Quokka (and queries about how to fix Python environment issues) that I would prefer if the Docker container were the default. Users who don't want that almost certainly know how to disable it.

BenWibking:
@chongchonghe I've added some missing Sphinx dependencies.

Everything works now in Codespaces, except for PDF preview (seems to be impossible). If it works for you in VSCode, can you approve?

BenWibking:
@chongchonghe When I run it locally in VSCode, I get this error:
```
ubuntu@dfa95a222bb4:/workspaces/quokka$ git submodule update --init
fatal: detected dubious ownership in repository at '/workspaces/quokka'
To add an exception for this directory, call:

        git config --global --add safe.directory /workspaces/quokka
```
Any ideas?

Update: it's this issue: https://github.com/microsoft/vscode-remote-release/issues/7923.

See also: https://www.kenmuse.com/blog/avoiding-dubious-ownership-in-dev-containers/

BenWibking:
Ok, if I create a new dev container from scratch, everything works fine locally. I think this is ready to merge.

chongchonghe (approved):

chongchonghe:
@BenWibking  Is it possible to configure a nvcc build container on a macOS host? If you have tried and made it work, can you share the Dockerfile?

BenWibking:
> @BenWibking Is it possible to configure a nvcc build container on a macOS host? If you have tried and made it work, can you share the Dockerfile?

I haven't tried this.

BenWibking:
@chongchonghe How do you get the built-in debugger in VSCode to work? Did you configure `launch.json` manually (https://code.visualstudio.com/docs/cpp/launch-json-reference)?

PR #690, opened 2024-07-30
Update citations

chongchonghe:
### Description

Update citation for the newly published multigroup RHD paper on arxiv. 

### Checklist


BenWibking:
I think it's a bit repetitive to have the citation repeated three times:
image

Maybe you could create a separate bibliography block for each bullet point?

chongchonghe:
> I think it's a bit repetitive to have the citation repeated three times: image
> 
> Maybe you could create a separate bibliography block for each bullet point?

I tried, but you'll get two bibliographies with the same citation key, since they have the same authors and year. I don't know a way to get away with it, so that's the best way I found that is concise and also displays the full article titles in the same page. 

If two times instead of three times is better, I can replace `cite:ts:` with `cite:`, which would remove the first occurrence of the names.

BenWibking (approved):
Ok, I think this is better.

PR #691, opened 2024-07-30
add equations to docs

BenWibking:
### Description
This adds the complete set of equations that are solved by Quokka to the documentation.

### Related issues
N/A

### Checklist

### DIFF
--- a/docs/equations.rst
+++ b/docs/equations.rst
@@ -3,14 +3,96 @@
 Equations
 ==========================
 
-*This page is still under construction.*
+Fluids and radiation
+-----------------
+
+Assuming the speed of light is not reduced (:math:`\hat{c} = c`),
+Quokka solves the system of conservation laws:
 
 .. math::
 
-    \frac{\partial E}{\partial t} + \nabla \cdot \vec{F} = -G^0 \\
-    \frac{1}{c^2} \frac{\partial F}{\partial t} + \nabla \cdot \vec{P} = -\vec{G}
+    \frac{\partial \vec{U}}{\partial t}+\nabla \cdot \vec{F}(\vec{U}) = \vec{S}(\vec{U}),
+
 ..
 
-where :math:`E` is the radiation energy density, :math:`F` is the radiation flux,
-:math:`P` is the radiation pressure tensor, and :math:`G` is the radiation four-force.
+.. math::
 
+    \vec{U} =\left[
+    \begin{array}{c}
+      \rho \\
+      \rho \vec{v} \\
+      E_{\rm gas} \\
+      \rho X_n \\
+      E_g \\
+      \vec{F}_g
+    \end{array}\right], \;
+    \vec{F}(U) = \left[
+    \begin{array}{c}
+      \rho \vec{v} \\
+      \rho \vec{v} \otimes \vec{v}+p \\
+      (E_{\rm gas} + p) \vec{v} \\
+      \rho X_n \vec{v} \\
+      \vec{F}_g \\
+      c^2 \boldsymbol{P}_g
+    \end{array}\right], \;
+    \vec{S}(U)=\left[
+    \begin{array}{c}
+      0 \\
+      \sum_g \vec{G}_g + \rho \vec{g} \\
+      c \sum_g G^0_{g} + \rho \vec{v} \cdot \vec{g} + \mathcal{H} - \mathcal{C} \\
+      \rho \dot{X}_n \\
+      - c G^0_{g} \\
+      - c^2 \vec{G}_g
+    \end{array}\right],
+
+..
+
+along with the non-conservative auxiliary internal energy equation:
+
+.. math::
+
+    \frac{\partial \rho e_{\text{aux}}}{\partial t} =
+    - \nabla \cdot (\rho e_{\text{aux}} \vec{v}) + p \nabla \cdot \vec{v},
+
+..
+
+and the gravitational Poisson equation:
+
+.. math::
+
+    \nabla^2 \phi = -4 \pi G \left( \rho + \sum_i \rho_i \right), \\
+    \vec{g} = -\nabla \phi,
+
+..
+
+where
+
+* :math:`\rho` is the gas density,
+* :math:`\vec{v}` is the gas velocity,
+* :math:`E_{\text{gas}}` is the total gas energy,
+* :math:`\rho e_{\text{aux}}` is the auxiliary gas internal energy,
+* :math:`X_n` is the fractional concentration of species :math:`n`,
+* :math:`\dot{X}_n` is the chemical reaction term for species :math:`n`,
+* :math:`\mathcal{H}` is the optically-thin volumetric heating term,
+* :math:`\mathcal{C}` is the optically-thin volumetric cooling term,
+* :math:`p(\rho, e)` is the gas pressure derived from a general convex equation of state,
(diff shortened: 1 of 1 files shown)


BenWibking:
The equations render like this:

Screenshot 2024-07-30 232530

BenWibking:
@chongchonghe @markkrumholz I am not sure how to write the radiation energy update for the auxiliary internal energy. It is not $\sum_g G^0_g$, as written above, since that includes work contributions... any ideas?

markkrumholz:
In practice we only use the auxiliary internal energy during the hydro update stage, so in some sense what you have written is technically correct, with no radiation terms at all on the RHS. The way radiation terms get incorporated is when the internal energy gets recomputed after the radiation stage. But I think it might actually be more sensible if we are trying to summarize the equations to omit the auxiliary internal energy equation, and instead think of that as simply an alternative numerical method for solving the actual energy equation. That is, the fundamental equations we are solving are the conservation equations, but we use two possible numerical methods to do so, one of which involves solving the auxiliary internal energy equation during the hydro stage only.

BenWibking:
> In practice we only use the auxiliary internal energy during the hydro update stage, so in some sense what you have written is technically correct, with no radiation terms at all on the RHS. The way radiation terms get incorporated is when the internal energy gets recomputed after the radiation stage. But I think it might actually be more sensible if we are trying to summarize the equations to omit the auxiliary internal energy equation, and instead think of that as simply an alternative numerical method for solving the actual energy equation. That is, the fundamental equations we are solving are the conservation equations, but we use two possible numerical methods to do so, one of which involves solving the auxiliary internal energy equation during the hydro stage only.

That is not quite true. I implemented the cooling routines ($\mathcal{H-C}$) such that they also update the auxiliary internal energy.

To be consistent, I had originally designed the radiation update to do likewise, but I lost track of what the current version of the code does in this respect.

(This is also what Castro does: https://amrex-astro.github.io/Castro/docs/Hydrodynamics.html#internal-energy-and-temperature)

I've updated the equations above to show this.

markkrumholz:
OK, not sure what @chongchonghe is doing here either. I think the method you implemented originally does not work with our new relativistically-correct formulation, because in this formulation you can't neatly divide internal and kinetic energy. They get mixed at order v/c. I do not recall what @chongchonghe wound up doing about that.

BenWibking:
> OK, not sure what @chongchonghe is doing here either. I think the method you implemented originally does not work with our new relativistically-correct formulation, because in this formulation you can't neatly divide internal and kinetic energy. They get mixed at order v/c. I do not recall what @chongchonghe wound up doing about that.

I think we can just add $\Delta S_{\text{rad}} = \int_{t_0}^{t_0 + \Delta t} S_{\text{rad}} \ dt = [ E_{gas}(t_0 + \Delta t) - KE(t_0 + \Delta t) ] - [ E_{gas}(t_0) - KE(t_0) ]$ to the auxiliary internal energy, right?

BenWibking:
Maybe we should add a Mach 50 radiative shock test problem to ensure we are doing this right.

markkrumholz:
I agree that your expression has to be correct by construction. But I guess my more basic issue is how the internal energy update interacts with the radiation update within a timestep. My thought was that you only used the internal energy during the hydro step, and that you decided whether to use the internal energy computed separately or just use the total energy, all prior to starting the radiation substep. And then you recompute the internal energy from the total energy and the start of the next hydro substep -- in which case modifications you make to the internal energy during the radiation substep have no effect. So I guess I'm wrong about the ordering somewhere. So where does internal energy that is updated during the radiation substep actually get used?

BenWibking:
> I agree that your expression has to be correct by construction. But I guess my more basic issue is how the internal energy update interacts with the radiation update within a timestep. My thought was that you only used the internal energy during the hydro step, and that you decided whether to use the internal energy computed separately or just use the total energy, all prior to starting the radiation substep. And then you recompute the internal energy from the total energy and the start of the next hydro substep -- in which case modifications you make to the internal energy during the radiation substep have no effect. So I guess I'm wrong about the ordering somewhere. So where does internal energy that is updated during the radiation substep actually get used?

The auxiliary internal energy and total energy are only synchronized after each hydro RK stage. So there is no sync at the start of the next hydro step. As it is done now, if there is a perturbation to the internal energy that is not reflected in the auxiliary internal energy, both the advection of auxiliary internal energy (due to the $F_{\rho X_n}$ flux term) in the RK stage 1 step and the next sync (at the end of the same hydro RK stage 1 update) will use inconsistent values of the auxiliary internal energy and total energy in determining whether and how to sync up.

BenWibking:
> And then you recompute the internal energy from the total energy and the start of the next hydro substep

The auxiliary internal energy is never unconditionally replaced by the total energy. It is always conditionally sync'd:
https://github.com/quokka-astro/quokka/blob/725c343e537527c75aafdf15d94ec3ada2497968/src/hydro/hydro_system.hpp#L841

Therefore, there is always the possibility that a perturbation to the total energy will cause the sync direction to change. So the auxiliary internal energy must always be kept consistent when adding energy source terms.

markkrumholz:
Ok, so I think we maybe need to change that. Once you go to a relativistically correct formulation of RHD, you can no longer sensibly separate the terms that apply to internal versus total energy. So I think we may need to sync at the end of the hydro and then just ignore the internal energy during the radiation substep. I don’t think there is a sensible way to write an internal energy equation for radiation that is also relativistically correct.

chongchonghe:
The precise update of the internal energy in the radiation-matter coupling step is this:

``` math
\begin{gather}
\Delta (\rho \vec{v}) = -\sum_g \Delta \vec{F}_g \\
S_{\rm rad} = \sum_g c G_g^0 = - 4 \pi \chi_{0 P} B + c \sum_g \chi_{0E,g} E_g - c^{-1} \sum_g \chi_{0F,g} v^i F_g^i \\
\Delta (\rho e) = S_{\rm rad} \Delta t - \frac{1}{2} \Delta \left(\rho v^2 \right) 
\end{gather}
```

Not sure how you would write this into one single equation.

For infinitesimal time step, those work term cancels, so you get the simple formula

``` math
\Delta (\rho e) = \left( - 4 \pi \chi_{0 P} B + c \sum_g \chi_{0E,g} E_g \right) \Delta t
```

so, in principle, you can write it as a differential equation, but that would hide how exact we solve those equations.

chongchonghe:
@BenWibking Do you want to update the docs to reflect that? All you wrote in the current PR is correct, and the only thing you need to add is exactly how $S_{\rm rad}$ is defined. The simple way to say it is $S_{\rm rad} = - 4 \pi \chi_{0 P} B + c \sum_g \chi_{0E,g} E_g$, although we solved the whole matter-radiation coupling differently: we include the work ($v/c$) term in the source when updating the gas total energy, then subtract the update of the kinetic energy from the total energy, and what remains is the update in internal energy.

BenWibking:
I would prefer to leave it as is, but we could specify that the time derivative $S_{\text{rad}}$ is never computed, only $\Delta S_{\text{rad}} = \int \sum c G^0_g \ dt - \frac{1}{2} \Delta \left(\rho v^2 \right)$ is computed.

An analogous integral over the timestep is also used in practice for, e.g., the gravitational work term.

BenWibking:
Ok, I am happy with this now. @chongchonghe does this look ok?

chongchonghe (approved):

PR #712, opened 2024-08-10
Metal advection problem

aditivijayan:
### Description
There was an issue in the test_sne.cpp, specifically in the addStrangSplitSources function.

Firstly, the momentum should have been referred as "HydroSystem<NewProblem>::x1Momentum_index" rather than "x1mom". 

Secondly, the state momentum variables need to be updated in two steps. Writing "state = state + rho*GradPhi" does not work. I am unsure why. 

The problem compiles and runs on gadi. 

### Related issues
none

### Checklist


src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+#if (AMREX_SPACEDIM >= 3)
+			posvec[2] = prob_lo[2] + (k + 0.5) * dx[2];
+#endif
+
+			GradPhi = HydroSystem<NewProblem>::GetGradFixedPotential(posvec);
+
+			x1mom_new = state(i, j, k, HydroSystem<NewProblem>::x1Momentum_index) + dt * (-rho * GradPhi[0]);
+			x2mom_new = state(i, j, k, HydroSystem<NewProblem>::x2Momentum_index) + dt * (-rho * GradPhi[1]);
+			x3mom_new = state(i, j, k, HydroSystem<NewProblem>::x3Momentum_index) + dt * (-rho * GradPhi[2]);
```
BenWibking: `state(i, j, k, HydroSystem<NewProblem>::x1Momentum_index)` can be replaced with `x1mom`,
`state(i, j, k, HydroSystem<NewProblem>::x2Momentum_index)` can be replaced with `x2mom`, and
`state(i, j, k, HydroSystem<NewProblem>::x3Momentum_index)` can be replaced with `x3mom`.

The whole string should be replaced here, not just the component index.

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+		// Calculate Gas Disk Potential
+
+		double Phigas;
+		// Interpolate to find the accurate g-value from array-- because linterp doesn't work on Setonix
+		// TODO - AV to find out why linterp doesn't work
+		size_t ii = 0;
+		double x_interp = std::abs(z);
+		amrex::GpuArray<amrex::Real, 100> xx = z_data;
+		amrex::GpuArray<amrex::Real, 100> yy = logphi_data;
```
BenWibking: This causes a copy into a local variable, which is probably not a good idea. You can make it a reference, or you can use `z_data` and `logphi_data` directly.

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,562 @@
...
+	grad_potential[0] = 0.0;
+	grad_potential[1] = 0.0;
+
+	double z = posvec[2];
+
+	// Interpolate to find the accurate g-value from array-- because linterp doesn't work on Setonix
+	size_t i = 0;
+	double x_interp = std::abs(z);
+	amrex::GpuArray<amrex::Real, 100> xx = z_data;
+	amrex::GpuArray<amrex::Real, 100> yy = logg_data;
```
BenWibking: This also makes a copy of a large array, which is not a good idea. You can change this to a reference, or you use `z_data` and `logg_data` directly.

src/problems/MetalAdvectionProblem/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+if (AMReX_SPACEDIM EQUAL 3)
+      add_executable(test_sne test_sne.cpp ${QuokkaObjSources})
+    if(AMReX_GPU_BACKEND MATCHES "CUDA")
+        setup_target_for_cuda_compilation(test_sne)
+    endif()
+
```
BenWibking: Can you add a CTest `add_test(...)` command here? You can look at the CMakeLists.txt in the other problems for examples.

src/problems/MetalAdvectionProblem/data_sets.dat
BenWibking, line 1: Since this file isn't used, I'm not sure if it's useful to include it here.

BenWibking (changes requested):

BenWibking:
In `state = state + rho*GradPhi`, I think the term should have a negative sign, since $g = -\nabla \phi$.
So I think it should be: `state += dt * (-rho*GradPhi)`.

But I think it makes sense to do it the way you've done it in the code, since you need the new momentum value in any case in order to compute the new total energy.

src/problems/MetalAdvectionProblem/test_sne.cpp
aditivijayan: I am a bit confused here. Because previously, it was - 
`x1mom_new = state(i, j, k, HydroSystem<NewProblem>::x1Momentum_index) + dt * (-rho * GradPhi[0]);`

and then - 
`state(i, j, k, x1mom) = x1mom_new`

-but this was the bit that was causing problem. For some reason, the density was turning negative between these sets of code. So, I switched it to `state(i, j, k, HydroSystem<NewProblem>::x1Momentum_index)` and it worked fine.

src/problems/MetalAdvectionProblem/data_sets.dat
aditivijayan: In case someone wants to reproduce non-Solar neighbourhood environments. I would prefer to have this file until I can fix the issue of having datasets being read from *h5 files on Setonix. I can add a header explaining the file and the data sets if you prefer.

src/problems/MetalAdvectionProblem/test_sne.cpp
BenWibking: This works:
```
const double x1mom_new = x1mom + dt * (-rho * GradPhi[0]);
state(i, j, k, HydroSystem<NewProblem>::x1Momentum_index) = x1mom_new;
```

src/problems/MetalAdvectionProblem/data_sets.dat
BenWibking: You could read it in from a ParmParse variable using `ParmParse::queryarr`.

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,558 @@
...
+    4.44181452e+21, 4.56350806e+21, 4.68520161e+21, 4.80689516e+21, 4.92858871e+21, 5.05028226e+21, 5.17197581e+21, 5.29366935e+21, 5.41536290e+21,
+    5.53705645e+21, 5.65875000e+21, 5.78044355e+21, 5.90213710e+21, 6.02383065e+21, 6.14552419e+21, 6.26721774e+21, 6.38891129e+21, 6.51060484e+21,
+    6.63229839e+21, 6.75399194e+21, 6.87568548e+21, 6.99737903e+21, 7.11907258e+21, 7.24076613e+21, 7.36245968e+21, 7.48415323e+21, 7.60584677e+21,
+    7.72754032e+21, 7.84923387e+21, 7.97092742e+21, 8.09262097e+21, 8.21431452e+21, 8.33600806e+21, 8.45770161e+21, 8.57939516e+21, 8.70108871e+21,
+    8.82278226e+21, 8.94447581e+21, 9.06616935e+21, 9.18786290e+21, 9.30955645e+21, 9.43125000e+21, 9.55294355e+21, 9.67463710e+21, 9.79633065e+21,
+    9.91802419e+21, 1.00397177e+22, 1.01614113e+22, 1.02831048e+22, 1.04047984e+22, 1.05264919e+22, 1.06481855e+22, 1.07698790e+22, 1.08915726e+22,
+    1.10132661e+22, 1.11349597e+22, 1.12566532e+22, 1.13783468e+22, 1.15000403e+22, 1.16217339e+22, 1.17434274e+22, 1.18651210e+22, 1.19868145e+22,
+    1.21085081e+22};
```
BenWibking: It might make sense to make each of these a variable that is read in using `amrex::ParmParse`. You can do it like this inside `problem_main()`:
```
amrex::ParmParse pp("metal_problem");
std::vector<double> z_vec, logphi_vec, logg_vec;
pp.queryarr("z_data", z_vec);
pp.queryarr("logphi_data", logphi_vec);
pp.queryarr("logg_data", logg_vec);
```

BenWibking: Then in the `metal_problem_regression.in` file, you have three lines like this:
```
metal_problem.z_data = 1.000e2 1.000e2 1.000e2
metal_problem.logphi_data = 1.000e2 1.000e2 1.000e2
metal_problem.logg_data =1.000e2 1.000e2 1.000e2
```
where instead of the values I've put above, you just copy and paste the real values from the arrays.

BenWibking: This can go in another PR. I'll resolve this for now.

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,542 @@
...
+	const double x1Mom_edge = consVar(i, j, kedge, HydroSystem<NewProblem>::x1Momentum_index);
+	const double x2Mom_edge = consVar(i, j, kedge, HydroSystem<NewProblem>::x2Momentum_index);
+	double x3Mom_edge = consVar(i, j, kedge, HydroSystem<NewProblem>::x3Momentum_index);
+	const double etot_edge = consVar(i, j, kedge, HydroSystem<NewProblem>::energy_index);
+	const double eint_edge = consVar(i, j, kedge, HydroSystem<NewProblem>::internalEnergy_index);
+	const double pscalar_edge = consVar(i, j, kedge, HydroSystem<NewProblem>::scalar0_index);
+
+	if ((x3Mom_edge * normal) < 0) { // gas is inflowing
+		x3Mom_edge = -1. * consVar(i, j, kedge, HydroSystem<NewProblem>::x3Momentum_index);
+	}
```
BenWibking, line 469: This implementation of the diode boundary condition is not correct.

When the gas is inflowing, you have to reflect all of the cells about the boundary face. For example:
* the `khi+1` ghost cell needs to be based on the `khi` cell, whereas
* the `khi+2` ghost cell needs to be based on the `khi-1` cell,
* the `khi+3` ghost cell needs to be based on the `khi-2` cell, and
* the `khi+4` ghost cell needs to be based on the `khi-3` cell.

BenWibking (changes requested):

BenWibking:
For future reference, I've successfully compiled and run commit `9bb4588a254745cb52a6c4484ad20deb482895b9` on Frontier with the `metal_problem_regression.in` input on 1 node (this is only a 32 x 32 x 256 problem).

Modules:
```
wibking@login03:/lustre/orion/ast146/scratch/wibking/quokka> module list

Currently Loaded Modules:
  1) Core/24.07           5) craype-network-ofi        9) rocm/5.7.1           13) cray-hdf5/1.12.2.9
  2) craype/2.7.31        6) PrgEnv-cray/8.5.0        10) cce/17.0.0           14) cmake/3.27.9
  3) cray-dsmml/0.2.2     7) craype-x86-trento        11) cray-libsci/23.12.5  15) emacs/29.3
  4) libfabric/1.15.2.0   8) craype-accel-amd-gfx90a  12) cray-mpich/8.1.28    16) cray-python/3.11.5
```

BenWibking:
I've also successfully run a 256 x 256 x 1024 problem on 32 nodes for 1e6 timesteps without any issues on Frontier. Same commit and modules as above.

BenWibking:
On Frontier, the `printf`s on the GPU all appear at the very end of the log file, after the simulation is finished running. To avoid this problem in the future, these `printfs` should probably be moved to the CPU.

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,542 @@
...
+
+				if (x0 < 0.5 * dx[0] && y0 < 0.5 * dx[1] && z0 < 0.5 * dx[2]) {
+					state(i, j, k, HydroSystem<NewProblem>::density_index) += rho_blast;
+					state(i, j, k, HydroSystem<NewProblem>::energy_index) += rho_eint_blast;
+					state(i, j, k, HydroSystem<NewProblem>::internalEnergy_index) += rho_eint_blast;
+					state(i, j, k, Physics_Indices<NewProblem>::pscalarFirstIndex) += scalar_blast;
+
+					printf("The total number of SN gone off=%d\n", cum_sn);
+					Real Rpds = 14. * std::pow(state(i, j, k, HydroSystem<NewProblem>::density_index) / Const_mH, -3. / 7.);
+					printf("Rpds = %.2e pc\n", Rpds);
```
BenWibking: On Frontier, `printf` from the GPU is buffered until the very end of the simulation. For future runs, it would help to move this outside of the `amrex::ParallelFor` (and therefore back onto the CPU).

BenWibking:
@aditivijayan I'm not worried about the boundary conditions anymore. Let me know if you still want to merge this. I think it would be good to merge.

aditivijayan:
Sorry! I did not see this notification. Yes, I am happy to merge this. Let me know what I need to do.

BenWibking:
> Sorry! I did not see this notification. Yes, I am happy to merge this. Let me know what I need to do.

Can you resolve the merge conflicts? Then the automated tests can run, and I can review it again.

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,540 @@
...
+		double Phist = prefac2 * (std::pow(1. + z * z / z_star / z_star, 0.5) - 1.);
+
+		// Calculate Gas Disk Potential
+
+		double Phigas;
+		// Interpolate to find the accurate g-value from array-- because linterp doesn't work on Setonix
+		// TODO - AV to find out why linterp doesn't work
+		size_t ii = 0;
+		double x_interp = std::abs(z);
+		while (ii < z_data.size() - 1 && x_interp > z_data[ii + 1]) {
```
BenWibking: Can you use the 1D interpolation routine in `math/interp.cpp` for this?

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,540 @@
...
+		state_cc(i, j, k, HydroSystem<NewProblem>::energy_index) = P / (gamma - 1.);
+		state_cc(i, j, k, Physics_Indices<NewProblem>::pscalarFirstIndex) = 1.e-5 / vol; // Injected tracer
+	});
+}
+
+void AddSupernova(amrex::MultiFab &mf, amrex::GpuArray<Real, AMREX_SPACEDIM> prob_lo, amrex::GpuArray<Real, AMREX_SPACEDIM> prob_hi,
+		  amrex::GpuArray<Real, AMREX_SPACEDIM> dx, SimulationData<NewProblem> const &userData, int level)
+{
+	// TODO for AV - ave (and restore) the RNG state in the metadata.yaml file
+	//  inject energy into cells with stochastic sampling
```
BenWibking, line 177: I think it would be good to do this, so the simulations are reproducible when restarted. It should help debugging a lot.

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,540 @@
...
+	double z = posvec[2];
+
+	// Interpolate to find the accurate g-value from array-- because linterp doesn't work on Setonix
+	size_t i = 0;
+	double x_interp = std::abs(z);
+	while (i < z_data.size() - 1 && x_interp > z_data[i + 1]) {
+		i++;
+	}
+
+	// Perform linear interpolation
```
BenWibking: Can you use `math/interpolate.cpp` for this instead?

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,540 @@
...
+
+	// set random state
+	const int seed = 42;
+	amrex::InitRandom(seed, 1); // all ranks should produce the same values
+
+	// Problem initialization
+	QuokkaSimulation<NewProblem> sim(BCs_cc);
+
+	sim.reconstructionOrder_ = 3; // 2=PLM, 3=PPM
+	sim.cflNumber_ = 0.3;	      // *must* be less than 1/3 in 3D!
```
BenWibking, line 507: This could be set in the input file instead. It's up to you.

src/problems/MetalAdvectionProblem/test_sne.cpp
```diff
@@ -0,0 +1,540 @@
...
+
+			Real Eint = RadSystem<NewProblem>::ComputeEintFromEgas(rho, x1mom, x2mom, x3mom, Egas);
+
+			posvec[0] = prob_lo[0] + (i + 0.5) * dx[0];
+
+#if (AMREX_SPACEDIM >= 2)
+			posvec[1] = prob_lo[1] + (j + 0.5) * dx[1];
+#endif
+
+#if (AMREX_SPACEDIM >= 3)
```
BenWibking: Since the CMakeLists.txt file only lets the user compile this problem for 3D anyway, these ifdefs are probably not necessary.

BenWibking (changes requested):
A few minor comments:

aditivijayan:
@BenWibking - the jobs seem to be failing because of dubious ownership issues. Could you suggest a fix?

BenWibking:
> @BenWibking - the jobs seem to be failing because of dubious ownership issues. Could you suggest a fix?

Where do you see this?

I don't see this when I look at the log files for the Github actions. It looks like clang-tidy-review, CodeQL, and warnings are all failing because of errors reported by the compiler, or because clang-tidy is suggesting fixes.

aditivijayan:
> > @BenWibking - the jobs seem to be failing because of dubious ownership issues. Could you suggest a fix?
> 
> Where do you see this?
> 
> I don't see this when I look at the log files for the Github actions. It looks like clang-tidy-review, CodeQL, and warnings are all failing because of errors reported by the compiler, or because clang-tidy is suggesting fixes.

This is what I get when I ask copilot to explain the error.

aditivijayan:
> > > @BenWibking - the jobs seem to be failing because of dubious ownership issues. Could you suggest a fix?
> > 
> > 
> > Where do you see this?
> > I don't see this when I look at the log files for the Github actions. It looks like clang-tidy-review, CodeQL, and warnings are all failing because of errors reported by the compiler, or because clang-tidy is suggesting fixes.
> 
> This is what I get when I ask copilot to explain the error.

Ah nevermind, it was complaining about CMake

BenWibking:
Those warnings from clang-tidy are not relevant. This PR will suppress them: https://github.com/quokka-astro/quokka/pull/997.

BenWibking:
I think Azure DevOps is configured to disallow running on forks. I don't immediately recall where that setting is, but @chongchonghe should have permissions to change it. Or you could move this to a branch in this repository and create a new PR.

chongchonghe:
@BenWibking I have changed the setting on azure dev to allow pipelines from fork-based PR. It has a secured rule, so it requires a team member's (you and me) comment before building a PR.

chongchonghe:
@aditivijayan The GPU test failed with `Erroneous arithmetic operation`. Please take a look. I also recommend putting your metal problem at the top of `src/problems/CMakeLists.txt` so it runs first.

aditivijayan:
> @aditivijayan The GPU test failed with `Erroneous arithmetic operation`. Please take a look. I also recommend putting your metal problem at the top of `src/problems/CMakeLists.txt` so it runs first.

I can't what/where the error is. Could you point me to it?

chongchonghe:
> > @aditivijayan The GPU test failed with `Erroneous arithmetic operation`. Please take a look. I also recommend putting your metal problem at the top of `src/problems/CMakeLists.txt` so it runs first.
> 
> I can't what/where the error is. Could you point me to it?

https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=11939&view=logs&j=0410d313-00ef-550d-37c8-facbf7a1cb10&t=62d0f0fc-52cd-5dfc-847e-751a25f0b5e6

It's `Erroneous arithmetic operation`, meaning there is some illegal float-point operations, e.g. division by zero, nan. You should be able to reproduce that with a CPU test on linux or macOS.

chongchonghe:
> > @aditivijayan The GPU test failed with `Erroneous arithmetic operation`. Please take a look. I also recommend putting your metal problem at the top of `src/problems/CMakeLists.txt` so it runs first.
> 
> I can't what/where the error is. Could you point me to it?

https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=11939&view=logs&j=0410d313-00ef-550d-37c8-facbf7a1cb10&t=62d0f0fc-52cd-5dfc-847e-751a25f0b5e6

It's `Erroneous arithmetic operation`, meaning there is some illegal float-point operations, e.g. division by zero, nan. You should be able to reproduce that with a CPU test on linux or macOS.

chongchonghe:
@aditivijayan Can you fix the `Erroneous arithmetic operation` issue and merge this PR?

chongchonghe:
I meant to push to my own chong/GasGrav (origin), but made a mistake and pushed it to the remote, which is this PR. Sorry.

However, all I did was I merged in the development branch and resolved some merge conflicts from recent commits. Once this branch compiles and pass the GPU test, it will be ready to merge.

BenWibking:
It looks like the GPU tests failed.

Also, is this going to be converted to use the new star particles infrastructure? I think that would make this PR a lot simpler, now that it's already in Quokka elsewhere.

chongchonghe:
I removed SN injection function and the test passed, so the error comes from the old SN function that causes the cooling function to fail. 
I agree we don't have to keep this old version of manual SN injection and should use the Stochastic particle types. It could be self-consistent star formation and SN, or reading pre-defined SN locations and time from txt file.

BenWibking:
@aditivijayan Are you still working on this PR? If not, can it be closed?

BenWibking:
Ok, I am closing this for now. Feel free to re-open this PR in the future.

PR #716, opened 2024-08-17
Make assertion work in the exchange step and dust temperature solution step 

### LINKED ISSUE #543: amrex::Abort does not unconditionally abort on GPU

chongchonghe:
### Description

A fix to #543 : make the Newton-Raphson iteration return a flag and check it on host. 

More changes:
- Change `maxIter` from 400 to 50. 50 iterations should be more than enough. If it doesn't converge in less than 50 steps, it probably never will. 

### Related issues
Fixes #543 

### Checklist

### DIFF
--- a/src/radiation/radiation_system.hpp
+++ b/src/radiation/radiation_system.hpp
@@ -205,7 +205,7 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
 				       amrex::Real time);
 
 	static void AddSourceTerms(array_t &consVar, arrayconst_t &radEnergySource, amrex::Box const &indexRange, amrex::Real dt, int stage,
-				   double dustGasCoeff);
+				   double dustGasCoeff, int *num_failed_coupling, int *num_failed_dust, int *p_num_failed_outer_ite);
 
 	static void balanceMatterRadiation(arrayconst_t &consPrev, array_t &consNew, amrex::Box const &indexRange);
 
@@ -1255,7 +1255,7 @@ AMREX_GPU_HOST_DEVICE auto RadSystem<problem_t>::ComputeDustTemperature(double c
 		}
 	}
 
-	AMREX_ASSERT_WITH_MESSAGE(ite_td < max_ite_td, "Newton iteration for dust temperature did not converge.");
+	AMREX_ASSERT_WITH_MESSAGE(ite_td < max_ite_td, "Newton iteration for dust temperature failed to converge.");
 	if (ite_td >= max_ite_td) {
 		T_d = -1.0;
 	}
@@ -1264,7 +1264,7 @@ AMREX_GPU_HOST_DEVICE auto RadSystem<problem_t>::ComputeDustTemperature(double c
 
 template <typename problem_t>
 void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEnergySource, amrex::Box const &indexRange, amrex::Real dt_radiation,
-					  const int stage, double dustGasCoeff)
+					  const int stage, double dustGasCoeff, int *p_num_failed_coupling, int *p_num_failed_dust, int *p_num_failed_outer_ite)
 {
 	arrayconst_t &consPrev = consVar; // make read-only
 	array_t &consNew = consVar;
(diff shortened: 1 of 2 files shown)


src/radiation/radiation_system.hpp
```diff
@@ -1678,6 +1695,10 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 				} // END NEWTON-RAPHSON LOOP
 
 				AMREX_ALWAYS_ASSERT_WITH_MESSAGE(n < maxIter, "Newton-Raphson iteration failed to converge!");
```
BenWibking: Is there a reason to leave the old abort statements in the code?

chongchonghe: I kept it there just to let abort earlier on CPU. This is particularly useful for development because you know exactly when and where it aborts.

BenWibking: Can you change it to `AMREX_ASSERT_WITH_MESSAGE`? Then it only has different behavior when `-DCMAKE_BUILD_TYPE=Debug`, rather than CPU vs GPU (which I think is confusing).

src/radiation/radiation_system.hpp
```diff
@@ -1962,6 +1983,15 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 			consNew(i, j, k, x3RadFlux_index + numRadVars_ * g) = Frad_t1[2][g];
 		}
 	});
+
+	const int nf_coupling = *(num_failed_coupling.copyToHost());
+	const int nf_dust = *(num_failed_dust.copyToHost());
```
BenWibking: I'm not sure whether `copyToHost` is asynchronous or not. Can you check this? If it's asynchronous, then there is a race condition here: we don't know whether the GPU will write to the memory location before the CPU reads it.

For performance reasons, it would be better to do this after all of the FABs have been processed, rather than after each FAB. Otherwise, the GPU will have to synchronize with the CPU after each FAB (in order to guarantee that the CPU is reading the up-to-date value of each variable). This synchronization can be relatively expensive.

src/radiation/radiation_system.hpp
chongchonghe: OK.

src/radiation/radiation_system.hpp
chongchonghe: I tested this PR on gadi GPU and it indeed aborted as expected when I set max_Iter to a small number.

chongchonghe: > For performance reasons, it would be better to do this after all of the FABs have been processed, rather than after each FAB.

I'll see if I can do that. My idea is to set a counter in `QuokkaSimulation<problem_t>::subcycleRadiationAtLevel` and pass it to `operatorSplitSourceTerms` then `AddSourceTerms`.

BenWibking: Ok, it's good to also check the source code: https://github.com/AMReX-Codes/amrex/blob/c061a14011d4d58755f330d15a20f7243863a90d/Src/Base/AMReX_GpuBuffer.H#L91.

There is still a potential performance issue. Maybe you could allocate the `GpuBuffer`s in the calling function, and then check the result there?

chongchonghe: Updated.

I'm not sure what you mean by "calling function". Now, `copyToHost` is called once per radiation cycling. This should be optimal enough. I also tested it on gadi GPU and the code aborted as expected when set to.

BenWibking (approved):

PR #720, opened 2024-08-26
improve convergence rate of the radiation Newton-Raphson solver

chongchonghe:
### Description

Changes:

1. Added a trick to make the Newton-Raphson iteration converge faster in some situations: when $\Delta T > \max(T_{\rm gas}, T_{\rm rad})$,  set $T_{\rm gas}$ to $T_{\rm rad}$ and $R$ to 0. The motivation for doing this trick is explained below.
2. Added `radiation.print_iteration_counts` parameter to enable counting the mean and max number of Newton-Raphson iterations in each radiation step. When set to `true`, the log file would look like:

```log
time_subcycle = 0, total number of Newton-Raphson solvings is 128, (mean, max) number of Newton-Raphson iterations are 1, 1

Coarse STEP 2 at t = 0.0046875 (0.9375%) starts ...
time_subcycle = 0.0046875, total number of Newton-Raphson solvings is 128, (mean, max) number of Newton-Raphson iterations are 1, 1

```

3. Defined a new radiation problem, `RadMarshakDust`. Initially, the gas is set with a density and temperature of 1, and the initial radiation energy density is 0. Radiation with a temperature $T_r = 0.01$ streams in from the left boundary. Before the dust model was introduced, we assumed gas and dust were perfectly coupled, and the gas would be heated by radiation where it could penetrate, and in regions without radiation, the gas temperature would cool to 0.003159 (solution of $1 = a_R T^4 + T$). In this test, however, we turned on the dust model and set the dust-gas interaction coefficient to a very small number. As a result, the dust and gas are decoupled: the dust temperature will reach the radiation temperature and stay in equilibrium, while the gas remains at its initial temperature. 

In a future PR, I will extend this test by incorporating a multigroup model. This will demonstrate how FUV radiation from the left boundary is absorbed by the dust, which then re-radiates in IR. 

### Related issues

None.

### Checklist


BenWibking:
We should not do this. The auxiliary internal energy should never be used as a primary source of truth for the internal energy. It should only be used when sync'ing with the total energy in the hydro update.

BenWibking:
The dominant source of error in the internal energy is due to *truncation* errors, not round off errors. These are not the same, and the distinction is *very* important.

BenWibking:
A better way to solve this would be to add a call to SyncInternalEnergy after each timestep for radiation-only test problems. In real problems, this is done already during the hydro update.

chongchonghe:
> The dominant source of error in the internal energy is due to _truncation_ errors, not round off errors. These are not the same, and the distinction is _very_ important.

I believe it is roundoff error in this case. What happens in this test is that the gas has an equilibrium temperature of 1 because I set the radiation temperature to 1. However, because the specific heat of radiation is very large, due to conservation of momentum, the gas momentum increases to 10^13 after the first step. Then, in the second time step, because the internal energy is derived from subtracting the total energy from the kinetic energy, the result is 0 due to roundoff error.

chongchonghe:
> We should not do this. The auxiliary internal energy should never be used as a primary source of truth for the internal energy. It should only be used when sync'ing with the total energy in the hydro update.

Why shouldn't we use the auxiliary internal energy?

chongchonghe:
> We should not do this. The auxiliary internal energy should never be used as a primary source of truth for the internal energy. It should only be used when sync'ing with the total energy in the hydro update.

The source of error in this test comes from the radiation subcycle it self and has nothing to do with hydro update. Here is the relavent code:

In the end of the first radiation subcycle:
```
consNew(i, j, k, gasInternalEnergy_index) = Egas_guess;
consNew(i, j, k, gasEnergy_index) = ComputeEgasFromEint(rho, x1GasMom1, x2GasMom1, x3GasMom1, Egas_guess);
```

In the beginning of the second radiation subcycle:
```
const double Egastot0 = consPrev(i, j, k, gasEnergy_index);
Egas0 = ComputeEintFromEgas(rho, x1GasMom0, x2GasMom0, x3GasMom0, Egastot0);
```

The internal energy is lost in this back-and-forth calculation. 

I don't think `SyncDualEnergy` will solve this problem. Here is the relavent code in `SyncDualEnergy`:

```cpp
		// Li et al. sync method
		// replace Eint with Eint_cons == (Etot - Ekin) if (Eint_cons / E) > eta
		if (Eint_cons > eta * Etot) {
			consVar[bx](i, j, k, internalEnergy_index) = Eint_cons;
		} else { // non-conservative sync
			consVar[bx](i, j, k, internalEnergy_index) = Eint_aux;
			consVar[bx](i, j, k, energy_index) = Eint_aux + Ekin;
		}
```

The problem is `(Eint_aux + Ekin) - Ekin = 0.0` when `Eint_aux << Ekin`. So, I think we have to use the auxiliary internal energy in the beginning of radiation update.

BenWibking:
> > The dominant source of error in the internal energy is due to _truncation_ errors, not round off errors. These are not the same, and the distinction is _very_ important.
> 
> I believe it is roundoff error in this case. What happens in this test is that the gas has an equilibrium temperature of 1 because I set the radiation temperature to 1. However, because the specific heat of radiation is very large, due to conservation of momentum, the gas momentum increases to 10^13 after the first step. Then, in the second time step, because the internal energy is derived from subtracting the total energy from the kinetic energy, the result is 0 due to roundoff error.

If it is roundoff error due to catastrophic cancellation, that means that the kinetic energy is $\sim 10^{15}$ times larger than the internal energy. The Mach number of that flow is unphysically enormous, and no code can deal with that (correctly). Not even molecular gas moving in a cosmological simulation has a Mach number that large (1000 km/s / 0.2 km/s ~ Mach 5000).

Unphysically enormous: a kinetic to internal ratio of $10^{15}$ implies a Mach number of order $10^{7.5}$. For molecular gas at 10 K, that implies a velocity of $\sim 20$ times the speed of light.

chongchonghe:
OK, that totally makes sense. I will change the parameters in this test to make the Mach number more realistic. 

However, I still don't understand why you can't use the auxiliary internal energy. The code I pasted above from `SyncDualEnergy` basically enforces that `consVar[bx](i, j, k, internalEnergy_index)` stores a value that is exactly what you would obtain from `ComputeEintFromEgas`, so, why not just use `consVar[bx](i, j, k, internalEnergy_index)` and save some calculation (and also avoid some roundoff error, although not significant)?

BenWibking:
> OK, that totally makes sense. I will change the parameters in this test to make the Mach number more realistic. 
> 
> 
> 
> However, I still don't understand why you can't use the auxiliary internal energy. The code I pasted above from `SyncDualEnergy` basically enforces that `consVar[bx](i, j, k, internalEnergy_index)` stores a value that is exactly what you would obtain from `ComputeEintFromEgas`, so, why not just use `consVar[bx](i, j, k, internalEnergy_index)` and save some calculation (and also avoid some roundoff error, although not significant)? 

You could, as long as the sync is done after the hydro step. But users have the option to disable the auxiliary internal energy completely with a runtime option, so it's probably not a good idea for other physics modules to assume it will always be synced.

chongchonghe:
> > OK, that totally makes sense. I will change the parameters in this test to make the Mach number more realistic.
> > However, I still don't understand why you can't use the auxiliary internal energy. The code I pasted above from `SyncDualEnergy` basically enforces that `consVar[bx](i, j, k, internalEnergy_index)` stores a value that is exactly what you would obtain from `ComputeEintFromEgas`, so, why not just use `consVar[bx](i, j, k, internalEnergy_index)` and save some calculation (and also avoid some roundoff error, although not significant)?
> 
> You could, as long as the sync is done after the hydro step. But users have the option to disable the auxiliary internal energy completely with a runtime option, so it's probably not a good idea for other physics modules to assume it will always be synced.

Oh, got it. I have brought back `ComputeEintFromEgas`. The other changes in this PR are still useful, so I'll keep it.

chongchonghe:
@BenWibking I've changed the title and description of this PR. The other changes are still useful. Can you review this again?

src/QuokkaSimulation.hpp
```diff
@@ -1652,8 +1656,16 @@ void QuokkaSimulation<problem_t>::subcycleRadiationAtLevel(int lev, amrex::Real
...
-			operatorSplitSourceTerms(stateNew, indexRange, time_subcycle, dt_radiation, 2, dx, prob_lo, prob_hi, p_num_failed_coupling,
-						 p_num_failed_dust, p_num_failed_outer);
+			operatorSplitSourceTerms(stateNew, indexRange, time_subcycle, dt_radiation, 2, dx, prob_lo, prob_hi, p_iteration_counter,
+						 p_num_failed_coupling, p_num_failed_dust, p_num_failed_outer);
+		}
+
+		if (radiation_enable_counter_) {
+			// Iteration counter
+			auto h_iteration_counter = iteration_counter.copyToHost();
+			const double iteration_mean = static_cast<double>(h_iteration_counter[1]) / static_cast<double>(h_iteration_counter[0]);
```
BenWibking: This only works here if you have 1 MPI rank. If you have more than 1 MPI rank, this is wrong. You need to perform an MPI reduction to find the maximum and mean number of iterations.

src/QuokkaSimulation.hpp
```diff
@@ -390,6 +391,7 @@ template <typename problem_t> void QuokkaSimulation<problem_t>::readParmParse()
 		rpp.query("reconstruction_order", radiationReconstructionOrder_);
 		rpp.query("cfl", radiationCflNumber_);
 		rpp.query("dust_gas_interaction_coeff", dustGasInteractionCoeff_);
+		rpp.query("enable_counter", radiation_enable_counter_);
```
BenWibking: Maybe this could be called `print_iteration_counts` or something slightly more precise? The counter is always enabled, it's just that it's not printed unless this is turned on.

src/problems/RadMarshakHot/test_radiation_marshak_hot.cpp
```diff
@@ -0,0 +1,265 @@
+//==============================================================================
+// TwoMomentRad - a radiation transport library for patch-based AMR codes
+// Copyright 2020 Benjamin Wibking.
+// Released under the MIT license. See LICENSE file included in the GitHub repo.
+//==============================================================================
+/// \file test_radiation_marshak_hot.cpp
+/// \brief Defines a test Marshak wave problem with extremely high specific heat capacity in radiation.
```
BenWibking: What is the motivation for this test problem?

src/problems/RadMarshakHot/test_radiation_marshak_hot.cpp
```diff
@@ -0,0 +1,265 @@
...
+	// 	sol_norm += std::abs(erad_exact[i]);
+	// }
+
+	// const double rel_err_norm = err_norm / sol_norm;
+	// const double rel_err_tol = 0.02;
+	// int status = 1;
+	// if (rel_err_norm < rel_err_tol) {
+	// 	status = 0;
+	// }
+	// amrex::Print() << "Relative L1 norm = " << rel_err_norm << std::endl;
```
BenWibking: Why did you comment this out? I think it's necessary to check the solution here.

src/radiation/radiation_system.hpp
```diff
@@ -1693,6 +1711,11 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 					amrex::Gpu::Atomic::Add(p_num_failed_coupling_local, 1);
 				}
 
+				// update iteration counter: (+1, +ite, max(self, ite))
+				amrex::Gpu::Atomic::Add(&p_iteration_counter_local[0], 1);
+				amrex::Gpu::Atomic::Add(&p_iteration_counter_local[1], n + 1);
+				amrex::Gpu::Atomic::Max(&p_iteration_counter_local[2], n + 1);
```
BenWibking: Can you add a comment explaining what each element of `p_iteration_counter_local` means?

BenWibking (changes requested):

BenWibking:
I think the title is a bit confusing. Maybe make it something like "improve convergence of radiation Newton-Raphson solver"?

src/problems/RadMarshakHot/test_radiation_marshak_hot.cpp
chongchonghe: It was intended for demonstrating that the updated Newton iteration converges faster. It's also used as a base for future multigroup test with dust and photoelectric heating.

src/QuokkaSimulation.hpp

src/problems/RadMarshakHot/test_radiation_marshak_hot.cpp
chongchonghe: Updated and now there is an exact solution to compare with.

src/radiation/radiation_system.hpp
```diff
@@ -1693,6 +1713,11 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 					amrex::Gpu::Atomic::Add(p_num_failed_coupling_local, 1);
 				}
 
+				// update iteration counter: (+1, +ite, max(self, ite))
+				amrex::Gpu::Atomic::Add(&p_iteration_counter_local[0], 1);     // total number of radiation updates
```
BenWibking, line 1717: Do we need a separate counter for this? This should be equal to the number of cells, no?

src/radiation/radiation_system.hpp
```diff
@@ -1447,7 +1447,7 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 				quokka::valarray<double, nGroups_> F_D{};
 
 				const double resid_tol = 1.0e-11; // 1.0e-15;
-				const int maxIter = 50;
+				const int maxIter = 100;
```
BenWibking: What is the motivation for changing this?

src/QuokkaSimulation.hpp
```diff
@@ -1652,8 +1656,27 @@ void QuokkaSimulation<problem_t>::subcycleRadiationAtLevel(int lev, amrex::Real
...
 			// update state_new_cc_[lev] in place (updates both radiation and hydro vars)
-			operatorSplitSourceTerms(stateNew, indexRange, time_subcycle, dt_radiation, 2, dx, prob_lo, prob_hi, p_num_failed_coupling,
-						 p_num_failed_dust, p_num_failed_outer);
+			operatorSplitSourceTerms(stateNew, indexRange, time_subcycle, dt_radiation, 2, dx, prob_lo, prob_hi, p_iteration_counter,
+						 p_num_failed_coupling, p_num_failed_dust, p_num_failed_outer);
+		}
+
+		if (print_rad_counter_) {
+			auto h_iteration_counter = iteration_counter.copyToHost();
+			long global_solver_count = h_iteration_counter[0];  // number of Newton-Raphson solvings
```
BenWibking, line 1665: This is just the number of cells, right?

BenWibking (commented):
Just two minor comments below.

src/QuokkaSimulation.hpp
chongchonghe: No, because there is an outer iteration to update the work term. If there is always only one outer iteration, `global_solver_count` would be 2x number of cells (2 comes from two IMEX stages). Therefore, `global_solver_count` also counts the average number of outer iterations per radiation update.

src/radiation/radiation_system.hpp
chongchonghe: With the dust-gas coupling model, it usually takes more iterations to converge, so I have to increase maxIter.

BenWibking (approved):

PR #731, opened 2024-09-02
Separate single- and multi-group radiation modules

chongchonghe:
### Description

The single-group radiation scheme has complex high-order terms, while the multi-group scheme has complexy opacity functions and Jacobian matrix. Previously, these two schemes are mixed together. It's a mess. In this PR, I splitted `AddSourceTerms` into `AddSourceTermsSingleGroup` and `AddSourceTermsMultiGroup`. In `AddSourceTermsSingleGroup`, whether or not the dust model is enabled, the Jacobian is a 2x2 matrix and can be inverted with a simple analytical expression. Second or higher order terms are supported. In `AddSourceTermsMultiGroup`, complexy multigroup opacity models are included but only 0th and 1st order terms are supported. The Jacobian is a (Ng + 1) by (Ng + 1) sparse matrix. 

### Related issues

None.

### Checklist


chongchonghe:
@BenWibking Both GPU tests failed due to `disk quota exceeded`. Everything else should be fine.

src/problems/RadDust/test_rad_dust.cpp
```diff
@@ -74,23 +74,14 @@ template <> AMREX_GPU_HOST_DEVICE auto RadSystem<DustProblem>::ComputeFluxMeanOp
...
-template <>
-AMREX_GPU_HOST_DEVICE auto RadSystem<DustProblem>::ComputeThermalRadiation(amrex::Real temperature, amrex::GpuArray<double, nGroups_ + 1> const &boundaries)
-    -> quokka::valarray<amrex::Real, nGroups_>
+template <> AMREX_GPU_HOST_DEVICE auto RadSystem<DustProblem>::ComputeThermalRadiationSingleGroup(amrex::Real temperature) -> amrex::Real
 {
-	auto radEnergyFractions = ComputePlanckEnergyFractions(boundaries, temperature);
-	const double power = radiation_constant_ * temperature;
-	auto Erad_g = power * radEnergyFractions;
-	return Erad_g;
+	return radiation_constant_ * temperature;
```
markkrumholz, line 80: The notation here is confusing. Is temperature really T^4? If so, please rename the variable to that. If not, can you clarify why the source term is only radiation_constant * T and not T^4?

src/problems/RadDust/test_rad_dust.cpp
```diff
@@ -74,23 +74,14 @@ template <> AMREX_GPU_HOST_DEVICE auto RadSystem<DustProblem>::ComputeFluxMeanOp
...
 
-template <>
-AMREX_GPU_HOST_DEVICE auto RadSystem<DustProblem>::ComputeThermalRadiationTempDerivative(
-    amrex::Real temperature, amrex::GpuArray<double, nGroups_ + 1> const &boundaries) -> quokka::valarray<amrex::Real, nGroups_>
+template <> AMREX_GPU_HOST_DEVICE auto RadSystem<DustProblem>::ComputeThermalRadiationTempDerivativeSingleGroup(amrex::Real /*temperature*/) -> amrex::Real
 {
-	auto radEnergyFractions = ComputePlanckEnergyFractions(boundaries, temperature);
-	const double d_power_dt = radiation_constant_;
-	return d_power_dt * radEnergyFractions;
+	return radiation_constant_;
```
markkrumholz, line 86: Same question as previous comment. I presume the reason it's just radiation constant is that you're taking the derivative with respect to T^4 and not T, but the variable naming and comments should make that clear.

markkrumholz (commented):
I left a couple of comments, and then there some unused variable warnings, but once those are resolved this looks fine and I can approve.

src/problems/RadDust/test_rad_dust.cpp
chongchonghe: I have to assume B = a_rad * T_d in this test in order to linearize this problem so that I can obtain an analytic solution. I added a comment to explain this. The motivation of doing this is also explained before Eq. 44 of the overleaf document quokka-thermochemistry.

chongchonghe:
> I left a couple of comments, and then there some unused variable warnings, but once those are resolved this looks fine and I can approve.

I had to assume B(T) = a_rad * T in that test in order to obtain an analytic solution. 

The unused variable warning are some false alarms and some unavoidable. They are used in several `if constexpr`, and if you compile the code with if constexpr false, it will think the variables are unused. I tried to use `amrex::ignore_unused` to suppress some of them.

src/radiation/radiation_system.hpp
```diff
@@ -2006,4 +1934,524 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
...
+						tau = dt * rho * kappaP * chat * lorentz_factor;
+						if (tau > 0.0) {
+							Erad_guess = kappaPoverE * (fourPiBoverC - (R - work) / tau);
+							if constexpr (force_rad_floor_in_iteration) {
+								if (Erad_guess <= 0.0) {
+									Egas_guess -= (c / chat) * (Erad_floor_ - Erad_guess);
+									Erad_guess = Erad_floor_;
+								}
+							} else {
+								// AMREX_ASSERT(Erad_guess > 0.0);
```
BenWibking: Is there a reason to keep this `else` statement?

src/radiation/radiation_system.hpp
```diff
@@ -2006,4 +1934,524 @@ void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEne
 	});
 }
 
+template <typename problem_t>
+void RadSystem<problem_t>::AddSourceTermsSingleGroup(array_t &consVar, arrayconst_t &radEnergySource, amrex::Box const &indexRange, Real dt_radiation,
```
BenWibking: The source term implementation has become very complicated. Maybe this function can go in a separate header file?

src/radiation/radiation_system.hpp
```diff
@@ -1261,10 +1295,12 @@ AMREX_GPU_HOST_DEVICE auto RadSystem<problem_t>::ComputeDustTemperature(double c
 }
 
 template <typename problem_t>
-void RadSystem<problem_t>::AddSourceTerms(array_t &consVar, arrayconst_t &radEnergySource, amrex::Box const &indexRange, amrex::Real dt_radiation,
-					  const int stage, double dustGasCoeff, int *p_iteration_counter, int *p_num_failed_coupling, int *p_num_failed_dust,
-					  int *p_num_failed_outer_ite)
+void RadSystem<problem_t>::AddSourceTermsMultiGroup(array_t &consVar, arrayconst_t &radEnergySource, amrex::Box const &indexRange, amrex::Real dt_radiation,
```
BenWibking: The multigroup source term implementation has become very complicated. Maybe this can go in a separate header file?

BenWibking:
> The unused variable warning are some false alarms and some unavoidable. They are used in several `if constexpr`, and if you compile the code with if constexpr false, it will think the variables are unused. I tried to use `amrex::ignore_unused` to suppress some of them.

Can you suppress them all with `amrex::ignore_unused`? Or is there an issue you are running into if you try to do this?

src/radiation/radiation_system.hpp
chongchonghe: No. I have deleted it.

chongchonghe:
> > The unused variable warning are some false alarms and some unavoidable. They are used in several `if constexpr`, and if you compile the code with if constexpr false, it will think the variables are unused. I tried to use `amrex::ignore_unused` to suppress some of them.
> 
> Can you suppress them all with `amrex::ignore_unused`? Or is there an issue you are running into if you try to do this?

I already have `amrex::ignore_unused` for all of them, but the warning still show up. It might be a bug of github-advanced-security. I'll ignore them for now.

src/radiation/radiation_system.hpp
chongchonghe: Maybe in a future PR?

src/radiation/radiation_system.hpp
chongchonghe: This is a good point. Maybe in a future PR?

BenWibking:
> > > The unused variable warning are some false alarms and some unavoidable. They are used in several `if constexpr`, and if you compile the code with if constexpr false, it will think the variables are unused. I tried to use `amrex::ignore_unused` to suppress some of them.
> > 
> > 
> > Can you suppress them all with `amrex::ignore_unused`? Or is there an issue you are running into if you try to do this?
> 
> I already have `amrex::ignore_unused` for all of them, but the warning still show up. It might be a bug of github-advanced-security. I'll ignore them for now.

Ok. You can dismiss them as false positives using the drop-down menu.

src/radiation/radiation_system.hpp
BenWibking: Is there a reason not to do it in this PR? It should take about 60 seconds to implement.

BenWibking: In this case, the right way to do it is probably with a recursive include. That is, at the end of `radiation_system.hpp`, you can do
```
#include "AddSourceTerms.hpp  // IWYU pragma: export
#include "AddSourceTermsMultigroup.hpp  // IWYU pragma: export
```
and then in each of those header files, just cut-and-paste the functions, and add this at the top of each file:
```
// IWYU pragma: private; include "radiation_system.h"
```
(See https://clangd.llvm.org/guides/include-cleaner#iwyu-pragmas)

chongchonghe: OK, this looks good. I've updated it. The code compiles and runs without any problem. The only drawback I'm experiencing is that vscode won't do variable indexing inside add_source_terms_multi_group.hpp, but I think this can be fixed.

chongchonghe:
@BenWibking Can you approve this? I'll merge it once the GPU tests finish.

BenWibking (dismissed):

chongchonghe:
I'll hold this for a little bit. I made some change to the code structure and fixed the clangd/IntelliSense problem.

chongchonghe:
This is an alternative approach I showed in the meeting today. This way the auto-complete and syntax highlighting works fine in VSCode. https://github.com/quokka-astro/quokka/tree/chong/separate-v2

BenWibking:
> This is an alternative approach I showed in the meeting today. This way the auto-complete and syntax highlighting works fine in VSCode. https://github.com/quokka-astro/quokka/tree/chong/separate-v2

Can you see if it works with a recursive include, e.g.:

radiation_system.hpp
```
// end of file
#include "source_terms.hpp" // IWYU: blahblah
```

source_terms.hpp
```
//start of file
// IWYU: blahblah1
#include "radiation_system.hpp" // IWYU: blahblah2
```

Since there are include guards for `radiation_system.hpp`, it should not actually cause an infinite loop, but it might be enough for VSCode to figure out where the symbols are coming from.

BenWibking:
This approach should be the same as "method 3" here, except that their version of `source_terms.hpp` (incorrectly) ends in `.cpp`: https://www.codeproject.com/Articles/48575/How-to-Define-a-Template-Class-in-a-h-File-and-Imp

If this still doesn't work, it should be reported as a bug in `clangd`: https://github.com/clangd/clangd/issues.

chongchonghe:
@BenWibking Works fine now.

BenWibking (approved):

PR #734, opened 2024-09-04
Restructure AddSourceTermsMultiGroup

chongchonghe:
### Description

In this PR I attempt to make the `AddSourceTermsMultiGroup` function more structured. 

Now, `AddSourceTermsMultiGroup` looks like this:

```c++
AddSourceTermsMultiGroup()
{
  // declare variables
  ...
    
  for outer loop {
    // declare variables
    ...
      
    // Compute the gas and radiation energy update. This also updates the opacities. When ite == 0, this also computes the work term.
    if (enable_dust_gas_thermal_coupling_model_) {
      updated_energy = SolveMatterRadiationEnergyExchange(..., &ComputeJacobianForGasAndDust, &ComputeDustTemperatureBateKeto);
    } else {
      updated_energy = SolveMatterRadiationEnergyExchange(..., &ComputeJacobianForGas, &ComputeDustTemperatureGasOnly);
    }
    
    // Compute radiation flux update
    ...
      
    // Deal with the work term.
    ...
  }
}
```

and this is the only place where `enable_dust_gas_thermal_coupling_model_` is used. 

Next, hopefully, I can implement more complex dust models and cooling/heating by defining new `ComputeJacobian` and new `ComputeDustTemperature`. 

### Related issues

N/A

### Checklist


src/radiation/radiation_system.cpp
BenWibking, line 1: We could actually just delete this file entirely. Nothing uses it.

src/radiation/radiation_system.hpp
```diff
@@ -88,6 +89,29 @@ struct RadPressureResult {
...
+
+// A struct to hold the results of ComputeJacobianForPureGas or ComputeJacobianForGasAndDust
+template <typename problem_t> struct JacobianResult {
+	double J00;
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> J0g;
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> Jg0;
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> Jgg;
+	double F0;
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> Fg;
+	double Fg_abs_sum;
```
BenWibking: It's usually best to group the struct members of the same type together for optimal memory packing, since C++ requires that struct members are laid out in memory in the order they appear in the definition. So I would put all of the `double`s first.

src/radiation/radiation_system.hpp
```diff
@@ -250,32 +272,59 @@ template <typename problem_t> class RadSystem : public HyperbolicSystem<problem_
...
 	AMREX_GPU_HOST_DEVICE static auto
 	ComputeThermalRadiationTempDerivativeMultiGroup(amrex::Real temperature,
 							amrex::GpuArray<double, nGroups_ + 1> const &boundaries) -> quokka::valarray<amrex::Real, nGroups_>;
 
-	AMREX_GPU_HOST_DEVICE static auto ComputeThermalRadiationTempDerivativeSingleGroup(amrex::Real temperature) -> Real;
+	AMREX_GPU_DEVICE static auto
+	ComputeDustTemperatureBateKeto(double T_gas, double T_d_init, double rho, quokka::valarray<double, nGroups_> const &Erad, double N_d, double dt,
+				       double R_sum, int n_step,
+				       amrex::GpuArray<double, nGroups_ + 1> const &rad_boundaries = amrex::GpuArray<double, nGroups_ + 1>{},
+				       amrex::GpuArray<double, nGroups_> const &rad_boundary_ratios = amrex::GpuArray<double, nGroups_>{}) -> double;
```
BenWibking, line 301: Ideally, I think the functions to compute a dust temperature should be in a separate class, but that could be done in a future PR.

src/radiation/radiation_system.hpp
```diff
@@ -88,6 +89,29 @@ struct RadPressureResult {
...
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> kappaPVec;		      // Planck mean opacity
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> kappaEVec;		      // energy mean opacity
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> kappaFVec;		      // flux mean opacity
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> work;		      // work term
+	amrex::GpuArray<double, Physics_Traits<problem_t>::nGroups> delta_nu_kappa_B_at_edge; // Delta (nu * kappa_B * B)
+};
+
+// A struct to hold the results of ComputeJacobianForPureGas or ComputeJacobianForGasAndDust
+template <typename problem_t> struct JacobianResult {
+	double J00;
```
BenWibking: Is this the (0, 0) element of the Jacobian? What variables does it correspond to? Can you add a comment that explains this?

src/radiation/source_terms_single_group.hpp
```diff
@@ -490,9 +497,9 @@ void RadSystem<problem_t>::AddSourceTermsSingleGroup(array_t &consVar, arraycons
 			}
 		} // end full-step iteration
 
-		AMREX_ASSERT_WITH_MESSAGE(ite < max_ite, "AddSourceTerms iteration failed to converge!");
+		AMREX_ASSERT_WITH_MESSAGE(ite < max_ite, "AddSourceTerms outer iteration failed to converge!");
 		if (ite >= max_ite) {
```
BenWibking, line 508: I think it would be more conventional to call these variables `iter` and `max_iter` to indicate iteration counts.

BenWibking (commented):
I think this is a good improvement and much better structured. I haven't gone through each line yet.

Did you intend to make the same changes to AddSourceTermsSingleGroup and AddSourceTermsMultiGroup? It looks like you've made similar changes, but I haven't compared them line-by-line.

There is a lot of commented-out code. Is it necessary to keep that? Otherwise, I think it would be a lot cleaner to just remove it.

src/radiation/radiation_system.cpp
chongchonghe: I had been thinking so.

src/radiation/radiation_system.hpp
chongchonghe: Done.

src/radiation/radiation_system.hpp
chongchonghe: Done.

chongchonghe:
> Did you intend to make the same changes to AddSourceTermsSingleGroup and AddSourceTermsMultiGroup? It looks like you've made similar changes, but I haven't compared them line-by-line.

No, I did not do the same thing to `AddSourceTermsSingleGroup` and I don't intend to. I think `AddSourceTermsSingleGroup` is simple enough and I won't make much change to it in the future, so It's probably not worth doing it. 

> There is a lot of commented-out code. Is it necessary to keep that? Otherwise, I think it would be a lot cleaner to just remove it.

I see only one commented-out code (just before using `SolveMatterRadiationEnergyExchange`) and I meant to keep it for future reference.

src/radiation/source_terms_single_group.hpp
chongchonghe: Done.

src/radiation/radiation_system.hpp
```diff
@@ -88,6 +89,29 @@ struct RadPressureResult {
...
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> kappaPVec;		      // Planck mean opacity
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> kappaEVec;		      // energy mean opacity
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> kappaFVec;		      // flux mean opacity
+	quokka::valarray<double, Physics_Traits<problem_t>::nGroups> work;		      // work term
+	amrex::GpuArray<double, Physics_Traits<problem_t>::nGroups> delta_nu_kappa_B_at_edge; // Delta (nu * kappa_B * B)
+};
+
+// A struct to hold the results of ComputeJacobianForPureGas or ComputeJacobianForGasAndDust
+template <typename problem_t> struct JacobianResult {
+	double J00;	   // (0, 0) component of the Jacobian matrix
```
BenWibking, line 109: What physical variables does this correspond to?

BenWibking (commented):
This looks good.

It would be useful to have the ordering of the variables used to define the Jacobian matrix listed in a comment next to the definition of `JacobianResult`.

src/radiation/radiation_system.hpp
chongchonghe: This `JacobianResult` can be used to hold any Jacobian matrix of similar size, so it depends on the problem we are solving. For pure gas or dust in the coupled regime, J00 is d F_G / d Egas. For dust in the decoupled regime, it's d F_G / d T_d. So, I wouldn't add comments to explain them here, but instead, explain all elements inside each `ComputeJacobian` functions.

BenWibking (approved):

PR #766, opened 2024-10-02
photoelectric heating

chongchonghe:
### Description
In this PR, I implement photoelectric heating within the current gas-radiation-dust interaction framework. PE can be enabled by setting `enable_photoelectric_heating = true` in `ISM_Traits`. The prerequisite is `nGroups > 1` and `enable_dust_gas_thermal_coupling_model = true`. 

Added two test problems, `RadiationMarshakDustPE-coupled` and `RadiationMarshakDustPE-decoupled`, to test photoelectric heating in cases where gas and dust are well coupled and decoupled. These two tests share the same problem code `test_radiation_marshak_dust_and_PE.cpp` but use different runtime parameters. 

Here is the code structure:

```
- AddSourceTermsMultiGroup
  - outer loop:
    - compute `v_times_F_term` 
    - solve gas-dust-radiation energy exchange:
      - if constexpr (!enable_dust_gas_thermal_coupling_model_)
        - `SolveGasRadiationEnergyExchange`:
          - The matter-radiation coupling scheme from the multigroup paper
      - else if constexpr (!enable_photoelectric_heating_)
        - `SolveGasDustRadiationEnergyExchange()`:
          - The gas-dust-radiation coupling scheme proposed in #733, which has two cases: the well-coupled case and decoupled case
      - else
        - `SolveGasDustRadiationEnergyExchangeWithPE()`:
          - Based on the gas-dust-radiation coupling scheme, but with modified Jacobian in the well-coupled case to include PE heating. In the decoupled case, PE heating is added via a forward-Euler update using the updated radiation energy density after the gas-dust-radiation step.
    - compute updated `v_times_F_term` 
    - if `v_times_F_term` has converged, break
  - `UpdateFlux()`
```

### Related issues
A modification to #733 to include PE heating. 

### Checklist


psharda:
PE heating will be included in chemistry. Do you still want to have a separate implementation here in radiation?

markkrumholz:
> PE heating will be included in chemistry. Do you still want to have a separate implementation here in radiation?

Yes -- @chongchonghe and I discussed this, and while we will include this with chemistry, there are also uses cases for a standalone implementation that does not require chemistry. For example, the method implemented in [Bate & Keto (2015)](https://ui.adsabs.harvard.edu/abs/2015MNRAS.449.2643B/abstract) does photoelectric heating without chemistry, and ours is loosely based on that. The main use case here is that if we want to simulate the atomic ISM without resolving the molecular phase, we can do the partition of the gas into hot and cold atomic phases with this implementation.

chongchonghe:
> PE heating will be included in chemistry. Do you still want to have a separate implementation here in radiation?

and speed is another big factor. I bet this implementation of multigroup RHD + dust + PE is at least 3 times faster than using the VODE solver in microphysics.

BenWibking:
This has the same code maintainability issues that I thought we were trying to avoid by using Microphysics for the radiation update. I would prefer to discuss this on the telecon tomorrow before abandoning our previous plan to use Microphysics.

However, regardless of whether we use Microphysics or not, in principle, I think RadiationSystem should be agnostic to the type of non-LTE effects one wants to model. That is, the code should be structured such that implementation details of those effects should be handled in a separate class.

chongchonghe:
> This has the same code maintainability issues that I thought we were trying to avoid by using Microphysics for the radiation update. I would prefer to discuss this on the telecon tomorrow before abandoning our previous plan to use Microphysics.
> 
> However, regardless of whether we use Microphysics or not, in principle, I think RadiationSystem should be agnostic to the type of non-LTE effects one wants to model. That is, the code should be structured such that implementation details of those effects should be handled in a separate class.

This is simple to do. Looking at the code structure I wrote above, it should be easy to put `SolveGasDustRadiationEnergyExchange` and `SolveGasDustRadiationEnergyExchangeWithPE` into a separate file. I'll create a issue to remind me to do it later. For this PR, I'll keep it here because this way it's easier to review.

BenWibking:
> > This has the same code maintainability issues that I thought we were trying to avoid by using Microphysics for the radiation update. I would prefer to discuss this on the telecon tomorrow before abandoning our previous plan to use Microphysics.
> > However, regardless of whether we use Microphysics or not, in principle, I think RadiationSystem should be agnostic to the type of non-LTE effects one wants to model. That is, the code should be structured such that implementation details of those effects should be handled in a separate class.
> 
> This is simple to do. Looking at the code structure I wrote above, it should be easy to put `SolveGasDustRadiationEnergyExchange` and `SolveGasDustRadiationEnergyExchangeWithPE` into a separate file. I'll create a issue to remind me to do it later. For this PR, I'll keep it here because this way it's easier to review.

I think it would actually be easier to review if those were in a separate file.

chongchonghe:
@BenWibking ROCm failed due to 'Disk quota exceeded'. Can you fix that?

markkrumholz:
Will wait to review this until the opacity simplification PR I just merged is synced with this, then come back and look at it.

chongchonghe:
OK. I have merged in the dev branch and also moved everything about dust and PE heating into a separate file `source_terms_imperfectly_coupled_dust.hpp`. Now, `source_terms_multi_group.hpp` is a lot simpler. 

Ready for review and merge.

src/QuokkaSimulation.hpp
```diff
@@ -1675,9 +1677,12 @@ void QuokkaSimulation<problem_t>::subcycleRadiationAtLevel(int lev, amrex::Real
...
 					const double global_solving_mean =
 					    static_cast<double>(global_solver_count) / static_cast<double>(n_cells) / 2.0; // 2 stages
-					amrex::Print() << "average number of Newton-Raphson solvings per IMEX stage is " << global_solving_mean
-						       << ", (mean, max) number of Newton-Raphson iterations are " << global_iteration_mean << ", "
-						       << global_iteration_max << "\n";
+					const double global_decoupled_iteration_mean =
+					    static_cast<double>(global_decoupled_iteration_sum) / static_cast<double>(global_solver_count);
+					amrex::Print()
+					    << "average number of Newton-Raphson solvings per IMEX stage is " << global_solving_mean
+					    << ", (mean, max) number of Newton-Raphson iterations are " << global_iteration_mean << ", " << global_iteration_max
```
markkrumholz: Printing this information out in a system that is not using decoupled dust will be confusing. Can we wrap this in an appropriate constexpr'ed if?

src/problems/RadMarshakDustPE/test_radiation_marshak_dust_and_PE.cpp
```diff
@@ -0,0 +1,311 @@
...
+/// \brief Defines a test Marshak wave problem with weak coupling between dust and gas.
+///
+
+#include "test_radiation_marshak_dust_and_PE.hpp"
+#include "AMReX.H"
+#include "QuokkaSimulation.hpp"
+#include "util/fextract.hpp"
+#include "util/valarray.hpp"
+
+struct StreamingProblem {
```
markkrumholz: Can we change the problem template name to something a bit less confusing, since this is not in fact a streaming problem?

markkrumholz:
I'm looking through this, and it looks like the Jacobian solver with decoupled dust + radiation and photoelectric heating is still sitting in radiation_system.hpp. I thought the plan was to pull this out into a separate file, so not only would the custom source terms sit in different files, the custom Jacabians would too, and we wouldn't wind up with a bunch of bespoke, model-specific Jacobian solvers sitting in that file. Am I misremembering?

chongchonghe:
> I'm looking through this, and it looks like the Jacobian solver with decoupled dust + radiation and photoelectric heating is still sitting in radiation_system.hpp. I thought the plan was to pull this out into a separate file, so not only would the custom source terms sit in different files, the custom Jacabians would too, and we wouldn't wind up with a bunch of bespoke, model-specific Jacobian solvers sitting in that file. Am I misremembering?

Yes, you are right. I'll make that change.

src/problems/RadMarshakDustPE/test_radiation_marshak_dust_and_PE.cpp
chongchonghe: OK. Done.

markkrumholz:
> > I'm looking through this, and it looks like the Jacobian solver with decoupled dust + radiation and photoelectric heating is still sitting in radiation_system.hpp. I thought the plan was to pull this out into a separate file, so not only would the custom source terms sit in different files, the custom Jacabians would too, and we wouldn't wind up with a bunch of bespoke, model-specific Jacobian solvers sitting in that file. Am I misremembering?
> 
> Yes, you are right. I'll make that change.

Great. Just ping me when I should look again.

src/QuokkaSimulation.hpp

chongchonghe:
@markkrumholz OK, all comments addressed.

src/radiation/radiation_system.hpp
```diff
@@ -29,6 +29,8 @@
 #include "radiation/planck_integral.hpp"
 #include "util/valarray.hpp"
 
+#define LARGE 1.e100
+
```
markkrumholz: Can you get rid of this #define here, or move it into a done of the source term files?

markkrumholz:
OK, another stylistic question. The design here seems a bit different than our normal paradigm, in that you have the various source and exchange models protected by ifdef brackets, and you forward-declare the various exchange and solver functions is radiation_system.hpp so that even if you turn off a particular exchange model by not #define'ing the appropriate flag to turn it on, you don't get compilation errors thanks to the forward declaration. This is contrary to our style thus far of mostly using if constexpr's and templates to manage conditional compilation. The paradigm that I would think would work would be remove the forward declarations from radiation_system.hpp and the protection ifdef's from the various model files, and just #include the various model files in radiation_system.hpp, but then select which model to use at compile time with a trait in the problem file. Is there a reason not to use this paradigm here?

src/radiation/radiation_system.hpp
chongchonghe: I moved it to the file for the dust model.

chongchonghe:
> OK, another stylistic question. The design here seems a bit different than our normal paradigm, in that you have the various source and exchange models protected by ifdef brackets, and you forward-declare the various exchange and solver functions is radiation_system.hpp so that even if you turn off a particular exchange model by not #define'ing the appropriate flag to turn it on, you don't get compilation errors thanks to the forward declaration. This is contrary to our style thus far of mostly using if constexpr's and templates to manage conditional compilation. The paradigm that I would think would work would be remove the forward declarations from radiation_system.hpp and the protection ifdef's from the various model files, and just #include the various model files in radiation_system.hpp, but then select which model to use at compile time with a trait in the problem file. Is there a reason not to use this paradigm here?

I also tried another paradigm and I just pushed it. In this paradigm, there are two main files in src/radiation: `radiation_system.hpp` and `radiation_dust_system.hpp`. The latter file contains all the definition of the functions used solely in dust or dust+PE models, and it includes `radiation_system.hpp`. Now, in most tests where there is no dust model, we include radiation_system.hpp and those functions for dust is never read by the compiler. In tests with dust, we include `radiation_dust_system.hpp` in the header. Does this resolve your concern?

Both of the two hpp files have to be protected by ifdef in this new paradigm.

src/radiation/radiation_dust_system.hpp
```diff
@@ -0,0 +1,876 @@
...
+
+template <typename problem_t>
+AMREX_GPU_HOST_DEVICE auto RadSystem<problem_t>::DefinePhotoelectricHeatingE1Derivative(amrex::Real const /*temperature*/, amrex::Real const num_density)
+    -> amrex::Real
+{
+	// const double epsilon = 0.05; // default efficiency factor for cold molecular clouds
+	// const double ref_J_ISR = 5.29e-14; // reference value for the ISR in erg cm^3
+	// const double coeff = 1.33e-24;
+	// return coeff * epsilon * num_density / ref_J_ISR; // s^-1
+	return 0.0;
```
markkrumholz, line 13: Can you explain what is going on with the commented-out code in this routine?

src/radiation/radiation_dust_system.hpp
```diff
@@ -0,0 +1,876 @@
...
+	result.Fg = Erad_diff - (Rvec + Src);
+	result.Fg_abs_sum = 0.0;
+	for (int g = 0; g < nGroups_; ++g) {
+		if (tau[g] > 0.0) {
+			result.Fg_abs_sum += std::abs(result.Fg[g]);
+			result.F0 += cscale * Rvec[g];
+		}
+	}
+
+	// const auto d_fourpiboverc_d_t = ComputeThermalRadiationTempDerivativeMultiGroup(T_d, radBoundaries_g_copy);
```
markkrumholz: Again, I'm a bit confused by the commented-out line. If this is dead code, can it be removed to avoid confusion?

markkrumholz: OK, I see -- this line was moved out of this routine and into the calling routine, since d_fourpiboverc_d_t is now evaluated there. That's fine, and if you want to leave this line here as a reminder of what this variable contains, that's fine, too, but please make it clear in the comment that is what you're doing.

src/radiation/radiation_dust_system.hpp
```diff
@@ -0,0 +1,876 @@
...
+	// compute Jacobian elements
+	// I assume (kappaPVec / kappaEVec) is constant here. This is usually a reasonable assumption. Note that this assumption
+	// only affects the convergence rate of the Newton-Raphson iteration and does not affect the converged solution at all.
+
+	auto dEg_dT = kappaPoverE * d_fourpiboverc_d_t;
+
+	result.J00 = 1.0;
+	result.J0g.fillin(cscale);
+	const double d_Td_d_T = 3. / 2. - T_d / (2. * T_gas);
+	// const double coeff_n = dt * dustGasCoeff_local * num_den * num_den / cscale;
```
markkrumholz: Same as my last comment: it looks from this comment like you've commented out something that should be happening here and that coeff_n is defined here, when in fact it is passed in from the outside.

src/radiation/radiation_dust_system.hpp
```diff
@@ -0,0 +1,876 @@
...
+
+	result.J00 = 1.0;
+	result.J0g.fillin(cscale);
+	if (tau[nGroups_ - 1] <= 0.0) {
+		result.J0g[nGroups_ - 1] = LARGE;
+	} else {
+		result.J0g[nGroups_ - 1] -= PE_heating_energy_derivative * d_Eg_d_Rg[nGroups_ - 1];
+	}
+	const double d_Td_d_T = 3. / 2. - T_d / (2. * T_gas);
+	// const double coeff_n = dt * dustGasCoeff_local * num_den * num_den / cscale;
```
markkrumholz: Same comment about the coeff_n comment as in the analogous line in the previous function.

src/radiation/source_terms_multi_group.hpp
```diff
@@ -250,14 +94,64 @@ RadSystem<problem_t>::ComputeModelDependentKappaFAndDeltaTerms(double const T, d
...
+	result.Fg = Erad_diff - (Rvec + Src);
+	result.Fg_abs_sum = 0.0;
+	for (int g = 0; g < nGroups_; ++g) {
+		if (tau[g] > 0.0) {
+			result.Fg_abs_sum += std::abs(result.Fg[g]);
+			result.F0 += cscale * Rvec[g];
+		}
+	}
+
+	// const auto d_fourpiboverc_d_t = ComputeThermalRadiationTempDerivativeMultiGroup(T_d, radBoundaries_g_copy);
```
markkrumholz: Again, a somewhat confusing comment

markkrumholz:
Thanks, that structure is much cleaner. I left a few minor stylistic comments (mostly about your comments!), but once those are resolved I'm happy to approve.

src/radiation/radiation_dust_system.hpp
chongchonghe: This is the cgs values from Bate & Keto, Eq. 26. Previously, I commented it out and replaced it with 0.0 as the default. I'll remove those commentted-out code and move it the PE test problem file.

src/radiation/radiation_dust_system.hpp
chongchonghe: I will remove it since we don't need it anymore. It's evaluated before the function call.

src/radiation/radiation_dust_system.hpp
chongchonghe: I will remove since it's not used anymore.

src/radiation/radiation_dust_system.hpp
chongchonghe: Same.

src/radiation/source_terms_multi_group.hpp
chongchonghe: Same. Will remove it.

chongchonghe:
OK. All done.

markkrumholz (approved):

PR #773, opened 2024-10-11
Line cooling and cosmic ray heating

chongchonghe:
### Description
This pull request implements line cooling and cosmic ray heating. The key changes include:
- Added line cooling to the single-group and multi-group radiation solvers when `enable_dust_gas_thermal_coupling_model = true` . Line cooling is enabled by defining `DefineNetCoolingRate` and `DefineNetCoolingRateTempDerivative` in the problem file. 
- Implemented cosmic ray heating in all solvers, which is enabled by defining `DefineCosmicRayHeatingRate` in the problem file. 
- Added tests for the new line cooling and cosmic ray heating functionality.
	- RadLineCooling: test line cooling + CR heating + PE heating for single-group radiation system.
	- RadLineCoolingMG: test line cooling + CR heating + PE heating for multi-group radiation system, for both well-coupled and decoupled gas-dust systems.

This is the last feature I have finished implementing a while ago and I want to merge into the development branch before we move to Microphysics. In the next PR, I'll try to structure the code in the format of `actual_rhs()`, `actual_jac()`, and`backward_euler_integrator()` so that we can readily start to use Microphysics API. 

##### RadLineCooling

Initial conditions: uniform static gas with a temperature $T_0 = 1$ and $C_V = 1$; the radiation energy density is 0. The dust opacity is 0. The line cooling rate is $\Lambda_{\rm line} = C_{\Lambda} T$, where $C_{\Lambda} = 0.1$, and the CR heating rate is $\Gamma_{\rm cr} = 0.03$. As the gas cools down, its temperature follows the following ODE:

$$
C_V \frac{d T}{d t} = - C_{\Lambda} T + \Gamma_{\rm cr}
$$

The solution is
```math
\begin{gather}
T(t) = C_{\Lambda}^{-1} e^{-\frac{C_{\Lambda
   }}{C_V} t} \left(\Gamma
   _{\text{cr}} e^{\frac{C_{\Lambda}}{C_V} t}+ T_0
   C_{\Lambda }-\Gamma_{\text{cr}}\right) \\
   E_{\rm rad}(t) = -(C_V T(t) - C_V T_0 - \Gamma_{\rm cr} t)
\end{gather}
```

In the end of the test, the numerical results of $T(t)$ and $E_{\rm rad}(t)$ are checked against the exact analytical solution.

##### RadLineCoolingMG

The initial condition is the same as that of RadLineCooling, except
1. There are four radiation bins. The first bin is the line emission and the last bin is photoelectric heating (FUV). 
2. The initial radiation energy density is 1 in the last bin and 0 in all other bins. The opacity is 0 in all bins.
3. PE heating is enabled and $\Gamma_{\rm pe} = 0.02 E_{\rm FUV} = 0.02$. 
The ODE is

$$
C_V \frac{d T}{d t} = - C_{\Lambda} T + \Gamma_{\rm cr} + \Gamma_{\rm pe}
$$

and the solution is the same as the previous one with two differences: (1) $\Gamma_{\rm cr}$ is replaced with $\Gamma_{\rm cr} + \Gamma_{\rm pe}$; (2) $E_{\rm rad}$ represents the energy density of the first bin. 

This test is run twice, one with well-coupled gas-dust model and one with decoupled gas-dust model.

### Related issues

### Checklist

### DIFF
--- a/src/math/interpolate.cpp
+++ b/src/math/interpolate.cpp
@@ -108,6 +108,9 @@ void interpolate_arrays(double *x, double *y, int len, double *arr_x, double *ar
 	/* Note: arr_x must be sorted in ascending order,
 		and arr_len must be >= 3. */
 
+	// check if x is within the range of arr_x
+	assert(x[0] >= arr_x[0] && x[len - 1] <= arr_x[arr_len - 1]);
+
 	int64_t j = 0;
 	for (int i = 0; i < len; i++) {
 		j = binary_search_with_guess(x[i], arr_x, arr_len, j);
--- a/src/math/interpolate.cpp
+++ b/src/math/interpolate.cpp
@@ -1,8 +1,8 @@
 #include <cassert>
 #include <cmath>
 
-#include "math/interpolate.hpp"
 #include "AMReX_BLassert.H"
+#include "math/interpolate.hpp"
 
 #define LIKELY_IN_CACHE_SIZE 8
 
--- a/src/math/interpolate.cpp
+++ b/src/math/interpolate.cpp
@@ -2,6 +2,7 @@
 #include <cmath>
 
 #include "math/interpolate.hpp"
+#include "AMReX_BLassert.H"
 
 #define LIKELY_IN_CACHE_SIZE 8
 
@@ -109,7 +110,7 @@ void interpolate_arrays(double *x, double *y, int len, double *arr_x, double *ar
 		and arr_len must be >= 3. */
 
 	// check if x is within the range of arr_x
-	assert(x[0] >= arr_x[0] && x[len - 1] <= arr_x[arr_len - 1]);
+	AMREX_ASSERT(x[0] >= arr_x[0] && x[len - 1] <= arr_x[arr_len - 1]);
 
 	int64_t j = 0;
 	for (int i = 0; i < len; i++) {
@@ -127,7 +128,7 @@ void interpolate_arrays(double *x, double *y, int len, double *arr_x, double *ar
 			const double slope = (arr_y[j + 1] - arr_y[j]) / (arr_x[j + 1] - arr_x[j]);
 			y[i] = slope * (x[i] - arr_x[j]) + arr_y[j];
 		}
-		assert(!std::isnan(y[i]));
+		AMREX_ASSERT(!std::isnan(y[i]));
 	}
 }
 
@@ -153,6 +154,6 @@ double interpolate_value(double x, double const *arr_x, double const *arr_y, int
 		const double slope = (arr_y[j + 1] - arr_y[j]) / (arr_x[j + 1] - arr_x[j]);
 		y = slope * (x - arr_x[j]) + arr_y[j];
 	}
-	assert(!std::isnan(y));
+	AMREX_ASSERT(!std::isnan(y));
 	return y;
 }
--- a/src/problems/RadLineCooling/test_rad_line_cooling.cpp
+++ b/src/problems/RadLineCooling/test_rad_line_cooling.cpp
@@ -22,6 +22,8 @@ constexpr double CR_heating_rate = 1.0;
 constexpr double line_cooling_rate = CR_heating_rate;
 constexpr amrex::GpuArray<double, 5> rad_boundaries_ = {1.00000000e-03, 1.77827941e-02, 3.16227766e-01, 5.62341325e+00, 1.00000000e+02};
 
+const double cooling_rate = 1.0e-1;
+
 constexpr double c = 1.0;
 constexpr double chat = c;
 constexpr double v0 = 0.0;
@@ -38,7 +40,7 @@ constexpr double T_equilibrium = 0.768032502191;
 constexpr double Erad_bar = a_rad * T0 * T0 * T0 * T0;
 constexpr double erad_floor = a_rad * 1e-20;
 
-constexpr double max_time = 1.0;
+const double max_time = 1.0;
 
 template <> struct quokka::EOS_Traits<PulseProblem> {
 	static constexpr double mean_molecular_weight = mu;
@@ -77,10 +79,20 @@ template <> struct ISM_Traits<PulseProblem> {
 };
 
 template <>
-AMREX_GPU_HOST_DEVICE auto RadSystem<PulseProblem>::DefineNetCoolingRate(amrex::Real const /*temperature*/, amrex::Real const /*num_density*/) -> quokka::valarray<double, nGroups_>
+AMREX_GPU_HOST_DEVICE auto RadSystem<PulseProblem>::DefineNetCoolingRate(amrex::Real const temperature, amrex::Real const /*num_density*/) -> quokka::valarray<double, nGroups_>
 {
 	quokka::valarray<double, nGroups_> cooling{};
 	cooling.fillin(0.0);
+	cooling[0] = cooling_rate * temperature;
+	return cooling;
+}
+
+template <>
+AMREX_GPU_HOST_DEVICE auto RadSystem<PulseProblem>::DefineNetCoolingRateTempDerivative(amrex::Real const temperature, amrex::Real const /*num_density*/) -> quokka::valarray<double, nGroups_>
+{
+	quokka::valarray<double, nGroups_> cooling{};
+	cooling.fillin(0.0);
+	cooling[0] = cooling_rate;
 	return cooling;
 }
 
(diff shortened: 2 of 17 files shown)


chongchonghe:
@markkrumholz This PR is ready for your review.

markkrumholz:
> @markkrumholz This PR is ready for your review.

Will get to it ASAP, but it may be a few days.

src/math/interpolate.cpp
```diff
@@ -108,6 +108,9 @@ void interpolate_arrays(double *x, double *y, int len, double *arr_x, double *ar
 	/* Note: arr_x must be sorted in ascending order,
 		and arr_len must be >= 3. */
 
+	// check if x is within the range of arr_x
```
markkrumholz, line 112: You should change this to one of the AMReX assert macros so that it functions properly on GPU or GPU, and in parallel. See https://amrex-codes.github.io/amrex/doxygen/AMReX__BLassert_8H.html

src/problems/RadLineCooling/test_rad_line_cooling.cpp
```diff
@@ -0,0 +1,295 @@
...
+constexpr double mu = 1.5; // mean molecular weight; so that C_V = 1.0
+constexpr double C_V = 1.0;
+constexpr double k_B = 1.0;
+
+constexpr double nu_unit = 1.0;
+constexpr double erad_floor = a_rad * 1e-20;
+
+const double max_time = 10.0;
+
+template <> struct SimulationData<PulseProblem> {
```
markkrumholz: Maybe call this something other than PulseProblem, since it appears to be unrelated to the actual radiation pulse problem described in our papers?

src/problems/RadLineCooling/test_rad_line_cooling.cpp
```diff
@@ -0,0 +1,295 @@
...
+	sim.plotfileInterval_ = -1;
+
+	// initialize
+	sim.setInitialConditions();
+
+	// evolve
+	sim.evolve();
+
+	// read output variables
+	auto [_, values] = fextract(sim.state_new_cc_[0], sim.Geom(0), 0, 0.0);
```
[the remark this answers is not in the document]
markkrumholz: Please either fix or NOLINT this.

src/problems/RadLineCoolingMG/test_rad_line_cooling_MG.cpp
```diff
@@ -0,0 +1,343 @@
...
+constexpr double Erad_floor_ = a_rad * 1e-20;
+constexpr double Erad_FUV = Erad_bar; // = 1.0
+
+const double max_time = 10.0;
+const int line_index = 0; // last group
+const double cooling_rate = 0.1;
+const double CR_heating_rate = 0.03;
+const double PE_rate = 0.02;
+
+template <> struct SimulationData<PulseProblem> {
```
markkrumholz: Same comment about the name of this test as with the non-multigroup version -- call it something more usefully descriptive.

src/problems/RadLineCoolingMG/test_rad_line_cooling_MG.cpp
```diff
@@ -0,0 +1,343 @@
...
+		const amrex::Real Egas_i = RadSystem<PulseProblem>::ComputeEintFromEgas(rho, x1GasMom, x2GasMom, x3GasMom, Etot_i);
+		userData_.Tgas_vec_.push_back(quokka::EOS<PulseProblem>::ComputeTgasFromEint(rho, Egas_i));
+		const double Erad_line_i = values.at(RadSystem<PulseProblem>::radEnergy_index + Physics_NumVars::numRadVars * line_index)[0];
+		userData_.Erad_line_vec_.push_back(Erad_line_i);
+	}
+}
+
+auto problem_main() -> int
+{
+	// This problem is a *linear* radiation diffusion problem, i.e.
```
markkrumholz: I think this comment does not apply to this problem, and is left over from the problem you cloned. Please correct or delete.

src/problems/RadLineCoolingMG/test_rad_line_cooling_MG.cpp
```diff
@@ -0,0 +1,343 @@
...
+	// initialize
+	sim.setInitialConditions();
+
+	// evolve
+	sim.evolve();
+
+	const bool is_coupled = sim.dustGasInteractionCoeff_ > 1.0;
+
+	// read output variables
+	auto [_, values] = fextract(sim.state_new_cc_[0], sim.Geom(0), 0, 0.0);
```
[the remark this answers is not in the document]
markkrumholz: Again, please fix or NOLINT

src/radiation/radiation_system.hpp
```diff
@@ -138,6 +139,10 @@ template <typename problem_t> struct FluxUpdateResult {
 	amrex::GpuArray<amrex::GpuArray<amrex::Real, Physics_Traits<problem_t>::nGroups>, 3> Frad; // radiation flux
 };
 
+// template <typename problem_t> struct SingleVariableState {
```
markkrumholz: Do we need to keep this commented-out code, or can it be deleted?

markkrumholz (commented):
I have left some comments on individual bits of code that you can look at, but there is also an overall design issue, which is the way you're using number density. The problem is that you don't specify the number density of what -- number of H nuclei, number of atomic H, number of H2, etc. -- and so this code becomes ambiguous and potentially incorrect as soon as we have chemistry. I think any time we need a number density, we need to be extremely careful to specify number density of which species, and we need to encapsulate the process of deriving the number density of that species from the density and mass scalars inside a function that can be swapped in or out depending on the chemistry network. Thus for example the current code that does things like `num_density = rho / mean_molecular_weight` could be encapsulated in a function called `derive_num_density_H0` which derives the number density of neutral hydrogen from the density and mass scalar array. If the chemistry trait you have then selects the network to be none, then the current implementation is turned on, but if another chemistry network is selected then this handled differently, depending on the species in that network.

What we don't want to do, though, is hardwire in a solution that works only when chemistry is off, and then have it break in subtle and hard-to-detect ways if anyone tries to use it with chemistry turned on.

chongchonghe:
I defined a function, `ComputeNumberDensityAtomicH(double rho, amrex::GpuArray<Real, nmscalars_> const &massScalars)`, to calculate atomic hydrogen number density. I use atomic hydrogen instead of neutral hydrogen because that's what's used in Bate & Keto. The cosmic ray heating might relate to neutral hydrogen density, but line cooling and photoelectric heating is related to atomic hydrogen (plus metal abundance). So, I'll use 'atomic hydrogen' number density to avoid confusion.

markkrumholz:
> I defined a function, `ComputeNumberDensityAtomicH(double rho, amrex::GpuArray<Real, nmscalars_> const &massScalars)`, to calculate atomic hydrogen number density. I use atomic hydrogen instead of neutral hydrogen because that's what's used in Bate & Keto. The cosmic ray heating might relate to neutral hydrogen density, but line cooling and photoelectric heating is related to atomic hydrogen (plus metal abundance). So, I'll use 'atomic hydrogen' number density to avoid confusion.

I'm a bit confused about the distinction you're making between "atomic hydrogen" and "neutral hydrogen", since usually they are used as synonyms. In general in the astrochemistry literature the notation is that n_H refers to the number density of H nuclei regardless of their chemical state, while n_H+, n_H0, and n_H2 refer to the number density of ionised hydrogen, neutral atomic hydrogen, and hydrogen molecules, so that we have the relationship n_H = n_H+ + n_H0 + 2 n_H2. I think it would be good to follow this notational convention in the code, so that we could have routines like ComputeNumberDensityH0, ComputeNumberDensityH, ComputeNumberDensityHp (for H+), etc. This will make the code more readable and less susceptible to mistakes.

chongchonghe:
> > I defined a function, `ComputeNumberDensityAtomicH(double rho, amrex::GpuArray<Real, nmscalars_> const &massScalars)`, to calculate atomic hydrogen number density. I use atomic hydrogen instead of neutral hydrogen because that's what's used in Bate & Keto. The cosmic ray heating might relate to neutral hydrogen density, but line cooling and photoelectric heating is related to atomic hydrogen (plus metal abundance). So, I'll use 'atomic hydrogen' number density to avoid confusion.
> 
> I'm a bit confused about the distinction you're making between "atomic hydrogen" and "neutral hydrogen", since usually they are used as synonyms. In general in the astrochemistry literature the notation is that n_H refers to the number density of H nuclei regardless of their chemical state, while n_H+, n_H0, and n_H2 refer to the number density of ionised hydrogen, neutral atomic hydrogen, and hydrogen molecules, so that we have the relationship n_H = n_H+ + n_H0 + 2 n_H2. I think it would be good to follow this notational convention in the code, so that we could have routines like ComputeNumberDensityH0, ComputeNumberDensityH, ComputeNumberDensityHp (for H+), etc. This will make the code more readable and less susceptible to mistakes.

OK, I see. Then, what I meant is density of H nuclei, n_H. I've changed it to `ComputeNumberDensityH`.

markkrumholz (approved):

PR #774, opened 2024-10-13
fix 'pure virtual method called' error

chongchonghe:
### Description
Move `DeviceVector` inside `setInitialConditionsOnGrid` to avoid the `pure virtual method called` error, first seen in #761 . 

This branch is based on dependabot/submodules/extern/amrex-6d9c25b . 

### Related issues
Fix issues in PR #761 

### Checklist

### DIFF
--- a/src/problems/RadTube/test_radiation_tube.cpp
+++ b/src/problems/RadTube/test_radiation_tube.cpp
@@ -97,10 +97,10 @@ template <> void QuokkaSimulation<TubeProblem>::setInitialConditionsOnGrid(quokk
 
 	// declare global variables
 	// initial conditions read from file
-	amrex::Gpu::HostVector<double> x_arr;
-	amrex::Gpu::HostVector<double> rho_arr;
-	amrex::Gpu::HostVector<double> Pgas_arr;
-	amrex::Gpu::HostVector<double> Erad_arr;
+	std::vector<double> x_arr;
+	std::vector<double> rho_arr;
+	std::vector<double> Pgas_arr;
+	std::vector<double> Erad_arr;
 
 	for (std::string line; std::getline(fstream, line);) {
 		std::istringstream iss(line);
--- a/src/problems/RadTube/test_radiation_tube.cpp
+++ b/src/problems/RadTube/test_radiation_tube.cpp
@@ -86,19 +86,7 @@ RadSystem<TubeProblem>::DefineOpacityExponentsAndLowerValues(amrex::GpuArray<dou
 	return exponents_and_values;
 }
 
-// declare global variables
-// initial conditions read from file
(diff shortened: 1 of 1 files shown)


chongchonghe:
@BenWibking This solves the problem, although I don't understand exactly why. It seems like the cause of the error is because `amrex::Gpu::DeviceVector` is declared globally and accessed inside the CPU function `preCalculateInitialCondition`. I moved everything in `preCalculateInitialCondition` to `setInitialConditionsOnGrid` and the issue is gone.

BenWibking (dismissed):

BenWibking:
> @BenWibking This solves the problem, although I don't understand exactly why. It seems like the cause of the error is because `amrex::Gpu::DeviceVector` is declared globally and accessed inside the CPU function `preCalculateInitialCondition`. I moved everything in `preCalculateInitialCondition` to `setInitialConditionsOnGrid` and the issue is gone.

Ok, that makes sense. The issue is that global objects are destroyed after `main()` ends, so it tries to deallocate the device memory (or maybe perform other operations) after `amrex::Finalize()` has been called. However, the GPU is not accessible after calling `amrex::Finalize()`. I'm not actually sure how it worked before.

We don't want to re-read the file for each grid, though, so we still need to do this inside `preCalculateInitialConditions`. You can move the `DeviceVectors` to `SimulationData<TubeProblem>` instead of global variables.

BenWibking:
> > @BenWibking This solves the problem, although I don't understand exactly why. It seems like the cause of the error is because `amrex::Gpu::DeviceVector` is declared globally and accessed inside the CPU function `preCalculateInitialCondition`. I moved everything in `preCalculateInitialCondition` to `setInitialConditionsOnGrid` and the issue is gone.
> 
> Ok, that makes sense. The issue is that global objects are destroyed after `main()` ends, so it tries to deallocate the device memory (or maybe perform other operations) after `amrex::Finalize()` has been called. However, the GPU is not accessible after calling `amrex::Finalize()`. I'm not actually sure how it worked before.
> 
> We don't want to re-read the file for each grid, though, so we still need to do this inside `preCalculateInitialConditions`. You can move the `DeviceVectors` to `SimulationData<TubeProblem>` instead of global variables.

BenWibking:
You can move the variables into `SimulationData` in the same way you did so here:
https://github.com/quokka-astro/quokka/blob/5f84397ede8fddc087985c8fe701ad6afe24c6bb/src/problems/RadDustMG/test_rad_dust_MG.cpp#L35

The only difference is that you need a `DeviceVector` instead of a `std::vector`.

chongchonghe:
@BenWibking @markkrumholz Should be ready to merge.

BenWibking (approved):

PR #804, opened 2024-12-02
fix openPMD plotfile output

### LINKED ISSUE #742: openPMD output is broken
**Describe the bug**
OpenPMD output doesn't work with the latest development commit. Discovered when testing https://github.com/quokka-astro/quokka/pull/741.

**To Reproduce**
1. Checkout commit 60a37f06084907f605f18b339d3dba179214eb80.
2. Run the HydroBlast3D problem
3. See error:
```
ubuntu@0301436d6d84:/workspaces/quokka/tests$ ../build/src/problems/HydroBlast3D/test_hydro3d_blast blast_32.in 
Initializing AMReX (24.09)...
AMReX (24.09) initialized
amrex::Abort::0:: (-2,-2,-2,6) is out of bound (0:31,0:31,0:31,0:8) !!!
SIGABRT
```
Backtrace:
```

 0: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x6fa4c0) [0xaaaad180a4c0]
    amrex::BLBackTrace::print_backtrace_info(_IO_FILE*) at /workspaces/quokka/extern/amrex/Src/Base/AMReX_BLBackTrace.cpp:203 (discriminator 1)

 1: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x6f9dd8) [0xaaaad1809dd8]
    amrex::BLBackTrace::handler(int) at /workspaces/quokka/extern/amrex/Src/Base/AMReX_BLBackTrace.cpp:101

 2: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x6b9b1c) [0xaaaad17c9b1c]
    amrex::ParallelDescriptor::Abort(int, bool) at /workspaces/quokka/extern/amrex/Src/Base/AMReX_ParallelDescriptor.cpp:219

 3: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x62d6a8) [0xaaaad173d6a8]
    amrex::Error_host(char const*, char const*) at /workspaces/quokka/extern/amrex/Src/Base/AMReX.cpp:252

 4: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x62d5c0) [0xaaaad173d5c0]
    amrex::Abort(char const*) at /workspaces/quokka/extern/amrex/Src/Base/AMReX.H:160
 (inlined by) amrex::Abort(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) at /workspaces/quokka/extern/amrex/Src/Base/AMReX.cpp:223

 5: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x36196c) [0xaaaad147196c]
    amrex::Array4<double>::index_assert(int, int, int, int) const at /workspaces/quokka/extern/amrex/Src/Base/AMReX_Array4.H:279 (discriminator 53)

 6: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x397750) [0xaaaad14a7750]
    double& amrex::Array4<double>::operator()<double, 0>(int, int, int, int) const at /workspaces/quokka/extern/amrex/Src/Base/AMReX_Array4.H:144
 (inlined by) AMRSimulation<SedovProblem>::AverageFCToCC(amrex::MultiFab&, amrex::MultiFab const&, int, int, int, int, int) const::{lambda(int, int, int, int)#1}::operator()(int, int, int, int) const at /workspaces/quokka/src/simulation.hpp:2049

 7: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x42d394) [0xaaaad153d394]
    std::enable_if<amrex::IsFabArray<amrex::MultiFab, void>::value, void>::type amrex::experimental::detail::ParallelFor<amrex::MultiFab, AMRSimulation<SedovProblem>::AverageFCToCC(amrex::MultiFab&, amrex::MultiFab const&, int, int, int, int, int) const::{lambda(int, int, int, int)#1}>(amrex::MultiFab const&, amrex::IntVectND<3> const&, amrex::IntVectND<3> const&, bool, AMRSimulation<SedovProblem>::AverageFCToCC(amrex::MultiFab&, amrex::MultiFab const&, int, int, int, int, int) const::{lambda(int, int, int, int)#1} const&) at /workspaces/quokka/extern/amrex/Src/Base/AMReX_MFParallelForC.H:26 (discriminator 3)

 8: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x3be0f8) [0xaaaad14ce0f8]
    std::enable_if<amrex::IsFabArray<amrex::MultiFab, void>::value, void>::type amrex::experimental::ParallelFor<amrex::MultiFab, AMRSimulation<SedovProblem>::AverageFCToCC(amrex::MultiFab&, amrex::MultiFab const&, int, int, int, int, int) const::{lambda(int, int, int, int)#1}>(amrex::MultiFab const&, amrex::IntVectND<3> const&, AMRSimulation<SedovProblem>::AverageFCToCC(amrex::MultiFab&, amrex::MultiFab const&, int, int, int, int, int) const::{lambda(int, int, int, int)#1}&&) at /workspaces/quokka/extern/amrex/Src/Base/AMReX_MFParallelFor.H:98

 9: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x397934) [0xaaaad14a7934]
    AMRSimulation<SedovProblem>::AverageFCToCC(amrex::MultiFab&, amrex::MultiFab const&, int, int, int, int, int) const at /workspaces/quokka/src/simulation.hpp:2053

10: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x3802b0) [0xaaaad14902b0]
    AMRSimulation<SedovProblem>::PlotFileMFAtLevel(int, int) at /workspaces/quokka/src/simulation.hpp:2097

11: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x37e514) [0xaaaad148e514]
    AMRSimulation<SedovProblem>::PlotFileMF(int) at /workspaces/quokka/src/simulation.hpp:2115 (discriminator 2)

12: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x36b3b0) [0xaaaad147b3b0]
    AMRSimulation<SedovProblem>::WritePlotFile() at /workspaces/quokka/src/simulation.hpp:2310

13: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x35d468) [0xaaaad146d468]
    AMRSimulation<SedovProblem>::setInitialConditions() at /workspaces/quokka/src/simulation.hpp:686

14: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x320d0c) [0xaaaad1430d0c]
    problem_main() at /workspaces/quokka/src/problems/HydroBlast3D/test_hydro3d_blast.cpp:267

15: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x5b066c) [0xaaaad16c066c]
    main at /workspaces/quokka/src/main.cpp:54 (discriminator 1)

16: /lib/aarch64-linux-gnu/libc.so.6(+0x284c4) [0xffffb8c684c4]

17: /lib/aarch64-linux-gnu/libc.so.6(__libc_start_main+0x98) [0xffffb8c68598]

18: ../build/src/problems/HydroBlast3D/test_hydro3d_blast(+0x31cbf0) [0xaaaad142cbf0]
    ?? ??:0


===== TinyProfilers ======
AMRSimulation::setInitialConditions()
AMRSimulation::WritePlotFile()
```

**Additional context**
We don't currently have a CI test for openPMD output.

BenWibking:
### Description
Fixes a segmentation fault during openPMD plotfile output when face-centered variables are enabled.

There was an implicit assumption in `AMRSimulation<problem_t>::AverageFCToCC()` that the number of face-centered ghost zones was the same as the number of cell-centered ghost zones. This PR allows the number of ghosts to differ.

I manually tested that this works inside the CUDA dev container by compiling Quokka with `-DQUOKKA_OPENPMD=ON`,  running the HydroBlast3D problem, and then examining the output files with [openPMD-viewer](https://github.com/openPMD/openPMD-viewer).

### Related issues
Fixes https://github.com/quokka-astro/quokka/issues/742.

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -2230,7 +2230,7 @@ template <typename problem_t> void AMRSimulation<problem_t>::doDiagnostics()
 	if (computeVars) {
 		for (int lev{0}; lev <= finestLevel(); ++lev) {
 			diagMFVec[lev] = std::make_unique<amrex::MultiFab>(grids[lev], dmap[lev], m_diagVars.size(), 1);
-			amrex::MultiFab const mf = PlotFileMFAtLevel(lev, nghost_cc_);
+			amrex::MultiFab const mf = PlotFileMFAtLevel(lev, std::min(nghost_cc_, nghost_fc_));
 			auto const varnames = GetPlotfileVarNames();
 
 			for (int v{0}; v < m_diagVars.size(); ++v) {
@@ -2291,7 +2291,8 @@ template <typename problem_t> void AMRSimulation<problem_t>::RenderAscent()
 	BL_PROFILE("AMRSimulation::RenderAscent()");
 
 	// combine multifabs
-	amrex::Vector<amrex::MultiFab> mf = PlotFileMF(nghost_cc_);
+	const int included_ghosts = std::min(nghost_cc_, nghost_fc_);
+	amrex::Vector<amrex::MultiFab> mf = PlotFileMF(included_ghosts);
 	amrex::Vector<const amrex::MultiFab *> mf_ptr = amrex::GetVecOfConstPtrs(mf);
 	amrex::Vector<std::string> varnames;
 	varnames.insert(varnames.end(), componentNames_cc_.begin(), componentNames_cc_.end());
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -2091,8 +2091,10 @@ void AMRSimulation<problem_t>::AverageFCToCC(amrex::MultiFab &mf_cc, const amrex
 	// iterate over the domain
 	auto const &state_cc = mf_cc.arrays();
 	auto const &state_fc = mf_fc.const_arrays();
-	int const ng = mf_cc.nGrow();
-	amrex::ParallelFor(mf_cc, amrex::IntVect(AMREX_D_DECL(ng, ng, ng)), [=] AMREX_GPU_DEVICE(int boxidx, int i, int j, int k) {
+	int const ng_cc = mf_cc.nGrow();
+	int const ng_fc = mf_fc.nGrow();
+	AMREX_ALWAYS_ASSERT(ng_cc <= ng_fc); // if this is false, we can't fill the ghost cells!
+	amrex::ParallelFor(mf_cc, amrex::IntVect(AMREX_D_DECL(ng_cc, ng_cc, ng_cc)), [=] AMREX_GPU_DEVICE(int boxidx, int i, int j, int k) {
 		for (int icomp = 0; icomp < srccomp_total; ++icomp) {
 			state_cc[boxidx](i, j, k, dstcomp_start + icomp) =
 			    0.5 * (state_fc[boxidx](i, j, k, srccomp_start + icomp) + state_fc[boxidx](i + di, j + dj, k + dk, srccomp_start + icomp));
@@ -2352,7 +2354,7 @@ template <typename problem_t> void AMRSimulation<problem_t>::WritePlotFile()
 #ifdef QUOKKA_USE_OPENPMD
 	int included_ghosts = 0;
 #else
-	int included_ghosts = nghost_cc_;
+	int included_ghosts = std::min(nghost_cc_, nghost_fc_);
 #endif
 	amrex::Vector<amrex::MultiFab> mf = PlotFileMF(included_ghosts);
 	amrex::Vector<const amrex::MultiFab *> mf_ptr = amrex::GetVecOfConstPtrs(mf);
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -216,8 +216,7 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 	void computeTimestep();
 	auto computeTimestepAtLevel(int lev) -> amrex::Real;
 
-	void AverageFCToCC(amrex::MultiFab &mf_cc, const amrex::MultiFab &mf_fc, int idim, int dstcomp_start, int srccomp_start, int srccomp_total,
-			   int nGrow) const;
+	void AverageFCToCC(amrex::MultiFab &mf_cc, const amrex::MultiFab &mf_fc, int idim, int dstcomp_start, int srccomp_start, int srccomp_total) const;
 
 	virtual void computeMaxSignalLocal(int level) = 0;
 	virtual auto computeExtraPhysicsTimestep(int lev) -> amrex::Real = 0;
@@ -2077,7 +2076,7 @@ template <typename problem_t> auto AMRSimulation<problem_t>::CustomPlotFileName(
 
 template <typename problem_t>
 void AMRSimulation<problem_t>::AverageFCToCC(amrex::MultiFab &mf_cc, const amrex::MultiFab &mf_fc, int idim, int dstcomp_start, int srccomp_start,
-					     int srccomp_total, int nGrow) const
+					     int srccomp_total) const
 {
 	int di = 0;
 	int dj = 0;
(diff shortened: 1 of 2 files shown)


BenWibking:
Out-of-bounds array access: https://app.circleci.com/pipelines/circleci/TJttRfCZuo9qQMru3cFqzy/SoCatADYnqm7LfdmA3zswt/1080/workflows/79cefde6-56f9-4d61-907d-d5d4c15e7315/jobs/2338

To be investigated.

BenWibking:
@markkrumholz Not urgent, but this is now ready for review.

markkrumholz (commented):
There seems to be a slight inconsistency in the design: as written now, the code will always fail if run for a case with more cell-centered than face-centered ghost zones, due to the AMREX_ALWAYS_ASSERT added to the routine that does the averaging to centers. However, both doDiagnostics and renderAscent seem to support having more cell-centered than face-centered ghost zones, and gracefully use a std::min to handle that. This is confusing. I think we should put similar assertions in every routine if we don't want to support having nghost_cc > nghost_fc, or we should edit the averaging to center routine to also support this case. One possible way to do this would be to have the routine take an optional argument that provides a fill value to add to ghost zones that cannot be filled because the corresponding face-centered cells do not exist. I don't have strong feelings about which of these options we pick, but I think the design should be consistent as to whether we do or don't support the case nghost_cc > nghost_fc.

BenWibking:
> There seems to be a slight inconsistency in the design: as written now, the code will always fail if run for a case with more cell-centered than face-centered ghost zones, due to the AMREX_ALWAYS_ASSERT added to the routine that does the averaging to centers. However, both doDiagnostics and renderAscent seem to support having more cell-centered than face-centered ghost zones, and gracefully use a std::min to handle that. This is confusing. I think we should put similar assertions in every routine if we don't want to support having nghost_cc > nghost_fc, or we should edit the averaging to center routine to also support this case. One possible way to do this would be to have the routine take an optional argument that provides a fill value to add to ghost zones that cannot be filled because the corresponding face-centered cells do not exist. I don't have strong feelings about which of these options we pick, but I think the design should be consistent as to whether we do or don't support the case nghost_cc > nghost_fc.

In the current code, if I am remembering correctly, averaging to centers is only done for outputs. Everywhere an output needs to do that, it provides an appropriate number of ghost zones to use (that may be less than nghost_cc). If we are not doing MHD, then we don't need to carry around extra ghost faces we don't need, and we only need to average them to centers for outputs. I don't see an issue with this design in principle.

I don't think it's a good idea to allow averaging to centers with a fill value. That is almost certain to produce bugs in the future. The assertion just forces anyone who uses the averaging function to specify something sensible upfront. In any case, I don't know where we would need to do this.

BenWibking:
Actually, ghost cells in outputs are only really needed for Ascent, due to the way it uses them to compute surfaces and streamlines in parallel without doing extra communication itself. For all other outputs, we could just get rid of the ghost cells. They are not supported in OpenPMD outputs at all, and in AMReX plotfiles I don't think they are used for anything.

Edit: I need to double check whether they are used by any of the diagnostics.

BenWibking:
Just to clarify- the case where nghost_cc > nghost_fc is tested as part of the SphericalCollapse test suite problem, since tracer particles without MHD enabled set nghost_fc = 2, while nghost_cc is always 4.

markkrumholz:
So my confusion is these lines (2092 - 2096) of `Simulation.hpp`:

`	auto const &state_cc = mf_cc.arrays();
	auto const &state_fc = mf_fc.const_arrays();
	amrex::ParallelFor(mf_cc, amrex::IntVect(AMREX_D_DECL(nGrow, nGrow, nGrow)), [=] AMREX_GPU_DEVICE(int boxidx, int i, int j, int k) {
	int const ng_cc = mf_cc.nGrow();
	int const ng_fc = mf_fc.nGrow();
	AMREX_ALWAYS_ASSERT(ng_cc <= ng_fc); // if this is false, we can't fill the ghost cells!`
	
It seems to me like you are grabbing the state vectors, extracting their number of ghost zones, and then throwing an assertion that will fail anytime the cell-centered state variables have more ghost zones than the face-centered ones. So it seems like this should fail any time `AverageFCToCC` is called, which it always is any time you're writing a plot file, and it seems like it should fail independent of how many cells you're trying to write to that plot file. Am I misunderstanding what this code does?

src/simulation.hpp
```diff
@@ -2094,7 +2091,10 @@ void AMRSimulation<problem_t>::AverageFCToCC(amrex::MultiFab &mf_cc, const amrex
 	// iterate over the domain
 	auto const &state_cc = mf_cc.arrays();
```
BenWibking, line 2092: This is not the global state multifab (that is `state_new_cc_` and `state_old_cc_`). This is just whatever multifab is passed in as an argument to this function as `mf_cc`.

BenWibking: @markkrumholz This variable is confusingly named. I think this is the source of the confusion.

markkrumholz: I see. OK, thanks for clarifying. In that case I'm happy to approve the PR.

markkrumholz (approved):

PR #806, opened 2024-12-06
[WIP] compile with C++20 support

BenWibking:
### Description
This PR changes:
* captures `this` explicitly (only used by CICDeposit functor)
* uses standard library function `std::format` instead of third-party library function `fmt::format`
* removes `fmt` submodule

C++20 benefits:
* can use `consteval` qualifier for compile-time functions
* can use `constinit` qualifier for global variables
* requires CUDA 12.0 or newer

References:
* https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0806r2.html
* https://en.cppreference.com/w/cpp/utility/format/format
* https://en.cppreference.com/w/cpp/language/consteval
* https://en.cppreference.com/w/cpp/language/constinit

### Related issues
N/A

### Checklist


BenWibking:
https://github.com/quokka-astro/quokka/blob/06470346e6b642008582dd6612e14d558c106d8e/src/particles/CICParticles.hpp#L41

PR #810, opened 2024-12-09
Add radiation-driven shell to regression tests

### LINKED ISSUE #743: add radhydro problem to nightly regression tests
**Describe the proposal**
We currently only test hydro + optically-thin cooling in the regression test suite (https://quokka-astro.github.io/regression-testing-results/).

We should add a radhydro problem to ensure that the output and performance does not change unexpctedly.

Leave regression tests as-is.

**Additional context**
We have many test problems with exact solutions for radiation, but we don't currently know if there are perturbations smaller than the error thresholds.

We also don't know whether there are performance improvements/regressions over time. The regression test suite automatically tells us this, e.g.: https://quokka-astro.github.io/regression-testing-results/RandomBlast-GPU-timings.html

BenWibking:
### Description
Adds this problem to the nightly regression test suite. There was also a bug in the initial conditions in the radiation flux in converting from spherical to Cartesian flux components (now fixed).

At late times, the simulation still has a slightly discrepancy w.r.t. the semi-analytic solution (similar to that shown in the original code paper). This is most likely due to initial shell thickness being quite significant. (There is not an easy solution to this, since there is no steady-state solution that satisfies the M1 closure for sufficiently thin shells.)

shell_velocity

https://github.com/user-attachments/assets/8851d7cd-9d92-4de6-ac7e-ab5fef035af5

### Related issues
Closes https://github.com/quokka-astro/quokka/issues/743.

### Checklist


BenWibking:
Hmm, it looks like there is a global variable that is misbehaving:
```
pure virtual method called
terminate called without an active exception
[moth:2945339] *** Process received signal ***
[moth:2945339] Signal: Aborted (6)
[moth:2945339] Signal code:  (-753879392)
```
Will fix.

BenWibking:
@markkrumholz @chongchonghe I would appreciate a review of this from both of you, if you have time.

markkrumholz:
Code changes look fine. As to why we're not exactly preserving the analytic solution, one thing I might check is whether the shell is really very thin -- the analytic solution assumes an infinitely thin shell, and if that assumption is not fully satisfied, then we shouldn't expect to match the solution exactly. It may be necessary to tweak test parameters to reduce shell thickness, for example by reducing the sound speed so that pressure support becomes less important.

markkrumholz (dismissed):

chongchonghe:
The RadShell test on GPU failed. Even if you fix the bug, I speculate it will take too long to run. I would recommend reducing `max_timesteps`, which is currently set to 5000. With $128^3 = 2M$ cells, suppose a performance of 10 Mzone/s/GPU, that implies 5 steps/s. It will take 1000 seconds to run 5000 timesteps. 

EDIT: the run time is also determined by `stopTime_`, which is set to `0.125 * t0_hydro`. I guess that is bigger than 5000 steps?

BenWibking:
@chongchonghe The error is this:
```
Coarse STEP 3203 at t = 7.844071263e+12 (81.33839288%) starts ...
	>> Using global timestep on this coarse step (estimated work ratio: 1).
[Level 0 step 3203] ADVANCE with time = 7.8440712630e+12 dt = 5.5257787970e+07
	>> CFL violation detected on level 0 with dt_lev = 55257787.97 and dt_cfl = 48548760.42
	   max_signal = 993202495.4
	>> WARNING: Hydro advance failed on level 0
	>> Re-trying hydro advance at level 0 with reduced timestep (nsubsteps = 2, dt_new = 27628893.98)
	Radiation substeps: 1	dt: 55257787.97
amrex::Abort::0::Newton-Raphson iteration for matter-radiation coupling failed to converge! !!!
SIGABRT
```
I think this indicates our current method is not robust enough.

Edit: Ok, nvm. Something has gone wrong. The timestep is too small here anyway.

Somehow it works on AMD GPU, though...
```
      Start 34: RadhydroShell
34/52 Test #34: RadhydroShell ......................   Passed  465.20 sec
```

BenWibking:
> Code changes look fine. As to why we're not exactly preserving the analytic solution, one thing I might check is whether the shell is really very thin -- the analytic solution assumes an infinitely thin shell, and if that assumption is not fully satisfied, then we shouldn't expect to match the solution exactly. It may be necessary to tweak test parameters to reduce shell thickness, for example by reducing the sound speed so that pressure support becomes less important.

Ah, right. The shell is initially quite thick. This is necessary because there is no steady-state solution of the radiation moment equations for a sufficiently-thin shell under the M1 closure. But that probably means we shouldn't expect exact agreement...

chongchonghe (dismissed):

BenWibking:
@chongchonghe I added a density floor, so it now works for 128^3, which is tested in the nightly regression tests. The 64^3 version is used for the `/azp run` tests for each PR.

markkrumholz (approved):

PR #811, opened 2024-12-10
Use RSLA in all dust tests

chongchonghe:
### Description
Since in all practical runs we will use RSLA, it's important to test that our RSLA method gives the exacted result. In this PR, I update RadDust, RadMarshakDust, and RadMarshakDustPE tests to use RSLA. The analytical solutions are updated to include RSLA as well. In all these tests, RSLA changes the results significantly because 1) in the Marshak tests, the problem is in the optically thin limit, so the propagation speed is reduced, or 2) in the RadDust test, the gas has an initial temperature of $T_0$ and cools to a temperature $T_1$ through thermal emission: $T_0$ $C_V T_0 = C_V T_1 + (\hat{c} / c) a_R T_1^4$ . 

This also prepares for the variable RSLA method I'm working on. 

### Related issues
None.

### Checklist


src/problems/RadDustMG/test_rad_dust_MG.cpp
```diff
@@ -51,14 +52,14 @@ template <> struct Physics_Traits<DustProblem> {
...
-	static constexpr double boltzmann_constant = 1.0;
+	static constexpr double boltzmann_constant = k_B;
 	static constexpr double gravitational_constant = 1.0;
 	static constexpr double c_light = c;
 	static constexpr double radiation_constant = 1.0;
 };
 
 template <> struct RadSystem_Traits<DustProblem> {
-	static constexpr double c_hat_over_c = 1.0;
+	static constexpr double c_hat_over_c = 0.8;
```
BenWibking: This is large value for $\hat c/c$. Would it make sense to use $0.1$ instead?

src/problems/RadMarshakDust/test_radiation_marshak_dust.cpp
```diff
@@ -20,6 +20,8 @@ AMREX_GPU_MANAGED double kappa1 = NAN; // dust opacity at IR
 AMREX_GPU_MANAGED double kappa2 = NAN; // dust opacity at FUV
 
 constexpr double c = 1.0; // speed of light
+constexpr double c_hat_over_c_ = 0.5;
```
BenWibking: This value also seems unrealistically large.

src/problems/RadMarshakDustPE/test_radiation_marshak_dust_and_PE.cpp
```diff
@@ -24,7 +24,10 @@ constexpr bool dust_on = 1;
 constexpr bool PE_on = 1;
 constexpr double gas_dust_coupling_threshold_ = 1.0e-4;
 
+constexpr double stop_time = 0.5;
 constexpr double c = 1.0; // speed of light
+constexpr double c_hat_over_c_ = 0.5;
```
BenWibking: Also very large.

BenWibking (changes requested):
This seems fine, but I don't think values of $\hat c / c$ of 0.5 or 0.8 are realistic. I don't think we'd ever run with a value greater than $0.1$. Otherwise, we might as well just use the full speed of light, or give up and use a fully implicit method.

chongchonghe:
> This seems fine, but I don't think values of c ^ / c of 0.5 or 0.8 are realistic. I don't think we'd ever run with a value greater than 0.1 . Otherwise, we might as well just use the full speed of light, or give up and use a fully implicit method.

OK. I set `chat = 0.1 c` in all these tests. Previously I was just testing if the RSLA method is producing results that agree with the analytical ones.

BenWibking (approved):
Ok, looks good to me assuming GPU tests pass.

In general, I think it's better to make the test problems as close to our production simulations as we can, unless it's not possible or there's another good reason not to.

PR #820, opened 2024-12-13
Implement PhysicsParticleDescriptor and PhysicsParticleRegister

chongchonghe:
### Description
This PR implements `PhysicsParticleDescriptor` and `PhysicsParticleRegister` (in PhysicsParticles.hpp) to register any particle types into `particleRegistry_`. Then, you can loop over all particles in `particleRegister_` to do actions (e.g. deposit gravity) on all particle types with a certain flag. 

#### Particle definition

Line 2068 of simulation.hpp:

```cpp
template <typename problem_t> void AMRSimulation<problem_t>::InitPhyParticles()
{
	if (do_rad_particles != 0) {
		AMREX_ASSERT(RadParticles == nullptr);

		// Create radiating particle descriptor
		// the indices are: massIndex_, lumIndex_, interactsWithHydro_
		auto radParticleDesc = std::make_unique<quokka::PhysicsParticleDescriptor>(-1, quokka::RadParticleLumIdx, false);

		// Create particle container
		RadParticles = std::make_unique<quokka::RadParticleContainer<problem_t>>(this);
		RadParticles->SetVerbose(0);

		// Set container in descriptor
		radParticleDesc->neighborParticleContainer_ = RadParticles.get();

		// Register with particle register
		particleRegister_->registerParticleType("Rad_particles", std::move(radParticleDesc));

		createInitialRadParticles();

		RadParticles->Redistribute();
	}

	if (do_cic_particles != 0) {
    // repeating the above for CICParticles
    ...
  }
}

```

#### Shared functions in PhysicsParticleRegister

`PhysicsParticleRegister` has the following shared functions that are used among all/some particle types: `depositRadiation, depositMass, redistribute, writePlotFile, writeCheckpoint`. 

Example 1: The following code deposit mass/gravity of all particles that belong to a particle type (e.g. 'CIC_particles') with `mass_idx >= 0`. This operates on all AMR levels.

```cpp
particleRegister_->depositMass(rhs, finest_level, Gconst_);
```

source code:

```cpp
void depositMass(amrex::Vector<amrex::MultiFab>& rhs, int finest_lev, amrex::Real Gconst)
{
for (const auto &[name, descriptor] : particleRegistry_) {
	if (descriptor->getMassIndex() >= 0) {
		// Get particle container and deposit mass
		auto *container = static_cast<RadParticleContainer<problem_t> *>(descriptor->getParticleContainer());
		if (container != nullptr) {
			amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev, 
				MassDeposition{Gconst, descriptor->getMassIndex(), 0, 1}, true);
		}
	}
}
}
```

Example 2: The following code write checkpoint files of all particles of all types, each in their own folder. 

```cpp
// write all particles in particleRegister_ to checkpoint file
particleRegister_->writeCheckpoint(checkpointname, true);
```

#### Test problems

`RadParticle` and `BinaryOrbitCIC`. Passing both tests. 

### Related issues

### Checklist


chongchonghe:
Next step:
- Test on defining a RadCICParticles type that have both luminosity and mass. 
- Update `kickParticlesAllLevels` and `driftParticlesAllLevels` to use the new particle architecture. 

A few things that can be improved (I don't have ideas yet):
- Function `InitPhyParticles()` has lots of repetitions. Ideally, the particle initialization process can be simplified as `initializeParticleType("CIC_particles", CICParticles, 0, -1, false)` and `initializeParticleType("Rad_particles", RadParticles, -1, quokka::RadParticleLumIdx, false)`. 
- Reading particle data from chkfiles can be simplified similarly. 

@BenWibking @markkrumholz Please take a look at this PR and let me know what you think.

src/simulation.hpp
```diff
@@ -2054,14 +2065,49 @@ template <typename problem_t> void AMRSimulation<problem_t>::InitParticles()
...
+		// Create radiating particle descriptor
+		// the indices are: mass, lum, hydro_interact
+		auto radParticleDesc = std::make_unique<quokka::PhysicsParticleDescriptor>(-1, quokka::RadParticleLumIdx, false);
+
+		// Create particle container
+		RadParticles = std::make_unique<quokka::RadParticleContainer<problem_t>>(this);
+		RadParticles->SetVerbose(0);
+
+		// Set container in descriptor
+		radParticleDesc->neighborParticleContainer_ = RadParticles.get();
```
BenWibking, line 2086: This does not transfer ownership of `RadParticles`. In your implementation, `RadParticles` will be deallocated on line 2091, and attempting to access `radParticleDesc->neighborParticleContainer_` will cause a segmentation fault.

This is why I am so insistent about using `std::unique_ptr` instead of raw pointers. You cannot have this bug if `radParticleDesc->neighborParticleContainer` is a `std::unique_ptr`.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+	int lumIndex_{-1};		 // index for radiation luminosity, -1 if not used
+	bool interactsWithHydro_{false}; // whether particles interact with hydro
+
+      public:
+	PhysicsParticleDescriptor(int mass_idx, int lum_idx, bool hydro_interact)
+	    : massIndex_(mass_idx), lumIndex_(lum_idx), interactsWithHydro_(hydro_interact)
+	{
+	}
+	virtual ~PhysicsParticleDescriptor() = default;
+	void *neighborParticleContainer_{}; // pointer to particle container, type-erased
```
BenWibking: There is no reason to do type-erasure here, and no reason to make this a raw pointer.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+	// Getters
+	[[nodiscard]] auto getMassIndex() const -> int { return massIndex_; }
+	[[nodiscard]] auto getLumIndex() const -> int { return lumIndex_; }
+	[[nodiscard]] auto getInteractsWithHydro() const -> bool { return interactsWithHydro_; }
+
+	// Virtual methods that can be overridden
+	virtual void hydroInteract() {} // Default no-op
```
BenWibking: There is no reason to make this a virtual function. You can just use `std::function` instead.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+	PhysicsParticleDescriptor &operator=(const PhysicsParticleDescriptor &) = delete;
+	PhysicsParticleDescriptor(PhysicsParticleDescriptor &&) = delete;
+	PhysicsParticleDescriptor &operator=(PhysicsParticleDescriptor &&) = delete;
+};
+
+// Registry for physics particles
+template <typename problem_t> class PhysicsParticleRegister
+{
+      private:
+	std::map<std::string, std::unique_ptr<PhysicsParticleDescriptor>> particleRegistry_;
```
BenWibking, line 132: There is no reason to make this a map of pointers. You can just make this `std::map<std::string, PhysicsParticleDescriptor>`.

BenWibking:
Here is a good explanation of why void* is bad in C++: https://www.reddit.com/r/cpp_questions/comments/jaeky1/void_pointers/

src/particles/PhysicsParticles.hpp
BenWibking: You can make this `std::unique_ptr<amrex::ParticleContainerBase>` and then `dynamic_cast` to the sub-types of particle container, as needed: https://amrex-codes.github.io/amrex/doxygen/classamrex_1_1ParticleContainerBase.html

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+		return nullptr;
+	}
+
+	// Deposit radiation from all particles that have luminosity
+	void depositRadiation(amrex::MultiFab &radEnergySource, int lev, amrex::Real current_time)
+	{
+		for (const auto &[name, descriptor] : particleRegistry_) {
+			if (descriptor->getLumIndex() >= 0) {
+				// Get particle container and deposit luminosity
+				auto *container = static_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);
```
BenWibking: Use `dynamic_cast` instead.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+
+	// Deposit mass from all particles that have mass for gravity calculation
+	void depositMass(amrex::Vector<amrex::MultiFab> &rhs, int finest_lev, amrex::Real Gconst)
+	{
+		for (const auto &[name, descriptor] : particleRegistry_) {
+			if (descriptor->getMassIndex() >= 0) {
+				// Get particle container and deposit mass
+				auto *container = static_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);
```
BenWibking: Use `dynamic_cast` instead.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+				}
+			}
+		}
+	}
+
+	// Run Redistribute(lev) on all particles in particleRegistry_
+	void redistribute(int lev)
+	{
+		for (const auto &[name, descriptor] : particleRegistry_) {
+			auto *container = static_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);
```
BenWibking: Use `dynamic_cast` instead.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+				container->Redistribute(lev);
+			}
+		}
+	}
+
+	// Run Redistribute(lev, ngrow) on all particles in particleRegistry_
+	void redistribute(int lev, int ngrow)
+	{
+		for (const auto &[name, descriptor] : particleRegistry_) {
+			auto *container = static_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);
```
BenWibking: Use `dynamic_cast` instead.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+				container->Redistribute(lev, container->finestLevel(), ngrow);
+			}
+		}
+	}
+
+	// Run WritePlotFile(plotfilename, name) on all particles in particleRegistry_
+	void writePlotFile(const std::string &plotfilename)
+	{
+		for (const auto &[name, descriptor] : particleRegistry_) {
+			auto *container = static_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);
```
BenWibking: Use `dynamic_cast` instead.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+				container->WritePlotFile(plotfilename, name);
+			}
+		}
+	}
+
+	// Run Checkpoint(checkpointname, name, true) on all particles in particleRegistry_
+	void writeCheckpoint(const std::string &checkpointname, bool include_header)
+	{
+		for (const auto &[name, descriptor] : particleRegistry_) {
+			auto *container = static_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);
```
BenWibking: Use `dynamic_cast` instead.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,206 @@
...
+	int massIndex_{-1};		 // index for gravity mass, -1 if not used
+	int lumIndex_{-1};		 // index for radiation luminosity, -1 if not used
+	bool interactsWithHydro_{false}; // whether particles interact with hydro
+
+      public:
+	PhysicsParticleDescriptor(int mass_idx, int lum_idx, bool hydro_interact)
+	    : massIndex_(mass_idx), lumIndex_(lum_idx), interactsWithHydro_(hydro_interact)
+	{
+	}
+	virtual ~PhysicsParticleDescriptor() = default;
```
BenWibking: No reason to make this virtual.

src/radiation/source_terms_multi_group.hpp
```diff
@@ -616,9 +616,10 @@ void RadSystem<problem_t>::AddSourceTermsMultiGroup(array_t &consVar, arrayconst
 
 		// load radiation energy source term
 		// plus advection source term (for well-balanced/SDC integrators)
+		// Note that radEnergySource should contain the luminosity volume density, L / V; unit: erg s^-1 cm^-3
 		quokka::valarray<double, nGroups_> Src;
 		for (int g = 0; g < nGroups_; ++g) {
-			Src[g] = dt * (chat * radEnergySource(i, j, k, g));
+			Src[g] = dt * (chat / c * radEnergySource(i, j, k, g));
```
BenWibking, line 622: This seems like an unrelated change that should go in a separate PR.

src/simulation.hpp
```diff
@@ -447,16 +449,22 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 #ifdef AMREX_PARTICLES
 	void InitParticles();	 // create tracer particles
 	void InitCICParticles(); // create CIC particles
+	void InitPhyParticles(); // create PhysicsParticles
 	int do_tracers = 0;
 	int do_cic_particles = 0;
+	int do_rad_particles = 0;
 	std::unique_ptr<amrex::AmrTracerParticleContainer> TracerPC;
 	std::unique_ptr<quokka::CICParticleContainer> CICParticles;
+	std::unique_ptr<quokka::RadParticleContainer<problem_t>> RadParticles;
```
BenWibking, line 458: I think this doesn't need to be a pointer. It can just be `quokka::RadParticleContainer`.

src/simulation.hpp
```diff
@@ -447,16 +449,22 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
...
 #endif
 
 	// external objects
 #ifdef AMREX_USE_ASCENT
 	Ascent ascent_;
 #endif
+
+	// Add PhysicsParticleRegister member
+	std::unique_ptr<quokka::PhysicsParticleRegister<problem_t>> particleRegister_;
```
BenWibking, line 468: This does not need to be a pointer. It can just be `quokka::PhysicsParticleRegister<problem_t>`.

src/simulation.hpp
BenWibking, line 458: We don't need separate objects for these. They can just live inside the `particleRegister_`.

src/radiation/source_terms_single_group.hpp
```diff
@@ -47,9 +47,12 @@ void RadSystem<problem_t>::AddSourceTermsSingleGroup(array_t &consVar, arraycons
 		const double Erad0 = consPrev(i, j, k, radEnergy_index);
 		AMREX_ASSERT(Erad0 > 0.0);
 
+		const double cscale = c / chat;
```
BenWibking, line 50: The changes in this file look unrelated to particles. Can you put them in a separate PR?

src/particles/PhysicsParticles.hpp
chongchonghe: particleRegistry_ has to be a map from string to std::unique_ptr because it's used to store a descriptor, which is a `std::unique_ptr<ParticleDescriptor>`

BenWibking: Ok, I don't follow. Why do the descriptors need to be `std::unique_ptr<ParticleDescriptor>` instead of just `ParticleDescriptor`?

src/radiation/source_terms_multi_group.hpp
chongchonghe: This is related. I found the original definition of the source term (which I adapted from your original code) confusing. There is a factor of `c` difference from my current implementation. I assume radEnergySource contain the luminosity density (erg s^-1 cm^-3), so the conversion from radEnergySource to Src should be `Src = dt * (chat/c) * radEnergySource`.

src/radiation/source_terms_single_group.hpp
chongchonghe: This is used later to convert radEnergySource to Src, as I explained in my other comments.

src/simulation.hpp
chongchonghe: This is the part I don't understand. I don't think defining RadParticles inside particleRegister_ will work, because that means RadParticles will have to be declared inside a constexpr if, which is not allowed. I haven't tried. 

I forgot to mention that I will change do_cic_particles, etc to constexpr's in Physics_Traits later.

chongchonghe:
@BenWibking I tried to make `neighborParticleContainer_` a `std::unique_ptr<amrex::ParticleContainerBase>`, but it didn't work. The best I can do is to make it a non-owning pointer and it worked: `amrex::ParticleContainerBase *neighborParticleContainer_{nullptr};`. The reason `std::unique_prt` didn't work is because in `InitPhyParticles()`, if we attempt to do `radParticleDesc->neighborParticleContainer_ = std::unique_ptr<amrex::ParticleContainerBase>(RadParticles.get());`, it's trying to 'own' the memory, but the original `RadParticles` also owns the memory. This causes problems. With non-owning pointer and `radParticleDesc->neighborParticleContainer_ = RadParticles.get();`, this is no such a problem. So, in the current framework where `RadParticles` is defined in AMRSimulation (not in particleRegister_), we have to use non-owning pointer.

src/particles/PhysicsParticles.hpp
chongchonghe: Done.

BenWibking:
> @BenWibking I tried to make `neighborParticleContainer_` a `std::unique_ptr<amrex::ParticleContainerBase>`, but it didn't work. The best I can do is to make it a non-owning pointer and it worked: `amrex::ParticleContainerBase *neighborParticleContainer_{nullptr};`. The reason `std::unique_prt` didn't work is because in `InitPhyParticles()`, if we attempt to do `radParticleDesc->neighborParticleContainer_ = std::unique_ptr<amrex::ParticleContainerBase>(RadParticles.get());`, it's trying to 'own' the memory, but the original `RadParticles` also owns the memory. This causes problems. With non-owning pointer and `radParticleDesc->neighborParticleContainer_ = RadParticles.get();`, this is no such a problem. So, in the current framework where `RadParticles` is defined in AMRSimulation (not in particleRegister_), we have to use non-owning pointer.

Ok, that makes sense. I think I missed that in my first review of the PR, sorry. If it's defined/owned by AMRSimulation, then using a raw pointer inside the descriptor (as long as it's not a void pointer) is totally fine by me.

chongchonghe:
There is a problem with my previous implementation. In the following `depositMass` function, I have to use `auto *container = dynamic_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);`, but this is not generic. There will be multiple particles types that will have mass (e.g. CICRadParticles), and the container won't be `RadParticleContainer`, but we have to know it in order to do dynamic_cast. 

```cpp
	// Deposit mass from all particles that have mass for gravity calculation
	void depositMass(amrex::Vector<amrex::MultiFab> &rhs, int finest_lev, amrex::Real Gconst)
	{
		for (const auto &[name, descriptor] : particleRegistry_) {
			if (descriptor->getMassIndex() >= 0) {
				// Get particle container and deposit mass
				auto *container = dynamic_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);
				if (container != nullptr) {
					amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev,
							      MassDeposition{Gconst, descriptor->getMassIndex(), 0, 1}, true);
				}
			}
		}
	}
```

We can do checks of dynamic_cast like in the examples here: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c146-use-dynamic_cast-where-class-hierarchy-navigation-is-unavoidable ,
```cpp
if (auto *container = dynamic_cast<CICParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_)) {
	amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev, MassDeposition{Gconst, descriptor->getMassIndex(), 0, 1}, true);
} else if (auto *container = dynamic_cast<CICRadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_)) {
	amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev, MassDeposition{Gconst, descriptor->getMassIndex(), 0, 1}, true);
} else if (auto *container = dynamic_cast<SinkParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_)) {
	amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev, MassDeposition{Gconst, descriptor->getMassIndex(), 0, 1}, true);
}
```

 but this is exactly the sort of thing we want to avoid by using this new particle architecture. 

-------------

My solution is to use type-erased pointer to the class operations, allowing it to work with any particle container type through a common interface. I created a new PR for this: #826 . Although it works, it adds a lot more complexity compared to this PR, so I don't know if I like it.

src/particles/PhysicsParticles.hpp
chongchonghe: I believe changing it to `std::map<std::string, PhysicsParticleDescriptor>` would make a copy of PhysicsParticleDescriptor, which should be avoided. 

In simulation.hpp, radParticleDesc is created as a pointer, so it makes sense to define `particleRegistry_` as a map from string to a pointer. 
```cpp
auto radParticleDesc = std::make_unique<quokka::PhysicsParticleDescriptor>(-1, quokka::RadParticleLumIdx, quokka::RadParticleBirthTimeIdx, false);
// registration
particleRegister_->registerParticleType("Rad_particles", std::move(radParticleDesc));
```

BenWibking: You don't have to create it as a pointer. If you then `std::move` it, it won't be copied.

BenWibking:
> There is a problem with my previous implementation. In the following `depositMass` function, I have to use `auto *container = dynamic_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);`, but this is not generic. There will be multiple particles types that will have mass (e.g. CICRadParticles), and the container won't be `RadParticleContainer`, but we have to know it in order to do dynamic_cast.
> 
> ```c++
> 	// Deposit mass from all particles that have mass for gravity calculation
> 	void depositMass(amrex::Vector<amrex::MultiFab> &rhs, int finest_lev, amrex::Real Gconst)
> 	{
> 		for (const auto &[name, descriptor] : particleRegistry_) {
> 			if (descriptor->getMassIndex() >= 0) {
> 				// Get particle container and deposit mass
> 				auto *container = dynamic_cast<RadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_);
> 				if (container != nullptr) {
> 					amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev,
> 							      MassDeposition{Gconst, descriptor->getMassIndex(), 0, 1}, true);
> 				}
> 			}
> 		}
> 	}
> ```
> 
> We can do checks of dynamic_cast like in the examples here: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c146-use-dynamic_cast-where-class-hierarchy-navigation-is-unavoidable ,
> 
> ```c++
> if (auto *container = dynamic_cast<CICParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_)) {
> 	amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev, MassDeposition{Gconst, descriptor->getMassIndex(), 0, 1}, true);
> } else if (auto *container = dynamic_cast<CICRadParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_)) {
> 	amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev, MassDeposition{Gconst, descriptor->getMassIndex(), 0, 1}, true);
> } else if (auto *container = dynamic_cast<SinkParticleContainer<problem_t> *>(descriptor->neighborParticleContainer_)) {
> 	amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev, MassDeposition{Gconst, descriptor->getMassIndex(), 0, 1}, true);
> }
> ```
> 
> but this is exactly the sort of thing we want to avoid by using this new particle architecture.
> 
> My solution is to use type-erased pointer to the class operations, allowing it to work with any particle container type through a common interface. I created a new PR for this: #826 . Although it works, it adds a lot more complexity compared to this PR, so I don't know if I like it.

No type-erasure is needed. I think there is a simple solution. See here: https://github.com/quokka-astro/quokka/pull/826#issuecomment-2551941829.

chongchonghe:
This PR doesn't work. The other two PRs: #826 and #827 work, one using type-erased ParticleOperations and the other using derived ParticleDescriptors.

BenWibking:
Ok. In that case, can you close this PR?

PR #826, opened 2024-12-18
New particle structure: type-erased ParticleOperations

chongchonghe:
### Description

This new particle system is built on top of AMReX's particle capabilities and provides several key abstractions:

1. Base particle types
2. Physics particle descriptors
3. A particle registry system
4. Common particle operations

#### Base Particle Types

There are four main particle types defined:

1. `CICParticle` (Cloud-In-Cell)
   - Basic particles with mass and velocity
   - Components: mass, vx, vy, vz

2. `RadParticle` (Radiation)
   - Particles that can emit radiation
   - Components: birth_time, death_time, luminosity (per radiation group)

3. `CICRadParticle` (Combined CIC and Radiation)
   - Particles with both mass/velocity and radiation properties
   - Combines features of both CIC and Rad particles

4. `TracerParticle` (not part of the physics particle system)
   - Simple particles for flow visualization

#### PhysicsParticleDescriptor

This is the base class that describes particle properties and operations:

```cpp
class PhysicsParticleDescriptor {
    int massIndex_;         // Index for gravity mass
    int lumIndex_;          // Index for radiation luminosity
    int birthTimeIndex_;    // Index for birth time
    bool interactsWithHydro_; 
    std::unique_ptr<ParticleOperations> operations_;
};
```

#### ParticleOperations

An interface that defines common operations that can be performed on particles:

```cpp
class ParticleOperations {
    virtual void redistribute(int lev) = 0;
    virtual void writePlotFile(...) = 0;
    virtual void writeCheckpoint(...) = 0;
    virtual void depositRadiation(...) = 0;
    virtual void depositMass(...) = 0;
    virtual void driftParticles(...) = 0;
    virtual void kickParticles(...) = 0;
};
```

1. **Deposition**
   - Mass deposition for gravity calculations
   - Radiation deposition for radiation transfer

2. **Movement**
   - `driftParticles`: Update particle positions based on velocities
   - `kickParticles`: Update particle velocities based on forces

3. **Data Management**
   - `redistribute`: Redistribute particles across MPI ranks
   - `writePlotFile`/`writeCheckpoint`: I/O operations

#### PhysicsParticleRegister

A registry that manages particle types and their operations. It acts as a container and coordinator for all particle types, providing a unified interface for particle operations:

```cpp
template <typename problem_t>
class PhysicsParticleRegister {
    std::map<std::string, std::unique_ptr<PhysicsParticleDescriptor>> particleRegistry_;
    // Methods for particle management and operations
};
```

The Particle Registration registers a new particle type with a unique name and takes ownership of the particle descriptor through a unique pointer:

```cpp
template <typename ParticleDescriptor>
void registerParticleType(const std::string &name, std::unique_ptr<ParticleDescriptor> descriptor)
```

The register also provides several global operations that apply to all registered particles:

```cpp
// Radiation deposition from all particles with luminosity
void depositRadiation(amrex::MultiFab &radEnergySource, int lev, amrex::Real current_time);

// Mass deposition for gravity calculations
void depositMass(amrex::Vector<amrex::MultiFab> &rhs, int finest_lev, amrex::Real Gconst);

// Movement operations for particles with mass
void driftParticlesAllLevels(amrex::Real dt);
void kickParticles(amrex::Real dt, const amrex::MultiFab &accel, 
                  const amrex::Vector<amrex::Geometry> &geom, int lev);

// Data management
void redistribute(int lev);
void redistribute(int lev, int ngrow);
void writePlotFile(const std::string &plotfilename);
void writeCheckpoint(const std::string &checkpointname, bool include_header);
```

These operations only affect particles that have the relevant properties. For example, `depositRadiation` only processes particles that have luminosity (`lumIndex_ >= 0`); `depositMass` only processes particles that have mass (`massIndex_ >= 0`). 

### How to Add a New Particle Type

To add a new particle type, follow these steps:

1. **Define the Particle Data Structure**
   ```cpp
   // Define particle data indices
   enum NewParticleDataIdx { 
       NewParticleProperty1Idx = 0,
       NewParticleProperty2Idx,
       // ... 
   };
   
   // Define number of components
   constexpr int NewParticleRealComps = N; // N = number of real components
   
   // Define container type
   using NewParticleContainer = amrex::AmrParticleContainer<NewParticleRealComps>;
   ```

3. **Register the Particle Type** 
   
   ```cpp
   // In your simulation setup
   auto descriptor = std::make_unique<quokka::PhysicsParticleDescriptor>(massIndex, lumIndex, birthTimeIndex, interactsWithHydro);
   
   // Create particle container
   RadParticles = std::make_unique<quokka::RadParticleContainer<problem_t>>(this);
   RadParticles->SetVerbose(0);
   
   // set particle container
   radParticleDesc->setParticleContainer(RadParticles.get());
   
   // register with particle register
   particleRegister_.registerParticleType("NewParticle", std::move(descriptor));
   
   createInitialRadParticles();
   ```

### Related issues

### Checklist


BenWibking:
I think `std::visit` may solve this problem in a cleaner way. I think you have implemented essentially what `std::variant` combined with `std::visit` does.

Here's an example:
```
#include <iostream>
#include <variant>

struct Fluid { };
struct LightItem { };
struct HeavyItem { };
struct FragileItem { };

template<class... Ts> struct overload : Ts... { using Ts::operator()...; };
template<class... Ts> overload(Ts...) -> overload<Ts...>; // line not needed in C++20...

int main() {
    std::variant<Fluid, LightItem, HeavyItem, FragileItem> package;

    std::visit(overload{
        [](Fluid& )       { std::cout << "fluid\n"; },
        [](LightItem& )   { std::cout << "light item\n"; },
        [](HeavyItem& )   { std::cout << "heavy item\n"; },
        [](FragileItem& ) { std::cout << "fragile\n"; }
    }, package);
}
```
In this case, `std::variant` could be `std::variant<CICParticles, RadParticles>`, and any additional particle containers can be added to the variant template arguments.

Let me know what you think. More examples in [1][2][3].

[1] https://levelup.gitconnected.com/understanding-std-visit-in-c-a-type-safe-way-to-traverse-variant-objects-dbeff9b47003
[2] https://www.cppstories.com/2018/09/visit-variants/
[2] https://en.cppreference.com/w/cpp/utility/variant/visit

BenWibking:
I don't understand the motivation for having both inheritance and a template class in this case. I think you could use either, but I don't see why you need both, since ParticleOperationsImpl is the only subclass of ParticleOperations. (This is not the case for AMRSimulation. It has two subclasses: QuokkaSimulation and LinearAdvectionSimulation. I don't think following that pattern makes sense here.)
```
template<typename ParticleContainerType>
class ParticleOperationsImpl : public ParticleOperations {
    ParticleContainerType* container_;
    // Implements all virtual methods using the concrete container type
};
```

BenWibking:
How does this code handle ParticleContainers that don't have luminosity? For example, CICParticles might represent dark matter particles. Does this code call ops->depositRadiation for all particle types? It looks like it does to me, but maybe I'm missing something.
```
	void depositRadiation(amrex::MultiFab& radEnergySource, int lev, amrex::Real current_time) {
		for (const auto& [name, descriptor] : particleRegistry_) {
			if (auto* ops = descriptor->getOperations()) {
				ops->depositRadiation(radEnergySource, lev, current_time, 
					descriptor->getLumIndex(), descriptor->getBirthTimeIndex(), Physics_Traits<problem_t>::nGroups);
			}
		}
	}
```

BenWibking:
Actually, I think everything can be handled with standard C++ virtual functions. As long as the object is created as the subclass you want, then everything should "just work":

```
class Base {
    public: virtual int thing() { return 0; }
};

class Derived : public Base {
    public: int thing() { return 1; }
};

int caller(Base &obj)
{
    return obj.thing(); // Returns 1 if obj actually references a Derived object
}

int main() {
    Base b;
    Derived d;
    
    caller(d); // 1
    caller(b); // 0
}
```

See explanation here: https://stackoverflow.com/a/59271362

BenWibking:
So in conclusion:
* if you want to do the *same* operation (that may have a different virtual derived implementation) on *different* ParticleContainers (that are all derived from the same base class), then there are two options:
  * **If all of the methods you want to call are marked virtual in the base class**, you can just call the method on a reference/pointer to the base class, and it will actually call the method of the derived class (since it knows it was created as an object of the derived class originally) (see example: https://github.com/quokka-astro/quokka/pull/826#issuecomment-2551934378). This works regardless of whether you have a reference or a pointer, but it only works when the method is virtual. **It fails in the example below (which is the case for some of the ParticleContainer methods, like `Redistribute()`):**
    ```
    struct B {
    void f (int);
    void f (double);
    };

    struct D : public B {
      void f (); // this hides f(int) and f(double) in base class
    };
  * You can perform `dynamic_cast<void*>(PointerToSomeDerivedParticleContainer)->method()` and this is *guaranteed* to call the method `SomeDerivedParticleContainer->method()`, regardless of what the derived particle container is. This is because `dynamic_cast<void*>(PointerToDerivedObject)` (very counterintuitively!) actually casts `PointerToDerivedObject` to a pointer to the *most derived* object type: "Otherwise, if target-type is “pointer to (possibly cv-qualified) void”, the result is a pointer to the [most derived object](https://en.cppreference.com/w/cpp/language/object#Subobject) pointed to by expression." This is the one case in C++ where using `void*` is okay :D
* If you want to do *different* operations on different ParticleContainers, you can use the `std::visit` pattern as demonstrated in the example code I posted above (e.g., call `PointerToSomeDerivedParticleContainer->explode()` where `explode()` is only defined for `ExplodingParticleContainer`).

chongchonghe:
> How does this code handle ParticleContainers that don't have luminosity? For example, CICParticles might represent dark matter particles. Does this code call ops->depositRadiation for all particle types? It looks like it does to me, but maybe I'm missing something.
> 
> ```
> 	void depositRadiation(amrex::MultiFab& radEnergySource, int lev, amrex::Real current_time) {
> 		for (const auto& [name, descriptor] : particleRegistry_) {
> 			if (auto* ops = descriptor->getOperations()) {
> 				ops->depositRadiation(radEnergySource, lev, current_time, 
> 					descriptor->getLumIndex(), descriptor->getBirthTimeIndex(), Physics_Traits<problem_t>::nGroups);
> 			}
> 		}
> 	}
> ```

`ParticleOperationsImpl` defines `depositRadiation` the following way, so `lumIndex >= 0` is checked.

```cpp
	void depositRadiation(amrex::MultiFab &radEnergySource, int lev, amrex::Real current_time, int lumIndex, int birthTimeIndex, int nGroups) override
	{
		if (container_ && lumIndex >= 0) {
			amrex::ParticleToMesh(*container_, radEnergySource, lev, RadDeposition{current_time, lumIndex, 0, nGroups, birthTimeIndex}, false);
		}
	}
```

chongchonghe:
> I don't understand the motivation for having both inheritance and a template class in this case. I think you could use either, but I don't see why you need both, since ParticleOperationsImpl is the only subclass of ParticleOperations. (This is not the case for AMRSimulation. It has two subclasses: QuokkaSimulation and LinearAdvectionSimulation. I don't think following that pattern makes sense here.)
> 
> ```
> template<typename ParticleContainerType>
> class ParticleOperationsImpl : public ParticleOperations {
>     ParticleContainerType* container_;
>     // Implements all virtual methods using the concrete container type
> };
> ```

The reason to do class inheritance here is because `ParticleOperations` is not templated and `ParticleOperationsImpl` is. So, by using `ParticleOperationsImpl` with different typename ParticleContainerType, we are essentially creating multiple inheritance of `ParticleOperations`. In `PhysicsParticleDescriptor`, `operations_` is a unique_ptr to the untemplated `ParticleOperations`: `std::unique_ptr<ParticleOperations> operations_;` Each `ParticleOperationsImpl` created at particle creation is a proper operation for that container type.

BenWibking:
> > ```
> > template<typename ParticleContainerType>
> > class ParticleOperationsImpl : public ParticleOperations {
> >     ParticleContainerType* container_;
> >     // Implements all virtual methods using the concrete container type
> > };
> > ```
> 
> The reason to do class inheritance here is because `ParticleOperations` is not templated and `ParticleOperationsImpl` is. So, by using `ParticleOperationsImpl` with different typename ParticleContainerType, we are essentially creating multiple inheritance of `ParticleOperations`. In `PhysicsParticleDescriptor`, `operations_` is a unique_ptr to the untemplated `ParticleOperations`: `std::unique_ptr<ParticleOperations> operations_;` Each `ParticleOperationsImpl` created at particle creation is a proper operation for that container type.

Ok, I don't think templates are useful for this use case. I think everything can be implemented via class inheritance and virtual functions.

PR #827, opened 2024-12-19
New particle interface using derived classes

chongchonghe:
### Description

The physics particle interface provides a flexible and extensible system for managing different types of particles in the simulation. It uses a combination of the Strategy pattern and Registry pattern to handle multiple particle types through a common interface.

##### 1. PhysicsParticleDescriptor (Base Class)

```cpp
class PhysicsParticleDescriptor {
		protected:
  int massIndex_{-1};	// index for gravity mass, -1 if not used
  int lumIndex_{-1};	// index for radiation luminosity, -1 if not used
  int birthTimeIndex_{-1}; // index for birth time, -1 if not used
  bool interactsWithHydro_{false}; // whether particles interact with hydro

		public:
	virtual void redistribute(int lev) = 0;
  virtual void writePlotFile(...) = 0;
  virtual void depositRadiation(...) = 0;
  virtual void depositMass(...) = 0;
  // ...
};
```

This abstract base class defines the interface that all particle descriptors must implement. It provides virtual methods for common particle operations like:
- Redistribution across processors
- Writing to plot files and checkpoints
- Depositing radiation and mass to the grid

##### 2. Concrete Descriptor Classes
Currently, the system provides three specialized descriptor classes. More can be added by subclassing `PhysicsParticleDescriptor`.
- `RadParticleDescriptor`: For radiation-emitting particles
- `CICParticleDescriptor`: For Cloud-In-Cell (gravitating) particles
- `CICRadParticleDescriptor`: For particles that both emit radiation and interact via CIC

##### 3. PhysicsParticleRegister
```cpp
template <typename problem_t>
class PhysicsParticleRegister {
    void registerParticleType(const std::string &name, 
                            std::unique_ptr<ParticleDescriptor> descriptor);
    void depositRadiation(...);
    void depositMass(...);
    // ...
};
```

A registry that manages all particle descriptors and provides high-level operations that apply to all registered particles.

##### 1. Creating and Registering Particles

See example in `src/simulation.hpp: InitParticles()`

```cpp
// Create particle descriptor
auto radParticleDesc = std::make_unique<RadParticleDescriptor<problem_t>>(
    -1,                // mass index
    RadParticleLumIdx, // luminosity index 
    RadParticleBirthTimeIdx, // birth time index
    false              // hydro interaction flag
);

// Create particle container
RadParticles = std::make_unique<RadParticleContainer<problem_t>>(this);

// Set container in descriptor
radParticleDesc->setParticleContainer(RadParticles.get());

// Register with particle register
particleRegister_->registerParticleType("Rad_particles", std::move(radParticleDesc));
```

##### 2. Using the Interface
Once particles are registered, you can perform operations on all particles through the register:

```cpp
// Deposit radiation from all particles that have luminosity
particleRegister_->depositRadiation(radEnergySource, lev, current_time);

// Redistribute particles after grid changes
particleRegister_->redistribute(lev);

// Write particle data to plot files
particleRegister_->writePlotFile(plotfilename);
```

#### Key Features

1. **Type Safety**: Each descriptor handles type-safe casting of its particle container internally.

2. **Extensibility**: New particle types can be added by:
   - Creating a new particle container class
   - Creating a new descriptor class that inherits from `PhysicsParticleDescriptor`
   - Implementing the required virtual methods

3. **Unified Interface**: All particle operations can be performed through the register without knowing the specific particle types.

4. **Selective Operations**: The system automatically handles operations only for particles that support them:
   ```cpp
   // Only particles with luminosity will deposit radiation
   if (descriptor->getLumIndex() >= 0) {
       descriptor->depositRadiation(...);
   }
   ```

#### Adding a New Particle Type

1. Create a new particle container and iterator:
```cpp
using XXXParticleContainer = amrex::AmrParticleContainer<XXXParticleRealComps>;
using XXXParticleIterator = amrex::ParIter<XXXParticleRealComps>;
```

2. Create a new descriptor class:
```cpp
class XXXParticleDescriptor : public PhysicsParticleDescriptor {
    void redistribute(int lev) override {
        if (auto *container = dynamic_cast<XXXParticleContainer*>(
            this->neighborParticleContainer_)) {
            container->Redistribute(lev);
        }
    }
    // Implement other virtual methods
};
```

3. Define a unique_ptr to the new particle container in AMRSimulation:
```cpp
std::unique_ptr<XXXParticleContainer> XXXParticles;
```

4. Register the new particle type in AMRSimulation::InitPhyParticles():
```cpp
auto newParticleDesc = std::make_unique<XXXParticleDescriptor>(...);
...
particleRegister_->registerParticleType("XXX_particles", std::move(newParticleDesc));
createInitialXXXParticles();
```

5. Register the new particle type in AMRSimulation::ReadCheckpointFile():
```cpp
particleRegister_->registerParticleType("XXX_particles", std::move(newParticleDesc));
```

### New test problems

This implementation is totally working, passing both `RadParticle` and `GravRadParticle3D` tests. 

### Related issues
None.

### Checklist


src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,365 @@
...
+	PhysicsParticleDescriptor &operator=(PhysicsParticleDescriptor &&) = delete;
+
+	// Setter for particle container
+	template <typename ParticleContainerType> void setParticleContainer(ParticleContainerType *container) { neighborParticleContainer_ = container; }
+
+	// Virtual operations that will be called on the particle container
+	virtual void redistribute(int lev) = 0;
+	virtual void redistribute(int lev, int ngrow) = 0;
+	virtual void writePlotFile(const std::string &plotfilename, const std::string &name) = 0;
+	virtual void writeCheckpoint(const std::string &checkpointname, const std::string &name, bool include_header) = 0;
```
BenWibking, line 158: redistribute, writePlotFile, and writeCheckpoint can all be implemented here in the base class. You can do `dynamic_cast<void*>(neighborParticleContainer_)->redistribute`, etc.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,365 @@
...
+	virtual void depositRadiation(amrex::MultiFab &radEnergySource, int lev, amrex::Real current_time, int lumIndex, int birthTimeIndex, int nGroups) = 0;
+	virtual void depositMass(amrex::Vector<amrex::MultiFab> &rhs, int finest_lev, amrex::Real Gconst, int massIndex) = 0;
```
BenWibking, line 160: I think `depositMass` could also be implemented here in the base class.

src/simulation.hpp
```diff
@@ -2795,11 +2871,36 @@ template <typename problem_t> void AMRSimulation<problem_t>::ReadCheckpointFile(
...
+		cicParticleDesc->setParticleContainer(CICParticles.get());
+		particleRegister_->registerParticleType("CIC_particles", std::move(cicParticleDesc));
 		CICParticles->Restart(restart_chkfile, "CIC_particles");
 	}
+
+	if (do_rad_particles != 0) {
+		AMREX_ASSERT(RadParticles == nullptr);
+		RadParticles = std::make_unique<quokka::RadParticleContainer<problem_t>>(this);
+		auto radParticleDesc =
+		    std::make_unique<quokka::RadParticleDescriptor<problem_t>>(-1, quokka::RadParticleLumIdx, quokka::RadParticleBirthTimeIdx, false);
```
BenWibking, line 2896: I don't think it's necessary to create a `unique_ptr<quokka::RadParticleContainer>` You can just create a `RadParticleContainer` and then `std::move` it to the particleRegister_.

src/particles/PhysicsParticles.hpp
```diff
@@ -0,0 +1,365 @@
...
+			amrex::ParticleToMesh(*container, amrex::GetVecOfPtrs(rhs), 0, finest_lev, MassDeposition{Gconst, massIndex, 0, 1}, true);
+		}
+	}
+};
+
+// Registry for physics particles
+template <typename problem_t> class PhysicsParticleRegister
+{
+      private:
+	std::vector<std::unique_ptr<PhysicsParticleDescriptor>> particles_;
```
BenWibking: We might want to provide a function to get a weak pointer to the PhysicsParticleDescriptor (or its underlying ParticleContainer) from the PhysicsParticleRegister based on its name. This could be done with `std::map<std::string, PhysicsParticleDescriptor>` instead.

BenWibking (changes requested):

BenWibking:
@chongchonghe  I think this is the right approach. I made a few comments that about changes that should reduce code duplication. I also think you should change the type of PhysicsParticleRegister::particles_ (see comments for the reason why) to `std::map<std::string, PhysicsParticleDescriptor>`.

Since it looks like they're not needed anymore, can you close the other two PRs?

src/particles/PhysicsParticles.hpp
chongchonghe: No, unfortunately. Please check the example I showed in the slack channel amrex-astro/random. Attempting to do this will result in this error: 'Member reference base type 'void' is not a structure or union'. We can't directly call methods on a void* pointer.

chongchonghe:
@BenWibking I think we have to redefine every single methods in every derived ParticleDescriptor class. This is because 
1. `dynamic_cast<void*>(neighborParticleContainer_)->redistribute(lev)` won't work because `void*` is not a class type and cannot have methods. `void*` is used to compare object identity or to cast back to original type, but this can't be done unless we know what the original type is. 
2. If we want to cast neighborParticleContainer_ to some base class, the deepest we can go is `ParticleContainer_impl`, because redistribute and WritePlotFile are defined there. However, `ParticleContainer_impl` has to take templated parameter `NArrayReal`, which is different for different ParticleContainers. 

The use of `void*` is demonstrated in the example cpp file I shared in the slack channel amrex-astro/random, which I copied below.

```cpp
#include <iostream>

class Base {
  virtual void dummy() {} // Need at least one virtual function for RTTI
public:
  int data = 1;
};

class Derived : public Base {
public:
  int data = 2;
};

template <int T> class TDerived : public Base {
public:
  int data = T;
};

int main() {
  Base *base_ptr = new Base();
  Derived *derived_ptr = new Derived();
  Base *derived_ptr_2 = new Derived();
  TDerived<3> *derived2_ptr = new TDerived<3>();

  // print the data of each object
  std::cout << "Original class data: "
            << "\n";
  std::cout << "Base data: " << base_ptr->data << "\n";
  std::cout << "Derived data: " << derived_ptr->data << "\n";
  std::cout << "Derived_ptr_2 data: " << derived_ptr_2->data << "\n";
  std::cout << "Templated Derived2 data: " << derived2_ptr->data << "\n";

  // Cast to void*
  void *base_void = dynamic_cast<void *>(base_ptr);
  void *derived_void = dynamic_cast<void *>(derived_ptr);
  void *derived_ptr_2_void = dynamic_cast<void *>(derived_ptr_2);
  void *derived2_void = dynamic_cast<void *>(derived2_ptr);

  // These void* pointers will point to the start of their complete objects
  // base_void points to Base object
  // derived_void points to Derived object
  // derived_ptr_2_void points to Base object
  // derived2_void points to TDerived<3> object

  // Cast back to original type
  std::cout << "Cast back to original type: "
            << "\n";
  Base *original_base = static_cast<Base *>(base_void);
  std::cout << "Base data: " << original_base->data << "\n";

  Derived *original_derived = static_cast<Derived *>(derived_void);
  std::cout << "Derived data: " << original_derived->data << "\n";

  Base *original_derived_ptr_2 = static_cast<Base *>(derived_ptr_2_void);
  std::cout << "Derived_ptr_2 data: " << original_derived_ptr_2->data << "\n";

  Base *original_derived2 = static_cast<Base *>(derived2_void);
  std::cout << "Derived2 data: " << original_derived2->data << "\n";

  TDerived<3> *original_derived2_2 = static_cast<TDerived<3> *>(derived2_void);
  std::cout << "Templated Derived2 data: " << original_derived2_2->data << "\n";

  // This is allowed
  Base *tmp = dynamic_cast<Base *>(derived2_ptr);
  std::cout << "tmp: " << dynamic_cast<TDerived<3> *>(tmp)->data << "\n";

  // This is not allowed, because void* points to Base, and it (1) cannot access
  // data, and (2) does not know about TDerived<3> unless you explicitly cast it
  // to TDerived<3>
  // std::cout << "tmp: " << dynamic_cast<void*>(tmp)->data << "\n";

  // Error message: "Member reference base type 'void' is not a structure or union"
}

// The output is:
/*
Original class data:
Base data: 1
Derived data: 2
Derived_ptr_2 data: 1
Templated Derived2 data: 3
Cast back to original type:
Base data: 1
Derived data: 2
Derived_ptr_2 data: 1
Derived2 data: 1
Templated Derived2 data: 3
tmp: 3
*/
```

src/particles/PhysicsParticles.hpp
chongchonghe: Unfortrunately, no. See my comments above.

src/particles/PhysicsParticles.hpp
chongchonghe: Done, and it has to be `std::map<std::string, std::unique_ptr<PhysicsParticleDescriptor>>` because `particleRegister_` is unique_ptr.

src/simulation.hpp
chongchonghe: Actually it is necessary. This is because particleRegister_ is a unique_ptr, and in registerParticleType, particleRegister_ 'emplace' std::move(descriptor), which has to be a unique_ptr.

chongchonghe: This does the automatic memory management as expected, and std::move transfers ownership of the descriptor to the registry.

BenWibking: You can `std::move` any object:
```
#include <iomanip>
#include <iostream>
#include <string>
#include <utility>
#include <vector>
 
int main()
{
    std::string str = "Salut";
    std::vector<std::string> v;
 
    // uses the push_back(const T&) overload, which means
    // we'll incur the cost of copying str
    v.push_back(str);
    std::cout << "After copy, str is " << std::quoted(str) << '\n';
 
    // uses the rvalue reference push_back(T&&) overload,
    // which means no strings will be copied; instead, the contents
    // of str will be moved into the vector. This is less
    // expensive, but also means str might now be empty.
    v.push_back(std::move(str));
    std::cout << "After move, str is " << std::quoted(str) << '\n';
 
    std::cout << "The contents of the vector are {" << std::quoted(v[0])
              << ", " << std::quoted(v[1]) << "}\n";
}
```

src/simulation.hpp
```diff
@@ -446,17 +448,24 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
...
+	std::unique_ptr<quokka::CICRadParticleContainer<problem_t>> CICRadParticles;
 #endif
 
 	// external objects
 #ifdef AMREX_USE_ASCENT
 	Ascent ascent_;
 #endif
+
+	// Add PhysicsParticleRegister member
+	std::unique_ptr<quokka::PhysicsParticleRegister<problem_t>> particleRegister_;
```
BenWibking: Can you make this just `quokka::PhysicsParticleRegister<problem_t>`? (no `unique_ptr`)?

chongchonghe: I thought it's more memory-safe to define particleRegister_ as unique_ptr, but maybe I'm wrong.

chongchonghe:
@BenWibking I've changed `particleRegister_` from `std::unique_ptr<quokka::PhysicsParticleRegister<problem_t>>` to `quokka::PhysicsParticleRegister<problem_t>` in my latest commit. Please have a look.

src/QuokkaSimulation.hpp
```diff
@@ -1721,7 +1721,7 @@ void QuokkaSimulation<problem_t>::subcycleRadiationAtLevel(int lev, amrex::Real
 			PrintRadEnergySource(radEnergySource);
 
 			// Deposit radiation from all particles that have luminosity. When there are no particles with luminosity, this will do nothing.
-			AMRSimulation<problem_t>::particleRegister_->depositRadiation(radEnergySource, lev, time_subcycle);
+			AMRSimulation<problem_t>::particleRegister_.depositRadiation(radEnergySource, lev, time_subcycle);
```
BenWibking: If you add `using AMRSimulation<problem_t>::particleRegister_;` at the start of the `QuokkaSimulation` class, you will be able to reference `particleRegister_` as a normal class variable here.

src/simulation.hpp
BenWibking: Ok, I see now why you did this. This makes sense if we have `AMRSimulation<problem_t>::CICParticles` as a member variable. But if it's a unique_ptr, that means we can't access it as a member variable once it's been moved. So it should be either a `shared_ptr`, or we should just get a weak pointer to it from the PhysicsParticleRegister whenever we need to access it directly.

src/simulation.hpp
BenWibking: It's more memory-safe compared to a raw pointer, but in this case, I don't think it needs to be a pointer at all.

src/QuokkaSimulation.hpp

BenWibking:
It looks like most of the clang-tidy warnings are due to missing direct header includes (https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html). Can you fix those?

BenWibking:
Since this PR is superceded by the newer one, can we close this PR?

BenWibking:
Since we've merged the particle infrastructure in another PR, I'm going to close this one.

PR #833, opened 2024-12-22
Redistribute particles after drift in the kick-drift-kick process

chongchonghe:
### Description
I believe it is necessary to redistribute particles after particle movement in the kick-drift-kick process. This is required by mass deposition in the second kick step. 

I came across the following error while testing our new particle architecture through the CICRadParticle test, and the changes in this PR fixed this error. In this test, the particles move across MultiFab box edges in the first time step.

```
ParticleContainer spread across MPI nodes - bytes (num particles): [Min: 176 (2), Max: 176 (2), Total: 176 (2)]
ParticleContainer::Redistribute() time: 8.5291e-05

amrex::Abort::0:: (47,30,31,0) is out of bound (31:64,31:64,31:64,0:0) !!!
```

Backtrace.0:
```
 8: quokka::CICRadParticleDescriptor<ParticleProblem>::depositMass(amrex::Vector<amrex::MultiFab, std::__1::allocator<amrex::MultiFab>>&, int, double, int) (in test_grav_rad_particle_3D) (PhysicsParticles.hpp:344)

 9: quokka::PhysicsParticleRegister<ParticleProblem>::depositMass(amrex::Vector<amrex::MultiFab, std::__1::allocator<amrex::MultiFab>>&, int, double) const (in test_grav_rad_particle_3D) (PhysicsParticles.hpp:462)

10: AMRSimulation<ParticleProblem>::calculateGpotAllLevels() (in test_grav_rad_particle_3D) (simulation.hpp:1120)
```

However, I don't understand why this was not caught earlier by the binary_orbit test. I have tried to modify the runtime parameters of the binary_orbit test so that `n_cell = 64` and `blocking_factor = 32`, and I was expecting it to fail similarly but it didn't.

### Related issues

### Checklist


markkrumholz (commented):
Agree that this redistribute is required. Once you fix the codeQL warnings, happy to approve.

markkrumholz:
Should add: I suspect the reason it doesn't fail for the binary orbit test has to do with the time step size. It looks like in this test the particle moved multiple gas cells, which let it get out of the ghost zone. If in a given test the particles don't move more than a cell, you won't get this out of bounds error, even if you get a slightly wrong answer.

BenWibking:
This is necessary in our current approach, but I'm worried this will increase communication costs significantly if we have a lot of particles.

I don't think Nyx or PeleC do this. Can you check what they do, or ask Andrew Myers? Maybe we should add a "particle CFL" timestep restriction.

BenWibking:
@atmyers Do you have advice for us on what to do here? TLDR: we don't currently limit the timestep based on particle velocity. Is the optimal thing to do to add a redistribute after drift, or there is a less-communication-intensive solution (i.e. adding a "particle CFL")?

chongchonghe:
@BenWibking That should be the only particle redistribution required, right? Can you go through all `redistribute` in `simulation.hpp` and check if they are really necessary? I suspect some `redistribute` in `timeStepWithSubcycling()` may not be required. 

"Particle CFL condition" may solve this out-of-bound issue, but it may not solve the erroneous result issue. When a particle moves to a boundary cell during drift, the gas in another multifab should feel its gravity. This would be particularly important when you have a binary star at multifab boundary.

BenWibking:
> @BenWibking That should be the only particle redistribution required, right? Can you go through all `redistribute` in `simulation.hpp` and check if they are really necessary? I suspect some `redistribute` in `timeStepWithSubcycling()` may not be required.

All of the current redistributes are necessary. It is currently done i) on initialization (to ensure consistency, even if users create particle in weird places in the problem setup), ii) on regrid (for obvious reasons), iii) after subcycling a level (particle can move on the same level *or* could move to a different AMR level, so this one is unavoidable). Those are the places redistribute is done in PeleC: https://github.com/AMReX-Combustion/PeleC/blob/development/Source/Particle.cpp.

> "Particle CFL condition" may solve this out-of-bound issue, but it may not solve the erroneous result issue. When a particle moves to a boundary cell during drift, the gas in another multifab should feel its gravity. This would be particularly important when you have a binary star at multifab boundary.

`amrex::ParticleToMesh` communicates the depositions from ghost cells to neighboring grids, so this should be handled correctly. If it didn't do this, there would be an error even if the particles didn't move but were close to the boundary. The only error is if it goes out of bounds.

BenWibking:
Here is how PelePhysics computes the timestep based on the particle velocities: https://github.com/AMReX-Combustion/PelePhysics/blob/760e2febb1573bc9f1a3dd9b70ac5a36ebc8ce5c/Source/Spray/SprayParticles.cpp#L92

BenWibking:
If we don't add a particle CFL < 1.0, then we can't use the communication-efficient version of redistribute: https://github.com/AMReX-Codes/amrex/blob/b3f67385e62f387b548389222840486c0fffca57/Src/Particle/AMReX_ParticleContainer.H#L499

The trajectory of the particle will also be somewhat inaccurate without a particle CFL restriction.

chongchonghe:
@BenWibking Do you know what could've caused the ROCm tests to fail? Is there a way to download the full log files?

BenWibking:
> @BenWibking Do you know what could've caused the ROCm tests to fail? Is there a way to download the full log files?

The new ROCm version is broken. It cannot start GPU executables. We have to wait until IT makes available the old version again.

chongchonghe:
@BenWibking In any case, while we discuss about optimal ways to do particle redistribute and CFL conditions, this PR is worth merging since we need to fix this bug for implementation of the new particle architecture.

BenWibking:
> @BenWibking In any case, while we discuss about optimal ways to do particle redistribute and CFL conditions, this PR is worth merging since we need to fix this bug for implementation of the new particle architecture.

Since this PR departs from what other AMReX codes do, can we hold off until we hear from Andrew or Weiqun? I don't want to accidentally introduce a new bug because there is something subtle about how this works that isn't in the documentation (or is out-of-date in the documentation or source code comments, which has happened before).

For instance, there appear to be subtle, but potentially significant, consequences to each argument to Redistribute:
```
 /**
    * \brief Redistribute puts all the particles back in the right places (for some value of right)
    *
    * Assigns particles to the levels, grids, and tiles that contain their current positions.
    * If periodic boundary conditions are used, those will be enforced here.
    *
    * If Redistribute is called with default arguments, all particles will be placed on the
    * finest level that covers their current positions.
    *
    * The lev_min, lev_max, and nGrow flags are used to do proper checking for subcycling particles.
    * The default values are fine for non-subcycling methods
    *
    * The local flag controls whether this is `local` or `global` Redistribute.
    * In a local Redistribute, particles can only have moved a certain distance since the last
    * time Redistribute() was called. Thus, communication only needs to happen between neighboring
    * ranks. In a global Redistribute, the particles can potentially go from any rank to any rank.
    * This usually happens after initialization or when doing dynamic load balancing.
    *
    * \param lev_min The minimum level consider. Particles on levels less than this will not be
    *                touched, and particles on finer levels will not be assigned to levels less
    *                than this, either.
    *                Default: 0.
    * \param lev_max The maximum level consider. Particles on levels greater than this will not be
    *                touched, and particles on coarser levels will not be assigned to levels greater
    *                than this, either. If negative, will use the finest level in the hierarchy.
    *                Default: -1.
    * \param nGrow If particles are within nGrow cells of their current box, they will not moved.
    *              This is useful for subcycling methods, when fine level particles need to be
    *              redistributed but are not necessarily at the same time as those on the coarse
    *              level.
    *              Default: 0
    * \param local If 0, this will be a non-local redistribute, meaning that particle can potentially
    *              go to any other box in the simulation. If > 0, this is the maximum number of cells
    *              a particle can have moved since the last Redistribute() call. Knowing this number
    *              allows an optimized MPI communication pattern to be used.
    */
    void Redistribute (int lev_min = 0, int lev_max = -1, int nGrow = 0, int local=0,
                       bool remove_negative=true);
```

BenWibking:
This is how Nyx does `Redistribute()` for neutrinos (since they are allowed to have particleCfl > 1):
https://github.com/AMReX-Astro/Nyx/commit/639655edd7899ef6bbc817e932afbe26e3f11b5a

See also: https://github.com/AMReX-Astro/Nyx/blob/development/Source/Particle/NyxParticleContainer.H

chongchonghe:
Okay, that's fine, since we are going to hold off new particle implementation anyway.

BenWibking:
Since we currently drift particles on all levels at once, this is okay.

For future reference, if we switch to CIC particles that support AMR subcycling, note that the `Redistribute()` call cannot be done during the level advance. This is because particles from a given level $l > 0$ can cross the coarse-fine boundary and no longer live on level $l$ after the first subcycle on level $l$. Then the particle is missed during the second subcycle kick-drift-kick. `Redistribute()` must instead be called after the recursive updates on levels $l' > l$, but before level $l$ returns control back to level $l-1$. In this scenario, a particle CFL restriction is also required for correct operation.

Once this passes GPU testing, I will approve.

BenWibking (dismissed):

BenWibking:
It looks like this needs to be update for the new particles infrastructure.

src/simulation.hpp
```diff
@@ -992,15 +992,12 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
...
-		// TODO(cch): I believe this is needed and missing this in the original code was a bug, but I don't understand why this was not caught earlier.
-		// Maybe the binary_orbit test was the only test for gravity and there was no multiple boxes in that test?
-		// for (int lev = 0; lev <= finest_level; ++lev) {
-		// 	// Use ngrow=0 since we only need particles in valid cells for radiation
-		// 	particleRegister_.redistribute(lev, 0);
-		// }
+		// Redistribute particles at all levels after movement. This ensures particles are in the correct cells/processors for mass deposition in
+		// ellipticSolveAllLevels()
+		// Use ngrow=0 since we only need particles in valid cells for radiation
+		particleRegister_.redistribute(0, 0);
```
BenWibking, line 999: I don't understand this. I thought the `ngrow` argument was not necessary:
```
    * \param nGrow If particles are within nGrow cells of their current box, they will not moved.
    *              This is useful for subcycling methods, when fine level particles need to be
    *              redistributed but are not necessarily at the same time as those on the coarse
    *              level.
    *              Default: 0
 ```

chongchonghe: Either `redistribute(0)` or `redistribute(0, 0)` is fine here. They are the same. I used the latter for better readability since I said ngrow=0 in the comments.

BenWibking: Right, I understand that they are the same. I don't understand your comment, though.

BenWibking:
Since we have a separate PR for implementing the particle CFL, I'll close this PR.

PR #843, opened 2025-01-09
Enforce minimum ROCm version >= 6.3

### LINKED ISSUE #394: AMD GPUs produce incorrect results

BenWibking:
### Description
Require ROCm 6.3 or newer to compile Quokka for AMD GPUs.

### Related issues
Closes https://github.com/quokka-astro/quokka/issues/394.

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,7 @@ if(AMReX_GPU_BACKEND MATCHES "HIP")
   string(REPLACE "HIP version: " "" hipcc_version ${hipcc_version})
   message(STATUS "HIP compiler version: ${hipcc_version}")
 
-  if(hipcc_version VERSION_LESS 6.4)
+  if(hipcc_version VERSION_LESS 6.3)
     message(FATAL_ERROR "You must use ROCm version >= 6.3 to compile Quokka. All previous ROCm versions have compiler bugs that may cause Quokka to produce incorrect results.")
   endif()
 endif(AMReX_GPU_BACKEND MATCHES "HIP")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -49,6 +49,25 @@ if(AMReX_GPU_BACKEND MATCHES "CUDA")
...
+      COMMAND hipcc --version
+      RESULT_VARIABLE hipcc_result
+      OUTPUT_VARIABLE hipcc_version
+      ERROR_VARIABLE hipcc_error
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+  if(NOT hipcc_result EQUAL 0)
+      message(FATAL_ERROR "Failed to detect HIP compiler version: ${hipcc_error}")
+  else()
+      message(STATUS "HIP compiler version: ${hipcc_version}")
```
BenWibking: `hipcc_version` is the string `HIP version: 6.3.42133-1b9c17779`. We need to parse this string.

markkrumholz (approved):

PR #846, opened 2025-01-12
Add midpoint convergence condition to Newton-Raphson iteration for radiation-matter coupling

chongchonghe:
### Description  
This introduces a midpoint convergence condition for radiation-matter coupling to terminate the Newton-Raphson iteration when the average of two consecutive steps shows a relative change of less than $10^{-8}$. 

This was discussed long time ago but was never implemented because it was not helpful for any of the previous tests. However, while incorporating radiation into Aditi's QED simulation, I observed that when the heat capacity of matter is significantly lower than that of radiation, the solution oscillates and fails to converge, so this midpoint convergence check is necessary. At $T = 10^5 \ K, n = 1~{\rm cm}^{-3}$, with radiation in thermal equilibrium with matter, the heat capacity of radiation ($a_R T^3$) is 16 orders of magnitude greater than that of matter. Any tiny temperature change would cause the source term to fluctuate significantly, resulting in unstable behavior. 

### Related issues
None.

### Checklist


BenWibking:
I don't think this is a good idea. If it converges slowly, it could pass this test but still be far away from the true solution. I think the convergence check always needs to be on the true residual.

BenWibking:
Ok, I will amend that statement. It is okay to use the relative difference between consecutive iterations to detect whether convergence has stalled. However, it is *not* ok to use that to terminate the iteration entirely and declare victory. Instead, you need to restart either with a smaller timestep *or* a different initial guess *or* a different root-finding method.

BenWibking:
One other thing I can think of that would be more robust would be to write down another function $g(x) = 0$ that has the same roots as the function that you care about $f(x)=0$, but whose residual norm $\Vert f(x) \Vert$ does *not* scale with the heat capacity. Then test for convergence based on the residual of $g(x) = 0$, rather than $f(x) = 0$.

chongchonghe:
Another way is to demand `F0 < 1e-11` and check midpoint convergence on `E_g`, because in the situation I described, the gas temperature is well converged and it's only the radiation energy densities fluctuates. This way it's guaranteed the gas temperature is correct.

BenWibking:
> Another way is to demand `F0 < 1e-11` and check midpoint convergence on `E_g`, because in the situation I described, the gas temperature is well converged and it's only the radiation energy densities fluctuates. This way it's guaranteed the gas temperature is correct.

Right, it works for your problem, but AFAIK we don't know whether there's a different situation where this would lead to a false conclusion about convergence.

chongchonghe:
> > Another way is to demand `F0 < 1e-11` and check midpoint convergence on `E_g`, because in the situation I described, the gas temperature is well converged and it's only the radiation energy densities fluctuates. This way it's guaranteed the gas temperature is correct.
> 
> Right, it works for your problem, but AFAIK we don't know whether there's a different situation where this would lead to a false conclusion about convergence.

Convergence on the relative change is generally a good method, and some previous code does that (e.g. the Fornax code with tolerance of 1e-6). If the relative change between consecutive iterations is below 1e-6 and it's monotonically converging to the correct solution, it means it would require at least 1 million iterations. This should not happen in any realistic problem. The only way this could happen is when the guess solution is fluctuating around some value, and this value is *probably* the correct solution (I'll try to figure out a robust check for this).

BenWibking:
> Convergence on the relative change is generally a good method, and some previous code does that (e.g. the Fornax code with tolerance of 1e-6). If the relative change between consecutive iterations is below 1e-6 and it's monotonically converging to the correct solution, it means it would require at least 1 million iterations. This should not happen in any realistic problem. The only way this could happen is when the guess solution is fluctuating around some value, and this value is _probably_ the correct solution (I'll try to figure out a robust check for this).

The problem is that in general this method is not robust. I provided an example of a real radiation transport problem where this happens (it converges monotonically but can require an arbitrary number of iterations) in the Slack channel.

BenWibking:
Ok, the issue with this method is discussed in detail in section 3 of Shampine (1980): https://epubs.siam.org/doi/10.1137/0901005.

The problem is that you need to know both the rate of convergence $r$ and the relative difference in iterates in order to bound the difference from the exact solution (see Equation 13 of Shampine 1980). The problem is that a rigorous bound on the rate of convergence is not computable from the iterates. Nonetheless, Shampine advocates that a reasonable robust estimate of the $r$ can be computed, if you do so carefully. Can you try implementing the algorithm discussed there?

chongchonghe:
It turns out that this is not necessary if we turn on imperfect dust-gas interaction. The situation I described above happens when the temperature is very high and the gas density is very low, in which case the dust is almost certainly poorly coupled to the gas. The dedicated solver for poorly coupled dust-gas-radiation system ( #733 ) would avoid convergence problem.

BenWibking:
Do you have a single-cell test problem for this case? I would like to save this away for later to look more deeply into why it fails to converge. I suspect there may be a numerical precision issue lurking here (because double precision cannot represent more than ~15 orders of magnitude in addition, e.g. 1 + 1e-16 = 1 in `double`).

For reference:
* https://herbie.uwplse.org/demo/1cfdd1990d42e8a612051f31af26f699996dd07a.1a83c9f5186db102b678a2d5f869f99992469071/graph.html
* https://github.com/uwplse/herbgrind
* https://herbgrind.ucsd.edu/using-herbgrind.html

BenWibking:
@chongchonghe Is this PR needed anymore? If not, can you close it?

BenWibking:
Can we close this PR, since it does not appear to be needed?

BenWibking:
I'm going to close this PR for now. If it's needed later, feel free to re-open.

PR #857, opened 2025-01-21
Temporary workaround for gcc 14 libstdc++ incompatibility with HIP

### LINKED ISSUE #856: gcc 14 C++ headers have unresolved incompatibilities with HIP
**Describe the bug**
There are lots of changes in the GCC 14 C++ headers that break HIP builds. These are usually debugging functions in math functions, but there are some other examples, such as in `operator*` for `std::optional` objects. Since the debugging functions are host-only, whenever one of these function is called from device code, the compiler aborts with an error.

**To Reproduce**
1. Use `std::clamp` or `operator*` for an `std::optional` object in device code and compile for AMDGPU on a system that bundles GCC 14 as the system compiler.
2. See error below.

```
2025-01-21T00:28:55.8893422Z /usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/optional:482:2: error: reference to __host__ function '__glibcxx_assert_fail' in __host__ __device__ function
2025-01-21T00:28:55.8894500Z   482 |         __glibcxx_assert(this->_M_is_engaged());
2025-01-21T00:28:55.8895148Z       |         ^
2025-01-21T00:28:55.8895962Z /usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/x86_64-linux-gnu/c++/14/bits/c++config.h:612:7: note: expanded from macro '__glibcxx_assert'
2025-01-21T00:28:55.8897145Z   612 |         std::__glibcxx_assert_fail();                                   \
2025-01-21T00:28:55.8897673Z       |              ^
2025-01-21T00:28:55.8898316Z /usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/optional:968:22: note: called by 'operator*'
2025-01-21T00:28:55.8899038Z   968 |       { return this->_M_get(); }
2025-01-21T00:28:55.8899565Z       |                      ^
2025-01-21T00:28:55.8900235Z /home/runner/work/quokka/quokka/src/hydro/EOS.hpp:327:27: note: called by 'ComputePressure'
2025-01-21T00:28:55.8901001Z   327 |                 const auto &massArray = *massScalars;
2025-01-21T00:28:55.8901445Z       |                                         ^
```

BenWibking:
### Description
Temporarily workaround incompatibilities between the C++ standard library headers shipped with GCC 14 and HIP by switching to using the C++ standard library headers shipped with Clang (an independent implementation).

This may not be viable in production, since (in principle) any third-party libraries that use C++ should be complied against the same C++ standard library in order to ensure compatibility, but it will at least allow the CI to pass again and unblock several PRs. Currently, the only C++ library dependency that is not built simultaneously alongside Quokka is ADIOS2, which is optional.

### Related issues
Fixes https://github.com/quokka-astro/quokka/issues/856.

### Checklist

### DIFF
--- a/.github/workflows/hip.yml
+++ b/.github/workflows/hip.yml
@@ -60,11 +60,8 @@ jobs:
             -DCMAKE_C_COMPILER=$(which clang)             \
             -DCMAKE_CXX_COMPILER=$(which clang++)         \
             -DCMAKE_CXX_STANDARD=17                       \
-            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-        #    -DCMAKE_CXX_FLAGS=-D_GLIBCXX_NO_ASSERTIONS    \
-        #    -DCMAKE_HIP_FLAGS=-D_GLIBCXX_NO_ASSERTIONS
-        # this is disabled, but may be re-enabled in the future:
-        #   -DCMAKE_CXX_FLAGS=-stdlib=libc++
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache          \
+            -DCMAKE_CXX_FLAGS=-stdlib=libc++
         cmake --build build -j 2
 
         ccache -s
--- a/.github/workflows/hip.yml
+++ b/.github/workflows/hip.yml
@@ -60,8 +60,9 @@ jobs:
             -DCMAKE_C_COMPILER=$(which clang)             \
             -DCMAKE_CXX_COMPILER=$(which clang++)         \
             -DCMAKE_CXX_STANDARD=17                       \
-            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache          \
-            -DCMAKE_CXX_FLAGS=-stdlib=libc++
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+        # this is disabled, but may be re-enabled in the future:
+        #   -DCMAKE_CXX_FLAGS=-stdlib=libc++
         cmake --build build -j 2
 
         ccache -s
--- a/.github/workflows/hip.yml
+++ b/.github/workflows/hip.yml
@@ -61,7 +61,8 @@ jobs:
             -DCMAKE_CXX_COMPILER=$(which clang++)         \
             -DCMAKE_CXX_STANDARD=17                       \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache          \
-            -DCMAKE_CXX_FLAGS=-D_GLIBCXX_NO_ASSERTIONS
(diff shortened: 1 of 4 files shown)


BenWibking:
Weiqun suggests adding `-D_GLIBCXX_NO_ASSERTIONS` to the compile flags:
https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_macros.html

Strangely, adding `-DCMAKE_CXX_FLAGS=-D_GLIBCXX_NO_ASSERTIONS` does not appear to work.

BenWibking:
~~See also: https://github.com/ROCm/composable_kernel/issues/1405#issuecomment-2316076040~~

Edit: this doesn't really work for us, since we have templated everything, almost every file includes everything else.

BenWibking:
Adding `-DCMAKE_CXX_FLAGS=-stdlib=libc++` to the CMake build flags is another verified workaround.

.github/workflows/hip.yml
```diff
@@ -60,7 +60,10 @@ jobs:
             -DCMAKE_C_COMPILER=$(which clang)             \
             -DCMAKE_CXX_COMPILER=$(which clang++)         \
             -DCMAKE_CXX_STANDARD=17                       \
-            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache          \
+            -DCMAKE_CXX_FLAGS=-D_GLIBCXX_NO_ASSERTIONS
```
BenWibking: Strangely, this did not work.

src/problems/PrimordialChem/test_primordial_chem.cpp
```diff
@@ -6,6 +6,10 @@
 /// \file test_primordial_chem.cpp
 /// \brief Defines a test problem for primordial chemistry (microphysics).
 ///
+#ifdef __HIP_DEVICE_COMPILE__
+#undef _GLIBCXX_ASSERTIONS
+#endif
+
```
BenWibking: This is an alternative workaround. Unfortunately, this preprocessor magic needs to be placed at the top of every *.cpp file that has compile errors.

.github/workflows/hip.yml
BenWibking: We need `-DCMAKE_HIP_FLAGS=-D_GLIBCXX_NO_ASSERTIONS`.

src/problems/PrimordialChem/test_primordial_chem.cpp
BenWibking: For some reason, this did not work.

markkrumholz (approved):

PR #896, opened 2025-02-21
Particle creator

chongchonghe:
### Description

The `createCICParticles` method in `PhysicsParticles.hpp` provides a framework for particle creation on the fly from fluid cells. We run `createCICParticles` in the end of each hydro timestep. It uses two key components:

1. `CICParticleChecker`: A functor that determines whether a particle should be created in a given cell based on:
   - State-based conditions (can be customized to check fluid properties)
   - Time conditions 
   - Spatial conditions (grid spacing)

2. `CICParticleCreator`: A functor that handles the actual particle creation and initialization, including:
   - Setting particle position (e.g. cell-centered)
   - Assigning particle properties (mass, velocities)
   - Conserving mass and momentum between fluid and particles
   - Updating the fluid state after particle creation

This separation of checker and creator allows for easy customization of both when/where particles are created and how they are initialized. Currently, `CICParticleChecker` is set to return false, meaning no particles are created. `CICParticleCreator` is left empty. In `gravity_3d.cpp`, I showed an example of how to override the default checker and creator.

This implementation is validated by the `gravity_3d` test on CPU, GPU, and MPI, which demonstrates:
1. correct orbital dynamics of the CIC particles under self-gravity.
2. gravitational interaction between fluid, CIC particles, and CICRadParticles.
3. energy conservation of radiation energy in the presence of RadParticles.
4. creation of CICParticles from fluid cells on the fly.

#### Future

1. Particle creation with access to neighbor cells can be supported by adding ghost cells to the multi-fab. 
2. Particle-fluid interaction can be implemented similarly with very little extra work. 
3. More work needs to be done to add non-gravitational particle-particle interaction. `NeighborParticleContainer` does not have a AMR implementation, as far as I know. 

### Related issues
None.

### Checklist


src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +330,87 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+					const auto &state_arr = state.array(mfi);
+					const auto &geom = container_->Geom(lev);
+					const auto dx = geom.CellSizeArray();
+					const auto plo = geom.ProbLoArray();
+
+					// Create temporary buffer for this tile's particles
+					amrex::Gpu::DeviceVector<typename ContainerType::ParticleType> new_particles;
+
+					// Fill the buffer with new particles
+					amrex::ParallelFor(box, [=, &new_particles] AMREX_GPU_DEVICE(int i, int j, int k) {
```
BenWibking: I don't think this will work on the GPU.

It is necessary to follow what WarpX does and first count the number of new particles to be created, then on the CPU, resize the particle tile, then in a separate ParallelFor, fill the new particles with the appropriate data.

See here: https://github.com/ECP-WarpX/WarpX/blob/346bebdb27928c1acad892c4ffee5251f3c9d5f5/Source/Particles/PhysicalParticleContainer.cpp#L1007

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +330,87 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+			p.pos(0) = x;
+			p.pos(1) = y;
+			p.pos(2) = z;
+			p.rdata(mass_idx) = 0.5 * cell_mass;	   // Half the cell's mass
+			p.rdata(mass_idx + 1) = px / cell_density; // Cell's velocity
+			p.rdata(mass_idx + 2) = py / cell_density;
+			p.rdata(mass_idx + 3) = pz / cell_density;
+
+			// Add particle to buffer
+			new_particles.push_back(p);
```
BenWibking: This cannot be called on the GPU.

src/particles/PhysicsParticles.hpp
chongchonghe: Done.

src/particles/PhysicsParticles.hpp
chongchonghe: Also solved.

chongchonghe:
Now, particle creation should work just fine on GPU. In `createCICParticles`, we count the number of particles to be created in each box, allocate memory for those particles, and finally loop through all the cells and create particles.

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +330,109 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+						amrex::Gpu::Buffer<int> particle_count({0});
+						// Get raw pointer to GPU counter for use in device code
+						int *p_particle_count = particle_count.data();
+
+						amrex::ParallelFor(box, [=] AMREX_GPU_DEVICE(int i, int j, int k) {
+							if (is_create_particle(state_arr, i, j, k, plo, dx, current_time, dt, particle_creation_time_1,
+									       particle_creation_time_2)) {
+
+								// Get particle index using atomic add
+								const int idx = amrex::Gpu::Atomic::Add(&p_particle_count[0], 1); // NOLINT
```
BenWibking: This is not a good idea. This will be _very_ slow if a large number of particles are created.

It is a better idea to call `amrex::Scan::ExclusiveSum()` before this kernel in order to pre-calculate the indices: https://github.com/ECP-WarpX/WarpX/blob/2e0f9ab6aa903cea5a8d6ed4d93f0742a7e71eee/Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H#L116

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +330,109 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+					const auto plo = geom.ProbLoArray();
+
+					// count particles in this box
+					amrex::Gpu::Buffer<amrex::Long> num_new_particles({0});
+					amrex::Long *p_num_new_particles = num_new_particles.data();
+
+					amrex::ParallelFor(box, [=] AMREX_GPU_DEVICE(int i, int j, int k) {
+						if (is_create_particle(state_arr, i, j, k, plo, dx, current_time, dt, particle_creation_time_1,
+								       particle_creation_time_2)) {
+							amrex::Gpu::Atomic::Add(&p_num_new_particles[0], 1L); // NOLINT
```
BenWibking: Atomics are slow on the GPU. This is not good if a large number of particles are created.

It is probably better to follow what WarpX does and create a temporary FAB of integers, where each value is the number of particles (or a `bool`) to be created in a given cell. This requires more memory, but it should be much faster.

Example: https://warpx.readthedocs.io/en/23.06/developers/qed.html#_CPPv4I_i0000000E28filterCreateTransformFromFAB5IndexR7DstTileR7DstTileKN5amrex3BoxERK4FABsK5IndexK5IndexRR10FilterFuncRR11CreateFunc1RR11CreateFunc2RR9TransFunc

BenWibking (changes requested):
There is a better way to do this that doesn't require using atomic operations. See comments.

src/particles/PhysicsParticles.hpp
chongchonghe: Oh, actually that’s what I did in the previous version (one commit back). I thought atomic operations would be faster and save some memory. since atomic operations are slow, i'll revert back to ExclusiveSum instead.

BenWibking: I think the WarpX developers tried both ways and found this was significantly faster. But you can always benchmark it yourself to be sure. It may have changed over the past few years.

chongchonghe:
The star_cluster.cpp test runs normally on a single CPU or GPU, but fails with the shown error message when using MPI + CPU, even with a single process `mpirun -np 1`. @BenWibking Do you have any idea what've caused that?

```
Performance figure-of-merit: 4.863445404 μs/zone-update [0.2056155497 Mupdates/s]
Zone-updates on level 0: 98304

Writing plotfile plt00003
Writing checkpoint chk00003
mpirun has exited due to process rank 0 with PID 0 on
node moth exiting improperly. There are three reasons this could occur:

1. this process did not call "init" before exiting, but others in
the job did. This can cause a job to hang indefinitely while it waits
for all processes to call "init". By rule, if one process calls "init",
then ALL processes must call "init" prior to termination.

2. this process called "init", but exited without calling "finalize".
By rule, all processes that call "init" MUST call "finalize" prior to
exiting or it will be considered an "abnormal termination"

3. this process called "MPI_Abort" or "orte_abort" and the mca parameter
orte_create_session_dirs is set to false. In this case, the run-time cannot
detect that the abort call was an abnormal termination. Hence, the only
error message you will receive is this one.

This may have caused other processes in the application to be
terminated by signals sent by mpirun (as reported here).

You can avoid this message by specifying -quiet on the mpirun command line.
```

BenWibking:
> The star_cluster.cpp test runs normally on a single CPU or GPU, but fails with the shown error message when using MPI + CPU, even with a single process `mpirun -np 1`. @BenWibking Do you have any idea what've caused that?
> 
> ```
> Performance figure-of-merit: 4.863445404 μs/zone-update [0.2056155497 Mupdates/s]
> Zone-updates on level 0: 98304
> 
> Writing plotfile plt00003
> Writing checkpoint chk00003
> --------------------------------------------------------------------------
> mpirun has exited due to process rank 0 with PID 0 on
> node moth exiting improperly. There are three reasons this could occur:
> 
> 1. this process did not call "init" before exiting, but others in
> the job did. This can cause a job to hang indefinitely while it waits
> for all processes to call "init". By rule, if one process calls "init",
> then ALL processes must call "init" prior to termination.
> 
> 2. this process called "init", but exited without calling "finalize".
> By rule, all processes that call "init" MUST call "finalize" prior to
> exiting or it will be considered an "abnormal termination"
> 
> 3. this process called "MPI_Abort" or "orte_abort" and the mca parameter
> orte_create_session_dirs is set to false. In this case, the run-time cannot
> detect that the abort call was an abnormal termination. Hence, the only
> error message you will receive is this one.
> 
> This may have caused other processes in the application to be
> terminated by signals sent by mpirun (as reported here).
> 
> You can avoid this message by specifying -quiet on the mpirun command line.
> ```

Definitely weird. Unfortunately, I haven't seen it fail like this before, so I have no idea.

chongchonghe:
This MPI error exists in the development branch before the new particle architecture is implemented, so I am gonna leave it aside for now.

chongchonghe:
Particle creator is now fully working on CPU, GPU, and MPI. 

The MPI bug I reported above (which I deleted) was a mistake on my side, but it also revealed MPI failure on some tests. In this PR, I'm using the Gravity3D test which doesn't have this issue, so I'm gonna ignore it for now and reported it as a issue ( #908 ).

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +424,116 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+	{
+		if (container_ != nullptr) {
+			const int mass_idx = this->getMassIndex();
+			if (mass_idx >= 0 && mass_idx + 3 < ContainerType::ParticleType::NReal) {
+				const double particle_creation_time_1 = this->getParam1();
+				const double particle_creation_time_2 = this->getParam2();
+
+				for (amrex::MFIter mfi = container_->MakeMFIter(lev); mfi.isValid(); ++mfi) {
+					// TODO(cch): I don't know whether this is needed. WarpX has this.
+					amrex::Gpu::synchronize();
```
BenWibking: This isn't necessary. It's done in WarpX in order to compute the time it takes to process the particles in each FAB.

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +424,116 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+
+					const auto &box = mfi.validbox();
+					const auto &state_arr = state.array(mfi);
+					const auto &geom = container_->Geom(lev);
+					const auto dx = geom.CellSizeArray();
+					const auto plo = geom.ProbLoArray();
+
+					// Count particles to be created in this box
+					amrex::Gpu::DeviceVector<amrex::Long> counts(box.numPts(), 0); // 1 if cell creates particle, 0 if not
```
BenWibking: I think the type of this can be `amrex::Gpu::DeviceVector<unsigned int>` instead.

BenWibking: Also, assigning a default value is unnecessary, since you explicitly assign `0` if no particle should be created in the kernel below.

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +424,116 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+							p.rdata(mass_idx + 2) = state_arr(i, j, k, HydroSystem<problem_t>::x2Momentum_index) / cell_density;
+							p.rdata(mass_idx + 3) = state_arr(i, j, k, HydroSystem<problem_t>::x3Momentum_index) / cell_density;
+
+							// Update cell density
+							state_arr(i, j, k, HydroSystem<problem_t>::density_index) = 0.5 * cell_density;
+						}
+					});
+
+					// TODO(cch): I don't know whether this is needed. WarpX has this.
+					amrex::Gpu::synchronize();
```
BenWibking: It's not needed. See comment above.

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +424,116 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+					amrex::Gpu::DeviceVector<amrex::Long> offset(box.numPts());    // Will store starting index for each cell's particle
+					auto *pcounts = counts.data();
+
+					// Count potential particles per cell
+					amrex::ParallelFor(box, [=] AMREX_GPU_DEVICE(int i, int j, int k) {
+						const amrex::IntVect iv(AMREX_D_DECL(i, j, k));
+						const auto index = box.index(iv);
+
+						// Check if we should create a particle at this location and time
+						const bool should_create = is_create_particle(state_arr, i, j, k, plo, dx, current_time, dt,
```
BenWibking: In the production implementation, it would be good to replace `is_create_particle` with a functor or lambda function that the user can provide.

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +424,116 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+							const amrex::Real cell_density = state_arr(i, j, k, HydroSystem<problem_t>::density_index);
+							const amrex::Real cell_mass = cell_density * cell_volume;
+
+							p.rdata(mass_idx) = 0.5 * cell_mass;
+							p.rdata(mass_idx + 1) = state_arr(i, j, k, HydroSystem<problem_t>::x1Momentum_index) / cell_density;
+							p.rdata(mass_idx + 2) = state_arr(i, j, k, HydroSystem<problem_t>::x2Momentum_index) / cell_density;
+							p.rdata(mass_idx + 3) = state_arr(i, j, k, HydroSystem<problem_t>::x3Momentum_index) / cell_density;
+
+							// Update cell density
+							state_arr(i, j, k, HydroSystem<problem_t>::density_index) = 0.5 * cell_density;
```
BenWibking: It should be possible for the user to provide a functor or lambda function to set the particle properties (and fluid properties in the same cell).

src/particles/PhysicsParticles.hpp
chongchonghe: My idea is to provide a default `is_create_particle` and `create_particle_and_update_fluid` function and the user can overwrite it.

chongchonghe: OK. Functor makes more sense here. I'll implement it right now.

chongchonghe: I've implemented two functors: particle_checker and particle_creator

src/particles/PhysicsParticles.hpp
chongchonghe: I've implemented two functors: particle_checker and particle_creator

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +490,71 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
 		}
 	}
 
+	// Implementation of CIC particle creation
+	void createCICParticles(amrex::MultiFab &state, int lev, amrex::Real current_time, amrex::Real dt) const override
+	{
+		if (container_ != nullptr) {
+			const int mass_idx = this->getMassIndex();
+			if (mass_idx >= 0 && mass_idx + 3 < ContainerType::ParticleType::NReal) {
+				CICParticleChecker<problem_t> particle_checker(this->getParam1(), this->getParam2());
```
BenWibking: Is the idea that users will then specialize the `CICParticleChecker` for their own problems?

src/particles/PhysicsParticles.hpp
```diff
@@ -323,6 +490,71 @@ template <typename ContainerType, ParticleType particleType> class PhysicsPartic
...
+					const int old_size = aos.size();
+					aos.resize(old_size + max_new_particles);
+
+					// Create the particles
+					auto *poffset = offset.data();
+					auto *pdata = aos.data() + old_size;
+					const int cpu_id = amrex::ParallelDescriptor::MyProc();
+
+					// Initialize particle creator functor
+					CICParticleCreator<problem_t> particle_creator(mass_idx, cpu_id, pid);
```
BenWibking: What is the idea for how users will customize this? Is the expectation that this will be specialized in each problem file that needs particle creation?

src/particles/PhysicsParticles.hpp
```diff
@@ -165,11 +167,13 @@ class PhysicsParticleDescriptorBase
 	int massIndex_{-1};		 // Index for particle mass (-1 if not used)
 	int lumIndex_{-1};		 // Index for radiation luminosity (-1 if not used)
 	int birthTimeIndex_{-1};	 // Index for birth time (-1 if not used)
+	double param1_{-1.0};		 // First particle parameter
+	double param2_{-1.0};		 // Second particle parameter
```
BenWibking: Are you planning to add all new particle properties here? Or can they go in subclasses?

BenWibking: Is it necessary to add these here just for the test problem?

src/particles/PhysicsParticles.hpp
```diff
@@ -165,11 +167,13 @@ class PhysicsParticleDescriptorBase
...
 	bool interactsWithHydro_{false}; // Whether particles interact with hydrodynamics
 
       public:
-	PhysicsParticleDescriptorBase(int mass_idx, int lum_idx, int birth_time_idx, bool hydro_interact)
-	    : massIndex_(mass_idx), lumIndex_(lum_idx), birthTimeIndex_(birth_time_idx), interactsWithHydro_(hydro_interact)
+	PhysicsParticleDescriptorBase(int mass_idx, int lum_idx, int birth_time_idx, double param1, double param2, bool hydro_interact)
+	    : massIndex_(mass_idx), lumIndex_(lum_idx), birthTimeIndex_(birth_time_idx), param1_(param1), param2_(param2), interactsWithHydro_(hydro_interact)
```
BenWibking: Is it necessary to modify the constructor just for the test problem?

src/particles/PhysicsParticles.hpp
```diff
@@ -185,11 +189,16 @@ class PhysicsParticleDescriptorBase
 	[[nodiscard]] AMREX_GPU_HOST_DEVICE auto getMassIndex() const -> int { return massIndex_; }
 	[[nodiscard]] AMREX_GPU_HOST_DEVICE auto getLumIndex() const -> int { return lumIndex_; }
 	[[nodiscard]] AMREX_GPU_HOST_DEVICE auto getBirthTimeIndex() const -> int { return birthTimeIndex_; }
+	[[nodiscard]] AMREX_GPU_HOST_DEVICE auto getParam1() const -> double { return param1_; }
+	[[nodiscard]] AMREX_GPU_HOST_DEVICE auto getParam2() const -> double { return param2_; }
```
BenWibking: Same issue as before -- does this class need to be modified just for the test problem?

BenWibking (commented):
I think this PR works well as an example. However, is this PR something you want to merge, or was this just intended as an example for other PRs?

In either case, it's somewhat unclear to me how the particle creation functors and `param1` and `param2` generalize to other particle types or problems. Does this need to be hard-coded for each particle creation model in the way you've currently modified `PhysicsParticleDescriptorBase` and `PhysicsParticleDescriptor`?

More generally, what is your vision for how users should go about adding particle creation for their specific problem or specific particle type? Can you handle the case where there are multiple possible particle creation methods for a single particle type, and users want to be able to choose between them for their problem (either at compile-time or runtime)? This might happen in practice, for say, star particles in a galaxy or cosmological simulation.

chongchonghe:
My idea, for the final version, is to provide a default `ParticleCreator` routine coded in PhysicsParticles.hpp for each particle types that supports particle creation. The user can choose to override `ParticleCreator` in the problem file if they want. An `InteractWithGas()` can be defined similarly. 

Here's what I've done in this PR:
I added a Particle_Traits to control which whether or not we want particle creation on the fly. I have left `ParticleCreator` empty for now and let `CICParticleChecker` return `false`. An example of overriding `CICParticleCreator` by the user is available in gravity_3d.cpp. 

I don't have plan to work on sink formation/accretion recipe yet, so I'll probably pause it here. I am going to work on thermochemistry and iRSLA. psli, Aditi, or others can start from here and implement a robust star formation/sink particle scheme.

@BenWibking @markkrumholz  If you like this design and this implementation, we can merge it into the dev branch so future development on sink particles can start from an up-to-date brach. Or, we can leave it as a draft. I'm fine with either way.

src/particles/PhysicsParticles.hpp
chongchonghe: Good point. I have moved these two parameters to the function call:
```cpp
particleRegister_.createCICParticles(state_new_cc_[0], 0, cur_time, dt_[0], particle_creator_param1, particle_creator_param2);
```

src/particles/PhysicsParticles.hpp
chongchonghe: See my reply to previous comments. I have moved them.

chongchonghe:
> More generally, what is your vision for how users should go about adding particle creation for their specific problem or specific particle type? Can you handle the case where there are multiple possible particle creation methods for a single particle type, and users want to be able to choose between them for their problem (either at compile-time or runtime)? This might happen in practice, for say, star particles in a galaxy or cosmological simulation.

I showed an example of overriding ParticleCreator in the problem file in gravity_3d.cpp. By doing so, the user can impose multiple particle creation methods for a single particle type either at compile time or runtime (via if clause inside the function). 

We can also do this. We provide a default particle creator, call it `StarClusterParticleCreator`. Then, we pass it to `CICParticleCreator` in `PhysicsParticles.hpp`, which defines the default CIC particle creation behaviour. The user can override the scheme in the problem file by overriding `CICParticleCreator`, which can call `StarClusterParticleCreator`, or some other creator provided by Quokka, or any other creator the user defines.

BenWibking (approved):
Looks good to me!

BenWibking:
I think it would be good to merge this as-is. Then we have a tested infrastructure for actual physics-based particle creation.

PR #911, opened 2025-02-27
Compile-time particle switch with Bitwise Operations

### LINKED ISSUE #815: Add PhysicsParticle architecture
**Describe the proposal**
This is a discussion of implementing a general, reusable particle architecture.

**Additional context**

chongchonghe:
### Description

This PR introduces a flexible architecture for particle creation, implementing a type-specific framework for handling multiple particle types with specialized behaviors.

##### Type-Safe Particle Switching

- Implemented a type-safe bitwise operation system for controlling particle types
- Added `ParticleSwitch` enum class with proper bitflag operations
- Enables combining multiple particle types in a type-safe manner

```cpp
enum class ParticleSwitch : unsigned int {
    None = 0U,           // No particles, = 0b0000
    CIC = bitflag<1>(),  // Cloud-In-Cell (gravitating) particles, = 0b0001
    Rad = bitflag<2>(),  // Radiation particles, = 0b0010
    CICRad = bitflag<3>() // Combined gravitating-radiating particles, = 0b0100
};
```

##### Improved Particle Traits System

- Added `Particle_Traits` template for configuring particle types per problem
- Simple interface for enabling multiple particle types:

```cpp
template <> struct Particle_Traits<BinaryOrbit> {
    static constexpr ParticleSwitch particle_switch = ParticleSwitch::CIC | ParticleSwitch::Rad;
};
```

- Type checking prevents common errors like using wrong enum types or raw integers

##### Flexible Particle Creation Framework

- Introduced `ParticleCreationTraits` to solve C++ partial template specialization limitations
- Each particle type can now have specialized creation behavior while maintaining the problem_t template
- Modular design with separate checker and creator components per particle type

#### Implementation Details about particle creation

The architecture uses a traits-based approach with specialized components:

1. **ParticleCreationTraits**: Template class for particle creation specialized for each particle type
2. **ParticleChecker & ParticleCreator**: Problem-specific implementations
3. **ParticleCreationImpl**: Common implementation logic for particle creation, including CPU, GPU, and MPI-related operations
4. **PhysicsParticleDescriptor::createParticlesFromState**: Interface for particle creation from simulation state for a specific particle type
5. **PhysicsParticleRegister::createParticlesFromState**: Registry of all particle types and their creation methods

#### Usage Example

To enable particle types in your problem:

```cpp
// Enable multiple particle types in your problem
template <> struct Particle_Traits<MyProblem> {
    static constexpr ParticleSwitch particle_switch = ParticleSwitch::CIC | ParticleSwitch::Rad;
};
```

To define creation routine for the particle types that support creation:

```cpp
namespace quokka
{
// Specialization for CIC particles
template <> struct ParticleCreationTraits<ParticleType::CIC> {
	// Specialized nested ParticleChecker for CIC particles
	template <typename problem_t> struct ParticleChecker {
		amrex::Real param1 = particle_param1;
		amrex::Real param2 = particle_param2;

		AMREX_GPU_DEVICE auto operator()(amrex::Array4<const amrex::Real> const &state_arr, int i, int j, int k,
						 amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &dx, amrex::Real current_time, amrex::Real dt) const -> bool
		{
			// Check if we should create a particle at this location and time
		}
	};

	// Specialized nested ParticleCreator for CIC particles
	template <typename problem_t> struct ParticleCreator {
		int mass_idx;
		int cpu_id;
		amrex::Long pid_start;
		amrex::Real param1 = particle_param1;
		amrex::Real param2 = particle_param2;

		AMREX_GPU_HOST_DEVICE
		ParticleCreator(int mass_index, int processor_id, amrex::Long particle_id_start)
		    : mass_idx(mass_index), cpu_id(processor_id), pid_start(particle_id_start)
		{
		}

		template <typename ParticleType, typename StateArray>
		AMREX_GPU_DEVICE void operator()(ParticleType &p, StateArray const &state_arr, int i, int j, int k,
						 amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &dx,
						 amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &plo, amrex::Long particle_offset) const
		{
			// Create a particle at the given location and time
		}
	};

	// Main method to create particles - uses the helper implementation
	template <typename problem_t, typename ContainerType>
	static void createParticles(ContainerType *container, int mass_idx, amrex::MultiFab &state, int lev, amrex::Real current_time, amrex::Real dt)
	{
		// Use the common implementation with our checker and creator types
		ParticleCreationImpl::createParticlesImpl<problem_t, ContainerType, ParticleCreationTraits<ParticleType::CIC>::template ParticleChecker,
							  ParticleCreationTraits<ParticleType::CIC>::template ParticleCreator>(container, mass_idx, state, lev,
															       current_time, dt);
	}
};
} // namespace quokka
```

### Related issues
Closes https://github.com/quokka-astro/quokka/issues/815.

### Checklist


src/particles/PhysicsParticles.hpp
```diff
@@ -23,6 +23,28 @@
...
+
+// This struct should be specialized by the user application code to configure particle behavior.
+// Determines which particle types are enabled using bitwise flags.
+// Examples:
+// - particle_switch = 0                     -> No particles enabled
+// - particle_switch = ParticleSwitch::CIC   -> Only CIC particles
+// - particle_switch = ParticleSwitch::CIC | ParticleSwitch::Rad -> Both CIC and Rad particles
+template <typename problem_t> struct Particle_Traits {
+	static constexpr int particle_switch = 0;		    // Determines which particle types are enabled using bitwise flags.
+	static constexpr bool is_particle_creation_enabled = false; // Controls whether particles can be created during simulation
```
markkrumholz: Is there a way to not have to turn particle creation on and off separately, but instead deduce whether particle creation is enabled based on the particle types present? For example, one way I could see of implementing that would be to define a new integer constant that has one bit assigned to each particle type, as in particle_switch, and then has a 0 or 1 in that bit depending on whether that particle type allows particle creation. One can then determine if particle creation is turned on in a given run just by doing a bitwise and between the value of particle_switch and that integer.

chongchonghe: The idea is to allow situations where I want CIC particles at the beginning and not creating it during the simulation, as well as situations where I want CIC particles to create on the fly. Particle traits (such as Nreal, mass_index, lum_index, etc) could operate independently from particle creator. However, the limitation of the current approach is that we can only toggle particle creation of ALL particles at once. Once way to resolve this limitation is to make `is_particle_creation_enabled` an integer like `particle_switch` and implement it as `is_particle_creation_enabled = ParticleSwitch::CIC | ParticleSwitch::Rad`. How does that sound?

chongchonghe: Or, is there a consideration that most particle types will need to support particle creation, making it easier to attach a creation switch directly to each particle type?

markkrumholz: Thinking about it more, we probably don't even need a separate switch at all, as long as each particle type contains information on whether it provides a particle creation routine. Then it becomes exactly like mass deposition: you loop over particle types, and if a particle type declares that it supports creation, you invoke its creation routine. There is no cost to doing this because you don't execute a loop over multifabs for particle types that declare they don't have a creation routine.

My general thinking is that 90% of the time whether a particle type should be created is associated with the nature of the particle, and isn't going to vary from problem to problem. That is, situations in which one will want to create CIC particles on the fly are vanishingly rare, and we shouldn't organise the code around them. If people do want to do such a thing, then they can either overwrite the CIC particle class with an implementation for their particular problem_t, or they can derive a new particle type from CIC particles that provides a creation routine. But in general I think the rule about whether a particular particle type provides a creation method should be attached to the particle type, not be set by hand at either compile or run time. That covers by far the most common use-cases, and avoids the errors that will inevitably happen if you have two different switches that each need to be set manually.

BenWibking:
I don't think this PR is a good idea. There should almost never be a need to do bitwise flag manipulation. The only exception is for things that happen in inner loops (i.e., inside ParallelFor) where there is no other way to do it performantly. That is not the case here.

I also agree with Mark that in most cases, it should just happen automatically based on particle type. There could, however, be a runtime parameter to disable particle creation on a particle type-by-type basis.

In particular, I think adding a runtime option to the PhysicsParticleRegister is the right place to put this.

markkrumholz:
I like the bitwise flag approach because it makes turning particles on and off extremely easy for users: they just add a line to the effect use_particles = cic_particles + stellar_pop_particles in their problem setup file, and it just automagically works.

Alternatives look somewhat ugly to me:

1. We could leave particle enabling each particle type as a runtime flag instead of a physics trait-like structure, as it stands now, but this breaks with the approach we have used for all other aspects of physics; we don't turn gravity or MHD on using run-time options, so why should particles be treated differently?
2. We could add physics traits for each particle type, but then as the number of particle types grows we wind up having to insert a very long physics trait at the start of every problem file, specifying exactly which types of particles are turned on and off for a given simulation. This is unwieldy if there are 20 different particle types available, which I could easily see there being at some point.

Given these alternatives, I thought that the bitwise flag option was the cleanest solution from a user perspective.

BenWibking:
> I like the bitwise flag approach because it makes turning particles on and off extremely easy for users: they just add a line to the effect use_particles = cic_particles + stellar_pop_particles in their problem setup file, and it just automagically works.

From a C++ implementation point of view, this can be done more cleanly without bitwise flags: you can instead create smart objects that override `operator+` and do the same thing, except without having to do low-level bit manipulation behind the scenes. This is clearer when reading the code, avoids bugs caused by accidentally setting the same bit twice, and works for more than 32 particle types.

> Alternatives look somewhat ugly to me:
> 
> 1. We could leave particle enabling each particle type as a runtime flag instead of a physics trait-like structure, as it stands now, but this breaks with the approach we have used for all other aspects of physics; we don't turn gravity or MHD on using run-time options, so why should particles be treated differently?

This is actually what I would propose, because I think we should move away from using physics traits altogether. This was not possible at first, but now AMReX has the `ParallelForCTO` mechanism that allows ParallelFors to take runtime parameters and then use them with `if constexpr` inside the ParallelFor (using template magic). So the original motivation for the Physics_Traits no longer exists. I think all new parameters should just be runtime parameters, and we should eventually convert the existing Physics_Traits to runtime parameters.

> 2. We could add physics traits for each particle type, but then as the number of particle types grows we wind up having to insert a very long physics trait at the start of every problem file, specifying exactly which types of particles are turned on and off for a given simulation. This is unwieldy if there are 20 different particle types available, which I could easily see there being at some point.

I agree that this option is a bad idea.

markkrumholz:
> > I like the bitwise flag approach because it makes turning particles on and off extremely easy for users: they just add a line to the effect use_particles = cic_particles + stellar_pop_particles in their problem setup file, and it just automagically works.
> 
> From a C++ implementation point of view, this can be done more cleanly without bitwise flags: you can instead create smart objects that override `operator+` and do the same thing, except without having to do low-level bit manipulation behind the scenes. This is clearer when reading the code, avoids bugs caused by accidentally setting the same bit twice, and works for more than 32 particle types.

I have no objection to an alternative solution based on C++ classes, although I think the resulting code will definitely be bulkier than the bitwise option. I guess I don't find bitwise flags particularly hard to read or understand, but that may just be because I cut my teeth as a C programmer.

> > Alternatives look somewhat ugly to me:
> > 
> > 1. We could leave particle enabling each particle type as a runtime flag instead of a physics trait-like structure, as it stands now, but this breaks with the approach we have used for all other aspects of physics; we don't turn gravity or MHD on using run-time options, so why should particles be treated differently?
> 
> This is actually what I would propose, because I think we should move away from using physics traits altogether. This was not possible at first, but now AMReX has the `ParallelForCTO` mechanism that allows ParallelFors to take runtime parameters and then use them with `if constexpr` inside the ParallelFor (using template magic). So the original motivation for the Physics_Traits no longer exists. I think all new parameters should just be runtime parameters, and we should eventually convert the existing Physics_Traits to runtime parameters.

This seems like a major code re-engineering, and I'm a bit wary about a solution that leaves particles treated differently than everything else on the basis that someday we will get around to converting everything else to work like particles.

chongchonghe:
> From a C++ implementation point of view, this can be done more cleanly without bitwise flags: you can instead create smart objects that override operator+ and do the same thing, except without having to do low-level bit manipulation behind the scenes. This is clearer when reading the code, avoids bugs caused by accidentally setting the same bit twice, and works for more than 32 particle types.

> "bugs caused by accidentally setting the same bit twice" 

This is already taken into consideration. I use bitwise OR instead of '+'. For example, `0b0001 | 0b0001 | 0b0100 = 0b0101`. There is no such danger here.

It's probably safe to assume we won't have more than 32 particle types. Even if we do get more than 32 types someday, we can also change to use Long and support 64 types. 

I do agree that the user may find this operation counter-intuitive, so if there is better and cleaner way, I'll use it.

chongchonghe:
> This is actually what I would propose, because I think we should move away from using physics traits altogether. This was not possible at first, but now AMReX has the ParallelForCTO mechanism that allows ParallelFors to take runtime parameters and then use them with if constexpr inside the ParallelFor (using template magic). So the original motivation for the Physics_Traits no longer exists. I think all new parameters should just be runtime parameters, and we should eventually convert the existing Physics_Traits to runtime parameters.

I don't quite understand how this could work. If you use runtime parameter `is_enable_rad` to switch on and off radiation, it's not possible to avoid compiling the whole radiation module (host functions which contains ParallelFor's) like what we could do with the current Physics_Traits approach. I might be missing something here; I have to read more about ParallelForCTO. However, even if it could work as you said, I agree with Mark that we should keep it consistent. 
When we want to switch to use ParallelForCTO, we make the change and turn all switches to runtime parameters. For now, I think constexpr fits the design better.

BenWibking:
> > From a C++ implementation point of view, this can be done more cleanly without bitwise flags: you can instead create smart objects that override operator+ and do the same thing, except without having to do low-level bit manipulation behind the scenes. This is clearer when reading the code, avoids bugs caused by accidentally setting the same bit twice, and works for more than 32 particle types.
> 
> 
> 
> > "bugs caused by accidentally setting the same bit twice" 
> 
> 
> 
> This is already taken into consideration. I use bitwise OR instead of '+'. For example, `0b0001 | 0b0001 | 0b0100 = 0b0101`. There is no such danger here.
> 
> 
> 
> It's probably safe to assume we won't have more than 32 particle types. Even if we do get more than 32 types someday, we can also change to use Long and support 64 types. 
> 
> 
> 
> I do agree that the user may find this operation counter-intuitive, so if there is better and cleaner way, I'll use it. 

The problem is we may end up having multiple bit flags for different things, and nothing prevents you from mixing them up. You might have bit flags for particles and a separate set of bit flags for physics modules to turn on (just as an example, not saying we want to do this) and the compile will let you nonsensically combine the two.

The other problem is that the compiler does not guarantee that each bit flag is unique. Someone could make a typo and create two bit flags for different particles that happen to accidentally use the same bit.

This can all be avoided with a bit flag wrapper class like this one: https://github.com/DmitrySokolov/bitflags

Or you could write you own version of this class.

markkrumholz:
> > > From a C++ implementation point of view, this can be done more cleanly without bitwise flags: you can instead create smart objects that override operator+ and do the same thing, except without having to do low-level bit manipulation behind the scenes. This is clearer when reading the code, avoids bugs caused by accidentally setting the same bit twice, and works for more than 32 particle types.
> > 
> > 
> > > "bugs caused by accidentally setting the same bit twice"
> > 
> > 
> > This is already taken into consideration. I use bitwise OR instead of '+'. For example, `0b0001 | 0b0001 | 0b0100 = 0b0101`. There is no such danger here.
> > It's probably safe to assume we won't have more than 32 particle types. Even if we do get more than 32 types someday, we can also change to use Long and support 64 types.
> > I do agree that the user may find this operation counter-intuitive, so if there is better and cleaner way, I'll use it.
> 
> The problem is we may end up having multiple bit flags for different things, and nothing prevents you from mixing them up. You might have bit flags for particles and a separate set of bit flags for physics modules to turn on (just as an example, not saying we want to do this) and the compile will let you nonsensically combine the two.
> 
> The other problem is that the compiler does not guarantee that each bit flag is unique. Someone could make a typo and create two bit flags for different particles that happen to accidentally use the same bit.
> 
> This can all be avoided with a bit flag wrapper class like this one: https://github.com/DmitrySokolov/bitflags
> 
> Or you could write you own version of this class.

This seems like a reasonable compromise that keeps the existing paradigm unified (turn physics on and off with compile-time options) while avoiding the issues @BenWibking raises with traditional C bit flags.

BenWibking:
A way to write your own class to do this would be to create a class that wraps `std::vector<bool>` where each element of the vector corresponds to a different bit. Then you can add member functions to set each bit uniquely, and have this guaranteed at compile time. Each set of bit flags would correspond to a different template parameter specialization for this object, eg `boolset<ParticleTypes>`

BenWibking:
> > This is actually what I would propose, because I think we should move away from using physics traits altogether. This was not possible at first, but now AMReX has the ParallelForCTO mechanism that allows ParallelFors to take runtime parameters and then use them with if constexpr inside the ParallelFor (using template magic). So the original motivation for the Physics_Traits no longer exists. I think all new parameters should just be runtime parameters, and we should eventually convert the existing Physics_Traits to runtime parameters.
> 
> I don't quite understand how this could work. If you use runtime parameter `is_enable_rad` to switch on and off radiation, it's not possible to avoid compiling the whole radiation module (host functions which contains ParallelFor's) like what we could do with the current Physics_Traits approach. I might be missing something here; I have to read more about ParallelForCTO. However, even if it could work as you said, I agree with Mark that we should keep it consistent. When we want to switch to use ParallelForCTO, we make the change and turn all switches to runtime parameters. For now, I think constexpr fits the design better.

You would have to compile everything. On the other hand, you would only have a compile a single binary for all problems, so the compile time for the test suite might be reduced (not guaranteed, of course). Then this opens the door to creating new problems without having to compile a new binary-- you could just set the runtime parameters and then the initial conditions with, e.g., a Python script, rather than a *.cpp problem file.

However, I realize I'm fighting a losing battle on this. I'm happy for this PR to just use a smart bitfield wrapper class.

markkrumholz:
> However, I realize I'm fighting a losing battle on this. I'm happy for this PR to just use a smart bitfield wrapper class.

Just to be clear: I'm not objecting to this as a long-term plan, and I can see the appeal. I'm just wary of starting that plan with this one particular piece of physics, because I worry that we won't get to where you're proposing in the immediate term, and we'll be left with a design that is half one thing and half the other for an extended period.

chongchonghe:
Seems like we come to an agreement. We'll use constexpr switches to turn on and off particle types at compile time, and use  type-safe bit operations to prevent using flags from different enums.

chongchonghe:
@BenWibking @markkrumholz Looks good to me now. Ready for your review.

src/particles/PhysicsParticles.hpp
```diff
@@ -16,9 +16,60 @@
...
+// Example: To enable both CIC and Rad particles, use:
+//   particle_switch = ParticleSwitch::CIC | ParticleSwitch::Rad  (= 0b00000011)
+// To check if CIC particles are enabled:
+//   if (particle_switch & ParticleSwitch::CIC) { ... }
+enum class ParticleSwitch : unsigned int {
+	None = 0U,	     // No particles
+	CIC = 0b00000001U,   // Cloud-In-Cell (gravitating) particles
+	Rad = 0b00000010U,   // Radiation particles
+	CICRad = 0b00000100U // Combined gravitating-radiating particles
+};
```
BenWibking: This requires setting the bit directly if future developers want to add new particle types, which is not ideal, because it's possible to accidentally re-use the same bit twice (or more).

It may be possible to automate this with a set of macros, like this implementation does: https://github.com/m-peko/bitflags?tab=readme-ov-file#quick-start

src/particles/PhysicsParticles.hpp
```diff
@@ -642,44 +695,67 @@ template <typename problem_t> class PhysicsParticleRegister
...
+	{
+		switch (type) {
+			case ParticleType::Rad:
+				return "Rad_particles";
+			case ParticleType::CIC:
+				return "CIC_particles";
+			case ParticleType::CICRad:
+				return "CICRad_particles";
+			default:
+				return "Unknown_particles";
```
BenWibking, line 709: This should probably call `amrex::Abort` if it reaches this case.

src/particles/PhysicsParticles.hpp
```diff
@@ -642,44 +695,67 @@ template <typename problem_t> class PhysicsParticleRegister
...
-			descriptor = std::make_unique<PhysicsParticleDescriptor<ContainerType, problem_t, ParticleType::Rad>>(mass_idx, lum_idx, birth_time_idx,
-															      hydro_interact, container);
-		}
+
+		// Create the appropriate descriptor based on the particle type
+		switch (type) {
+			case ParticleType::Rad:
+				descriptor = std::make_unique<PhysicsParticleDescriptor<ContainerType, problem_t, ParticleType::Rad>>(
+				    mass_idx, lum_idx, birth_time_idx, hydro_interact, allows_creation, container);
+				break;
```
BenWibking: I would prefer to avoid `switch`/`case` in Quokka in favor of using `if` / `else if`, because it is easy to forget the `break` statement after any one of the cases. If a developer forgets this, the behavior of the code is almost always not what is intended.

src/particles/PhysicsParticles.hpp
```diff
@@ -746,19 +822,33 @@ template <typename problem_t> class PhysicsParticleRegister
...
 	{
-		auto descriptor = getParticleDescriptor("CIC_particles");
-		if (descriptor != nullptr) {
-			descriptor->createCICParticles(state, lev, current_time, dt, param1, param2);
+		for (const auto &[type, descriptor] : particleRegistry_) {
+			// Only create particles if the descriptor allows creation
+			if (descriptor->getAllowsCreation()) {
+				// Dispatch to the appropriate particle type based on the particle type
+				switch (type) { // NOLINT
+					case ParticleType::CIC:
```
BenWibking: I thought the templated design of the PhysicsParticleDescriptor allows us to avoid all of these kinds of explicit enumerations of particle types (except possibly when creating each descriptor).

Can we instead provide a specialization of `descriptor->createParticles` for each particle type (which should also be a template parameter)?

src/particles/PhysicsParticles.hpp
BenWibking, line 418: Can this be switch statement be replaced by a member function of PhysicsParticleRegister that is specialized based on particle type?

BenWibking (changes requested):
Some design changes I think are needed.

BenWibking:
> Compile-time switches for better performance.

The performance cost at the ParticleContainer level is negligible. These flags are never checked inside a ParallelFor, so it is not the case there is any meaningful performance advantage of compile-time vs runtime flags.

src/particles/PhysicsParticles.hpp
chongchonghe: I tried the simpliest solution possible. Should get rid of this concern.

chongchonghe:
OK, I've made some significant change to the particle creation routine. I'm very happy with it. 

Our original idea was to use template specialization for particle creation, but I found out C++ does not allow *partial* template specialiazation. For example, in the following demo `TDerived<T, 1>` doesn't work. You have to specify all templates if you want to specialize a function outside of the class, like `TDerived<1, 1>::getData()`. However, in Quokka code, `problem_t` must be kept as a template. 

```cpp
#include <iostream>
#include <vector>

class Base {
public:
  int data = 0;
  int datan = 0;
  virtual int getData() const = 0;
};

template <int T, int N>
class TDerived : public Base {
public:
  int data = T;
  int datan = N;
  int getData() const override { return data + datan; }
};

// attempt partial specialization for getData; this does not work
template <int T>
int TDerived<T, 1>::getData() const {
  return data + datan;
}

// this works
template <>
int TDerived<1, 1>::getData() const {
  return data + datan;
}

int main() {
  std::vector<Base *> vector_ptrs;
  Base *base20_ptr = new TDerived<2, 0>();
  Base *base21_ptr = new TDerived<2, 1>();
  Base *base22_ptr = new TDerived<2, 2>();
  vector_ptrs.push_back(base20_ptr);
  vector_ptrs.push_back(base21_ptr);
  vector_ptrs.push_back(base22_ptr);

  for (const auto *ptr : vector_ptrs) {
    std::cout << "ptr->getData(): " << ptr->getData() << '\n';
  }
}
```

My solution is to use traits class for particle creation. The traits class `ParticleCreationTraits` is a class templated by ParticleType that contains the corresponding checker, creator, and creation routine.

```cpp
template <ParticleType particleType> struct ParticleCreationTraits {
	template <typename problem_t> struct ParticleChecker {};
	template <typename problem_t> struct ParticleCreator {};
	template <typename problem_t, typename ContainerType>
	static void createParticles(ContainerType *container, int mass_idx, amrex::MultiFab &state, int lev, amrex::Real current_time, amrex::Real dt, amrex::Real param1, amrex::Real param2)
	{
		// Use the common implementation with our checker and creator types
	}
}
```

We can now specialize the checker, creator, and creation routine for each particle type. As mentioned before, because we have to keep `problem_t` as a template, we must specialize all three in the same specialization.

```cpp
template <> struct ParticleCreationTraits<ParticleType::CIC> {
	template <typename problem_t> struct ParticleChecker {};
	template <typename problem_t> struct ParticleCreator {};
	template <typename problem_t, typename ContainerType>
	static void createParticles(ContainerType *container, int mass_idx, amrex::MultiFab &state, int lev, amrex::Real current_time, amrex::Real dt, amrex::Real param1, amrex::Real param2)
	{
		// Use the common implementation with our checker and creator types
		ParticleCreationImpl::createParticlesImpl<problem_t, ContainerType, ParticleCreationTraits<ParticleType::CIC>::template ParticleChecker,
							  ParticleCreationTraits<ParticleType::CIC>::template ParticleCreator>(
		    container, mass_idx, state, lev, current_time, dt, param1, param2);
	}
};
```

The `ParticleCreationImpl` is a helper class that contains the common implementation of the creation routine. It is templated by the problem type, container type, checker, and creator.

At last, the `createParticlesFromState` method in `PhysicsParticleDescriptor` calls the traits class to implement the specialized behavior:

```cpp
void createParticlesFromState(amrex::MultiFab &state, int lev, amrex::Real current_time, amrex::Real dt, amrex::Real param1,
						amrex::Real param2) const override
{
	// Use the traits class to implement the specialized behavior
	ParticleCreationTraits<particleType_>::template createParticles<problem_t, ContainerType>(container_, this->getMassIndex(), state, lev,
													current_time, dt, param1, param2);
}
```

and the `createParticlesFromState` method in the `PhysicsParticleRegister` class loops over all particle descriptors and calls the traits class to implement the specialized behavior:

```cpp
void createParticlesFromState(amrex::MultiFab &state, int lev, amrex::Real current_time, amrex::Real dt, amrex::Real param1, amrex::Real param2)
{
	for (const auto &[type, descriptor] : particleRegistry_) {
		// Only create particles if the descriptor allows creation
		if (descriptor->getAllowsCreation()) {
			// Call the appropriate particle creation method based on the particle type
			descriptor->createParticlesFromState(state, lev, current_time, dt, param1, param2);
		}
	}
}
```

src/particles/particle_types.hpp
```diff
@@ -0,0 +1,151 @@
...
+// Example: To enable both CIC and Rad particles, use:
+//   particle_switch = ParticleSwitch::CIC | ParticleSwitch::Rad  (= 0b00000011)
+// To check if CIC particles are enabled:
+//   if (particle_switch & ParticleSwitch::CIC) { ... }
+enum class ParticleSwitch : unsigned int {
+	None = 0U,	      // No particles, = 0b0000
+	CIC = bitflag<1>(),   // Cloud-In-Cell (gravitating) particles, = 0b0001
+	Rad = bitflag<2>(),   // Radiation particles, = 0b0010
+	CICRad = bitflag<3>() // Combined gravitating-radiating particles, = 0b0100
+};
```
BenWibking, line 22: Nice solution.

src/particles/particle_creation.hpp
```diff
@@ -0,0 +1,216 @@
...
+	}
+}
+} // namespace ParticleCreationImpl
+
+// Traits class for specializing particle creation behavior
+template <ParticleType particleType> struct ParticleCreationTraits {
+	// Default nested ParticleChecker - determines if a particle should be created at a location
+	template <typename problem_t> struct ParticleChecker {
+		double param1;
+		double param2;
```
BenWibking: Do these parameters need to be hard-coded here?

src/particles/particle_creation.hpp
```diff
@@ -0,0 +1,216 @@
...
+		}
+	};
+
+	// Default nested ParticleCreator - initializes a particle's properties
+	template <typename problem_t> struct ParticleCreator {
+		int mass_idx;
+		int cpu_id;
+		amrex::Long pid_start;
+		amrex::Real param1;
+		amrex::Real param2;
```
BenWibking: Do these parameters need to be hard-coded here?

src/particles/particle_creation.hpp
```diff
@@ -0,0 +1,216 @@
...
+	template <typename problem_t> struct ParticleChecker {
+		double param1;
+		double param2;
+		AMREX_GPU_HOST_DEVICE ParticleChecker(double t1, double t2) : param1(t1), param2(t2) {}
+
+		AMREX_GPU_DEVICE auto operator()(amrex::Array4<const amrex::Real> const &state_arr, int i, int j, int k,
+						 amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &dx, amrex::Real current_time, amrex::Real dt) const -> bool
+		{
+			// A simple demonstration of particle creation
+			// Could check density threshold or other state-based conditions
```
BenWibking: Does it make more sense to put this in the problem file, like before?

BenWibking (changes requested):
This is a nice solution. I have a few comments about the test problem-specific particle creator routine that you put in `particles/particle_creation.hpp`.

BenWibking:
Ah, right, I forgot about the partial template specialization rules. It turns out that you _can_ partially specialize class templates, but that requires you to redeclare every member function and variable of the class (which is not helpful in this case). But other partial specializations (e.g., of member functions of class templates) are not allowed.

src/particles/particle_types.hpp
```diff
@@ -146,6 +154,20 @@ template <typename problem_t> using CICRadParticleIterator = amrex::ParIter<CICR
...
+// should be treated as the same function. This is a common pattern for small utility
+// functions defined in header files.
+inline void particleParmParse()
+{
+	// Parse particle parameters
+	const amrex::ParmParse pp("particles");
+	pp.query("param1", particle_param1);
+	pp.query("param2", particle_param2);
+}
+
```
BenWibking, line 170: Is this a placeholder for future parameters that will replace these? Or can this code also go in `gravity_3d.cpp`?

BenWibking: @chongchonghe I still don't understand this. I think it would make sense to just put this parmparse stuff in gravity_3d.cpp, unless this is a placeholder for code that you intend to add in a future PR that will generalize to all particles, rather than just a specific test problem.

src/particles/particle_types.hpp
```diff
@@ -0,0 +1,173 @@
...
+	CICRad // Gravitating radiation particles
+};
+
+// Global particle parameters
+// The 'inline' keyword is used here to avoid multiple definition errors when this header
+// is included in multiple source files. It ensures that all translation units that include
+// this header will refer to the same instance of these variables, rather than creating
+// their own copies.
+inline amrex::Real particle_param1 = -1.0; // NOLINT
+inline amrex::Real particle_param2 = -1.0; // NOLINT
```
BenWibking, line 80: Same comment here as above.

BenWibking (changes requested):
Still confused about the test-problem-specific parameters that are in particle_types.hpp:

chongchonghe:
> Still confused about the test-problem-specific parameters that are in particle_types.hpp:

Yes, this is a placeholder for future implementation. We absolutely want inline void particleParmParse() in particle_types.hpp because particleParmParse is called in simulation.hpp to do the parsing at the begging of a simulation. In the future, we can simply rename these parameters to stellar_population_particle_density_thresh and use it in the stellar population particle creator.

BenWibking (dismissed):

chongchonghe:
@BenWibking  I don't understand the review comment system on github. I replied to your comment, but it says 'pending' and it seems like you can't see it. See attached image. I can't see it in this page if I log out of github. Do you know what's going on?

CleanShot 2025-03-05 at 12 30 12

BenWibking:
No, I can't see the first of the comments that you see in the screenshot.

I have seen Github PRs malfunction before when they have a very large number of comments like this PR does. In the worst case, you can always close this PR and open a new one based on the same branch.

chongchonghe:
@BenWibking @markkrumholz Need another approval after merging in the development branch.

markkrumholz (approved):

PR #912, opened 2025-02-28
Particle CFL condition

### LINKED ISSUE #884: add particle CFL condition
**Describe the proposal**
Limit the hydro timestep based on the particle velocity (for non-tracer particles).

This is also required to handle particles correctly with AMR subcycling.

We could perform a non-local Redistribute on the particles (see https://github.com/quokka-astro/quokka/pull/833), but that is computationally expensive and less accurate.

**Additional context**
Needed to fix the crashes seen here: https://github.com/quokka-astro/quokka/pull/833.

chongchonghe:
### Description
Add particle speed to the calculation of hydro CFL condition.

1. Defined a function `computeMaxParticleSpeed` in `PhysicsParticleRegister` to compute the maximum speed of all particles of all particle types that have velocity components. 
2. Defined particle CFL number `particleCflNumber_`, which can be specified via runtime parameter `particle_cfl`. The default value is 0.99
3. Run `computeMaxParticleSpeed` every timestep to compute the maximum particle speed. Then, `particle_dt` is calculated: `particle_dt = particleCflNumber_ * (dx_min / max_particle_speed)`, where `particleCflNumber_` is the particle CFL number and is a runtime parameter. 

### Related issues
Replace #833 and resolve #884 . 

### Checklist


src/particles/PhysicsParticles.hpp
```diff
@@ -574,6 +575,51 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
+					auto *pData = particles().data();
+					const amrex::Long np = pti.numParticles();
+
+					if (np > 0) {
+						// Allocate device memory for reduction
+						amrex::Gpu::DeviceScalar<amrex::Real> ds_max(0.0);
+						amrex::Real *p_max = ds_max.dataPtr();
+
+						// Compute maximum speed in parallel using reduction
+						amrex::ParallelFor(np, [=] AMREX_GPU_DEVICE(amrex::Long i) {
```
BenWibking: You should use the built-in AMReX functions for reductions instead of writing your own. It will be significantly faster.

src/simulation.hpp
```diff
@@ -146,15 +146,16 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
...
-	amrex::Real dtToleranceFactor_ = 1.1; // default
+	amrex::Vector<int> istep;	       // which step?
+	amrex::Vector<int> nsubsteps;	       // how many substeps on each level?
+	amrex::Vector<amrex::Real> tNew_;      // for state_new_cc_
+	amrex::Vector<amrex::Real> tOld_;      // for state_old_cc_
+	amrex::Vector<amrex::Real> dt_;	       // timestep for each level
+	amrex::Vector<int> reductionFactor_;   // timestep reduction factor for each level
+	amrex::Real stopTime_ = 1.0;	       // default
+	amrex::Real cflNumber_ = 0.3;	       // default
+	amrex::Real particleCflNumber_ = 0.99; // default
```
BenWibking: I think this should be more conservative. Nyx uses a value of `0.5`, for example.

BenWibking (changes requested):
Two comments:

src/particles/PhysicsParticles.hpp
BenWibking: See the example for hydro signal speed here: https://github.com/quokka-astro/quokka/blob/9e3a4210042a2f3a4118da28be3da9d490ddf370/src/hydro/hydro_system.hpp#L203

BenWibking:
I don't think step 3 in the PR description corresponds to what you actually do in the code. You actually compute a separate `dt_particle`. Since it has a possibly different CFL factor, I think that's the right thing to do. Can you update the description so it matches the code you've written?

src/particles/PhysicsParticles.hpp
BenWibking: Hm, for some reason `ParReduce` only works for FABs, not particles. You can instead follow what Nyx does here: https://github.com/AMReX-Astro/Nyx/blob/51a5f8e55ef70ac488b13a376629e9027bc2e5d6/Source/Particle/NyxParticleContainer.H#L388

BenWibking: Andrew Myers mentioned that there are the `ParticleReduce` functions here: https://amrex-codes.github.io/amrex/doxygen/namespaceamrex.html#a8d32380d1bccbedf1b24bd2c1bb7a1da

src/simulation.hpp
chongchonghe: A smaller number makes sense. Once potential problem with particle CFL close to 1 is that in the kick-drift-kick step, a kick could raise the speed to a large value such that the particle moves more than a grid cell during drift. Although, we are not garanteed to avoid this with a smaller number.

BenWibking: That's a good point. This might be why Nyx uses a conservative value.

I think if it is kicked more than more cell away, then the right thing to do in principle is to go back and re-do the coarse step with a smaller CFL number, similar to the way the level advance is re-done. Re-doing the level advance is not sufficient when doing subcycling, though, since the particle may be kicked off of all the grids on a given level.

This may be tricky to solve in a way that guarantees it works in absolutely all cases. We may have to implement a check for this and have the code crash if this happens for now. Then users can manually restart the code with a smaller particle CFL.

src/particles/PhysicsParticles.hpp
chongchonghe: This is awesome. I didn't know it exists.

chongchonghe:
> I don't think step 3 in the PR description corresponds to what you actually do in the code. You actually compute a separate `dt_particle`. Since it has a possibly different CFL factor, I think that's the right thing to do. Can you update the description so it matches the code you've written?

Good catch! I updated the design in recent commits but forgot to update the documentation.

BenWibking (dismissed):

BenWibking:
Let's do a new Quokka release (either 25.02 or 25.03) after this is merged, since this fixes a significant bug in the 25.01 release.

src/simulation.hpp
chongchonghe: I would expect the code to crash when this happens in the current implementation. When a particle moves out of a tile during drift, it tries to deposit mass/radiation and the code will complain, as shown in #833 .

BenWibking: Does that happen in Release mode, or only if compiled with debugging enabled?

BenWibking:
Are the GPU tests failing because of division by zero (in this case, the max particle speed)? A lot of tests on moth fail with this error:
```
max_particle_speed = 0
```

chongchonghe:
> Are the GPU tests failing because of division by zero (in this case, the max particle speed)? A lot of tests on moth fail with this error:
> 
> ```
> max_particle_speed = 0
> Erroneous arithmetic operation
> ```

It happens in `ComputeEddingtonTensor`, caught by `amrex.fpe_trap_invalid=1`. Somehow `f > 0.` is invalid (NaN), but `AMREX_ASSERT(!std::isnan(f));` won't catch it. 

```cpp
	auto f = std::sqrt(fx * fx + fy * fy + fz * fz);
	std::array<amrex::Real, 3> fvec = {fx, fy, fz};

	// angle between interface and radiation flux \hat{n}
	// If direction is undefined, just drop direction-dependent
	// terms.
	std::array<amrex::Real, 3> n{};

	for (int ii = 0; ii < 3; ++ii) {
		n[ii] = (f > 0.) ? (fvec[ii] / f) : 0.;
	}
```

https://github.com/quokka-astro/quokka/blob/6a45666b1a5b174b1c8bfedc88ef517bfefb52ff/src/radiation/radiation_system.hpp#L924

chongchonghe:
The same tests in the development branch, compiled for 1D, fail on moth with the same error. That's weird.

BenWibking:
> > Are the GPU tests failing because of division by zero (in this case, the max particle speed)? A lot of tests on moth fail with this error:
> > ```
> > max_particle_speed = 0
> > Erroneous arithmetic operation
> > See Backtrace.0 file for details
> > ```
> 
> It happens in `ComputeEddingtonTensor`, caught by `amrex.fpe_trap_invalid=1`. Somehow `f > 0.` is invalid (NaN), but `AMREX_ASSERT(!std::isnan(f));` won't catch it.
> 
> ```c++
> 	auto f = std::sqrt(fx * fx + fy * fy + fz * fz);
> 	std::array<amrex::Real, 3> fvec = {fx, fy, fz};
> 
> 	// angle between interface and radiation flux \hat{n}
> 	// If direction is undefined, just drop direction-dependent
> 	// terms.
> 	std::array<amrex::Real, 3> n{};
> 
> 	for (int ii = 0; ii < 3; ++ii) {
> 		n[ii] = (f > 0.) ? (fvec[ii] / f) : 0.;
> 	}
> ```
> 
> https://github.com/quokka-astro/quokka/blob/6a45666b1a5b174b1c8bfedc88ef517bfefb52ff/src/radiation/radiation_system.hpp#L924

Are you compiling in Debug mode? `AMREX_ASSERT` only works if debug mode is used to compile. If you are, then I'm not sure what's going on.

BenWibking:
If you manually comment out the YAML output in WriteMetadataFile, does it still have the same problem?

BenWibking:
This is very odd. I don't understand why the tests on moth are failing.

chongchonghe:
> If you manually comment out the YAML output in WriteMetadataFile, does it still have the same problem?

Same problem after commenting out 
```
// WriteMetadataFile(plotfilename + ".yaml");
amrex::WriteMultiLevelPlotfile(plotfilename, finest_level + 1, mf_ptr, varnames, Geom(), tNew_[0], istep, refRatio());
// WriteMetadataFile(plotfilename + "/metadata.yaml");
```

How's this related to the 'Erroneous arithmetic operation' problem?

chongchonghe:
@BenWibking Something on moth must have changed recently. I checked out Quokka to a commit from a month ago, ran a 1D CPU test (e.g. RadStreaming), and it failed with the same error.

chongchonghe:
Apparently, `if (0 < 1.0e-10)` is `Erroneous arithmetic operation` 🤷‍♂️ . 

CleanShot 2025-03-03 at 15 03 14

LastTest.log:

```
f = 1
f = 0
addr2line: DWARF error: invalid or unhandled FORM value: 0x25
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 8.
```

Backtrace.0:

```
 1: /data/mash/cche/softwares/quokka/quokka/build/build-1D-debug.mpicc/src/problems/RadStreaming/test_radiation_streaming() [0x3ed9d1]
    amrex::BLBackTrace::handler(int)
/data/mash/cche/softwares/quokka/quokka/extern/amrex/Src/Base/AMReX_BLBackTrace.cpp:101:9

 2: /lib64/libc.so.6(+0x3e730) [0x7fa6dca3e730]

 3: /data/mash/cche/softwares/quokka/quokka/build/build-1D-debug.mpicc/src/problems/RadStreaming/test_radiation_streaming() [0x2ebed8]
    RadSystem<StreamingProblem>::ComputeEddingtonTensor(double, double, double)
/data/mash/cche/softwares/quokka/quokka/src/radiation/radiation_system.hpp:926:6
```

chongchonghe:
A work-around is to turn off `amrex.fpe_trap_invalid` on moth.

src/radiation/radiation_system.hpp
```diff
@@ -912,18 +912,47 @@ AMREX_GPU_DEVICE auto RadSystem<problem_t>::ComputeEddingtonTensor(const double
...
+	int inf_count = 0;
 	for (int ii = 0; ii < 3; ++ii) {
-		n[ii] = (f > 0.) ? (fvec[ii] / f) : 0.;
+		if (std::isinf(fvec[ii])) {
+			inf_count++;
+		}
 	}
 
+	if (inf_count > 0) {
+		// If we have infinite components, set them to 1/sqrt(inf_count) and others to 0
```
BenWibking: How do you get an `inf` as one of the components of `fvec`? Is something going wrong prior to this?

Why does setting them to 1/sqrt(inf_count) make sense?

chongchonghe: I speculate that it happens in 1D or 2D cases where fy or fz are not properly initialized.

chongchonghe: > Why does setting them to 1/sqrt(inf_count) make sense?

Em, it doesn't. I'll do a dimension check.

chongchonghe:
This has to be a compiler bug. I compiled it with gcc/g++ on moth and the same tests didn't fail.

markkrumholz:
OK, thanks for checking. In that case I would recommend that we temporarily disable just this test on moth so that we can merge this PR, and that at the same time we try other versions of the AMD compiler to see if we can find a version that doesn't show the bug.

chongchonghe:
> OK, thanks for checking. In that case I would recommend that we temporarily disable just this test on moth so that we can merge this PR, and that at the same time we try other versions of the AMD compiler to see if we can find a version that doesn't show the bug.

Sounds good. 

For the tests that previously failed (almost all radiation tests, incl. RadStreaming), although it might be due to a compiler bug as well, I avoided the error with a more robust calculation of flux normal vector in `ComputeEddingtonTensor`. Now, in the most recent commit, SphericalCollapse failed with `Erroneous arithmetic operation` during the calculation of `particle_dt`. I added an extra layer of check and that code should be very robust, but it failed with HIP compiler (for CPU) on moth. However, it runs successfully with GNU compiler, configured like this:

```
module load mpi/openmpi-x86_64

# compiler environment hints
export CC=$(which gcc)
export CXX=$(which g++)
export CFLAGS=""
export CXXFLAGS=""
export LDFLAGS=""
```

So, I reckon it's a compiler bug and we can fix it later by updating the compilers.

BenWibking:
Would it be possible to just disable FPE checks for the tests that fail on moth, rather than skipping the test entirely? I think that would be the most robust approach while avoiding whatever compiler weirdness is happening with the latest version of Clang.

src/radiation/radiation_system.hpp
BenWibking, line 1: Can you put the radiation system changes in a separate PR? They are not really related to the particle stuff. If we ever have to go back and bisect changes that change the FPE behavior in the code, having a separate PR will make it a lot easier to do so.

BenWibking (changes requested):

src/radiation/radiation_system.hpp
chongchonghe: Good point. I'll do that.

chongchonghe:
> Would it be possible to just disable FPE checks for the tests that fail on moth, rather than skipping the test entirely? I think that would be the most robust approach while avoiding whatever compiler weirdness is happening with the latest version of Clang.

I think that's possible.

chongchonghe:
OK. I disabled FPE checks for the SphericalCollapse test on moth to avoid a unfixable FPE failure (potentially a compiler bug). This PR also relies on changing radiation_system.hpp in order to avoid FPE failure on moth. I made this change in #921 . Once that PR is merged into the dev branch, we won't see changes to radiation_system.hpp in this PR.

src/problems/SphericalCollapse/CMakeLists.txt
```diff
@@ -4,5 +4,9 @@ if (AMReX_SPACEDIM EQUAL 3)
         setup_target_for_cuda_compilation(spherical_collapse)
     endif()
 
-    add_test(NAME SphericalCollapse COMMAND spherical_collapse SphericalCollapse.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
+    if(AMReX_GPU_BACKEND MATCHES "HIP")
+        add_test(NAME SphericalCollapse COMMAND spherical_collapse SphericalCollapse.in WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
```
BenWibking, line 9: Can you add a comment here that explains why you're doing this? Just as a reminder to someone who looks at this months/years from now and wonders why this is handled differently from all of the other tests.

BenWibking (changes requested):
Only one comment:

chongchonghe:
OK, all tests passed. I'm going to add a comment as Ben suggested and it should be ready to merge.

src/problems/SphericalCollapse/CMakeLists.txt
chongchonghe: OK.

chongchonghe:
@BenWibking Can you review this?

BenWibking (approved):

PR #921, opened 2025-03-04
add safety check in ComputeEddingtonTensor to avoid FPE error

chongchonghe:
### Description
Add safety check in ComputeEddingtonTensor to avoid FPE error:
1. Check for large flux limiting violation: `if (fvec[ii] > 2.0) { # set n = 0.0 }`
2. set `f_floor = 1e-10` and `if (f <= f_floor) { # set n = 0.0 }`

### Related issues

### Checklist

### DIFF
--- a/src/radiation/radiation_system.hpp
+++ b/src/radiation/radiation_system.hpp
@@ -912,16 +912,29 @@ AMREX_GPU_DEVICE auto RadSystem<problem_t>::ComputeEddingtonTensor(const double 
 	// AMREX_ASSERT(f < 1.0); // there is sometimes a small (<1%) flux
 	// limiting violation when using P1 AMREX_ASSERT(f_R < 1.0);
 
-	auto f = std::sqrt(fx * fx + fy * fy + fz * fz);
-	std::array<amrex::Real, 3> fvec = {fx, fy, fz};
+	const double f = std::sqrt((fx * fx) + (fy * fy) + (fz * fz));
+	const std::array<amrex::Real, 3> fvec = {fx, fy, fz};
+	const double f_floor = 1e-10;
 
 	// angle between interface and radiation flux \hat{n}
-	// If direction is undefined, just drop direction-dependent
-	// terms.
+	// If direction is undefined, just drop direction-dependent terms.
 	std::array<amrex::Real, 3> n{};
 
+	// if fvec has a inf component, set n to zero
+	bool has_inf = false;
 	for (int ii = 0; ii < 3; ++ii) {
-		n[ii] = (f > 0.) ? (fvec[ii] / f) : 0.;
+		if (fvec[ii] > 1.1) {
+			has_inf = true;
+			break;
+		}
+	}
+
+	if (has_inf || (f <= f_floor)) {
+		n = {0.0, 0.0, 0.0};
+	} else {
+		for (int ii = 0; ii < 3; ++ii) {
+			n[ii] = fvec[ii] / f;
+		}
 	}
 
 	// compute radiation pressure tensors
--- a/src/radiation/radiation_system.hpp
+++ b/src/radiation/radiation_system.hpp
@@ -920,16 +920,17 @@ AMREX_GPU_DEVICE auto RadSystem<problem_t>::ComputeEddingtonTensor(const double 
 	// If direction is undefined, just drop direction-dependent terms.
 	std::array<amrex::Real, 3> n{};
 
-	// if fvec has a inf component, set n to zero
-	bool has_inf = false;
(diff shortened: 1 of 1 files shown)


BenWibking (approved):
This seems like a very sensible solution to me.

For the old gray solver, I never saw flux limiting violations above 1.01 or so, and it was always with P_1 instead of M_1. What test problems do you see this on? Is it only for multigroup problems?

chongchonghe:
> This seems like a very sensible solution to me.
> 
> 
> 
> For the old gray solver, I never saw flux limiting violations above 1.01 or so, and it was always with P_1 instead of M_1. What test problems do you see this on? Is it only for multigroup problems?

I don't think it actually happened. This trick prevents the FPE error on moth for almost all radiation tests, single group or multi group. I can't figure out why this worked. A debugger on moth doesn't tell me how FPE occurred.

BenWibking:
Hmm, odd. It may be the case that these are spurious FPEs caused by the compiler vectorizing the code. That can happen because the input array may not be an exact multiple of the hardware vector length, and then the unused vector lanes can have NaNs or zeros in them. Even though the values in them are unused, they can still trigger FPEs. It used to be that only the Intel compiler would do this, but the Clang compiler might do this now. If this is the case, there is an additional compiler flag that should suppress this.

chongchonghe:
> Hmm, odd. It may be the case that these are spurious FPEs caused by the compiler vectorizing the code. That can happen because the input array may not be an exact multiple of the hardware vector length, and then the unused vector lanes can have NaNs or zeros in them. Even though the values in them are unused, they can still trigger FPEs. It used to be that only the Intel compiler would do this, but the Clang compiler might do this now. If this is the case, there is an additional compiler flag that should suppress this.

From what you described, this seems to be what's happening in computing max particle velocity which causes us having to disable FPE for the SphericalCollapse test. I ran a debugger on it, and an odd thing occurred: the variable `amrex::Real max_speed` changed from `= 0.0` to `<Invalid address size for DW_OP_deref_size: 16>` at some point, and it caused `amrex.fpe_trap_zero` error.

chongchonghe:
> > Hmm, odd. It may be the case that these are spurious FPEs caused by the compiler vectorizing the code. That can happen because the input array may not be an exact multiple of the hardware vector length, and then the unused vector lanes can have NaNs or zeros in them. Even though the values in them are unused, they can still trigger FPEs. It used to be that only the Intel compiler would do this, but the Clang compiler might do this now. If this is the case, there is an additional compiler flag that should suppress this.
> 
> From what you described, this seems to be what's happening in computing max particle velocity which causes us having to disable FPE for the SphericalCollapse test. I ran a debugger on it, and an odd thing occurred: the variable `amrex::Real max_speed` changed from `= 0.0` to `<Invalid address size for DW_OP_deref_size: 16>` at some point, and it caused `amrex.fpe_trap_zero` error.

What is the compiler flag to suppress this? @BenWibking

BenWibking:
You may have to disable vectorization: `-fno-vectorize -fno-slp-vectorize`.

More details: https://groups.google.com/g/cloudy-dev/c/4MkNUmBtFQI

BenWibking:
Ah, I forgot, there is also `-ffp-exception-behavior=maytrap`:

From https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffp-exception-behavior

>     Specify the floating-point exception behavior.
> 
>     Valid values are: ignore, maytrap, and strict. The default value is ignore. Details:
> 
>         ignore The compiler assumes that the exception status flags will not be read and that floating point exceptions will be masked.
>         maytrap The compiler avoids transformations that may raise exceptions that would not have been raised by the original code. Constant folding performed by the compiler is exempt from this option.
>         strict The compiler ensures that all transformations strictly preserve the floating point exception semantics of the original code.
> 

If that doesn't work, you can try `strict`.

BenWibking:
So it turns out that I had this problem before, and I added `-ffp-exception-behavior=strict` for the Apple Clang compiler CI:
https://github.com/quokka-astro/quokka/blob/8852ff326a05748bbcaed1e408f1b06d6519eba4/.github/workflows/cmake-macos.yml#L60

This may be necessary for all Clang-based compilers.

BenWibking:
@chongchonghe Actually, this PR probably just causes the compiler to not vectorize the loop where this problem appears. So we should probably revert this PR and instead add `-ffp-exception-behavior=strict` to moth and the HIP CI build.

chongchonghe:
OK, I believe `-ffp-exception-behavior=maytrap` is the solution. I'll try that and create another PR to add it and revert this PR.

chongchonghe:
Both `maytrap` and `strict` worked. I guess `strict` is prefered?

PR #927, opened 2025-03-06
Particle destructor

chongchonghe:
### Description
Support particle destruction. 

Changes:
- Define `ParticleDestructionTraits`. The user define their own particle destructor by specializing the `ParticleChecker` method in `ParticleDestructionTraits`, which return true or false to mark a particle for destruction. As a placeholder, `particles.param3` is used in particle destructor. 
- Define `ParticleDestructionImpl`, which loops over all particles at a level and mark particles for destruction. 
- Add a `getNumParticles` function to `PhysicsParticleDescriptor` to get the total number of *valid* particles of that type.
- Add a new particle parameter `particles.verbose` to turn on printing particle logistics at each time step. Currently, it prints the number of particles of each type. 

The new particle destructor feature is tested in gravity_3d.cpp. 

Next, I'll add a `particleDestruction` method in `ParticleDestructionTraits`, which modifies the hydro state while removing a particle. 

### Related issues
None. 

### Checklist

### DIFF
--- a/src/particles/particle_destruction.hpp
+++ b/src/particles/particle_destruction.hpp
@@ -55,24 +55,15 @@ static void destroyParticlesImpl(ContainerType *container, int mass_idx, int lev
 template <ParticleType particleType> struct ParticleDestructionTraits {
 	// Default nested ParticleChecker - determines if a particle should be destroyed
 	template <typename problem_t> struct ParticleChecker {
-		// amrex::Real param1 = ;
-
-		// AMREX_GPU_HOST_DEVICE ParticleChecker(amrex::Real param1) : param1(param1) {}
 
 		template <typename ParticleType>
 		AMREX_GPU_DEVICE auto operator()(ParticleType& p, int mass_idx, amrex::Real current_time, amrex::Real dt) const -> bool
 		{
 			// Default implementation: destroy particles with mass < 1.0
-			amrex::ignore_unused(current_time, dt);
-			
-			// Check if mass is below threshold
-			// return (p.rdata(mass_idx) < 1.0); // Destroy this particle
-
-			const double t_destroy = 0.0015;
-			const bool is_small_mass = (p.rdata(mass_idx) < 2.0e-20);
-			const bool is_time = (current_time <= t_destroy && current_time + dt > t_destroy);
-			return is_small_mass && is_time;
+			amrex::ignore_unused(p, mass_idx, current_time, dt);
+			return false;
 		}
+
 	};
 
 	// Main method to destroy particles - uses the helper implementation
--- a/src/particles/particle_destruction.hpp
+++ b/src/particles/particle_destruction.hpp
@@ -20,20 +20,20 @@ static void destroyParticlesImpl(ContainerType *container, int mass_idx, int lev
 
 			// Iterate through all particles at this level
 			for (typename ContainerType::ParIterType pti(*container, lev); pti.isValid(); ++pti) {
-				auto& particles = pti.GetArrayOfStructs();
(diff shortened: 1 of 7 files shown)


src/particles/particle_destruction.hpp
```diff
@@ -0,0 +1,80 @@
...
+						// Mark particle as invalid by negating its ID
+						// This is the AMReX way to mark particles for removal
+						// They will be removed during the next Redistribute call
+						p.id() = -1;
+					}
+				});
+			}
+
+			// Redistribute particles to actually remove the invalid particles
+			container->Redistribute(lev);
```
BenWibking, line 50: This may not work as intended if particle destruction is allowed to happen during an AMR subcycle on $l > 0$. Maybe add a comment here just in case particles are extended to work with AMR subcycling in the future.

BenWibking (changes requested):
Overall design looks good to me.

src/particles/particle_destruction.hpp
chongchonghe: `destroyParticles` in simulation.hpp is called on level 0, so it only works when AMR is turned off. The same for `createParticlesFromState`. So, I added a note to remind myself ot take care of AMR subcycling in the future. 

However, I believe the `destroyParticlesImpl` routine in particle_destruction.hpp should work in any level, right?

BenWibking: If AMR subcycling is enabled, and `destroyParticlesImpl` is called after the hydro advance on that level, the answer is in fact _no_.

This is for the following reason: a particle may leave the valid cells of level $l$ and travel onto the valid cells of level $l-1$ during the first substep for level $l$. If there is a redistribute here, then the second substep on level $l$ does not  kick that particle.  At first, this appears to be the correct behavior, but it's not. This is because when AMR subcycling is turned on, level $l$ must kick that particle twice, since it was *not* kicked by level $l-1$. If it is not kicked by both substeps at level $l$, it only receives effectively half of the force it was supposed to.

BenWibking: To elaborate further: in the second substep on level $l$, the particle receives a kick using the ghost cell values on level $l$. This is why a particle CFL is required for AMR subcycling.

BenWibking: This probably means that particle destruction can only happen at the end of a coarse step, even when AMR subcycling is turned on.

chongchonghe: So, since particle CFL is enforced, this is okay?

Oh, wait. If a particle is in the ghost cell in the beginning of the second substep on level `l`, it may move out of the ghost cell during drift, which is not okay. However, we have more than 1 ghost cells, and our particle CFL number is always smaller than 1, so it's probably fine?

chongchonghe: In either case, let's leave the AMR stuff for the future and merge this PR if there is nothing else need to be changed.

BenWibking: No, the issue is that once the particle moves into the ghost cells, `Redistribute` will move it from level $l$ to level $l-1$, preventing it from being kicked (as it should be) on the second substep of level $l$.

This is only an issue when doing AMR subcycling, so it doesn't have this issue for now. But I think it would be helpful to leave a comment about this so it's not forgotten.

chongchonghe: Sounds good. I'll add a comment to describe this in detail.

chongchonghe: Done.

BenWibking (approved):

PR #938, opened 2025-03-13
Add Stellar Evolution attribute and SNDeposition

chongchonghe:
### Description

This PR adds a new attribute `evolutionStage` to the particle data and a new function `SNDeposition` to deposit mass, momentum, and energy from particles to the grid.

Changes:

1. Add `Test` particle type for testing all particle features. 

2. Define an integer component `evolutionStage` to Test_particles. It is used to track the evolution stage of the particle. The evolution stage is defined as an enum class as listed below. For particle types that support stellar evolution (`evolutionStageIndex_ >= 0`), each star is created either as a low mass star or as a SN progenitor. When a SN progenitor reaches the end of its life, it executes `SNDeposition()` and evolves into a SN remnant.

```cpp
enum class StellarEvolutionStage {
	LowMassStar,  // Low mass star stage
	SNProgenitor,  // Supernova progenitor stage
	SNRemnant     // Supernova remnant stage
};
```

3. Defined `SNDeposition` to deposit mass, momentum, and energy from particles to the grid. At each time step, we find the particle types that support stellar evolution, and check if each particle is a SN progenitor and if it has reached the end of its life. If so, we run `SNDeposition()` and update the particle's evolution stage to SNRemnant. This also works as a framework for other interact-with-mesh operations. Unlike `interp.ParticleToMesh` which uses linear interpolation to deposit N particle components to N mesh components on a one-to-one basis, `SNDeposition` calculates the exact position of the particle in the mesh and deposit particle properties to arbitrary components of neighboring cells, using any interpolation kernel. Note that the maximum number of cells allowed to use in SNDeposition (the `stencil_width`) is `nghost_cc_ = 4`. 

The new code is fully tested via the gravity_3d test on CPU, GPU, and MPI, where we check the creation, stellar evolution, and destruction of the particles, as well as the total mass in SN ejecta. 

### Checklist

### DIFF
--- a/src/particles/PhysicsParticles.hpp
+++ b/src/particles/PhysicsParticles.hpp
@@ -467,9 +467,10 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 				// Get the units data for this particle type
 				const auto &unitsData = quokka::get_units_data();
 				if (unitsData.find(particleType_) == unitsData.end()) {
-					amrex::Abort("Error: Particle type not defined in units data map. Please add units for this particle type in get_units_data().");
+					amrex::Abort(
+					    "Error: Particle type not defined in units data map. Please add units for this particle type in get_units_data().");
 				}
-				
+
 				const auto &typeData = unitsData.at(particleType_);
 				if (!typeData.empty()) {
 					outFile << "# field: [M, L, T, Θ]\n";
--- a/src/particles/PhysicsParticles.hpp
+++ b/src/particles/PhysicsParticles.hpp
@@ -465,7 +465,12 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 				}
 
 				// Get the units data for this particle type
-				const auto &typeData = quokka::get_units_data().at(particleType_);
+				const auto &unitsData = quokka::get_units_data();
+				if (unitsData.find(particleType_) == unitsData.end()) {
+					amrex::Abort("Error: Particle type not defined in units data map. Please add units for this particle type in get_units_data().");
+				}
+				
+				const auto &typeData = unitsData.at(particleType_);
 				if (!typeData.empty()) {
 					outFile << "# field: [M, L, T, Θ]\n";
 					// Write each field's units to the YAML file
--- a/src/particles/PhysicsParticles.hpp
+++ b/src/particles/PhysicsParticles.hpp
@@ -441,6 +441,43 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 			container_->Checkpoint(checkpointname, name, include_header);
 		}
 	}
+
+	// Implementation of particle data output to units file
+	void writeUnitsFile(const std::string &snapshot_name, const std::string &name) override
+	{
+		if (container_ != nullptr) {
+			// Only write on rank 0
+			if (amrex::ParallelDescriptor::IOProcessor()) {
+				// Create the full path for the Fields.yaml file
+				std::string filename;
+#ifdef QUOKKA_USE_OPENPMD
+				// For OpenPMD, write the YAML file alongside the OpenPMD file
+				filename = snapshot_name + "_" + name + ".yaml";
+#else
+				// For standard output, write the YAML file in the particle directory
+				filename = snapshot_name + "/" + name + "/Fields.yaml";
+#endif
+
+				// Open the file for writing
+				std::ofstream outFile(filename);
+				if (!outFile) {
+					amrex::Abort("Error opening file for writing: " + filename);
+				}
+
+				// Get the units data for this particle type
+				const auto &typeData = quokka::get_units_data().at(particleType_);
+				if (!typeData.empty()) {
+					outFile << "# field: [M, L, T, Θ]\n";
+					// Write each field's units to the YAML file
+					for (const auto &[fieldName, units] : typeData[0]) {
+						outFile << fieldName << ": [" << units[0] << ", " << units[1] << ", " << units[2] << ", " << units[3] << "]\n";
+					}
+				}
+
+				outFile.close();
+			}
+		}
+	}
 };
 
 // New class for star particles that adds stellar evolution capabilities
--- a/src/particles/PhysicsParticles.hpp
+++ b/src/particles/PhysicsParticles.hpp
@@ -78,7 +78,7 @@ class PhysicsParticleDescriptorBase
 	[[nodiscard]] virtual auto getNumParticles() const -> int = 0;
 #if AMREX_SPACEDIM == 3
 	virtual void depositMass(const amrex::Vector<amrex::MultiFab *> &rhs, int finest_lev, amrex::Real Gconst) = 0;
-	virtual void depositSN(amrex::MultiFab &state, int lev, amrex::Real current_time)
+	virtual void depositSN(amrex::MultiFab &state, int lev, amrex::Real step_end_time)
 	{ /* Default empty implementation */
 	}
 	virtual void driftParticles(int lev, amrex::Real dt) const = 0;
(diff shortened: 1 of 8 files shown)


src/particles/PhysicsParticles.hpp
```diff
@@ -34,11 +35,13 @@ class PhysicsParticleDescriptorBase
 	bool interactsWithHydro_{false}; // Whether particles interact with hydrodynamics
 	bool allowsCreation_{false};	 // Whether particles can be created during simulation
 	bool allowsDestruction_{false};	 // Whether particles can be destroyed during simulation
+	int evolutionStageIndex_{-1};	 // Index for evolution stage (-1 if not used)
```
BenWibking, line 39: We could end up with a lot of component indices here... I'm not sure whether there's a better way to handle this or not. It may be useful in the future to move the indices to their own data structure.

src/particles/particle_deposition.hpp
```diff
@@ -62,6 +63,77 @@ struct MassDeposition {
...
+				// Find the cell containing the particle
+				int base_i = static_cast<int>(amrex::Math::floor((p.pos(0) - plo[0]) * dxi[0]));
+				int base_j = static_cast<int>(amrex::Math::floor((p.pos(1) - plo[1]) * dxi[1]));
+				int base_k = static_cast<int>(amrex::Math::floor((p.pos(2) - plo[2]) * dxi[2]));
+
+				// Calculate the volume factor for normalization (5³ cells)
+				const int num_cells = 125; // 5³ cells
+				const amrex::Real vol_factor = (AMREX_D_TERM(dxi[0], *dxi[1], *dxi[2])) / num_cells;
+
+				static constexpr int stencil_width = 2;
```
BenWibking: Does this have to be less than or equal to the number of ghost cells for it to work correctly?

BenWibking (commented):
I think the basic design looks fine. I think it would be useful to document (as a comment in the code) whether there are any restrictions on the `stencil_width` used for deposition: e.g., does it have to be <= nghost_cc?

My only (minor) design concern is that the number of particle component indices is getting to be nontrivial. In the future, the way the particle component indexing is done may get too unwieldly. I'm not sure how to fix that in a fully satisfying way.

src/particles/PhysicsParticles.hpp
markkrumholz: Should this be in the PhysicsParticleDescriptorBase class, or shoudl it only be included in a derived class that handles SN output? It's going to be meaningless for all but a small subset of possible particle types.

src/particles/PhysicsParticles.hpp
```diff
@@ -237,6 +243,19 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 		}
 	}
 
+	// Implementation of supernova energy and momentum deposition from particles to grid
+	void depositSN(amrex::MultiFab &state, int lev, amrex::Real current_time) override
```
markkrumholz: Again, I don't think we want to hang this off the main particle class, I think it should be implemented in a derived class.

src/particles/PhysicsParticles.hpp
BenWibking: I agree with this. Putting it in a derived class used for SN feedback would resolve my concern about the proliferation of indices in the base class.

chongchonghe: Almost all future component indices will have default values like `-1` or `false` for all previous particle types, so we don't need to change anything in registration step of previous particle types. 

Derived class is a good idea. I'll try to implement that.

markkrumholz: I would prefer if indices that are specific to the physics of a particular particle type weren't included in the base class at all. I think the base class should only contain generic indicators like mass and luminosity index that will be shared by many derived classes. This will avoid the proliferation of indices that @BenWibking was worried about.

src/particles/PhysicsParticles.hpp
chongchonghe: Make sense. Done.

chongchonghe:
To reply your comments on the design of the particle component indexing, I'd like to first explain the design choices. The indices/flags have to be in the base class so that they can be accessed by the particle register, because `particleRegistry_` is a map to `PhysicsParticleDescriptorBase`. For the same reason, any method that needs to access particle properties has to be a member of `PhysicsParticleDescriptorBase`, as long as it's accessed through `particleRegister_` in `simulation.cpp`. The methods in the base class can be pure virtual that has to be implemented by the derived class, or empty, like `depositSN`, that is implemented only for some particle types.

Then, in the templated derived class `PhysicsParticleDescriptor`, we define the methods that most particle types have in common, like `redistribute` and `kickParticles`. I also added a new class `StarParticleDescriptor` that derives from `PhysicsParticleDescriptor` and defines the methods that are specific to star particles, like `depositSN`.

In the `PhysicsParticleRegister`, we register the regular 'physics' particles via `registerParticleType`, and the star particles via `registerStarParticleType` (currently only supports Test particles).

src/particles/particle_deposition.hpp
chongchonghe: I tested it with MPI on CPU and found `stencil_width` can be as high as 6 without causing the code to crash. I haven't checked the accuracy of the results yet.

BenWibking: Huh, that's interesting. I think we need to check with the AMReX developers to understand why this is.

BenWibking: Ok, I looked at the code more carefully, and I don't see how this could work correctly for the case where `stencil_width > nghost_cc_`. I don't think checking whether it crashes or not is a sufficient test of this.

Is there a test particle placed in the outermost valid cell of one of the boxes? Can you check the output?

chongchonghe: I confirm that when you place particles at the edge of the tiles, `stencil_width = 4` is the maximum you can have while getting the correct result. With `stencil_width = 5`, the code sometimes crashes with a 'out of bound' error, sometimes produces a wrong result -- underestimating SN remnants, as expected.

src/particles/PhysicsParticles.hpp
chongchonghe: > I would prefer if indices that are specific to the physics of a particular particle type weren't included in the base class at all. I think the base class should only contain generic indicators like mass and luminosity index that will be shared by many derived classes. This will avoid the proliferation of indices that @BenWibking was worried about.

I don't think this is possible. See my comments below. But I would love ideas on how to make that work.

chongchonghe:
This new feature is fully tested and is ready for a review.

BenWibking:
One last GPU test with the MPI fix, just in case.

src/particles/particle_deposition.hpp
```diff
@@ -62,6 +65,107 @@ struct MassDeposition {
...
+				is_sn_progenitor = (p.idata(evolutionStageIndex) == static_cast<int>(StellarEvolutionStage::SNProgenitor));
+			}
+
+			if (is_sn_progenitor && step_end_time > p.rdata(birthTimeIndex) + SN_time) {
+				// Find the cell containing the particle
+				int base_i = static_cast<int>(amrex::Math::floor((p.pos(0) - plo[0]) * dxi[0]));
+				int base_j = static_cast<int>(amrex::Math::floor((p.pos(1) - plo[1]) * dxi[1]));
+				int base_k = static_cast<int>(amrex::Math::floor((p.pos(2) - plo[2]) * dxi[2]));
+
+				// Note: stencil_width <= nghost_cc_ is required!!!
```
BenWibking: Can you add a `static_assert` here to guarantee this is always this case? Then compiling will just fail if this the stencil is made too big.

src/simulation.hpp
```diff
@@ -1015,6 +1017,10 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 		// TODO(cch): Need to take care of AMR subscycling
 		particleRegister_.createParticlesFromState(state_new_cc_[0], 0, cur_time, dt_[0]);
 
+		// Stellar evolution and SN deposition
+		// TODO(cch): Need to take care of AMR subscycling
```
BenWibking: Typo: should be "subcycling"

src/simulation.hpp
```diff
@@ -1015,6 +1017,10 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
...
+		particleRegister_.depositSN(state_new_cc_[0], 0, cur_time + dt_[0]);
+
 		// Use the new type-aware particle destruction method
 		// TODO(cch): Need to take care of AMR subscycling
```
BenWibking: Typo: "subcycling"

BenWibking (changes requested):
It looks good, except I'm still not sure what to do about the particle component indices. Maybe we just ignore that for this PR.

src/particles/particle_deposition.hpp
chongchonghe: Done.

BenWibking (approved):

PR #964, opened 2025-04-01
Add tests of SNR deposition on multiple levels

chongchonghe:
### Description

In this PR, all particle operations (creation, destruction and SNR deposition) are fully tested on multiple levels. Assuming particles always live in the finest level, SNR deposition is accurate to machine accuracy even if they are near the coarse-fine boundary. Particles in coarser grids near coarse-fine boundary might work just fine, although I haven't tested it because the gravity solver doesn't work in this situation. 

Now we have two problems to test particle operations:

1. `Gravity3D`. This problem tests particle creation from fluid cells, stellar evolution (from SNProgenitor to SNRemnant), supernova deposition, and stellar destruction with fully refined grids (lev=1 in all domain). In the end of the simulation, we check (1) the number of Test particles, (2) the SNR mass, and (3) the orbit of the CIC particles (binary star in circular orbit), which is checked agains analytic solution. The Test particles are located at the corners of MultiFab boxes where eight boxes intersect. SNR mass is accurate to machine accuracy. 
2. `Gravity3DAMR`. This is the same test problem but with half of the domain refined to lev=1 and the other half in lev=0. The CIC particle is also removed due to the failure of the gravity solver when a particle crosses a coarse-fine boundary. The Test particles are located at the intersections between MultiFab boxes, and also at the coarse-fine boundary. SNR deposition is accurate to machine accuracy at the step right after deposition, but the relative error increases to ~1e-6 several steps after. Is mass a conserved quantity when AMR is turned on? 
2. Rewrite `getParticleData()` (now named `getParticleDataAtLevel()`) to make it work on any AMR level.
4. Removed `getParticlePositions()` since `getParticleDataAtLevel()` provides the same (and more) functionality. 

### Checklist


BenWibking:
I think we need to test the correctness of the deposition in this PR, otherwise it doesn't solve the original problem. Maybe you could read the output with a Python script using yt and check that it's correct that way?

BenWibking:
I think the most important thing to check is that particle deposition is conservative even when the particle position is exactly at a coarse-fine AMR boundary, since ParticleToMesh does special tricks to handle that case:
https://github.com/AMReX-Codes/amrex/blob/8534cf2685c3288d41ac161e34d78c06aabf5a20/Src/AmrCore/AMReX_AmrParticles.H#L208

So it should be possible to check this by doing MultiFab reductions over all levels, and manually checking that the total mass is gas mass + mass of all particles.

chongchonghe:
> I think the most important thing to check is that particle deposition is conservative even when the particle position is exactly at a coarse-fine AMR boundary, since ParticleToMesh does special tricks to handle that case: https://github.com/AMReX-Codes/amrex/blob/8534cf2685c3288d41ac161e34d78c06aabf5a20/Src/AmrCore/AMReX_AmrParticles.H#L208
> 
> So it should be possible to check this by doing MultiFab reductions over all levels, and manually checking that the total mass is gas mass + mass of all particles.

Check total mass is easy and can be done with a single `state.sum(...)`. I have done that in this test. The only thing that is lacking is to read the positions of the particles and check if they agree with the expectation or if the orbit maintains some property. I am going to check it with a Python script with YT and also add a regression test to do that. I'll use that regression test to validate all future particle operations as well.

chongchonghe:
@BenWibking https://github.com/AMReX-Codes/amrex/pull/4396#event-17136994157 this PR is merged. Can you bump it to Quokka?

BenWibking:
Feel free to merge https://github.com/quokka-astro/quokka/pull/974 once the GPU tests pass.

chongchonghe:
@BenWibking @markkrumholz Ready for your review.

chongchonghe:
Particle creation, stellar evolution, SN deposition, and particle destruction are fully tested on multiple levels, including at refinement boundaries or box boundaries. The following images show SN deposition at the corner of MultiFab boxes (left) and at refinement boundary (right). 

</img> </img>

markkrumholz (dismissed):

BenWibking:
@chongchonghe It looks like there is now a merge conflict that you will have to resolve manually.

chongchonghe:
@BenWibking should be ready now.

BenWibking (approved):

BenWibking:
Feel free to merge once the tests pass.

src/problems/Gravity3D/gravity_3d.cpp
```diff
@@ -313,98 +350,141 @@ auto problem_main() -> int
...
+			}
+		}
 
-		// ----- Check SN remnant mass -----
+		// check SN remnant mass
+		double SNR_mass_rel_err_tol = 1.0e-12; // should be close to machine precision
+		if (refine_half_domain) {
+			// SNR mass is not conserved to machine precision due to AMR interpolation.
+			// The relative error caused by AMR interpolation at coarse-fine boundary is exceptionally large. The relative error of SNR mass right
+			// after the SN explosion is within machine precision. However, it grows to 1e-4 after 4 steps, and goes to 1e-2 after 40 steps.
```
BenWibking, line 472: @chongchonghe Somehow I missed this part of this PR. The AMR interpolation itself is conservative. That part of AMReX has been used and tested for >30 years. The non-conservation has to be coming from somewhere else in the code.

PR #965, opened 2025-04-01
add metadata.yaml to slice outputs

chongchonghe:
### Description

This adds a metadata.yaml file to slicez_pltxxxxx folders. This is helpful because when yt loads that folder (it can, very sweet), yt will recognize it as a Quokka simulation output and do all the customization for Quokka. 

Note that this yt customization for Quokka is implemented and tested in [my yt fork](https://github.com/chongchonghe/yt) and we are in the process of publishing it before the next yt release. Demo of using the new YT frontend is here: https://github.com/Rongjun-ANU/README-of-yt-frontend-for-QUOKKA/blob/main/README.ipynb

Other changes: 
- this also adds `tests/slice*` to .gitignore

### Related issues
None.

### Checklist

### DIFF
--- a/src/io/DiagFramePlane.cpp
+++ b/src/io/DiagFramePlane.cpp
@@ -259,29 +259,15 @@ void DiagFramePlane::processDiag(int a_nstep, const amrex::Real &a_time, const a
 			diagfile = m_diagfile + std::to_string(a_time);
 		}
 		amrex::Vector<int> const step_array(nlevs, a_nstep);
-		Write2DMultiLevelPlotfile(diagfile, nlevs, GetVecOfConstPtrs(planeData), m_fieldNames, pltGeoms, a_time, step_array, ref_ratio);
-
-		// Write metadata file
-		if (amrex::ParallelDescriptor::IOProcessor()) {
-			std::string const MetadataFileName(diagfile + "/metadata.yaml");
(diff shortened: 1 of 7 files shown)


src/io/DiagFramePlane.cpp
```diff
@@ -310,6 +313,23 @@ void DiagFramePlane::Write2DMultiLevelPlotfile(const std::string &a_pltfile, int
 
 		PlaneFile.flush();
 		PlaneFile.close();
+
+		// Write metadata file
+#ifdef QUOKKA_USE_OPENPMD
+		std::string const MetadataFileName(a_pltfile + ".metadata.yaml");
+#else
+		std::string const MetadataFileName(a_pltfile + "/metadata.yaml");
```
BenWibking, line 319: The slices are always in AMReX plotfile format, so we can always use `/metadata.yaml`.

BenWibking (changes requested):
One minor change requested:

src/io/DiagFramePlane.cpp
chongchonghe: OK, good.

chongchonghe: Changed.

BenWibking (approved):

PR #506, opened 2024-01-26
Isolated disk galaxy

BenWibking:
### Description
Initializes a isolated disk galaxy simulation with collisionless particles and gas for a user-specified disk scale height, disk scale length, and disk mass.

Static mesh refinement is specified in terms of a maximum cylindrical radius $R_{max}$ and a maximum height $z_{max}$. The user must also provide i) an ASCII file with the particle initial conditions and ii) an ASCII file with the circular velocity tabulated as a function of radius (used to initialize the gas velocities).

_Caveat:_ These initial conditions do not have a realistic CGM (it is at a constant density and temperature).

This problem is added to the nightly regression tests.

### Related issues
N/A

### Checklist
_Before this pull request can be reviewed, all of these tasks should be completed. Denote completed tasks with an `x` inside the square brackets `[ ]` in the Markdown source below:_
- [x] I have added a description (see above).
- [x] I have added a link to any related issues see (see above).
- [x] I have read the [Contributing Guide](https://github.com/quokka-astro/quokka/blob/development/CONTRIBUTING.md).
- [ ] I have added tests for any new physics that this PR adds to the code.
- [x] I have manually triggered the GPU tests with the magic comment `/azp run`.

PR #966, opened 2025-04-01
Stochastic Stellar Population

aditivijayan:
### Description
This branch includes a specialisation for the particles. We check cells for Jeans stability condition and if it is Jeans unstable, we add a population of particles. The population comprises one particle representing low mass star and a random number of high mass particles. The number of high mass particles is determined using the Chabrier IMF, which is log normal for M<Msun and a power law with a slope of -2.35 above that. It is assumed that the IMF covers a mass range 0.08-120 Msun. The fate and lifetime of the high mass particles are interpolated using tables from SLUG.

Once added, the high mass stars gain a momentum weighted over neighbouring particles while the low mass particle gains an opposite momentum. At the end of particle updates, the density, total energy and the internal energy density of the gas cell. 

### Related issues
N/A

### Checklist


src/particles/stellarpop_data.hpp
```diff
@@ -0,0 +1,128 @@
...
+	    3.81049975e+06, 3.72820650e+06, 3.66312475e+06, 3.59110975e+06, 3.52811250e+06, 3.46507475e+06, 3.40187575e+06, 3.33257625e+06, 3.26777575e+06,
+	    3.15637975e+06, 2.95494125e+06, 2.92330350e+06, 2.87439675e+06, 2.80228975e+06, 2.73674900e+06, 2.68266550e+06
+
+	};
+
+	// Interpolate to find the accurate g-value from array-- because linterp doesn't work on Setonix
+
+	size_t ii = 0;
+	double x_interp = mass_star / Msun;
+	while (ii < AGE_ARR_SIZE - 1 && x_interp > interp_mass_star[ii + 1]) {
```
BenWibking: Linear search should be avoided. You can use the existing 1D and 2D interpolation functions in `math/`.

src/particles/stellarpop_data.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include "AMReX_Array.H"
+#include "AMReX_BLassert.H"
+#include "AMReX_Extension.H"
+#include "AMReX_GpuQualifiers.H"
+#include "AMReX_REAL.H"
+
+using Real = amrex::Real;
+static constexpr int FATE_ARR_SIZE = 200;
+static constexpr int AGE_ARR_SIZE = 196;
+static constexpr Real Msun = 1.989e33;
```
BenWibking: You can use the existing constant in Microphysics `C::M_solar`.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +137,219 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+			auto engine = amrex::RandomEngine();
+			const amrex::Real cell_volume = AMREX_D_TERM(dx[0], *dx[1], *dx[2]);
+			const amrex::Real cell_density = state_arr(i, j, k, HydroSystem<problem_t>::density_index);
+			const amrex::Real intenergy = state_arr(i, j, k, HydroSystem<problem_t>::x2Momentum_index) / cell_density;
+
+			const amrex::Real cs = HydroSystem<problem_t>::ComputeSoundSpeed(state_arr, i, j, k);
+			const amrex::Real LambdaJ = cs / std::sqrt(C::Gconst * cell_density);
+			int num_star = 0;
+
+			if (LambdaJ < J * dx[0]) {
```
BenWibking: Do you always want to create a stellar population if it's Jeans unstable? I think this might be undesirable, because then we can't control the star formation rate per free-fall time. In TIGRESS, I think they also check for flow convergence (and maybe alsohave a density threshold?) in order to avoid creating particles from cells that than only Jeans unstable for a couple of timesteps.

BenWibking (changes requested):

src/particles/stellarpop_data.hpp
aditivijayan: I get this linker error when I use interpolate_value from math/interpolate.hpp - 
`FAILED: src/problems/StellarPop/stellarpop 
: && /opt/nci/bin/c++ -O3 -DNDEBUG -pthread src/problems/StellarPop/CMakeFiles/stellarpop.dir/stellarpop.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/main.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/io/DiagBase.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/io/DiagFilter.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/io/DiagFramePlane.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/io/DiagPDF.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/io/projection.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/cooling/GrackleLikeCooling.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/cooling/GrackleDataReader.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/cooling/TabulatedCooling.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/cooling/CloudyDataReader.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/__/extern/Microphysics/interfaces/eos_data.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/__/extern/Microphysics/interfaces/network_initialization.cpp.o src/problems/StellarPop/CMakeFiles/stellarpop.dir/__/__/extern_parameters.cpp.o -o src/problems/StellarPop/stellarpop  -lstdc++fs  amrex/Src/libamrex_3d.a  fmt/libfmt.a  /apps/hdf5/1.12.2/lib/libhdf5.so  yaml-cpp/libyaml-cpp.a  -pthread  /apps/openmpi-mofed5.6-pbs2021.1/4.1.4/lib/libmpi_cxx.so  /apps/openmpi-mofed5.6-pbs2021.1/4.1.4/lib/libmpi.so && :
src/problems/StellarPop/CMakeFiles/stellarpop.dir/stellarpop.cpp.o: In function `void quokka::ParticleCreationTraits<(quokka::ParticleType)3>::ParticleCreator<StellarPop>::operator()<amrex::Particle<6, 1>, amrex::Array4<double> >(amrex::Particle<6, 1>*, int, amrex::Array4<double> const&, int, int, int, amrex::GpuArray<double, 3u> const&, amrex::GpuArray<double, 3u> const&, long) const':
stellarpop.cpp:(.text._ZNK6quokka22ParticleCreationTraitsILNS_12ParticleTypeE3EE15ParticleCreatorI10StellarPopEclIN5amrex8ParticleILi6ELi1EEENS7_6Array4IdEEEEvPT_iRKT0_iiiRKNS7_8GpuArrayIdLj3EEESK_l[_ZNK6quokka22ParticleCreationTraitsILNS_12ParticleTypeE3EE15ParticleCreatorI10StellarPopEclIN5amrex8ParticleILi6ELi1EEENS7_6Array4IdEEEEvPT_iRKT0_iiiRKNS7_8GpuArrayIdLj3EEESK_l]+0xe07): undefined reference to `interpolate_value(double, double const*, double const*, int)'
collect2: error: ld returned 1 exit status`

BenWibking: Currently, you have to modify the `CMakeLists.txt` for your problem to add `interpolate.cpp` manually. We should probably change this so it's not necessary. But for now you can follow the example here:
https://github.com/quokka-astro/quokka/blob/c5cba08b8781913c64642d4f42fd9e3262ebc9ff/src/problems/HydroShocktube/CMakeLists.txt#L1

src/particles/particle_creation.hpp
aditivijayan: Is there self-gravity in the gas? My understanding was that we don't yet have sink particles.

BenWibking: There is in my disk galaxy test problem: https://github.com/quokka-astro/quokka/pull/506. I was thinking we would have self-gravity for everything in the INCITE simulations as well.

aditivijayan: Right. In that case, I guess I need to chat with @markkrumholz.

markkrumholz: Happy to chat about this, as there are a few options. 

For the sort of galaxy-style star particles we're implementing here, I don't think there is really a need to check for flow convergence -- that's something people generally do for sink particles in higher resolution simulations, not in galaxy simulations, and I don't think that there's really much advantage to it. We could certainly add a flat density threshold in addition to something Jeans-based, but I don't see that as really necessary or important to do. I don't really see what they add beyond the Jeans-based criterion, which for dense gas that returns to thermal equilibrium quickly is almost identical. I basically see Jeans-based particle creation as just a labor-saving device to avoid the need to calculate an appropriate density threshold from your resolution and cooling curve by hand.

The question of controlling the star formation efficiency per free-fall time is more interesting. I had suggested implementing this as a rule that you take a convert a fixed fraction eps_* of the mass in a Jeans-violating to stars, which effectively does set an eps_ff -- the logic here is that, if make the criterion for star creation that the local Jeans length lambda_J > J dx for some Jeans number J and cell size dx, and if this condition is met you convert a fraction eps_* of the gas mass to stars in a time step of size dt, you can work out that the effective value of eps_ff for a cell at threshold is eps_ff = eps_* sqrt(3/32) J dx / (c_s dt), where c_s is the gas sound speed. So effectively eps_* gives you a knob by which you can adjust eps_ff. But it might be more desirable to set eps_ff directly, in which case one can simply say that the target stellar mass to be created should be eps_ff M_cell dt / t_ff. In order to avoid creating too many tiny particles when eps_ff is small, one might then want to also specify a minimum eps_* value and implement star formation stochastically, in which case the rule becomes that in any Jeans-violating cell you have a probability p_sf = 1 - exp(-eps_ff dt / eps_* t_ff) of converting a fraction eps_* of the cell mass to star particles during that time step. This might be a better implemention, and minimal extra work to change to.

BenWibking: Ok, having a pure Jeans threshold sounds fine.

I do think it would be a significant improvement to compute p_sf with a minimum stellar mass particle threshold. It would help us avoid having a huge number low-mass particles that I don't think we would care about for feedback anyway.

markkrumholz: I would prefer to implement this as a threshold eps_* rather than a threshold star particle mass, because this avoids needing to re-tune the implementation manually if you change the resolution, and provides a nice resolution-independent way of adjusting the characteristic star particle mass to match your resolution. But other than this detail I think we are agreed.

BenWibking: Yes, I think we agree except for this point. I think a threshold star particle mass would be preferable since it's possible for star particles to form on levels $l$ < level_max. Or were you thinking of also requiring the gas to maximally refined for star formation to occur?

markkrumholz: Yes, I think we definitely want to enforce a requirement that star particles be created only at max level, because if you're Jeans-violating on a level other than max, the correct thing to do is to refine, not to spawn a star.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +137,236 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+		amrex::Real param1 = particle_param1;
+		amrex::Real param2 = particle_param2;
+
+		AMREX_GPU_HOST_DEVICE ParticleChecker(amrex::Real current_time, amrex::Real dt) : current_time(current_time), dt(dt) {}
+
+		AMREX_GPU_DEVICE auto operator()(amrex::Array4<const amrex::Real> const &state_arr, int i, int j, int k,
+						 amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &dx) const -> int
+		{
+			// Estimate the Jeans length(LambaJ) and create particle if LambaJ < J*dx
+			amrex::ignore_unused(state_arr, dx);
```
chongchonghe: remove this line since state_arr and dx are used.

src/particles/particle_creation.hpp
chongchonghe: I agree that setting a constant value of eps_ff is more robust since it's a well-understood number and resolution-independent. However, we need to set a cap on the SFE at a step since eps_ff * dt / t_ff > 1 could happen and we don't want to bring the gas density to zero. I like Mark's idea of stochastical SF which avoids this problem and avoid too many low-mass stars. My only worry about a formula like this one `p_sf = 1 - exp(-eps_ff dt / eps_* t_ff)` is that `p_sf` is never 1, then you have the problem that a cell may collapse to beyond Jeans instability and still not forming stars. I suggest we have a formula that gives `p_sf = 1` when `eps_ff * dt / t_ff > 1` or  `eps_ff dt / t_ff > eps_*`.

markkrumholz: Agreed, and in fact when @aditivijayan and I met to talk about this today, we agreed to exactly that modification.

BenWibking: Yes, that would have avoided a lot of problems with my GIZMO simulations...

BenWibking: Thinking about this further, I am less certain that guaranteeing star formation above a certain density (or equivalent threshold) makes sense for an AMR code. In GIZMO, the main problem this caused was that there was no way to halt the collapse toward ever-higher densities without converting gas particles to star particles, so this could cause a runaway time step problem for a small number of particles.

Since we have a maximum AMR level, we don't have this problem. We will get numerical fragmentation if we don't form stars, but we could add a pressure floor to counteract that. The main concern I have is that the actual star formation efficiency per free-fall time will no longer be the nominal value of `eps_ff` once a "guaranteed star formation" ceiling is imposed.

markkrumholz: I actually don't think this is a concern; it's something @aditivijayan went through together in person. So, a few points:

-- I agree that we don't have to worry about the issue in GIZMO of particles running away to very high densities if we don't have high eps_ff. In an AMR code this won't happen, because you are always softening gravity and forces on the grid scale. You will of course violate Jeans and get artificial fragmentation, but that is not really a problem in a galaxy simulation. Artificial fragmentation is bad if you care about the mass distribution of the collapsed objections, so you definitely want to avoid it in simulations where the science goal is the IMF, binaries, or really anything related to individual star formation. That is not our situation here, however, so I do not worry about artificial fragmentation in a galaxy simulation.

-- It is helpful to actually work through the effective eps_ff produced by different star formation prescriptions. Suppose we have a N_cell identical cells of mass M and density rho (corresponding to free-fall time t_ff), and we have a star formation prescription that says that, for Jeans-violating cells, we have a probability P of spawning star particles of total mass eps_* M in each cell per time step dt. In this case the total gas mass is N_cell M and the expected star formation rate is SFR = N_cell eps_* M P / dt, so the effective eps_ff during the time step is eps_ff = SFR / (N_cell M/t_ff) = P eps_* t_ff / dt. Thus a valid choice is P = (eps_ff / eps_*) (dt / t_ff). Now there is some subtlety in whether you want to make the SFR match eps_ff averaged over the time step or instantaneously; this calculation is for the latter, but we could equally well have said that we want to solve for the probability per unit time dp/dt of forming stars, in which case the instantaneous star formation rate is N_cell eps_* M dp/dt, and the condition that we get the right eps_ff becomes eps_ff = eps_* t_ff dp/dt. Integrating this equation over time dt gives rise to the alternative form of the probability, P = 1 - exp[-(eps_ff/eps_*) (dt / t_ff)]. So we can see that the two alternative forms just amount to different time centerings that both implement the condition that the star formation efficiency per free-fall time be eps_ff -- P =  (eps_ff / eps_*) (dt / t_ff) amounts to requiring this condition integrated over the time step, while P = 1 - exp[-(eps_ff/eps_*) (dt / t_ff)] amounts to requiring this instantaneously at the start of the time step. For time steps dt << t_ff, they are the same.

-- This also gets to the issue of whether guaranteed star formation is a problem. This situation in which the first prescription guarantees that star formation occur is when (eps_ff / eps_*) (dt / t_ff) > 1, or equivalently when dt > (eps_* / eps_ff) t_ff. In this case, we guarantee that we will spawn a star of mass eps_* in very cell, and thus SFR = eps_* M N_cell / dt, and eps_ff = eps_* t_ff / dt; for big dt this will indeed be < eps_ff -- but we can see now see that this is simply because you were computing eps_ff using the mass that existed at the start of the time step. That is, in evaluating eps_ff = SFR / (M / t_ff), the M you plugged into the denominator was the mass at the start of the time step. But that can't be the right thing to do when dt >> t_ff / eps_ff, because in that case you expect to consume a majority of the mass over a single time step. And in fact in that case you shouldn't keep t_ff constant either, since the density is depleting. In this situation the second prescription for P is no better, because it is also evaluated using the free-fall time as it existed at the start of the time step. So my conclusion is that the two forms for P are basically interchangeable when your time steps resolve the star formation time, and that when you don't neither one of them really gives you the right answer, because they are both evaluated using conditions as they existed at the start of the time step.

BenWibking: Ok, that seems reasonable. However, I thought things were operator split such that it evaluates the particle creation criteria using the state at the _end_ of the time step, no? I don't think this affects your conclusion, though.

BenWibking: It would be good to carefully document this, since (AFAIK) every other code considers this as the differential star formation probability. At the resolution we are running it, it won't matter since the time step will be very small, but at coarser resolution, it could give surprisingly different answers from other codes (although still wrong, as you point out).

BenWibking: Maybe this PR could include a documentation page, e.g. `docs/docs/star_formation.md`?

markkrumholz: Agree, I was being cavalier about before versus after the time step. And it occurs to me that if one wanted to be really anal, we could in fact solve the equation dM_* / dt = eps_ff M / t_ff properly accounting for both the time evolution of the mass and free-fall time over the time step to figure out the mass of stars created -- for fixed volume, the exact solution is that the stellar mass after a time t is
`M_0 [1 - 4 / (2 + tau eps_ff)^2]`, where `tau = t / t_ff` and `M_0` is the initial gas mass in the cell. So this gives the expected mass in stars that should be created after any amount of time; we could therefore say that we take the probability to spawn a star of mass `M_* = eps_* M_0` is `P = [1 - 4 / (2 + tau eps_ff)^2] / eps_*`. This correctly implements the outcome of star formation including both the change in the gas mass and the gas density (which the standard exponential solution does not, since it implicitly assumes constant free-fall time). But I still think that you will not really get the right answer no matter what if your time step is much bigger than the free-fall time, because then you're not resolving the time scale on which gravity brings new mass into the cell. In this limit, your star formation rate is not going to be limited by your star formation prescription, it's going to be limited by the maximum rate at which you can add gas to a cell given your time step.

@BenWibking, I also think this means that your concern about what happens when the time step is big at coarse resolution is not really a problem. That is because if you're forming stars, you presumably have stuff moving at near free-fall speeds, in which case you are necessarily going to resolving the free-fall timescale at least marginally. Thus as long as `eps_ff` is not too close to unity, and you're not starting from an artificial condition where star formation is occurring but somehow all the gas velocities are much smaller than the free-fall velocity, the CFL alone will guarantee that you always resolve the star formation timescale well.

docs/docs/gallery.md
BenWibking, line 1: It looks like the tests won't run because there's a merge conflict for this file. Did you mean to change it? If not, you should just be able to checkout this file from the development branch and commit.

aditivijayan:
I need to use dt for estimating the probability. I've added dt to stellar pop specialisation but it crashes because of conflicts with other problems. Should I just go ahead fix them or is there a way to access in ParticleChecker without changing the way it is defined right now?

BenWibking:
> I need to use dt for estimating the probability. I've added dt to stellar pop specialisation but it crashes because of conflicts with other problems. Should I just go ahead fix them or is there a way to access in ParticleChecker without changing the way it is defined right now?

I think it would make sense to add it to all of the ParticleCheckers. The particle checker is not part of QuokkaSimulation, so I don't think there is another way to do it.

aditivijayan:
I am unable to debug this linking error.

BenWibking:
> I am unable to debug this linking error.

Can you post the error? I don't see it.

aditivijayan:
> [100%] Linking CXX executable gravity_3d
/usr/bin/ld: CMakeFiles/gravity_3d.dir/gravity_3d.cpp.o: in function `void quokka::ParticleCreationImpl::createParticlesImpl<BinaryOrbit, amrex::AmrParticleContainer_impl<amrex::Particle<5, 1>, 0, 0, std::allocator, amrex::DefaultAssignor>, quokka::ParticleCreationTraits<(quokka::ParticleType)3>::ParticleChecker, quokka::ParticleCreationTraits<(quokka::ParticleType)3>::ParticleCreator>(amrex::AmrParticleContainer_impl<amrex::Particle<5, 1>, 0, 0, std::allocator, amrex::DefaultAssignor>*, int, amrex::MultiFab&, int, double, double, int, int)':
gravity_3d.cpp:(.text._ZN6quokka20ParticleCreationImplL19createParticlesImplI11BinaryOrbitN5amrex25AmrParticleContainer_implINS3_8ParticleILi5ELi1EEELi0ELi0ESaNS3_15DefaultAssignorEEENS_22ParticleCreationTraitsILNS_12ParticleTypeE3EE15ParticleCheckerENSB_15ParticleCreatorEEEvPT0_iRNS3_8MultiFabEiddii[_ZNK6quokka25PhysicsParticleDescriptorIN5amrex25AmrParticleContainer_implINS1_8ParticleILi5ELi1EEELi0ELi0ESaNS1_15DefaultAssignorEEE11BinaryOrbitLNS_12ParticleTypeE3EE24createParticlesFromStateERNS1_8MultiFabEidd]+0x1c24): undefined reference to `interpolate_value(double, double const*, double const*, int)'
/usr/bin/ld: gravity_3d.cpp:(.text._ZN6quokka20ParticleCreationImplL19createParticlesImplI11BinaryOrbitN5amrex25AmrParticleContainer_implINS3_8ParticleILi5ELi1EEELi0ELi0ESaNS3_15DefaultAssignorEEENS_22ParticleCreationTraitsILNS_12ParticleTypeE3EE15ParticleCheckerENSB_15ParticleCreatorEEEvPT0_iRNS3_8MultiFabEiddii[_ZNK6quokka25PhysicsParticleDescriptorIN5amrex25AmrParticleContainer_implINS1_8ParticleILi5ELi1EEELi0ELi0ESaNS1_15DefaultAssignorEEE11BinaryOrbitLNS_12ParticleTypeE3EE24createParticlesFromStateERNS1_8MultiFabEidd]+0x1c85): undefined reference to `interpolate_value(double, double const*, double const*, int)'
collect2: error: ld returned 1 exit status
gmake[2]: *** [src/problems/Gravity3D/CMakeFiles/gravity_3d.dir/build.make:316: src/problems/Gravity3D/gravity_3d] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:5177: src/problems/Gravity3D/CMakeFiles/gravity_3d.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
Error: Process completed with exit code 2. 

This is the error.

BenWibking:
> > [100%] Linking CXX executable gravity_3d
> > /usr/bin/ld: CMakeFiles/gravity_3d.dir/gravity_3d.cpp.o: in function `void quokka::ParticleCreationImpl::createParticlesImpl<BinaryOrbit, amrex::AmrParticleContainer_impl<amrex::Particle<5, 1>, 0, 0, std::allocator, amrex::DefaultAssignor>, quokka::ParticleCreationTraits<(quokka::ParticleType)3>::ParticleChecker, quokka::ParticleCreationTraits<(quokka::ParticleType)3>::ParticleCreator>(amrex::AmrParticleContainer_impl<amrex::Particle<5, 1>, 0, 0, std::allocator, amrex::DefaultAssignor>*, int, amrex::MultiFab&, int, double, double, int, int)': gravity_3d.cpp:(.text._ZN6quokka20ParticleCreationImplL19createParticlesImplI11BinaryOrbitN5amrex25AmrParticleContainer_implINS3_8ParticleILi5ELi1EEELi0ELi0ESaNS3_15DefaultAssignorEEENS_22ParticleCreationTraitsILNS_12ParticleTypeE3EE15ParticleCheckerENSB_15ParticleCreatorEEEvPT0_iRNS3_8MultiFabEiddii[_ZNK6quokka25PhysicsParticleDescriptorIN5amrex25AmrParticleContainer_implINS1_8ParticleILi5ELi1EEELi0ELi0ESaNS1_15DefaultAssignorEEE11BinaryOrbitLNS_12ParticleTypeE3EE24createParticlesFromStateERNS1_8MultiFabEidd]+0x1c24): undefined reference to `interpolate_value(double, double const*, double const*, int)'
> > /usr/bin/ld: gravity_3d.cpp:(.text._ZN6quokka20ParticleCreationImplL19createParticlesImplI11BinaryOrbitN5amrex25AmrParticleContainer_implINS3_8ParticleILi5ELi1EEELi0ELi0ESaNS3_15DefaultAssignorEEENS_22ParticleCreationTraitsILNS_12ParticleTypeE3EE15ParticleCheckerENSB_15ParticleCreatorEEEvPT0_iRNS3_8MultiFabEiddii[_ZNK6quokka25PhysicsParticleDescriptorIN5amrex25AmrParticleContainer_implINS1_8ParticleILi5ELi1EEELi0ELi0ESaNS1_15DefaultAssignorEEE11BinaryOrbitLNS_12ParticleTypeE3EE24createParticlesFromStateERNS1_8MultiFabEidd]+0x1c85): undefined reference to `interpolate_value(double, double const*, double const*, int)'
> > collect2: error: ld returned 1 exit status
> > gmake[2]: *** [src/problems/Gravity3D/CMakeFiles/gravity_3d.dir/build.make:316: src/problems/Gravity3D/gravity_3d] Error 1
> > gmake[1]: *** [CMakeFiles/Makefile2:5177: src/problems/Gravity3D/CMakeFiles/gravity_3d.dir/all] Error 2
> > gmake: *** [Makefile:146: all] Error 2
> > Error: Process completed with exit code 2.
> 
> This is the error.

Ok, I see the issue. Since you use the `interpolate_value` function in a particles header, `math/interpolate.cpp` has to be included as a source file for all problems that use particles. So you can just add `../../math/interpolate.cpp` to `src/problems/Gravity3D/CMakeLists.txt`.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,242 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+	static constexpr amrex::Real m_imf_break = 8.0 * C::M_solar; // IMF is lognormal below this mass and powerlaw above
+	static constexpr amrex::Real imf_disp = 0.55;		     // dispersion of the lognormal IMF
+	static constexpr amrex::Real imf_mu = 32.599;		     //=log10(0.2 * C::M_solar), mean of the lognormal IMF, avoid compiler error
+	static constexpr amrex::Real alpha = 2.35;		     // slope of the powerlaw
+
+	static double fstar_high;      // fstar is the fraction of number of high mass stars from the IMF
+	static double m_star_high_avg; // average mass of high mass stars
+
+	ParticleCreationTraits()
+	{
```
markkrumholz: I would probably be good to add some comments to this routine to explain the math that is being done here, just so that other people (including your future self 6 months from now) can understand what is being calculated.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,242 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+		amrex::Real dt;
+		amrex::Real param1 = particle_param1;
+		amrex::Real param2 = particle_param2;
+
+		AMREX_GPU_HOST_DEVICE ParticleChecker(amrex::Real current_time, amrex::Real dt) : current_time(current_time), dt(dt) {}
+
+		AMREX_GPU_DEVICE auto operator()(amrex::Array4<const amrex::Real> const &state_arr, int i, int j, int k,
+						 amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &dx) const -> int
+		{
+			// Estimate the Jeans length(LambaJ) and create particle if LambaJ < J*dx
```
markkrumholz: This comment should probably be changed, since it looks like it dates from before creation was stochastic.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,241 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+							}
+						}
+						sigma_sq_x = numx / denominator;
+						sigma_sq_y = numy / denominator;
+						sigma_sq_z = numz / denominator;
+
+						p.rdata(mass_idx + 1) = (std::abs(vx) / vx) * amrex::RandomNormal(std::abs(vx), std::sqrt(sigma_sq_x), engine);
+						p.rdata(mass_idx + 2) = (std::abs(vy) / vy) * amrex::RandomNormal(std::abs(vy), std::sqrt(sigma_sq_y), engine);
+						p.rdata(mass_idx + 3) = (std::abs(vz) / vz) * amrex::RandomNormal(std::abs(vz), std::sqrt(sigma_sq_z), engine);
+
```
markkrumholz: I don't understand why you need a loop here. You can just draw a single value and guarantee that the mass falls between min and max.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,241 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+									numz += rho_adj * (vz_adj - (vz)) * (vz_adj - (vz));
+
+									denominator += rho_adj;
+								}
+							}
+						}
+						sigma_sq_x = numx / denominator;
+						sigma_sq_y = numy / denominator;
+						sigma_sq_z = numz / denominator;
+
```
markkrumholz: Again, it would be good to add some comments here to explain the logic of what you're doing -- i.e., that you're estimating the velocity dispersion from this cell plus its 8 neighbors, and using that to assign a random velocity to the particles.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,241 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+				}
+
+				if (num_particles > 1) { // Update momentum of the low mass star if there is(are) high mass star(s) in the cell
+					const int p_idx = 0;
+					auto &plow = particles[p_idx]; // NOLINT
+					plow.rdata(mass_idx + 1) = -total_momx / plow.rdata(mass_idx);
+					plow.rdata(mass_idx + 2) = -total_momy / plow.rdata(mass_idx);
+					plow.rdata(mass_idx + 3) = -total_momz / plow.rdata(mass_idx);
+				}
+				state_arr(i, j, k, HydroSystem<problem_t>::density_index) = (cell_mass - particle_mass) / cell_volume;
```
markkrumholz: It looks like you're not updating the cell momentum here, which is not correct -- if you reduce the mass while leaving the momentum unchanged, the velocity will increase.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,241 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+				vx = state_arr(i, j, k, HydroSystem<problem_t>::x1Momentum_index) / state_arr(i, j, k, HydroSystem<problem_t>::density_index);
+				vy = state_arr(i, j, k, HydroSystem<problem_t>::x2Momentum_index) / state_arr(i, j, k, HydroSystem<problem_t>::density_index);
+				vz = state_arr(i, j, k, HydroSystem<problem_t>::x3Momentum_index) / state_arr(i, j, k, HydroSystem<problem_t>::density_index);
+				state_arr(i, j, k, HydroSystem<problem_t>::internalEnergy_index) *=
+				    cell_density / state_arr(i, j, k, HydroSystem<problem_t>::density_index);
+				state_arr(i, j, k, HydroSystem<problem_t>::energy_index) =
```
markkrumholz: You don't need to separate out the velocities. The cell energy should just decrease in proportion to the mass decrease, so that you leave the energy per unit mass of the gas unchanged.

src/particles/stellarpop_data.hpp
```diff
@@ -0,0 +1,107 @@
...
+	    16.,  16.1, 16.2, 16.3, 16.4, 16.5,	 16.6, 16.7,  16.8, 16.9,  17.,	 17.1,	17.2, 17.3,  17.4, 17.5,  17.6, 17.7, 17.8, 17.9, 18.,	18.1, 18.2,
+	    18.3, 18.4, 18.5, 18.6, 18.7, 18.8,	 18.9, 19.,   19.1, 19.2,  19.3, 19.4,	19.5, 19.6,  19.7, 19.8,  19.9, 20.,  20.1, 20.2, 20.3, 20.4, 20.5,
+	    20.6, 20.7, 20.8, 20.9, 21.,  21.1,	 21.2, 21.3,  21.4, 21.5,  21.6, 21.7,	21.8, 21.9,  22.,  22.1,  22.2, 22.3, 22.4, 22.5, 22.6, 22.7, 22.8,
+	    22.9, 23.,	23.1, 23.2, 23.3, 23.4,	 23.5, 23.6,  23.7, 23.8,  23.9, 24.,	24.1, 24.2,  24.3, 24.4,  24.5, 24.6, 24.7, 24.8, 24.9, 25.,  25.1,
+	    25.2, 25.3, 25.4, 25.5, 25.6, 25.7,	 25.8, 25.9,  26.,  26.1,  26.2, 26.3,	26.4, 26.5,  26.6, 26.7,  26.8, 26.9, 27.,  27.1, 27.2, 27.3, 27.4,
+	    27.5, 27.6, 27.7, 27.8, 27.9, 28.,	 28.1, 28.2,  28.3, 28.4,  28.5, 28.6,	28.7, 28.8,  28.9, 29.,	  29.1, 29.2, 29.3, 29.4, 29.5, 29.6, 29.7,
+	    29.8, 29.9, 30.,  31.,  32.,  33.,	 35.,  40.,   45.,  50.,   55.,	 60.,	70.,  80.,   100., 120.};
+
+	const amrex::GpuArray<Real, FATE_ARR_SIZE> interp_star_fate{
```
markkrumholz, line 40: I would name this variable more specifically (say stellar_end_state or something like that), and use an enum for the possible fates, because in the future we may want to distinguish more possible fates, e.g., type II SN versus pair instability SN, different types of white dwarf, etc.

src/particles/stellarpop_data.hpp
```diff
@@ -0,0 +1,107 @@
...
+	    2.38e+00, 2.40e+00, 2.42e+00, 2.44e+00, 2.46e+00, 2.48e+00, 2.50e+00, 2.52e+00, 2.54e+00, 2.56e+00, 2.58e+00, 2.60e+00, 2.62e+00, 2.64e+00,
+	    2.66e+00, 2.68e+00, 2.70e+00, 2.72e+00, 2.74e+00, 2.76e+00, 2.78e+00, 2.80e+00, 3.00e+00, 3.20e+00, 3.40e+00, 3.60e+00, 3.80e+00, 4.00e+00,
+	    4.20e+00, 4.40e+00, 4.60e+00, 4.80e+00, 5.00e+00, 5.20e+00, 5.40e+00, 5.60e+00, 5.80e+00, 6.00e+00, 6.20e+00, 6.40e+00, 6.60e+00, 6.80e+00,
+	    7.00e+00, 7.20e+00, 7.40e+00, 7.60e+00, 7.80e+00, 8.00e+00, 9.00e+00, 1.00e+01, 1.10e+01, 1.20e+01, 1.30e+01, 1.40e+01, 1.50e+01, 1.60e+01,
+	    1.70e+01, 1.80e+01, 1.90e+01, 2.00e+01, 2.20e+01, 2.40e+01, 2.60e+01, 2.80e+01, 3.00e+01, 3.20e+01, 3.40e+01, 3.60e+01, 3.80e+01, 4.00e+01,
+	    4.50e+01, 5.00e+01, 5.50e+01, 6.00e+01, 6.50e+01, 7.00e+01, 7.50e+01, 8.00e+01, 8.50e+01, 9.00e+01, 9.50e+01, 1.00e+02, 1.05e+02, 1.10e+02,
+	    1.15e+02, 1.20e+02, 1.25e+02, 1.30e+02, 1.35e+02, 1.40e+02, 1.45e+02, 1.50e+02, 1.75e+02, 2.00e+02, 2.25e+02, 2.50e+02, 2.75e+02, 3.00e+02};
+
+	const amrex::GpuArray<Real, AGE_ARR_SIZE> interp_death_time{
```
markkrumholz, line 77: You should specify that these times are in yr, since this is a different unit than the default code time unit.

src/particles/stellarpop_data.hpp
```diff
@@ -0,0 +1,107 @@
...
+	    4.21476160e+07, 4.03531400e+07, 3.32273580e+07, 2.65855740e+07, 2.27899720e+07, 1.99802960e+07, 1.78570140e+07, 1.61082790e+07, 1.46735110e+07,
+	    1.35130840e+07, 1.25814310e+07, 1.17604370e+07, 1.10608340e+07, 1.04301560e+07, 9.43675700e+06, 8.75512700e+06, 8.42390100e+06, 7.88027700e+06,
+	    7.64807500e+06, 7.25954850e+06, 6.99782700e+06, 6.54252700e+06, 6.29859250e+06, 6.06362300e+06, 5.84142800e+06, 5.36775850e+06, 4.89408950e+06,
+	    4.68576750e+06, 4.49476350e+06, 4.38545700e+06, 4.21975200e+06, 4.09634625e+06, 4.15417150e+06, 4.04259400e+06, 3.96940425e+06, 3.89933150e+06,
+	    3.81049975e+06, 3.72820650e+06, 3.66312475e+06, 3.59110975e+06, 3.52811250e+06, 3.46507475e+06, 3.40187575e+06, 3.33257625e+06, 3.26777575e+06,
+	    3.15637975e+06, 2.95494125e+06, 2.92330350e+06, 2.87439675e+06, 2.80228975e+06, 2.73674900e+06, 2.68266550e+06
+
+	};
+
+	// Interpolate to find the accurate g-value from array-- because linterp doesn't work on Setonix
```
markkrumholz: You probably want to convert this to code time units before returning.

markkrumholz (changes requested):
I left some comments in the code where I think there is a need for some changes.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,250 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+		{
+			// A simple demonstration of particle creation
+
+			auto engine = amrex::RandomEngine();
+			if (mass_idx + 3 < ParticleType::NReal) {
+				// Calculate common values for all particles
+				const amrex::Real cell_density = state_arr(i, j, k, HydroSystem<problem_t>::density_index);
+				const amrex::Real cell_volume = AMREX_D_TERM(dx[0], *dx[1], *dx[2]);
+				const amrex::Real cell_mass = cell_volume * cell_density;
+				amrex::Real vx = state_arr(i, j, k, HydroSystem<problem_t>::x1Momentum_index) / cell_density;
```
chongchonghe: vx vy vz can also be declared as const (you will get tidy warnings)

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,250 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+									rho_adj = state_arr(ii, jj, kk, HydroSystem<problem_t>::density_index);
+									numx += rho_adj * (vx_adj - (vx)) * (vx_adj - (vx));
+									numy += rho_adj * (vy_adj - (vy)) * (vy_adj - (vy));
+									numz += rho_adj * (vz_adj - (vz)) * (vz_adj - (vz));
+
+									denominator += rho_adj;
+								}
+							}
+						}
+						sigma_sq_x = numx / denominator;
```
chongchonghe: For better readability, you can do the following since `sigma_sq_x` is only assigned once. Similarly for other variables.
```suggestion
						const double sigma_sq_x = numx / denominator;
```

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,250 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+
+				if (num_particles > 1) { // Update momentum of the low mass star if there is(are) high mass star(s) in the cell
+					const int p_idx = 0;
+					auto &plow = particles[p_idx]; // NOLINT
+					plow.rdata(mass_idx + 1) = -total_momx / plow.rdata(mass_idx);
+					plow.rdata(mass_idx + 2) = -total_momy / plow.rdata(mass_idx);
+					plow.rdata(mass_idx + 3) = -total_momz / plow.rdata(mass_idx);
+				}
+
+				double factor = NAN;
```
chongchonghe:
```suggestion
				const double factor = (cell_mass - particle_mass) / cell_volume / state_arr(i, j, k, HydroSystem<problem_t>::density_index);
```

docs/markdown/star_formation.md
```diff
@@ -0,0 +1,33 @@
+# Star Formation Recipe 
+
+## Checking for Jeans Violation
+The star formation recipe is implemented via the Stochastic Stellar Population specialisation. Every cell is checked for Jeans violation and star particles are added if $\lambda_J = c_s/\sqrt{G\rho} < J \times dx$, where J is $0.5$. 
+
+Adding star particle to every Jeans violating cell can lead to a large number of small mass stars which can be cumbersome. To alleviate this issue, we control the star formation rate through $\epsilon_{\rm ff}$ and $\epsilon_{*}$. $\epsilon_{\rm ff}$ is the efficiency per freefall time given by $=\dot{M}_* \cdot t_{\rm ff} / M_{\rm cell} \cdot dt $, while $\epsilon_{*}$ is the fraction of cell mass that gets converted into stars.
```
BenWibking: It looks like there is a LaTeX formatting issue in this paragraph:
Screenshot 2025-04-16 at 8 22 59 PM

docs/markdown/star_formation.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+## Estimating number and type of stars
+
+Once we establish that a star will form in a cell, we need to find out the stellar population looks like. In our implementation, the population comprises at least particle representing low mass star ($M<8M_{\odot}$) population and a random number of particles, each represeting a high mass star.
+
+The number of high mass stars is determined by the initial mass function, taken to be Chabrier03. The IMF is represented by a log normal for $M<M_{\odot}$ and a power law with a slope of $2.35$ beyond that. From the IMF, we can estimate the fraction of high mass stars, $f_{*,\rm{high}}$, and the average mass of high mass stars, $<m>_{*,\rm{high}}$. The expectation value of the number of high mass stars, \textit{num_high_mass_stars}, then is the ratio of the total mass in high mass stars and $<m>_{*,\rm{high}}$. 
+
+The total number of star particles to be spwaned in a cell is $1+$ Poission distribution with an expectation value of \textit{num_high_mass_stars}. A fraction $1-f_{*,\rm{high}}$ goes into the particle representing low mass star while the mass of the high mass star particles is randomly drawn from the high mass part of the IMF.
```
BenWibking: There is a LaTeX formatting issue in this paragraph:
Screenshot 2025-04-16 at 8 23 28 PM

BenWibking (dismissed):
There are a few minor LaTeX formatting issues in the documentation that should be fixed.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,250 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+				// Update the cell momentum to make sure velocities don't change
+				state_arr(i, j, k, HydroSystem<problem_t>::x1Momentum_index) *= factor;
+				state_arr(i, j, k, HydroSystem<problem_t>::x2Momentum_index) *= factor;
+				state_arr(i, j, k, HydroSystem<problem_t>::x3Momentum_index) *= factor;
+
+				// Update internal energy to relect mass change
+				state_arr(i, j, k, HydroSystem<problem_t>::internalEnergy_index) *= factor;
+
+				// Update total energy
+				state_arr(i, j, k, HydroSystem<problem_t>::energy_index) *= factor;
```
chongchonghe, line 407: I know this is mathematically correct, but it's usually more robust to compute total energy from internal and kinetic energy to avoid round-off errors and mismatch in cases where, say, the internal energy is not a linear function of temperature or density.

src/particles/particle_creation.hpp
chongchonghe: or `const double factor = (cell_mass - particle_mass) / (cell_volume * cell_density);`

chongchonghe:
I left some minor comments on the coding style and readability.

A general comment/request is that we should include a test problem for this model. It can be as simple as creating stochastic stellar pop particles from a single cell and counting the number of particles and checking for conservation laws in the end.

src/particles/particle_creation.hpp
BenWibking: As long as the EOS is an ideal gas or mixture of ideal gases, I think this should be fine as-is.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,250 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+					plow.rdata(mass_idx + 1) = -total_momx / plow.rdata(mass_idx);
+					plow.rdata(mass_idx + 2) = -total_momy / plow.rdata(mass_idx);
+					plow.rdata(mass_idx + 3) = -total_momz / plow.rdata(mass_idx);
+				}
+
+				double factor = NAN;
+				factor = (cell_mass - particle_mass) / cell_volume / state_arr(i, j, k, HydroSystem<problem_t>::density_index);
+
+				// Update the cell density to reflect mass conversion into stars
+				state_arr(i, j, k, HydroSystem<problem_t>::density_index) *= factor;
```
BenWibking, line 396: The partial densities / mass scalars (if they exist) also need to be updated here.

aditivijayan:
> I left some minor comments on the coding style and readability.
> 
> A general comment/request is that we should include a test problem for this model. It can be as simple as creating stochastic stellar pop particles from a single cell and counting the number of particles and checking for conservation laws in the end.

The test problem will be added once this PR has merged.

chongchonghe:
There are four open comments from Mark, Ben, and me. Can you address them? @aditivijayan 

In case this could be helpful. If you have installed 'GitHub Pull Requests' extension in VSCode, you can see all open comments/requested changes at the bottom "Comments" tab.

src/particles/stellarpop_data.hpp
chongchonghe: We are already using enum for fates, but they have to be stored in particles as intergers. So, as long as we have a comment here saying something like "stellar fate 1: LowMassStar, 2: SNProgenitor", it should be fine.

aditivijayan:
> There are four open comments from Mark, Ben, and me. Can you address them? @aditivijayan
> 
> In case this could be helpful. If you have installed 'GitHub Pull Requests' extension in VSCode, you can see all open comments/requested changes at the bottom "Comments" tab.

I am working through the comments one by one. If you would like me to address something urgently, you can let me know.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,254 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+
+									denominator += rho_adj;
+								}
+							}
+						}
+						const double sigma_sq_x = numx / denominator;
+						const double sigma_sq_y = numy / denominator;
+						const double sigma_sq_z = numz / denominator;
+
+						p.rdata(mass_idx + 1) = (std::abs(vx) / vx) * amrex::RandomNormal(std::abs(vx), std::sqrt(sigma_sq_x), engine);
```
chongchonghe: You need to take care of the case when vx = 0. It's easier to use a if else clause to get the sign here.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,254 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+						mass_of_star = std::pow(mass_of_star, 1. / (1. - alpha));
+						p.rdata(mass_idx) = mass_of_star;
+
+						total_momx += p.rdata(mass_idx + 1) * p.rdata(mass_idx);
+						total_momy += p.rdata(mass_idx + 2) * p.rdata(mass_idx);
+						total_momz += p.rdata(mass_idx + 3) * p.rdata(mass_idx);
+
+						p.idata(evolution_stage_index) = interpolate_fate(p.rdata(mass_idx));
+						p.rdata(birth_time_index + 1) = interpolate_death_time(p.rdata(mass_idx));
+						;
```
chongchonghe: what is this ';'?

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,254 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+
+				if (num_particles > 1) { // Update momentum of the low mass star if there is(are) high mass star(s) in the cell
+					const int p_idx = 0;
+					auto &plow = particles[p_idx]; // NOLINT
+					plow.rdata(mass_idx + 1) = -total_momx / plow.rdata(mass_idx);
+					plow.rdata(mass_idx + 2) = -total_momy / plow.rdata(mass_idx);
+					plow.rdata(mass_idx + 3) = -total_momz / plow.rdata(mass_idx);
+				}
+
+				const double factor = (cell_mass - particle_mass) / cell_volume / state_arr(i, j, k, HydroSystem<problem_t>::density_index);
```
chongchonghe: you have a cell_mass variable, so why not use it here?

chongchonghe:
@aditivijayan I left a few more comments. One of them is about division-by-zero error and is essential to resolve. The others are just about styles.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,254 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+		// Use the common implementation with our checker and creator types
+		ParticleCreationImpl::createParticlesImpl<problem_t, ContainerType,
+							  ParticleCreationTraits<ParticleType::StochasticStellarPop>::template ParticleChecker,
+							  ParticleCreationTraits<ParticleType::StochasticStellarPop>::template ParticleCreator>(
+		    container, mass_idx, state, lev, current_time, dt, evolution_stage_index, birth_time_index);
+	}
+}; // ParticleCreationTraits<ParticleType::StochasticStellarPop>
+
+double ParticleCreationTraits<ParticleType::StochasticStellarPop>::fstar_high;	    // NOLINT
+double ParticleCreationTraits<ParticleType::StochasticStellarPop>::m_star_high_avg; // NOLINT
```
BenWibking: Global variables for this kind of data are fine, but they can't be in header files: if two different *.cpp files include this header, they will have two different versions of this variable implicitly defined. This problem is generically referred to as an "ODR violation." (This is an unfortunate "feature" of C++...)

These variables either need to be moved into a *.cpp file, or you need to add `inline` to these variables.

src/particles/particle_creation.hpp
```diff
@@ -135,6 +139,254 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+				state_arr(i, j, k, HydroSystem<problem_t>::x3Momentum_index) *= factor;
+
+				// Update internal energy to relect mass change
+				state_arr(i, j, k, HydroSystem<problem_t>::internalEnergy_index) *= factor;
+
+				// Update total energy
+				state_arr(i, j, k, HydroSystem<problem_t>::energy_index) *= factor;
+
+				// Update mass scalars
+				if (nscalars > 0) {
```
BenWibking, line 410: I think we should probably just update the mass scalars, since those are by definition scalars that can be interpreted as a partial density. That may not be the case for passive scalars in general. Can you update this so it only updates the mass scalars?

src/particles/particle_creation.hpp
BenWibking: The (very) technical details of why this is a problem are here: https://en.cppreference.com/w/cpp/language/definition

BenWibking:
@aditivijayan Can you fix the remaining warnings?

I would like to merge this as soon as possible so that I can run a simulation to show at the AMReX meeting.

aditivijayan:
> @aditivijayan Can you fix the remaining warnings?
> 
> I would like to merge this as soon as possible so that I can run a simulation to show at the AMReX meeting.

Working on it now! What about the passive scalar issue? Should we let it be for now ?

markkrumholz (dismissed):

src/particles/stellarpop_data.hpp
BenWibking: @aditivijayan Is this comment resolved?

BenWibking (commented):
It looks like there is one unresolved conversation that needs to be marked as 'resolved'.

Is there a test problem that exercises the stochastic star formation machinery? I don't see one.

chongchonghe (dismissed):

chongchonghe:
@markkrumholz  still need your approval to merge

src/particles/particle_creation.hpp
markkrumholz: Now that I read the comments, I realize I think there may a problem -- either that or I am misunderstanding and getting confused by your variable names. Your say that fstar_high is the fraction of mass in high mass stars, but then you compute this as num_high_mass_stars / total_stars, which makes me think you're setting fstar_high to be the *number fraction* rather than the *mass fraction*. These two are not the same.

The issue is a bit easier to see with a pure powerlaw functional form, so let's consider a Salpeter-like IMF of the form dn/dm ~ m^-alpha for m in some range [m_min, m_max]. In this case the total fraction of stars by *number* with mass > m_0 is given by

f_num( > m_0) = \int_{m_0}^{m_max} (dn/dm) dm / \int_{m_min}^{m_max} (dn/dm) dm = [m_max^{1-alpha} - m_0^{1-alpha}] / [m_max^{1-alpha)} - m_min^{1-alpha}],

but the total fraction of stars by *mass* with mass > m_0 is

f_mass( > m_0) = \int_{m_0}^{m_max} m (dn/dm) dm / \int_{m_min}^{m_max} m (dn/dm) dm = [m_max^{2-alpha} - m_0^{2-alpha}] / [m_max^{2-alpha)} - m_min^{2-alpha}].

See the difference -- for the fraction by mass you need an extra factor of m inside the integrals, which leads to the exponent being 2 - alpha instead if 1 - alpha. Your num_high_mass and total_stars calculations look like fractions by number, not by mass, and I therefore think they are not correct.

markkrumholz:
Just looked, but I think there is a math error, as explained in the comments. This needs to be fixed before merging.

src/particles/particle_creation.hpp
```diff
@@ -151,6 +156,300 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+	static constexpr double mass_highmass_stars = pow_alpha_m_imf_max - pow_alpha_m_star_high;
+
+	// // fstar is the fraction of number of high mass stars from the IMF
+	static constexpr double fstar_high = mass_highmass_stars / total_star_mass;
+	;
+
+	static constexpr double m_star_high_avg = m_imf_max * ((alpha - 1.0) / (alpha - 2.0)) * (1. - gcem::pow(m_star_high / m_imf_max, 2.0 - alpha)) /
+						  (1. - gcem::pow(m_star_high / m_imf_max, 1.0 - alpha)); // average mass of high mass stars
+
+	ParticleCreationTraits()
```
chongchonghe:
```suggestion
ParticleCreationTraits() = default;
```

chongchonghe:
Should pass all tests now. @markkrumholz can you review this again? If Aditi's fix to the math error is fine and all tests passed, we can merge it.

docs/markdown/star_formation.md
```diff
@@ -0,0 +1,33 @@
+# Star Formation Recipe 
+
+## Checking for Jeans Violation
+The star formation recipe is implemented via the Stochastic Stellar Population specialisation. Every cell is checked for Jeans violation and star particles are added if $\lambda_J = c_s/\sqrt{G\rho} < J \cdot dx$, where J is $0.5$. 
+
+Adding star particle to every Jeans violating cell can lead to a large number of small mass stars which can be cumbersome. To alleviate this issue, we control the star formation rate through $\epsilon_{\rm ff}$ and $\epsilon_{*}$. $\epsilon_{\rm ff}$ is the efficiency per freefall time given by $\frac{\dot{M}_{*} \cdot t_{\rm ff}}{M_{\rm cell} \cdot dt}$, while $\epsilon_{*}$ is the fraction of cell mass that gets converted into stars. 
+
+There are two equivalent ways of converting a certain mass from cell into stars. 
+
+In the first, we assume that every cell that is Jeans unstable definitely forms a star. In this scenario, $M_{*} = \epsilon_{\rm ff} \cdot dt \cdot M/t_{\rm ff}$ is the mass that gets converted into stars from a cell mass of M. This will lead to the issue mentioned earlier, that is of a large number of stars being created in a galaxy-scale simulation.
```
BenWibking: Can we delete this paragraph? Since we don't use this method, I think it is better to remove it for the sake of clarity.

docs/markdown/star_formation.md
```diff
@@ -0,0 +1,33 @@
...
+
+We circumvent this by associating a probability P to the process of spawning a star in the cell. A choice for P $=\epsilon_{\rm ff} \cdot dt/\epsilon_{*}$, which is ratio of the expected from the star formation rate in the cell over the timestep dt ($\epsilon_{\rm ff} dt/t_{\rm ff}$) to the fraction of mass available for star formation ($\epsilon_{\rm ff} M$). It should be noted that this is the probability for $\epsilon_{\rm ff}$ averaged over the timestep rather than for the instantaneous star formation rate, which should be integrated over the timestep dt. 
+
+Both the descriptions of P, from assuming a star formation rate averaged over timestep or an instantaneous star formation rate, lead to identical expression of P if $dt<t_{\rm ff}$. However if we are in the regime of $dt>t_{\rm ff}$ neither prescriptions of P offer the correct answer because they rely on conditions at the beginning of the timestep.
```
BenWibking: They rely on conditions at the *end* of the timestep, not the beginning of the timestep. I think it is better to say: they are still wrong in the unresolved $t_{ff}$ case because they are operator split.

docs/markdown/star_formation.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+## Estimating number and type of stars
+
+Once we establish that a star will form in a cell, we need to find out the stellar population looks like. In our implementation, the population comprises at least particle representing low mass star ($M<8M_{\odot}$) population and a random number of particles, each represeting a high mass star.
+
+The number of high mass stars is determined by the initial mass function, taken to be Chabrier03. The IMF is represented by a log normal for $M<M_{\odot}$ and a power law with a slope of $2.35$ beyond that. From the IMF, we can estimate the fraction of high mass stars, $f_{*,\rm{high}}$, and the average mass of high mass stars, $\langle m \rangle _{*,\rm{high}}$. The expectation value of the number of high mass stars, *num_high_mass_stars*, then is the ratio of the total mass in high mass stars and $\langle m \rangle _{*,\rm{high}}$.
```
BenWibking: @markkrumholz @aditivijayan $f_{*,\rm{high}}$ should be the mass in high mass stars, not the number, right? I think this paragraph might be incorrect for the reason discussed on the telecon yesterday. If not, I think the current wording is confusing.

docs/markdown/star_formation.md
```diff
@@ -0,0 +1,33 @@
+# Star Formation Recipe 
+
+## Checking for Jeans Violation
+The star formation recipe is implemented via the Stochastic Stellar Population specialisation. Every cell is checked for Jeans violation and star particles are added if $\lambda_J = c_s/\sqrt{G\rho} < J \cdot dx$, where J is $0.5$. 
+
+Adding star particle to every Jeans violating cell can lead to a large number of small mass stars which can be cumbersome. To alleviate this issue, we control the star formation rate through $\epsilon_{\rm ff}$ and $\epsilon_{*}$. $\epsilon_{\rm ff}$ is the efficiency per freefall time given by $\frac{\dot{M}_{*} \cdot t_{\rm ff}}{M_{\rm cell} \cdot dt}$, while $\epsilon_{*}$ is the fraction of cell mass that gets converted into stars.
```
BenWibking: Can you change this to "Adding a star particle"?

docs/markdown/star_formation.md
```diff
@@ -0,0 +1,33 @@
+# Star Formation Recipe 
+
+## Checking for Jeans Violation
+The star formation recipe is implemented via the Stochastic Stellar Population specialisation. Every cell is checked for Jeans violation and star particles are added if $\lambda_J = c_s/\sqrt{G\rho} < J \cdot dx$, where J is $0.5$. 
+
+Adding star particle to every Jeans violating cell can lead to a large number of small mass stars which can be cumbersome. To alleviate this issue, we control the star formation rate through $\epsilon_{\rm ff}$ and $\epsilon_{*}$. $\epsilon_{\rm ff}$ is the efficiency per freefall time given by $\frac{\dot{M}_{*} \cdot t_{\rm ff}}{M_{\rm cell} \cdot dt}$, while $\epsilon_{*}$ is the fraction of cell mass that gets converted into stars. 
+
+There are two equivalent ways of converting a certain mass from cell into stars.
```
BenWibking: Can you change this to "from the cell into stars"?

docs/markdown/star_formation.md
```diff
@@ -0,0 +1,33 @@
...
+
+Both the descriptions of P, from assuming a star formation rate averaged over timestep or an instantaneous star formation rate, lead to identical expression of P if $dt<t_{\rm ff}$. However if we are in the regime of $dt>t_{\rm ff}$ neither prescriptions of P offer the correct answer because they rely on conditions at the beginning of the timestep. 
+
+
+## Estimating number and type of stars
+
+Once we establish that a star will form in a cell, we need to find out the stellar population looks like. In our implementation, the population comprises at least particle representing low mass star ($M<8M_{\odot}$) population and a random number of particles, each represeting a high mass star.
```
BenWibking: Can you change this to "comprises at least a particle representing the low mass star ($M<8M_{\odot}$) population and a random number of particles, each representing a high mass star."?

docs/markdown/star_formation.md
```diff
@@ -0,0 +1,33 @@
...
+
+The number of high mass stars is determined by the initial mass function, taken to be Chabrier03. The IMF is represented by a log normal for $M<M_{\odot}$ and a power law with a slope of $2.35$ beyond that. From the IMF, we can estimate the fraction of high mass stars, $f_{*,\rm{high}}$, and the average mass of high mass stars, $\langle m \rangle _{*,\rm{high}}$. The expectation value of the number of high mass stars, *num_high_mass_stars*, then is the ratio of the total mass in high mass stars and $\langle m \rangle _{*,\rm{high}}$. 
+
+The total number of star particles to be spwaned in a cell is $1+$ Poission distribution with an expectation value of *num_high_mass_stars*. A fraction $1-f_{*,\rm{high}}$ goes into the particle representing low mass star while the mass of the high mass star particles is randomly drawn from the high mass part of the IMF.
```
BenWibking: Can you change this to "The total number of star particles to be spawned" and "A fraction $1-f_{*,\rm{high}}$ goes into the particle representing low mass stars"?

BenWibking:
> Should pass all tests now. @markkrumholz can you review this again? If Aditi's fix to the math error is fine and all tests passed, we can merge it.

There are a few issues with the documentation that should be addressed before merging.

src/particles/particle_creation.hpp
```diff
@@ -151,6 +157,298 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+	// Constants for the Chabrier IMF
+	static constexpr amrex::Real m_star_high = 8.0 * C::M_solar; // all stars above this mass are considered high mass stars
+	static constexpr amrex::Real m_imf_min = 0.08 * C::M_solar;  // lower limit of the IMF
+	static constexpr amrex::Real m_imf_max = 120.0 * C::M_solar; // high mass limit of the IMF
+	static constexpr amrex::Real m_imf_break = 1.0 * C::M_solar; // IMF is lognormal below this mass and powerlaw above
+	static constexpr amrex::Real imf_disp = 0.55;		     // dispersion of the lognormal IMF
+	static constexpr amrex::Real imf_mu = 32.599;		     //=log10(0.2 * C::M_solar), mean of the lognormal IMF, avoid compiler error
+	static constexpr amrex::Real alpha = 2.35;		     // slope of the powerlaw
+
+	// Write out expression because compiler error for nested gcem
```
markkrumholz, line 178: I think it is super-confusing to have half the calculation done here this way, which is sort of halfway between writing out the answer and doing the calculation all on the fly with constexpr's. I suggest just writing out the values of fstar_high and m_star_high_avg explicitly, with comments giving the algebraic expressions from which they are derived? It would make the code shorter and much more readable.

src/particles/particle_creation.hpp
```diff
@@ -151,6 +157,298 @@ template <ParticleType particleType> struct ParticleCreationTraits {
...
+	static constexpr double mass_highmass_stars = pow_alpha_m_imf_max - pow_alpha_m_star_high;
+
+	// // fstar is the fraction of number of high mass stars from the IMF
+	static constexpr double fstar_high = mass_highmass_stars / total_star_mass;
+	;
+
+	static constexpr double m_star_high_avg = m_imf_max * ((alpha - 1.0) / (alpha - 2.0)) * (1. - gcem::pow(m_star_high / m_imf_max, 2.0 - alpha)) /
+						  (1. - gcem::pow(m_star_high / m_imf_max, 1.0 - alpha)); // average mass of high mass stars
+
+	ParticleCreationTraits() = default;
```
markkrumholz, line 211: I suggest deleting the commented-out incorrect code to avoid confusion.

markkrumholz (commented):
@aditivijayan, I left some comments on improving code readability, and it looks like there is an unresolved comment from @BenWibking on the documentation. Can you address these and then I can approve?

aditivijayan:
> @aditivijayan, I left some comments on improving code readability, and it looks like there is an unresolved comment from @BenWibking on the documentation. Can you address these and then I can approve?

I have replied to the comments made by you and @BenWibking .

markkrumholz:
> > @aditivijayan, I left some comments on improving code readability, and it looks like there is an unresolved comment from @BenWibking on the documentation. Can you address these and then I can approve?
> 
> I have replied to the comments made by you and @BenWibking .

Where? I don't see any responses, and the code looks the same.

chongchonghe:
It seems like you changed one line of code but not everything. I believe what Mark meant is you can update lines 175-186 to use `gcem::pow` and `gcem::log` and do the calculation inline.

markkrumholz:
> It seems like you changed one line of code but not everything. I believe what Mark meant is you can update lines 175-186 to use `gcem::pow` and `gcem::log` and do the calculation inline.

Yes -- I was basically proposing to just delete all of the gcem stuff entirely, and just hand-enter the values of fstar_high and mstar_high_avg, because the math written out here is super-complicated and hard-to-follow, which makes the code hard to read and maintain. It would be much easier to read if you simply gave the numerical values of these quantities directly, and gave their definitions in comments so that people can see how they were calculated. If we could easily compute them inline using constexpr's from the basic quantities that define the IMF that would be better, but it seems that we can't easily, and the current approach of hand-entering a bunch of values and the using those intermediate products to do an inline calculation is the worst of both worlds: not fully automatic when it comes to changing the IMF parameters (since you'd then still have to update those intermediate quantities) but also not easy to read.

aditivijayan:
> > > @aditivijayan, I left some comments on improving code readability, and it looks like there is an unresolved comment from @BenWibking on the documentation. Can you address these and then I can approve?
> > 
> > 
> > I have replied to the comments made by you and @BenWibking .
> 
> Where? I don't see any responses, and the code looks the same.

Screenshot 2025-05-01 at 3 27 39 pm

Is this not visible?

aditivijayan:
> > It seems like you changed one line of code but not everything. I believe what Mark meant is you can update lines 175-186 to use `gcem::pow` and `gcem::log` and do the calculation inline.
> 
> Yes -- I was basically proposing to just delete all of the gcem stuff entirely, and just hand-enter the values of fstar_high and mstar_high_avg, because the math written out here is super-complicated and hard-to-follow, which makes the code hard to read and maintain. It would be much easier to read if you simply gave the numerical values of these quantities directly, and gave their definitions in comments so that people can see how they were calculated. If we could easily compute them inline using constexpr's from the basic quantities that define the IMF that would be better, but it seems that we can't easily, and the current approach of hand-entering a bunch of values and the using those intermediate products to do an inline calculation is the worst of both worlds: not fully automatic when it comes to changing the IMF parameters (since you'd then still have to update those intermediate quantities) but also not easy to read.

I understood @markkrumholz 's first comment and these comments as well. Since for some reason my reply isn't visible I will explain why I chose to write it this way - 


If a user wants to change the params of the IMF, they would have recalculate everything which means writing a script first including complicated functions of erf for f_star_high.

So, I made a very conscious choice to not calculate every quantity because I personally feel making an error in copying out the complicated function for f_star_high is much more likely to occur and trickier to debug than estimating log and powers of masses -- keeping track of +/- and factors of 2 is my personal nightmare. At least we have at least some intuitive idea what the log of mass would be, 33-ish, rather than erf of (weird looking quantity).

I am strongly in favour of keeping this as is. Yes, the code is clunky and not aesthetically appealing. I can restructure to make it look better if that helps readability.


True --- fstar_high has complicated mathematical expression but I personally prefer seeing the expression coded out and explaining it further in a paper or the docs rather than having it hard coded in. In this case, if I change the break mass or the max or min mass of IMF i can be sure that I am estimating f_star_high correctly without having to redo the calculation somewhere else. Again, in my opinion, plugging in the values of log of mass and power of masses is easier than estimating erf of log(mass) - mu etc.

markkrumholz:
> > > It seems like you changed one line of code but not everything. I believe what Mark meant is you can update lines 175-186 to use `gcem::pow` and `gcem::log` and do the calculation inline.
> > 
> > 
> > Yes -- I was basically proposing to just delete all of the gcem stuff entirely, and just hand-enter the values of fstar_high and mstar_high_avg, because the math written out here is super-complicated and hard-to-follow, which makes the code hard to read and maintain. It would be much easier to read if you simply gave the numerical values of these quantities directly, and gave their definitions in comments so that people can see how they were calculated. If we could easily compute them inline using constexpr's from the basic quantities that define the IMF that would be better, but it seems that we can't easily, and the current approach of hand-entering a bunch of values and the using those intermediate products to do an inline calculation is the worst of both worlds: not fully automatic when it comes to changing the IMF parameters (since you'd then still have to update those intermediate quantities) but also not easy to read.
> 
> I understood @markkrumholz 's first comment and these comments as well. Since for some reason my reply isn't visible I will explain why I chose to write it this way -
> 
> If a user wants to change the params of the IMF, they would have recalculate everything which means writing a script first including complicated functions of erf for f_star_high.
> 
> So, I made a very conscious choice to not calculate every quantity because I personally feel making an error in copying out the complicated function for f_star_high is much more likely to occur and trickier to debug than estimating log and powers of masses -- keeping track of +/- and factors of 2 is my personal nightmare. At least we have at least some intuitive idea what the log of mass would be, 33-ish, rather than erf of (weird looking quantity).
> 
> I am strongly in favour of keeping this as is. Yes, the code is clunky and not aesthetically appealing. I can restructure to make it look better if that helps readability.
> 
> True --- fstar_high has complicated mathematical expression but I personally prefer seeing the expression coded out and explaining it further in a paper or the docs rather than having it hard coded in. In this case, if I change the break mass or the max or min mass of IMF i can be sure that I am estimating f_star_high correctly without having to redo the calculation somewhere else. Again, in my opinion, plugging in the values of log of mass and power of masses is easier than estimating erf of log(mass) - mu etc.

Sorry, I indeed did not see your response before -- don't know why it's not showing up for me, and my apologies.

In any event: ok, I don't feel super-strongly about this. I guess I didn't think it would be hard to replicate because I would never do that sort of calculation by hand, I'd do it in mathematica, precisely because I agree that keeping track fo all th factors of 2 and erf's and stuff like that is a pain. But since I don't feel very strongly, I am willing to defer to your preference, and thus I'll consider this conversation resolved.

Can you also send along your responses to Ben's comments, since for some reason I can't see them either. If you have responded to them I can look at those responses and then hopefully merge this PR.

docs/markdown/star_formation.md
markkrumholz: I think it's fine to leave it, at least for now.

markkrumholz (dismissed):

markkrumholz (approved):

BenWibking:
I also don't see the replies to my comments. I think Github PRs just kind of malfunction when there are too many comments.

In any case, I see the updated documentation text, and it looks good to me.

PR #985, opened 2025-04-09
Use FaceLinear AMR interpolation for particle accelerations

### LINKED ISSUE #984: not enough ghost cells for particle accelerations?
**Describe the bug**
We don't have enough ghost cells to properly compute particle accelerations. We currently one have 1:
https://github.com/quokka-astro/quokka/blob/a30891d669fb51afa61194afaa825ce88d37767f/src/simulation.hpp#L1291

Because particles are redistributed, then drifted, then kicked, particles can move into the first layer of ghost cells and receive kicks there. That means we need at least 2 ghost cells for the acceleration multifab.

**To Reproduce**
1. Compile the BinaryOrbitCIC problem
2. Run the BinaryOrbitCIC problem with refinement only covering one particle.
3. Run the simulation and see what happens as the particle crosses the coarse-fine boundary.

**Additional context**
Nyx uses 3 ghost cells for the cell-centered acceleration multifab: https://github.com/AMReX-Astro/Nyx/blob/51a5f8e55ef70ac488b13a376629e9027bc2e5d6/Source/TimeStep/Nyx_advance_particles.cpp#L62.

See also: https://github.com/AMReX-Astro/Nyx/blob/development/Source/Gravity/Gravity.cpp#L814

BenWibking:
### Description
This extends the ghost cells in the particle acceleration MultiFab, and uses `amrex::FaceLinear` AMR interpolation (linear in the normal component, constant for the transverse components) for the face-centered accelerations, which are then averaged to cell centers before kicking particles.

We also extend the ghost cells in the `rhs` MultiFab when solving the Poisson equation (in order to allow for particle drift since the last redistribute).

### Related issues
Fixes https://github.com/quokka-astro/quokka/issues/984.

### Checklist


src/simulation.hpp
```diff
@@ -1191,12 +1191,15 @@ template <typename problem_t> void AMRSimulation<problem_t>::calculateGpotAllLev
 		phi.resize(finest_level + 1);
 		// solve Poisson equation with open b.c. using the method of James (1977)
 		amrex::Vector<amrex::MultiFab> rhs(finest_level + 1);
-		const int nghost = 1;
-		const int ncomp = 1;
+		constexpr int nghost_phi = 1;
+		constexpr int nghost_deposit = 1; // CIC deposition requires 1 ghost cell
+		constexpr int nghost_drift = 1;	  // particle can drift up to 1 cell
+		constexpr int nghost_rhs = nghost_deposit + nghost_drift;
```
BenWibking, line 1202: We also need 2 ghost cells to deposit mass, since particles can drift up to 1 cell + CIC deposition requires 1 ghost cell.

chongchonghe:
The error comes from `particleRegister_.depositMass`. The error message seems to suggest that it's trying to access the ghost cells (boundaries) of the whole domain from a Fab that does not touch that boundary.

BenWibking:
> The error comes from `particleRegister_.depositMass`. The error message seems to suggest that it's trying to access the ghost cells (boundaries) of the whole domain from a Fab that does not touch that boundary.

Right, I don't understand how this is possible. I will try to debug it tomorrow.

chongchonghe:
Adding a `particleRegister_.redistribute(0);` before `ellipticSolveAllLevels(dt_[0]);` avoids this error, but the following assertion failed:

```cpp
AMREX_ALWAYS_ASSERT(!phi[lev].contains_nan()); // NOTE: this fails when multiple levels are fully refined
```

This could be helpful to identify the problem.

BenWibking:
The particle has `NaN` position values. Most likely this is because `phi` is NaN, but this should be caught inside the poisson solve...

```
(const amrex::Particle<4> &) 0x00006000014239c0: {
  amrex::ParticleBase<double, 4, 0> = {
    m_pos = ([0] = NaN, [1] = NaN, [2] = NaN)
    m_rdata = ([0] = 1.9999999999999999E+34, [1] = NaN, [2] = NaN, [3] = NaN)
    m_idcpu = 9223372036888330240
  }
}
```

BenWibking:
@chongchonghe I think I fixed it. Can you review this?

BenWibking:
Orbit plot for BinaryOrbitAMR.in:

orbit

BenWibking:
@markkrumholz Can you also review this? (This is required to merge since you are the code owner for some of the files.)

BenWibking:
After asking Weiqun about FaceLinear, I think we actually want to use CellQuadratic on the potential directly. So I'm putting this back as a work-in-progress.

BenWibking:
FLASH does quadratic interpolation of the potential (see [documentation](https://flash.rochester.edu/site/flashcode/user_support/flash4_ug_4p71/node128.html)). The lowest-order self-force error disappears with quadratic interpolation.

However, so we can run the scaling tests for the galaxy simulations now, this will be tested in a future PR.

chongchonghe:
@BenWibking  This failure on avatar GPU is probably also from my previous mistake and should be fixed by #994 . Sorry about that.

chongchonghe:
Is this PR intending to fix the erroneous orbit at coarse-fine boundary? Unfortrunately, the orbit at coarse-fine boundary is still not correct. The particles get bounced back (see images below). 

However, I believe the FaceLinear interpolation should give better results, and this PR also fix the error related to particle dirft and redistribute.

chongchonghe (approved):

chongchonghe:
You can use this PR to run the binary_orbit test with half of the domain refined:  https://github.com/quokka-astro/quokka/tree/chong/particles/binary-orbit-amr-test-half-domain 

I believe we do need to implement the correction technique proposed in the paper you showed during the previous Quokka meeting in order to fix this error. I'm currently focusing on writing the SN and sink particle module, so I'm afraid I don't have time to work on this in the next few weeks.

BenWibking:
@markkrumholz This PR is also ready to be reviewed.

markkrumholz (approved):

PR #988, opened 2025-04-10
Add an example of manually setting particles' integer components

chongchonghe:
### Description

`InitFromAsciiFile` does not support integer components (see [here](https://github.com/AMReX-Codes/amrex/blob/e24e4a8c4df84ed71f961efe7637f86ba1c2ce62/Src/Particle/AMReX_ParticleInit.H#L24)). In this PR, I added an example showing how to initialize the integer components of particles in the `createInitialTestParticles` function in `gravity_3d.cpp`. We loop over all Test particles and set the integer component manually. This is helpful, for example, when the user want to initialize StachasticStellerPop particles from file and set all of them to `LowMassStar`, or assign their StellarEvolutionStage according to their mass and age. 

More changes:
- Defined `createInitialTestParticles`
- Added `StochasticStellarPopParticleDeathTimeIdx` and `TestParticleDeathTimeIdx`

I also created an [issue](https://github.com/AMReX-Codes/amrex/issues/4415) on AMReX repo to request adding the functionality to initialize particles with integer components from file. 

### Checklist


BenWibking:
Maybe we could open a PR to modify amrex::InitFromAsciiFile to support integer components? I think that would be the ideal solution.

chongchonghe:
Actually, we don't need this new function `InitSetPhyParticles`, because we can simply do that inside `createInitialTestParticles` (updated).

> Maybe we could open a PR to modify amrex::InitFromAsciiFile to support integer components? I think that would be the ideal solution. 

I created a issue for that in AMReX repo. I'll work on that when I got time. Before that, this PR is a quick workaround. Aditi's simulation night need this.

BenWibking (approved):

PR #990, opened 2025-04-11
Print the cell that sets the timestep on each level

BenWibking:
### Description
When `amr.v = 1`, this will print the cell that sets the candidate timestep for hydro (and also for particles, if applicable) for each level, as well as which level sets the overall coarse timestep.

Example output:
```
Coarse STEP 74 at t = 638990.8046 (1.277981609%) starts ...
...[level 0] estimated hydro timestep: 1.009208e+04
...[level 0]    hydro timestep limited at cell [16, 17, 15] with signal speed = 4.644731e+07
...[level 0]    cell density = 9.641857e-22, |v| = 3.344731e+07, cs = nan
...[level 0] estimated particle timestep: 7.274247e+04
...[level 0]    max particle velocity: 1.073994e+07
...[level 0]    particle timestep limited at position [1.724770e+12, -2.473514e+12, -9.395845e+03]
...coarse timestep set by level 0
```

**Note that for Cray MPICH, standard output is not ordered with respect to different ranks, so the cell density, velocity and sound speed may be printed later in the output.** (We could copy the cell to rank 0 and then print, but that will incur a performance cost.)

This PR also gets rid of old timestepping code that was used when the hydro update was insufficiently stable to take timesteps longer than the cooling time for some problems. Since that is no longer necessary, that code is removed.

### Related issues
Implementation based on https://github.com/AMReX-Astro/Castro/pull/2273.

### Checklist

### DIFF
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -182,7 +182,7 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 
 	void checkHydroStates(amrex::MultiFab &mf, char const *file, int line);
 	void computeMaxSignalLocal(int level) override;
-	auto computeExtraPhysicsTimestep(int lev) -> amrex::Real override;
+	auto computeExtraPhysicsTimestep(int lev) -> amrex::ValLocPair<amrex::Real, amrex::IntVect> override;
 	void preCalculateInitialConditions() override;
 	void setInitialConditionsOnGrid(quokka::grid const &grid_elem) override;
 	void setInitialConditionsOnGridFaceVars(quokka::grid const &grid_elem) override;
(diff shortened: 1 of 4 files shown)


BenWibking:
Need to debug: `Gravity3D` fails on avatargpu. I cannot reproduce this in Debug or Release mode on my laptop.
```
Initializing AMReX (25.04)...
Initializing CUDA...
CUDA initialized with 1 device.
AMReX (25.04) initialized
Redistributing from processor 0 to 0
Total number of particles: 2
Redistributing from processor 0 to 0
Total number of particles: 2
Segfault
```

BenWibking:
It turns out that Cray MPICH buffers all output from ranks > 0 until MPI_Finalize() is called. So we always need to print from rank 0.

chongchonghe:
Gravity3D test on avatargpu still failed. I'll take a look.

chongchonghe:
> Gravity3D test on avatargpu still failed. I'll take a look.

That was my bad and I fixed it. I'll create a PR for this fix.

chongchonghe:
@BenWibking avatargpu error should be fixed by #994

chongchonghe:
BTW, do you intend to update amrex in this PR? @BenWibking

BenWibking:
> BTW, do you intend to update amrex in this PR? @BenWibking

Yes, it is necessary in order to do `amrex::ParallelAllReduce::Max` on a `amrex::ValLocPair<amrex::Real, amrex::RealVect>`.

chongchonghe:
Looks good to me. The only minor comment is that your example output has `cs = nan`, and I wonder if that is expected.

BenWibking:
> Looks good to me. The only minor comment is that your example output has `cs = nan`, and I wonder if that is expected.

It's because quokka::EOS<problem_t> does not handle isothermal EOS consistently with other EOSes. So for isothermal EOS, it will always say 'nan'. We should fix that, but I would prefer to do that in a different PR.

chongchonghe (approved):

BenWibking:
@markkrumholz Can you review this PR?

markkrumholz (approved):

PR #995, opened 2025-04-15
A new SN feedback model

chongchonghe:
### Description

In this PR I implement a new SNe feedback model that smoothly transitions between thermal feedback, kinetic feedback, and momentum feedback, depending on how well the cooling/shell formation phase of the SNR evolution is resolved. An innovation of this scheme is that we do a two stage deposition scheme to account for SNR overlapping and get rid of depedence on the order that SNe events occur. 

I have tested three models:
1. Model 0: Pure thermal deposition (for comparison only, not for production)
2. Model 1: Thermal energy deposition in the resolved regime, thermal + kinetic in the intermediate regime, and terminal momentum in the unresolved regime.
3. Model 2:  Kinetic energy deposition in the resolved regime, thermal + kinetic in the intermediate regime, and terminal momentum in the unresolved regime. 

At a SN event, we distribute mass, momentum, and energy uniformly among the cells that intercect with s sphere of radius `3 dx` centered at the center of the cell that the SN is located in. No background smoothing is conducted before deposition.

The results looks great! See figure below. Both model 1 and 2 produces 'correct' and almost identical results. The only difference is the maximum temperature (defined as the maximum temperature in any cell throughout the simulation) in model 2 is slightly lower than that in model 1. In model 2, right after SN explosion, the gas is cool but moves fast, and the shock wave immediately heat the gas to comparable temperature to that in the thermal model. After some experiments, I am convinced that the thermal model and kinetic model (in the resolved regime) has no difference in accuracy or performance related to timestep. This is because, imagine a star explodes into vacuum, q simple calculation will show that the sound speed of the hot SNR in the pure thermal model is about $1/\sqrt{3}$ times the bulk velocity of the cool SNR in the pure kinetic model (microscopic thermal motion has three degrees of freedom while kinetic motion has one). The thermal model turns out to have slightly longer timestep overall. In the TIGRESS model they advocate thermal model because they have over cooling issue, but we don't have this issue at all (thanks to Ben's awesome Grackle cooling module which has a robust ODE solver). 

compare

> Reproducing Figure 6 of Kim & Ostriker 2017. The spacial resolution and SN ejecta mass/energy/momentum in our simulations are the same as theirs.

### Checklist


chongchonghe:
The next step is to test overlapping and non-uniform background. I would need to set up a non-uniform initial condition to mimick realistic ISM.  @BenWibking Do you have any idea how to quickly produce such an I.C.?

BenWibking:
> The next step is to test overlapping and non-uniform background. I would need to set up a non-uniform initial condition to mimick realistic ISM. @BenWibking Do you have any idea how to quickly produce such an I.C.?

I would just run it until you have multiple SN going off in a periodic box, like the RandomBlast problem. Then the background will be very non-uniform :D

BenWibking:
Regarding model 2, for supernovae that stay confined to the disk, I think it's a fine model.  However, I am slightly concerned that this model might artificially inhibit the buoyancy of superbubbles. Buoyancy-driven winds can be significant for some galaxies.

So I think the best test to compare would be of models 1 and 2 in a stratified box, where the Brunt-Vaisala timescale is less than or comparable to the cooling timescale.

@markkrumholz Do you have any thoughts on this particular issue?

chongchonghe:
I've finished a test of our model 2 in the draft, or `SNScheme::SN_thermal_or_thermal_momentum` in the code, in a more realistic ISM environment. The result is satisfactory: the terminal momentum agrees well with the classical fitting formula at the given density. 

The initial condition was created from a uniform background with 27 SN explosions at t=0 and ran till t = 80 Myr to come to equilibrium. SNR overlapping did not cause unexpected behaviour, nor cause the code to crash, thanks to the limiter in our scheme. I'm still testing the accuracy/convergence of our method when SNR overlaps. 

CleanShot 2025-04-18 at 02 35 27@2x

CleanShot 2025-04-18 at 02 36 01@2x

BenWibking:
That looks nice!

The only possible improvement that I can think of is to do the deposition using a spherical kernel, such as one of the Wendland kernels (https://link.springer.com/article/10.1007/BF02123482).

There is an implementation of Wendland's $C^2$ kernel already in Quokka here:
https://github.com/quokka-astro/quokka/blob/371117febb8446c280efe5b18060fc64e981c0b4/src/math/quadrature.hpp#L13

chongchonghe:
> Regarding model 2, for supernovae that stay confined to the disk, I think it's a fine model. However, I am slightly concerned that this model might artificially inhibit the buoyancy of superbubbles. Buoyancy-driven winds can be significant for some galaxies.
> 
> So I think the best test to compare would be of models 1 and 2 in a stratified box, where the Brunt-Vaisala timescale is less than or comparable to the cooling timescale.
> 
> @markkrumholz Do you have any thoughts on this particular issue?

The model 1 and model 2 in my draft are basically the same model, the only difference being model 2 does a smooth transition from unresolved to resolved phase. Therefore, model 2 is always (slightly) better than model 1. I also observe that in my tests. I don't get how that model could "artificially inhibit the buoyancy of superbubbles".

BenWibking:
> > Regarding model 2, for supernovae that stay confined to the disk, I think it's a fine model. However, I am slightly concerned that this model might artificially inhibit the buoyancy of superbubbles. Buoyancy-driven winds can be significant for some galaxies.
> > So I think the best test to compare would be of models 1 and 2 in a stratified box, where the Brunt-Vaisala timescale is less than or comparable to the cooling timescale.
> > @markkrumholz Do you have any thoughts on this particular issue?
> 
> The model 1 and model 2 in my draft are basically the same model, the only difference being model 2 does a smooth transition from unresolved to resolved phase. Therefore, model 2 is always (slightly) better than model 1. I also observe that in my tests. I don't get how that model could "artificially inhibit the buoyancy of superbubbles".

Your description of the models was:

> Model 1: Thermal energy deposition in the resolved regime, thermal + kinetic in the intermediate regime, and terminal momentum in the unresolved regime.
> Model 2: Kinetic energy deposition in the resolved regime, thermal + kinetic in the intermediate regime, and terminal momentum in the unresolved regime.

So if you only inject kinetic energy in the resolved regime, then you are not directly increasing the entropy of the gas. The entropy of the gas will increase after it shocks, but it may be delayed compared to when the ISM entropy should increase, so any high-entropy bubbles may not rise as far as they should.

BenWibking:
Your plot of the maximum temperature with this model suggests there could be an effect like this, since the maximum temperature is less for Model 2 compared to Model 1 in the resolved phase.

chongchonghe:
> Your plot of the maximum temperature with this model suggests there could be an effect like this, since the maximum temperature is less for Model 2 compared to Model 1 in the resolved phase.

Oh, I was referring to a different 'model 1' and 'model 2'. Never mind. I will give a thorough reply tomorrow.

chongchonghe:
I am now convinced that model 1, "Thermal energy deposition in the resolved regime, ...", is as good as model 2, as long as we have a reliable cooling module like the one in Quokka. There is no difference in timesteps between the two models, as I argued above. The only advantage of model 2 occurs when SNe are ejected into vacuum and the temperature goes to 10^9 K; an inaccurate cooling module (e.g. a single Backward Euler) could lead to overcooling or an inaccurate cooling curve. Besides, while both models have similar timesteps, model 1 places more pressure on the cooling step, which could negatively impact performance if it requires many iterations. 

model 1 and model2 are only different when RM < 0.027, which occurs at n <~ 0.1 cm-3. The realistic ISM test I did above has a background density of ~0.5 cm-3, so these two models make no difference. Considering this, I reckon model 2 does nothing to prevent superbubbles, because pure kinetic deposition is only activated when the shell formation is well resolved (dx < 0.1 R_sf. Note that we use 3 cells and 0.3^3 = 0.027), and in this regime, the ambient gas gets immediately shock-heated and the subsequent evolution is very similar between the two models. See temporal profiles below and pay attention to the last panel -- the max temperature v.s. time in log-log scale.

</img> 
</img>

> Fig1: Temporal profile of the Sedov-Taylor phase of a SN event at a background density of n = 0.1, using model 1. Fig2: same as Fig1 but with model 2.

BenWibking:
Ok, the time it takes to reach T_max is much shorter than the buoyancy timescale, so it's fine.

BenWibking:
Actually, just to be sure, can you plot the [entropy function](https://en.wikipedia.org/wiki/Entropy_(astrophysics)), defined as $K = kT / n^{2/3}$, as a function of time?

chongchonghe:
> That looks nice!
> 
> The only possible improvement that I can think of is to do the deposition using a spherical kernel, such as one of the Wendland kernels (https://link.springer.com/article/10.1007/BF02123482).
> 
> There is an implementation of Wendland's C 2 kernel already in Quokka here:

I think the key here is not a kernel function like Wendland's. TIGRESS use a uniform kernel for energy deposition and $\propto r^2$ for momentum deposition. Note it's positively correlated with $r$, unlike any kernel function in SPH. This makes sense because the momentum is concentrated on the shell.

What's more important is to account for partial occupation, similar to the CIC deposition algorithm, but with a uniform spherical kernel with a radius of 3 dx (k = 1 if r < 3 dx else 0). There is complicated geometry on computing the overlapping volume when a cube lives at the boundary of a sphere. 

What I need is a formula that gives the overlapping volume between a sphere with radius r at position (x,y,z) and a unit cube at the origin, i.e. $x,y,z \in [-0.5, 0.5]$. I'm sure there are approximate formulas for that. @BenWibking Do you know any?

chongchonghe:
> Actually, just to be sure, can you plot the [entropy function](https://en.wikipedia.org/wiki/Entropy_(astrophysics)), defined as K = k T / n 2 / 3 , as a function of time?

What do you mean "as a function of time"? That requires a single number at a snapshot, but how do you integrate 'Entropy' over volume? Entropy is not a volume density of anything. Perhaps it's because I'm not familiar with Entropy in astrophysics.

chongchonghe:
> What I need is a formula that gives the overlapping volume between a sphere with radius r at position (x,y,z) and a unit cube at the origin, i.e. $x,y,z \in [-0.5, 0.5]$.

I am thinking of a possible approximation in my mind. We could do this: 

I only care about the case where r > 1. In our algorithm, r = 3.
1. Determine the point within the cube that is closest to (x,y,z) and the point that is furthest to (x,y,z). Their distances to (x,y,z) are `d_min` and `d_max`.
2. `V = 0` if `d_min > 3`. `V = 1` if `d_max < 3`. When `d_min < 3 < d_max`, approximate `V` with a smooth function that goes from 0 (when `d_min = 3`) to 1 (when `d_max = 3`). One example is `V = (3 - d_min) / (d_max - d_min)`. 

How does this simple algorithm sound?

BenWibking:
There is no closed form expression for this, due to the many special cases involved. I think it should be possible to compute the volume integral numerically with marginal cost. I have started a PR in Microphysics to do Gaussian quadrature: https://github.com/AMReX-Astro/Microphysics/pull/1783

BenWibking:
I need to think about this in more detail, but I think this approach might be correct:
```
    codex
    I added volume_overlap.cpp implementing the analytic formula for volume overlap of an axis-aligned cube and sphere using nested
    adaptive quadrature:

        * `ChordLengthZ`: Computes the length of chord in z direction where circle intersects cube z-limit for given y.
        * `IntegrandY`: Integrates chord length over y clipped to circle and cube bounds.
        * `IntegrandX`: Integrates previous result over x clipped by sphere and cube, where radius in yz plane depends on x.
        * `volume_overlap_adaptive`: Top-level function that clips input cube bounds to sphere limits and calls adaptive quadrature on
    `IntegrandX`.

    All integration is done adaptively using the provided adaptive_quadrature from util/quadrature.H assuming it takes a functor,
    limits, and tolerance.

    This is the closest efficient analytic-based calculation feasible. Let me know if you want me to help integrate or test it within
    your codebase.
user
Print the code without the patch formatting.

    codex
    // volume_overlap.cpp
        // Computes the volume overlap between axis-aligned cube and sphere
        // using analytic decomposition and adaptive quadrature from util/quadrature.H

        #include <cmath>
        #include "util/quadrature.H"

        namespace geometry {

        // Functor for chord length in z direction for fixed x and y
        struct ChordLengthZ {
            double z_min_, z_max_;
            double c_z_;
            double radius_;
            double y_, c_y_;

            ChordLengthZ(double z_min, double z_max, double c_z, double radius, double y, double c_y)
                : z_min_(z_min), z_max_(z_max), c_z_(c_z), radius_(radius), y_(y), c_y_(c_y) {}

            double operator()(double /*unused*/) const {
                double dy = y_ - c_y_;
                double r_y = radius_;
                double val = r_y * r_y - dy * dy;
                if (val <= 0.0) return 0.0; // no chord because outside circle
                double sqrt_val = std::sqrt(val);
                double chord_low = c_z_ - sqrt_val;
                double chord_high = c_z_ + sqrt_val;
                double low = std::max(chord_low, z_min_);
                double high = std::min(chord_high, z_max_);
                return (high > low) ? (high - low) : 0.0;
            }
        };

        // Functor for integration over y at fixed x
        struct IntegrandY {
            double y_min_, y_max_;
            double z_min_, z_max_;
            double c_y_, c_z_;
            double radius_;

            IntegrandY(double y_min, double y_max,
                       double z_min, double z_max,
                       double c_y, double c_z, double radius)
                : y_min_(y_min), y_max_(y_max), z_min_(z_min), z_max_(z_max),
                  c_y_(c_y), c_z_(c_z), radius_(radius) {}

            double operator()(double y) const {
                ChordLengthZ chord(z_min_, z_max_, c_z_, radius_, y, c_y_);
                return chord(0.0);
            }
        };

        // Functor for integration over x
        struct IntegrandX {
            double x_min_, x_max_;
            double y_min_, y_max_;
            double z_min_, z_max_;
            double c_x_, c_y_, c_z_;
            double radius_;
            double tol_;

            IntegrandX(double x_min, double x_max,
                       double y_min, double y_max,
                       double z_min, double z_max,
                       double c_x, double c_y, double c_z,
                       double radius, double tol)
                : x_min_(x_min), x_max_(x_max),
                  y_min_(y_min), y_max_(y_max),
                  z_min_(z_min), z_max_(z_max),
                  c_x_(c_x), c_y_(c_y), c_z_(c_z),
                  radius_(radius), tol_(tol) {}

            double operator()(double x) const {
                double dx = x - c_x_;
                if (dx*dx > radius_*radius_) {
                    return 0.0; // Outside sphere radius, chord area zero
                }
                double r_yz = std::sqrt(radius_ * radius_ - dx * dx);
                // effective integration bounds in y clipped by circle and cube
                double y_low = std::max(y_min_, c_y_ - r_yz);
                double y_high = std::min(y_max_, c_y_ + r_yz);
                if (y_high <= y_low) {
                    return 0.0; // no overlap in y
                }

                IntegrandY integrandY(y_low, y_high, z_min_, z_max_, c_y_, c_z_, r_yz);
                return adaptive_quadrature(integrandY, y_low, y_high, tol_);
            }
        };

        double volume_overlap_adaptive(
            double x_min, double x_max,
            double y_min, double y_max,
            double z_min, double z_max,
            double c_x, double c_y, double c_z,
            double radius,
            double tol = 1e-6)
        {
            // Clip x integration bounds by sphere radius and cube limits
            double x_low = std::max(x_min, c_x - radius);
            double x_high = std::min(x_max, c_x + radius);
            if (x_high <= x_low) return 0.0; // no overlap in x dimension

            IntegrandX integrandX(x_low, x_high, y_min, y_max, z_min, z_max,
                                  c_x, c_y, c_z, radius, tol);
            return adaptive_quadrature(integrandX, x_low, x_high, tol);
        }

        } // namespace geometry
```

chongchonghe:
I really don't think it's worth it to do a complex integral for this. The piecewise constant function in 3D that I'm currently use is already decent for the purpose. A simple approximation formula like the one I proposed above should be good enough. 

I'm going to try it in a simple Python script, and maybe also try the algorithm you recommended and compare them.

BenWibking:
> I really don't think it's worth it to do a complex integral for this. The piecewise constant function in 3D that I'm currently use is already decent for the purpose. A simple approximation formula like the one I proposed above should be good enough.

I think we have to do something that guarantees conservation across coarse-fine boundaries. So some kind of (at least approximate) volume integral is necessary. Of course, I would be happy if this turns out to be wrong. Then it would be a lot easier.

BenWibking:
> > Actually, just to be sure, can you plot the [entropy function](https://en.wikipedia.org/wiki/Entropy_(astrophysics)), defined as K = k T / n 2 / 3 , as a function of time?
> 
> What do you mean "as a function of time"? That requires a single number at a snapshot, but how do you integrate 'Entropy' over volume? Entropy is not a volume density of anything. Perhaps it's because I'm not familiar with Entropy in astrophysics.

You could just compute the maximum entropy (over space) as a function of time.

chongchonghe:
> > I really don't think it's worth it to do a complex integral for this. The piecewise constant function in 3D that I'm currently use is already decent for the purpose. A simple approximation formula like the one I proposed above should be good enough.
> 
> I think we have to do something that guarantees conservation across coarse-fine boundaries. So some kind of (at least approximate) volume integral is necessary. Of course, I would be happy if this turns out to be wrong. Then it would be a lot easier.

We use enough ghost cells for SN deposition, so we always deposit into uniform cells. Therefore, there is no issue in the conservation. Am I missing something? 

Oh, conservation of momentum is an issue, but it's not due to refinement boundaries. Due to the asymmetry of my current algorithm, the net momentum of the SNR region is usually not zero.

BenWibking:
> We use enough ghost cells for SN deposition, so we always deposit into uniform cells. Therefore, there is no issue in the conservation. Am I missing something?

No, I forgot about this... Then it's fine.

Actually, if we assume the SN always goes off in the center of the cell, we can just precompute the values of the integral, and store them in a lookup table.

chongchonghe:
> > We use enough ghost cells for SN deposition, so we always deposit into uniform cells. Therefore, there is no issue in the conservation. Am I missing something?
> 
> No, I forgot about this... Then it's fine.
> 
> Actually, if we assume the SN always goes off in the center of the cell, we can just precompute the values of the integral, and store them in a lookup table.

That would conserve momentum, which is pretty nice. However, that means the SN position essentially becomes a step function of the particle position. I can't think of any significant issue with that. I guess it's probably fine.

chongchonghe:
@BenWibking Ready for your review! (Mark is on vacation)

but I won't merge it yet. I'll wait for Mark's review.

BenWibking:
It looks like the GPU tests fail because this variable is accessed on device:
```
/tmp/bwibking/_work/1/s/src/particles/particle_types.hpp:91:17: note: host variable declared here
   91 | inline SNScheme SN_scheme = SNScheme::SN_thermal_only; // NOLINT
      |                 ^
```

BenWibking:
It looks like all of the Gravity3D tests failed due to a hydro timestep issue. Did the test particles explode? If so, that might be why those tests are failing.

chongchonghe:
With the new SN feedback model, the original gravity_3d.cpp no longer works because it relies on the old SN model where we only did mass deposition. I have to rewrite this test in CGS units and do a proper check for mass deposition (m_ej = 10 M_sun instead of particle mass)

chongchonghe:
> It looks like all of the Gravity3D tests failed due to a hydro timestep issue. Did the test particles explode? If so, that might be why those tests are failing.

I am exploding energy in CGS units into hydro cells with dimensionless units. There is a >50 orders of magnitude difference!

chongchonghe:
@BenWibking I tried on my computer and all the tests that use GrackleLikeCooling failed with `SIGILL Invalid` caught by `amrex.fpe_trap_invalid=1`. Can you take a look? 

I think we really need to add one of the test problems with GrackleLikeCooling into the unit tests. Currently none of those problems are activated. We can reduce max_timesteps to make ti run fast. That will help catch any problem caused by Grackle cooling module.

chongchonghe:
I have to temporarily disable FPE in test_SN.

Another option is to disable Grackle cooling for test_SN.cpp. The SN feedback module does not rely on cooling anyway.

chongchonghe:
This PR is well tested and should work just fine. @BenWibking  If you want to use this for a test run for your talk or proposal, go for it.

BenWibking:
> This PR is well tested and should work just fine. @BenWibking If you want to use this for a test run for your talk or proposal, go for it.

A star formation model to create the star particles is still needed though, right?

chongchonghe:
> > This PR is well tested and should work just fine. @BenWibking If you want to use this for a test run for your talk or proposal, go for it.
> 
> A star formation model to create the star particles is still needed though, right?

Right. We'll need Aditi's StochasticStellarPop module.

markkrumholz:
Just want to check on the status of this code, since I'm now back online. Is there stuff I should review, or @BenWibking are you happy to do so? I will also go look at the paper draft in any event.

chongchonghe:
> Just want to check on the status of this code, since I'm now back online. Is there stuff I should review, or @BenWibking are you happy to do so? I will also go look at the paper draft in any event.

@markkrumholz  Oh. I specifically requested waiting until your review before we merge this branch. Ben has reviewed it and I have addressed all his comments.

BenWibking (approved):
Looks good to me.

PR #996, opened 2025-04-15
Add the ability to split massive particles

BenWibking:
### Description
This adds the ability to split each massive particle into an integer number `splitFactor` new particles of equal mass.

This is needed in order to scale the disk galaxy problem to higher resolutions.

### Related issues
N/A

### Checklist


BenWibking:
On avatargpu, it reports the wrong number of CICParticles:
```
Position error: 0.5971851841
Position norm: 16.43533701
Relative error: 0.03633543893
Expected number of CIC particles: 16
Actual number of CIC particles: 0
Expected number of test particles: 8
Actual number of test particles: 8
```
Need to debug.

src/problems/Gravity3D/gravity_3d.cpp
```diff
@@ -253,6 +251,16 @@ template <> void QuokkaSimulation<BinaryOrbit>::createInitialCICParticles()
 	const int nreal_extra = 4; // mass vx vy vz
 	CICParticles->SetVerbose(1);
 	CICParticles->InitFromAsciiFile("Gravity3D.txt", nreal_extra, nullptr);
+
+	// test particle splitting
+	// (this is intended to only be used when restarting at a higher resolution)
+	if (do_split_particles) {
+		amrex::Print() << "Splitting CICParticles using split_factor = " << split_factor << "\n";
+		for (int lev = 0; lev <= CICParticles->finestLevel(); ++lev) {
```
chongchonghe: If you do splitParticles in createInitialCICParticles, all the existing particles must be in the base level so you don't have to loop over all levels.

chongchonghe:
A design consideration: `particleRegister_.splitParticles` would split all massive particles. If this is intended behaviour, I would write a separate function to do `splitParticles` at [here](https://github.com/quokka-astro/quokka/blob/ec126d3f315e5f1c6a4848a526902cf14e3b7045/src/simulation.hpp#L760) inside simulation.hpp, right after `InitPhyParticles()` and `ReadCheckpointFile()`, so that you can split particles read from restarting a simulation. Splitting all particle types inside `createInitialCICParticles` is a bit confusing.

src/particles/PhysicsParticles.hpp
```diff
@@ -359,6 +362,74 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 		}
 	}
 
+	void splitParticles(int const lev, int const splitFactor) override
+	{
+		if (container_ != nullptr && this->getMassIndex() >= 0) {
+			for (amrex::MFIter mfi = container_->MakeMFIter(lev); mfi.isValid(); ++mfi) {
```
chongchonghe: I would recommend using `typename ContainerType::ParIterType pIter(*container_, lev)` like in `kickParticles` to loop over all tiles that have particles in it. `MakeMFIter` actually make all tiles to include all hydro cells at a level, which is probably less efficient.

BenWibking:
> A design consideration: `particleRegister_.splitParticles` would split all massive particles. If this is intended behaviour, I would write a separate function to do `splitParticles` at [here](https://github.com/quokka-astro/quokka/blob/ec126d3f315e5f1c6a4848a526902cf14e3b7045/src/simulation.hpp#L760) inside simulation.hpp, right after `InitPhyParticles()` and `ReadCheckpointFile()`, so that you can split particles read from restarting a simulation. Splitting all particle types inside `createInitialCICParticles` is a bit confusing.

The reason for doing this is because we don't want to do a restart for the galaxy problem when rescaling to a higher resolution. We want to generate a new set of (static-)AMR grids in the initial conditions generator, copy level 0 from the plotfile onto the base grid, then interpolate from the coarse grid and copy from a lower-res plotfile for all refined levels sequentially. So my thinking was that it didn't make sense to use the restart machinery for this case, but to implement it as a problem-specific set of functionality.

The machinery to interpolate grids already exists in AMReX, but the machinery to split particles does not, so I thought that it made the most sense to put the particle splitting code in PhysicsParticles.

src/particles/PhysicsParticles.hpp
BenWibking: That's a good point. I'll do that.

BenWibking:
See here:
https://github.com/quokka-astro/quokka/blob/ca270fb54010bb87484e3a78d4f3e6ed1128165c/src/problems/AgoraGalaxy/galaxy.cpp#L299

and here:
https://github.com/quokka-astro/quokka/blob/ca270fb54010bb87484e3a78d4f3e6ed1128165c/src/problems/AgoraGalaxy/galaxy.cpp#L308

chongchonghe:
> The reason for doing this is because we don't want to do a restart for the galaxy problem when rescaling to a higher resolution. We want to generate a new set of (static-)AMR grids in the initial conditions generator, copy level 0 from the plotfile onto the base grid, then interpolate from the coarse grid and copy from a lower-res plotfile for all refined levels sequentially. So my thinking was that it didn't make sense to use the restart machinery for this case, but to implement it as a problem-specific set of functionality.
> 
> The machinery to interpolate grids already exists in AMReX, but the machinery to split particles does not, so I thought that it made the most sense to put the particle splitting code in PhysicsParticles.

OK, then we should only split CIC particles inside `createInitialCICParticles`. `particleRegister_.getParticleDescriptor(ParticleType:CIC).splitParticles(lev, split_factor);` 

Currently, it will split TestParticles if `createInitialTestParticles` is called before `createInitialCICParticles` and won't otherwise. <s>This is a potential cause of the GPU errors.</s> It won't cause the number of particles to be 0, though. Nonetheless, it may cause inconsistent results.

BenWibking:
> > The reason for doing this is because we don't want to do a restart for the galaxy problem when rescaling to a higher resolution. We want to generate a new set of (static-)AMR grids in the initial conditions generator, copy level 0 from the plotfile onto the base grid, then interpolate from the coarse grid and copy from a lower-res plotfile for all refined levels sequentially. So my thinking was that it didn't make sense to use the restart machinery for this case, but to implement it as a problem-specific set of functionality.
> > The machinery to interpolate grids already exists in AMReX, but the machinery to split particles does not, so I thought that it made the most sense to put the particle splitting code in PhysicsParticles.
> 
> OK, then we should only split CIC particles inside `createInitialCICParticles`. `particleRegister_.getParticleDescriptor(ParticleType:CIC).splitParticles(lev, split_factor);`
> 
> Currently, it will split TestParticles if `createInitialTestParticles` is called before `createInitialCICParticles` and won't otherwise. This is a potential cause of the GPU errors.

Ok, good point. Thanks for catching that. I'll change it to only split CICParticles.

src/particles/PhysicsParticles.hpp
BenWibking: Done.

src/problems/Gravity3D/gravity_3d.cpp
BenWibking: Done.

BenWibking:
Hmm, I looked over the code again and I don't understand the source of the GPU error. It is exactly analogous to the code for counting TestParticles, which works.

src/particles/PhysicsParticles.hpp
```diff
@@ -619,7 +685,7 @@ template <typename problem_t> class PhysicsParticleRegister
 #endif // AMREX_SPACEDIM == 3
 
 	// Retrieve a particle descriptor by type
-	[[nodiscard]] auto getParticleDescriptor(ParticleType type) const -> const PhysicsParticleDescriptorBase *
+	[[nodiscard]] auto getParticleDescriptor(ParticleType type) const -> PhysicsParticleDescriptorBase *
```
BenWibking, line 817: We have to un-`const` this pointer in order to call `splitParticles`, since it is not a `const` function.

BenWibking:
@chongchonghe I think there is a missing GPU synchronization in `amrex::ParticleContainer_impl<>::NumberOfParticlesAtLevel`. I will try to fix it.

BenWibking:
@chongchonghe I merged your changes and I now don't understand what the code is doing. It says there are 16 CICParticles, but it returns an empty vector of `real_data`:
```
Actual number of CIC particles: 16
Expected number of CIC particles: 16
Size of particle data: 0
Particle positions and data are: 
Exact positions are: 
0.999998, 0.001999998667, 0
Position error: 0
Position norm: 0
Relative error: nan
Test failed.
```
Do you have any idea what is going on?

src/problems/Gravity3D/gravity_3d.cpp
```diff
@@ -388,6 +399,11 @@ auto problem_main() -> int
 	// particle actions must be called on all ranks
 	const int max_level = sim.maxLevel();
 	auto [real_data, int_data] = sim.particleRegister_.getParticleDescriptor(quokka::ParticleType::CIC)->getParticleDataAtLevel(max_level);
+	const int n_particle_CIC = sim.particleRegister_.getParticleDescriptor(quokka::ParticleType::CIC)->getNumParticles();
+	const int n_expected_CIC_particles = do_split_particles ? 2 * split_factor : 2;
+	amrex::Print() << "Actual number of CIC particles: " << n_particle_CIC << "\n";
+	amrex::Print() << "Expected number of CIC particles: " << n_particle_CIC << "\n";
+	amrex::Print() << "Size of particle data: " << real_data.size() << "\n";
```
BenWibking: This prints `0`?

BenWibking:
Still fails:
```
Actual number of CIC particles: 0
Expected number of CIC particles: 0
Particle positions and data are: 
Position: -0.9782711181, 0.01103623949, -0.006089550914 | Mass: 0.125 | Velocities: -0.1931962808, -0.5648063309, 0.02163001356
Position: -0.9773249723, -0.0008343579107, -0.01845878541 | Mass: 0.125 | Velocities: -0.1917807818, -0.4646255008, 0.1406204983
Position: -1.022831509, 0.0231140174, -0.0008332088827 | Mass: 0.125 | Velocities: 0.2133437747, -0.6773534777, -0.01771827135
Position: -0.9763014943, 0.02204988269, 0.02544700878 | Mass: 0.125 | Velocities: -0.1451534713, -0.6235802095, -0.2049613422
Position: -1.005155595, -0.01050886194, 0.002493407054 | Mass: 0.125 | Velocities: 0.0717928188, -0.3422498384, -0.06708735388
Position: -1.011299585, 0.02157706893, 0.00130005336 | Mass: 0.125 | Velocities: 0.1143160282, -0.6842112281, -0.03789041449
Position: -1.027272229, -0.01010840468, -0.01199704492 | Mass: 0.125 | Velocities: 0.2626214578, -0.3904430738, 0.06540315917
Position: -0.9788955285, -0.03016859474, -0.01869952234 | Mass: 0.125 | Velocities: -0.1238785887, -0.2527515249, 0.1000511809
Position: 1.01584565, -0.02455897898, 0.03035832554 | Mass: 0.125 | Velocities: -0.1232808465, 0.674648505, -0.1742098934
Position: 1.02456557, 0.00749767712, -0.005487828395 | Mass: 0.125 | Velocities: -0.2662420454, 0.4440458232, 0.1100542597
Position: 0.9849509855, 0.005340176935, 0.02331488655 | Mass: 0.125 | Velocities: 0.09628991004, 0.4618039477, -0.1497797442
Position: 1.005730547, 0.01753520683, 0.004764281333 | Mass: 0.125 | Velocities: -0.08738493125, 0.3300609141, 0.03720878295
Position: 0.9733451881, 0.01543161532, 0.01658410091 | Mass: 0.125 | Velocities: 0.194824315, 0.3933430665, -0.06004233133
Position: 1.007927858, -0.0161234033, -0.02842929821 | Mass: 0.125 | Velocities: -0.068512187, 0.617523475, 0.294793978
Position: 0.977416383, 0.01420999936, 0.009582778271 | Mass: 0.125 | Velocities: 0.1734829474, 0.3921066253, -0.002785021827
Position: 0.9872927039, -0.02074344084, 0.01530670923 | Mass: 0.125 | Velocities: 0.07275781547, 0.6864886188, -0.05528771429
Exact positions are: 
0.999998, 0.001999998667, 0
Position error: 0.7551461693
Position norm: 16.42441163
Relative error: 0.04597706062
Test failed.
...wrong number of CIC particles.
```

chongchonghe:
The particle position data shows that there are indeed 16 particles, as expected. However, `getNumParticles()` returns 0.

Apparently, the number of CIC particles in the end is correct, so your particle splitter worked. It seems like there is a bug in `getNumParticles()`, either in the code I wrote, or in AMReX's `TotalNumberOfParticles()` function.

src/problems/Gravity3D/gravity_3d.cpp
```diff
@@ -291,6 +289,14 @@ template <> void QuokkaSimulation<TestParticle>::createInitialCICParticles()
 	const int nreal_extra = 4; // mass vx vy vz
 	CICParticles->SetVerbose(1);
 	CICParticles->InitFromAsciiFile("Gravity3D.txt", nreal_extra, nullptr);
+
+	// test particle splitting
+	// (this is intended to only be used when restarting at a higher resolution)
+	if (do_split_particles) {
+		amrex::Print() << "Splitting CICParticles using split_factor = " << split_factor << "\n";
+		int const lev = 0; // all CICParticles are on level 0
```
chongchonghe, line 297: I believe all CICParticles are on level 1 when the whole domain is refined.

src/problems/Gravity3D/gravity_3d.cpp
```diff
@@ -389,12 +400,17 @@ auto problem_main() -> int
 
 	// particle actions must be called on all ranks
 	const int max_level = sim.maxLevel();
-	auto [real_data, int_data] = sim.particleRegister_.getParticleDescriptor(quokka::ParticleType::CIC)->getParticleDataAtLevel(max_level);
+	const int n_particle_CIC = sim.particleRegister_.getParticleDescriptor(quokka::ParticleType::CIC)->getNumParticles();
+	const int n_expected_CIC_particles = do_split_particles ? 2 * split_factor : 2;
+	amrex::Print() << "Actual number of CIC particles: " << n_particle_CIC << "\n";
+	amrex::Print() << "Expected number of CIC particles: " << n_particle_CIC << "\n";
```
chongchonghe: This should be `n_expected_CIC_particles`?

chongchonghe:
Oh, I found the issue. It's because `container_->TotalNumberOfParticles(true, false)` only counts valid particles, meaning particles with positive indices. However, when the particle IDs are not being set correctly when creating the particles, causing `getNumParticles()` to return 0. 

One option is to not require valid IDs by changing it to `TotalNumberOfParticles(false, false)`. I tried it and the split test passed. The other option is to manually set particle ID right after `InitFromAsciiFile()`. Which option makes more sense to you?

Since we always do redistribute to remove invalid particles, the first option seems the more preferable one.

src/problems/Gravity3D/gravity_3d.cpp
```diff
@@ -451,9 +467,9 @@ auto problem_main() -> int
 				// particle positions are known to be incorrect if a particle crosses the refinement boundary
 				pos_max_err_tol = std::numeric_limits<double>::infinity();
 			} else {
-				if (sim.tNew_[0] < 0.011) {
+				if (sim.tNew_[0] < 0.011 && !do_split_particles) {
```
chongchonghe, line 470: We don't want to check particle positions when we do particle split, so I would wrap everything about particle position inside `if (!do_split_particles)`.

BenWibking:
> Oh, I found the issue. It's because `container_->TotalNumberOfParticles(true, false)` only counts valid particles, meaning particles with positive indices. However, when the particle IDs are not being set correctly when creating the particles, causing `getNumParticles()` to return 0.

Do you know why the particle IDs aren't valid? I set the IDs following your code to create new particles.

src/problems/Gravity3D/gravity_3d.cpp
BenWibking: I think we want to check the positions, but with a looser tolerance. They should be close to the positions of the original particles if they haven't evolved very far.

chongchonghe:
> > Oh, I found the issue. It's because `container_->TotalNumberOfParticles(true, false)` only counts valid particles, meaning particles with positive indices. However, when the particle IDs are not being set correctly when creating the particles, causing `getNumParticles()` to return 0.
> 
> Do you know why the particle IDs aren't valid? I set the IDs following your code to create new particles.

Oh, that's the reason. You set it wrong. We should do something like

```
					// Set particle ID and CPU
					p.id() = pid_start + base_offset + p_idx;
					p.cpu() = cpu_id;
```

see examples in `ParticleCreator()` in `gravity_3d.cpp`. You set p.id() to cpu_id, which is 0.

src/particles/PhysicsParticles.hpp
```diff
@@ -456,6 +457,69 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
+						auto &p_new = new_particles[idx_new]; // NOLINT
+						// copy old particle properties
+						for (int rc = 0; rc < ContainerType::ParticleType::NReal; ++rc) {
+							p_new.rdata(rc) = p_old.rdata(rc);
+						}
+						for (int ic = 0; ic < ContainerType::ParticleType::NInt; ++ic) {
+							p_new.idata(ic) = p_old.idata(ic);
+						}
+						// set new particle properties
+						p_new.id() = cpu_id;
```
chongchonghe: This is not correct. `p_new.id()` should be different for each new particle.

BenWibking: Oops, that was a typo... That would certainly cause weird things to happen.

src/particles/PhysicsParticles.hpp
```diff
@@ -456,6 +457,70 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
+						// copy old particle properties
+						for (int rc = 0; rc < ContainerType::ParticleType::NReal; ++rc) {
+							p_new.rdata(rc) = p_old.rdata(rc);
+						}
+						for (int ic = 0; ic < ContainerType::ParticleType::NInt; ++ic) {
+							p_new.idata(ic) = p_old.idata(ic);
+						}
+						// set new particle properties
+						p_new.cpu() = cpu_id;
+						p_new.id() = pid + idx * splitFactor + idx_new;
```
BenWibking, line 511: @chongchonghe does this look right?

src/particles/PhysicsParticles.hpp
chongchonghe: That's why I like to let LLM write and I proofread 😉

src/particles/PhysicsParticles.hpp
chongchonghe: looks right to me

BenWibking:
@markkrumholz This should be ready for you to review now.

chongchonghe (approved):
Looks good to me assuming the GPU tests will pass.

chongchonghe:
@BenWibking Can you bypass code owner review and merge this?

BenWibking:
> @BenWibking Can you bypass code owner review and merge this?

I added the admin role to the bypass list, but it still doesn't let me force merge it for some reason.

BenWibking:
Well, I can bypass the branch protection rules on other PRs, but not my own ... That's a bit of a catch-22.

markkrumholz:
Now that I'm back online I can review this.

markkrumholz (commented):
OK, I have looked through this, and in general it looks fine, but I would like to suggest one minor change: as written this process will not exactly conserve center of mass position and angular momentum, because there is no constraint on the split particle positions. It would be easy to modify the procedure to fix this: just modify the loop that creates new particle properties to go to N-1 instead of N, and keep track of the cumulative sum of particle positions. Then for the Nth new particle to be created, don't set its position randomly, set it so that the center of mass of the split particles is exactly equal to the position of the previous particle.

BenWibking:
> OK, I have looked through this, and in general it looks fine, but I would like to suggest one minor change: as written this process will not exactly conserve center of mass position and angular momentum, because there is no constraint on the split particle positions. It would be easy to modify the procedure to fix this: just modify the loop that creates new particle properties to go to N-1 instead of N, and keep track of the cumulative sum of particle positions. Then for the Nth new particle to be created, don't set its position randomly, set it so that the center of mass of the split particles is exactly equal to the position of the previous particle.

@markkrumholz This is easy to do, but I'm not sure if I see the point. What is a use case where this would matter?

markkrumholz:
> > OK, I have looked through this, and in general it looks fine, but I would like to suggest one minor change: as written this process will not exactly conserve center of mass position and angular momentum, because there is no constraint on the split particle positions. It would be easy to modify the procedure to fix this: just modify the loop that creates new particle properties to go to N-1 instead of N, and keep track of the cumulative sum of particle positions. Then for the Nth new particle to be created, don't set its position randomly, set it so that the center of mass of the split particles is exactly equal to the position of the previous particle.
> 
> @markkrumholz This is easy to do, but I'm not sure if I see the point. What is a use case where this would matter?

If you wanted to track the total angular momentum content of a system, the current algorithm does not exactly conserve that, so there would be artificial jumps in the angular momentum at splitting. But this is probably not a huge deal given that we're not exactly angular momentum-conserving anyway, but guaranteeing exact angular momentum conservation would be the reason to do it. I'm happy to approve the PR as-is, and we can revisit this if it ever becomes relevant.

markkrumholz (approved):

chongchonghe:
After #1005 this PR won't be able to merge because the Gravity3D test was removed. We should move particle splitting test to the binary_orbit test problem.

BenWibking:
> After #1005 this PR won't be able to merge because the Gravity3D test was removed. We should move particle splitting test to the binary_orbit test problem.

Yes, I agree. Feel free to go ahead and do this if you want. Otherwise, I will get to this after the AMReX conference.

BenWibking:
Superceded by #1104.

PR #1002, opened 2025-04-19
TotalNumberOfParticles(false, false)

chongchonghe:
### Description

Changed `TotalNumberOfParticles(true, false)` to `TotalNumberOfParticles(false, false)`. 

`container_->TotalNumberOfParticles(true, false)` only counts valid particles, which are particles with positive IDs. However, the particle IDs are not being set correctly when creating the particles, causing `getNumParticles()` to return 0. 

By setting the first argument of `TotalNumberOfParticles` to `false`, we do not require valid IDs when counting. Since we always do redistribute to remove invalid particles, this is a fine choice. 

### Related issues
Fixes a bug that caused failure in #996 . 

### Checklist

### DIFF
--- a/src/particles/PhysicsParticles.hpp
+++ b/src/particles/PhysicsParticles.hpp
@@ -354,11 +354,13 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 		return {real_data, int_data}; // Empty vectors on non-root ranks
 	}
 
-	// Get the number of particles in the container
+	// Get the number of particles in the container on all levels
 	[[nodiscard]] auto getNumParticles() const -> int override
 	{
 		if (container_ != nullptr) {
-			return static_cast<int>(container_->TotalNumberOfParticles(true, false));
+			// only_valid=false: count invalid particles as well
+			// only_local=false: count particles on all ranks
+			return static_cast<int>(container_->TotalNumberOfParticles(false, false));
 		}
 		return 0;
 	}


BenWibking:
I don't think we want this, since any invalid particles will be deleted on the next Redistribute(), no?

chongchonghe:
> I don't think we want this, since any invalid particles will be deleted on the next Redistribute(), no?

Exactly, invalid particles will be deleted, so we are not going to count them anyway (as long as it's called after Redistribute()). This PR is a simple hack to make `getNumParticles()` return the correct answer right after `createInitialCICParticles()`.

BenWibking:
> > I don't think we want this, since any invalid particles will be deleted on the next Redistribute(), no?
> 
> Exactly, invalid particles will be deleted, so we are not going to count them anyway (as long as it's called after Redistribute()). This PR is a simple hack to make `getNumParticles()` return the correct answer right after `createInitialCICParticles()`.

I think I need to find out why the particles are invalid. It would be useful to have getNumParticles take an optional argument so you could get the invalid particles too, if desired, but maybe the default should be only the valid particles?

Something like: `auto getNumParticles(bool only_valid=true) const -> int`

chongchonghe:
After `p_new.id()` is set correctly in your split function, your test should pass without this PR. Then, we can close this PR.

BenWibking:
Ok, since you discovered my mistake in the other PR was the root cause of this, do you want to close this PR?

chongchonghe:
Yes, I'll close it later.

Issue #666, opened 2024-07-01
add support for running entirely in single precision


BenWibking:
**Describe the proposal**
Support running `amrex::Real` as `float` (compile with `-DAMReX_PRECISION=SINGLE`) and replace all uses of `double` with `amrex::Real`. Ensure that all tests still pass.

**Describe alternatives you've considered**
We have so far relied on users running on GPUs with hardware support for double precision arithmetic.

**Additional context**
Due to AI, future-generation GPUs may completely drop hardware support for double precision arithmetic.

chongchonghe:
I totally agree and support this plan. I recently learned about another code under development that uses mixed precision and achieves extremely high speed, about 100 Mupdates/s/GPU on game cards (RTX 3080/3090). They use single precision for the approximate Riemann solver but use double precision in the conservation step.

BenWibking:
That's interesting.

Over the next couple of years, the code needs to run without double precision at all. If we need extra precision, we can do Kahan summation or represent numbers as the sum of two single precision values (then overload arithmetic operators such as `+` and `*` to emulate double precision).

chongchonghe:
Now that Claude is so powerful, it can probably create a PR to do this.

chongchonghe:
@claude Evaluate the feasibility of creating a PR to do the task proposed in this issue. Further note that, since all unannotated floating point constants are double in C++, we also have to add the suffix _rt to every floating point constant that we use in the code.

chongchonghe:
@BenWibking @markkrumholz I think Claude can probably do this job with some human help, since it's all systematic. There is a design consideration. If we decide to use single precision in the future, do we want to use single precision for everything, or mixed precision between hydro, radiation, and particles. I know AMReX has a separate parameter for particle precision, but I don't know if it allows setting two precision types in hydro and radiation, although we can always define our own (like `quokka::Real_rad`.

BenWibking:
That's a good idea. We'd have to create our own type for radiation precision.

PR #1003, opened 2025-04-19
Be Real

chongchonghe:
Let's use amrex::Real consistently. This is part of the effort to prepare for a potential transition to single precision. More importantly, this is for coding consistency. Currently, the use of `double` and `amrex::Real` is mixed and inconsistent. One barrier to using amrex::Real is that it's cumbersome to type. Therefore, I added a line `using Real = amrex::Real;` in physics_info.hpp which every file imports, so we can use `Real` instead of `amrex::Real` anywhere. 

### Related issues
Related to #666 

### Checklist

### DIFF
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -133,13 +133,13 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 	static constexpr int ncompHyperbolic_ = RadSystem<problem_t>::nvarHyperbolic_;
 	static constexpr int nstartHyperbolic_ = RadSystem<problem_t>::nstartHyperbolic_;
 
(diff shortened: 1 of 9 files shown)


BenWibking:
Unfortunately, it's a bit more complicated. Since all unannotated floating point constants are `double` in C++, we also have to add the suffix `_rt` to every floating point constant that we use in the code. If we don't do this, then the complier will _implicitly cast_ the result to double. Because this is incredibly annoying to do, I did not originally bother to strictly use `amrex::Real` everywhere.

See: https://warpx.readthedocs.io/en/22.11/developers/faq.html#what-is-0-0-rt

chongchonghe:
I see. However, I suggest we use either `Real` or `double` consistently for better readability. I prefer `Real` because it has fewer letters than `double` and is convenient for a future transition.

PR #1005, opened 2025-04-20
New ParticleCreation test to replace Gravity3D

chongchonghe:
### Description

The previous `Gravity3D` problem tested too many things simultaneously, making it hard to track specific module functions. Consequently, I have replaced it with a new test called `ParticleCreation`, which tests particle creation from cell, stellar evolution, and particle destruction. It's unnecessary to test gravity in this problem because gravity is tested in `BinaryOrbit`. Similarly, SN feedback is tested in the `SN` test in #995. Additionally, `Gravity3D` is badly named. 

More changes:

- Add logging of the number of particles created inside `createParticlesImpl`
- Add logging of the number of particles removed at each level inside `destroyParticlesImpl`
- `updateEvolutionStage` updates a particle's stage based on its `deathtime`. 
- Add an extra evolution stage: `Removed`. Only those particles will be removed.
- Updated the format of `printParticleStatistics` 

### Checklist
_Before this pull request can be reviewed, all of these tasks should be completed. Denote completed tasks with an `x` inside the square brackets `[ ]` in the Markdown source below:_
- [x] I have added a description (see above).
- [x] I have added a link to any related issues (if applicable; see above).
- [x] I have read the [Contributing Guide](https://github.com/quokka-astro/quokka/blob/development/CONTRIBUTING.md).
- [x] I have added tests for any new physics that this PR adds to the code.
- [x] *(For quokka-astro org members)* I have manually triggered the GPU tests with the magic comment `/azp run`.

PR #1009, opened 2025-04-22
Allow to force some particles to always live in finest cells

chongchonghe:
### Description
We want to force some particles to always live on the finest cells. Those particles could be sink particles that need to accrete, or `SNProgenitor` particles that can go SN explosion at any time. This is also important for getting around the issue with the gravity solver on refinment boundaries. 

The user can turn on manually turn on this feature for a specific particle type by running the following line of code in `problem_main()`

```cpp
sim.particleRegister_.getParticleDescriptor(quokka::ParticleType::Test)->setForceFinestLevel(true);
```

Here is a demonstration with the `ParticleCreation` test. It seems that by tagging a single cell at the particle location, the code actually refines at least 4+ cells around it. This happens to be the desired behaviour. Lovely. 

plt00004_Slice_z_density

Note: This branch is set to merge to #1005 because it's based on that branch. This is just for showing the actual changes made in this PR. Once we merge #1005 , I'll update this PR to merge to dev branch. 

### Related issues
None. 

### Checklist


BenWibking:
Whether or not this works depends on the value of `amr.n_error_buf`: https://amrex-codes.github.io/amrex/docs_html/GridCreation.html.

It might be a good idea to check that n_error_buf is >= 3 when this is turned on.

chongchonghe:
> Whether or not this works depends on the value of `amr.n_error_buf`: https://amrex-codes.github.io/amrex/docs_html/GridCreation.html.
> 
> It might be a good idea to check that n_error_buf is >= 3 when this is turned on.

Nice. We should check `amr.n_error_buf >= 4` for SN and sink particles. The particle can drift one cell before the code calls `depositSN`. Perhaps we should `redistribute` the particles after the second kick; then we can reduce this number to 3.

BenWibking:
> > Whether or not this works depends on the value of `amr.n_error_buf`: https://amrex-codes.github.io/amrex/docs_html/GridCreation.html.
> > It might be a good idea to check that n_error_buf is >= 3 when this is turned on.
> 
> Nice. We should check `amr.n_error_buf >= 4` for SN and sink particles. The particle can drift one cell before the code calls `depositSN`. Perhaps we should `redistribute` the particles after the second kick; then we can reduce this number to 3.

Ah right. We should just set it to 4. I don't think we can add more redistributes if we want to generalize it to work with AMR subcycling in the future.

src/particles/PhysicsParticles.hpp
```diff
@@ -617,6 +623,39 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
+			const auto tag = tags.array(pti);
+
+			amrex::ParallelFor(np, [=] AMREX_GPU_DEVICE(int64_t idx) {
+				auto &p = pData[idx]; // NOLINT(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+				// Find the cell containing the particle
+				const int ix = static_cast<int>(amrex::Math::floor((p.pos(0) - plo[0]) * dxi[0]));
+				const int iy = static_cast<int>(amrex::Math::floor((p.pos(1) - plo[1]) * dxi[1]));
+				const int iz = static_cast<int>(amrex::Math::floor((p.pos(2) - plo[2]) * dxi[2]));
+
+				tag(ix, iy, iz) = amrex::TagBox::SET;
```
BenWibking, line 659: There is a race condition here, since multiple particles could live in the same cell. I'm not sure whether it's guaranteed to work if all of the overlapping writes all set the same value. Might have to check the CUDA and/or HIP documentation.

BenWibking (changes requested):

src/particles/PhysicsParticles.hpp
chongchonghe: There is likely a problematic race condition here. We can use atomic max to avoid this.

chongchonghe: problem solved.

src/particles/PhysicsParticles.hpp
```diff
@@ -617,6 +623,40 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
+
+			amrex::ParallelFor(np, [=] AMREX_GPU_DEVICE(int64_t idx) {
+				auto &p = pData[idx]; // NOLINT(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+				// Find the cell containing the particle
+				const int ix = static_cast<int>(amrex::Math::floor((p.pos(0) - plo[0]) * dxi[0]));
+				const int iy = static_cast<int>(amrex::Math::floor((p.pos(1) - plo[1]) * dxi[1]));
+				const int iz = static_cast<int>(amrex::Math::floor((p.pos(2) - plo[2]) * dxi[2]));
+
+				// Use atomic operation to safely set the tag
+				amrex::Gpu::Atomic::Max(&tag(ix, iy, iz), static_cast<char>(amrex::TagBox::SET));
```
BenWibking: Why is the cast to `char` necessary here?

src/particles/PhysicsParticles.hpp
chongchonghe: and this is logical. There might be a situation where one particle wants to set `BUF` and another wants to set `SET` to a cell. In this situation, we want to set it to `SET`, which is the max of the two.

src/particles/PhysicsParticles.hpp
chongchonghe: that's because atomic max does not recognize TagBox data type. I believe by setting tag to a char, it gets automatically converted to TagBox.

BenWibking: Just to be sure, can you ask on the AMReX slack about this?

chongchonghe: The test on my computer works perfectly fine. Do you think that's sufficient?

chongchonghe: Oh, if you meant asking if an atomic operation is necessary... I think that's worth asking. I just checked Castro source code and they did not do atomic operation.

BenWibking: Right, but I think they don't have particles.

chongchonghe: see 
https://github.com/AMReX-Astro/Castro/blob/1470c5449e5a240e9454838ae4760a4d34261f81/Source/driver/Castro.cpp#L3546 
and
https://github.com/AMReX-Astro/Castro/blob/1470c5449e5a240e9454838ae4760a4d34261f81/Exec/hydro_tests/rotating_torus/problem_tagging.H#L27

chongchonghe: Oh, they are doing it in parallelfor over indices I,j,k, so there is no possible racing condition. Nevermind. That does not prove we don't need atomic operations.

BenWibking: Ok, Andrew said we don't need to do this with an atomic operation. It should be fine to just set it directly, since the buffer tags are added in a separate step.

chongchonghe:
> > > Whether or not this works depends on the value of `amr.n_error_buf`: https://amrex-codes.github.io/amrex/docs_html/GridCreation.html.
> > > It might be a good idea to check that n_error_buf is >= 3 when this is turned on.
> > 
> > 
> > Nice. We should check `amr.n_error_buf >= 4` for SN and sink particles. The particle can drift one cell before the code calls `depositSN`. Perhaps we should `redistribute` the particles after the second kick; then we can reduce this number to 3.
> 
> Ah right. We should just set it to 4. I don't think we can add more redistributes if we want to generalize it to work with AMR subcycling in the future.

Wait a second. If a particle tag 2 cells, they become 4 cells after refinement, so the optimal `n_error_buf` is 2. Does this make sense? @BenWibking

BenWibking:
> > Ah right. We should just set it to 4. I don't think we can add more redistributes if we want to generalize it to work with AMR subcycling in the future.
> 
> Wait a second. If a particle tag 2 cells, they become 4 cells after refinement, so the optimal `n_error_buf` is 2. Does this make sense? @BenWibking

Ah, yes. That's true. But I think it's better to have a larger buffer anyway.

chongchonghe:
`n_error_buf = 2` v.s. `n_error_buf = 4`:

</img></img>

The structure of buf=2 and buf=4 are vastly different. There are too many wasted cells when set buf=4. buf=3 is probably fine since `2*3+1 = 7 <= 8`, but `2*4+1 = 9 <= 16`.

src/particles/PhysicsParticles.hpp
chongchonghe: Another fact that helps is that the result is deterministic even if there is a race. It should be safe to set it directly.

BenWibking:
It looks like the Advection2D test failed for some reason.

chongchonghe:
@BenWibking Looks good now.

chongchonghe:
@BenWibking @markkrumholz Can you approve this PR if it looks fine?

BenWibking (approved):

PR #1013, opened 2025-04-23
Support setting the maximum number of binary output files

BenWibking:
### Description
Support the ParmParse options `amr.plot_nfiles` and `amr.checkpoint_nfiles`. These set the maximum number of binary files written on disk per MultiFab.

This is necessary to optimize I/O performance on some Lustre filesystems. These should usually be set to either:
 * the number of MPI ranks, or, for extreme scale simulations, or
 * the number of nodes.

A value of `-1` sets it to the number of MPI ranks (which is the default).

The striping on Lustre filesystems should usually be set using the command:
```
lfs setstripe -c 1 -S 16M my_sim_working_dir
```
where `my_sim_working_dir` is the directory where the simulation is running and will write plotfiles and checkpoints.

### Related issues
N/A

### Checklist


src/simulation.hpp
```diff
@@ -2571,6 +2583,8 @@ template <typename problem_t> void AMRSimulation<problem_t>::WritePlotFile()
 	quokka::OpenPMDOutput::WriteFile(varnames, finest_level + 1, mf_ptr, Geom(), plot_file, tNew_[0], istep[0]);
 	WriteMetadataFile(plotfilename + ".yaml");
 #else
+	// sets the maximum number of binary files per MultiFab
+	amrex::VisMF::SetNOutFiles(plot_nfiles);
```
chongchonghe: This seems like a universal action and it might set the maximum number of binary files per MultiFab for all output processes. Am I right? I can't think of anything wrong about this, through.

chongchonghe (approved):

src/simulation.hpp
chongchonghe: actually there is a concern if this is true. by setting one parameter to 1, you essentially have set another parameter to 1 as well.

chongchonghe (dismissed):
see my other comments. maybe reset to -1 afterwards?

src/simulation.hpp
BenWibking: That's a good point. I've asked in the AMReX slack if there's a better way to set this only for a single write. If not, I can just restore the original value after doing the write.

BenWibking: We have to reset it manually. I will create an RAII wrapper to do this on destruction.

src/io/io_utils.hpp
BenWibking, line 1: @chongchonghe I added this helper class that will save and restore the previous value of NFiles when it goes out of scope.

BenWibking:
I think the Sonarcloud complaints are wrong here.

src/io/io_utils.hpp
chongchonghe: Looks good to me.

chongchonghe:
@BenWibking Do you want to set one of the existing test problems to test this? I recommend we test it in some way. Apart from that, this PR looks good to me.

BenWibking:
> @BenWibking Do you want to set one of the existing test problems to test this? I recommend we test it in some way. Apart from that, this PR looks good to me.

I can set this for the checkpoint/restart test and add a check for the number of binary files per level there.

tests/checkpoint_restart_test.sh
```diff
@@ -1,10 +1,39 @@
...
+# set number of binary output files per level
+# (NOTE: AMReX *never* outputs more binary files than the number of MPI ranks,
+#  so a non-trivial test requires running with NPROC > NFILES.)
+NFILES=2
+NPROC=4
 
+# run to generate checkpoint
+mpirun -np $NPROC $BUILD_DIR/src/problems/HydroBlast3D/test_hydro3d_blast blast_32.in max_walltime=0:00:10 plotfile_interval=100 checkpoint_interval=100 amr.plot_nfiles=$NFILES amr.checkpoint_nfiles=$NFILES
+
+# [amr.plot_nfiles test] verify that the last plotfile contains two binary files per level
```
BenWibking, line 13: Added test for `amr.plot_nfiles` here.

tests/checkpoint_restart_test.sh
```diff
@@ -1,10 +1,39 @@
...
+plotfile=`ls -1drt plt* | head -1`
+nfiles_plt_actual=`ls -1 $plotfile/Level_0/Cell_D_* | wc -l`
+if [ "$nfiles_plt_actual" = "$NFILES" ]; then
+    echo "amr.plot_nfiles working as expected."
+else
+    echo "TEST FAILED: Wrong number of binary cell data files in plotfiles!"
+    exit 1
+fi
+
+# [amr.checkpoint_nfiles test] verify that the last checkpoint contains two binary files per level
```
BenWibking, line 23: Added test for `amr.checkpoint_nfiles` here.

BenWibking:
> @BenWibking Do you want to set one of the existing test problems to test this? I recommend we test it in some way. Apart from that, this PR looks good to me.

@chongchonghe I've added a test for this as part of the CheckpointRestart action.

BenWibking:
@chongchonghe Can you review this? I want to get all of the dependencies of https://github.com/quokka-astro/quokka/pull/506 merged in if they are ready.

BenWibking:
@chongchonghe I think this one is ready to merge.

markkrumholz (approved):

PR #1016, opened 2025-04-26
Dockerfile for CI agents on moth and avatargpu

chongchonghe:
### Description

Add two Dockerfiles for CI agents on moth (ROCm) and avatargpu (CUDA). This has transitioned the azp GPU CI tests to containers. 

### Related issues
None.

### Checklist


.devcontainer/rocm-azp-agent-in-docker/rocm-azp-agent.dockerfile
```diff
@@ -0,0 +1,78 @@
+# Use the AMD ROCm image as the base image
+FROM rocm/dev-ubuntu-24.04@sha256:708bbb9e2031a81e949a0847f0655c4fb1a790f8ac7cd0031116f56b04ada7d5
```
BenWibking: You can set this to the ROCm devcontainer. Then it will take less time to build, and you can remove everything else below except for the azp agent steps.

BenWibking:
@chongchonghe Now that the proposal is submitted, I would like to transfer responsibility for maintaining this to you. I have turned off the daemons that were running, so you can start them inside the containers.

chongchonghe:
@BenWibking OK. The Docker-based avatargpu agent is working. The moth agent is not working properly. It successfully compiles QUOKKA but failed to run the test. I'll try to fix that.

.devcontainer/cuda-azp-agent-in-docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+USER agent
+# Another option is to run the agent as root.
+# ENV AGENT_ALLOW_RUNASROOT="true"
+
+# WORKDIR /home/agent
+# RUN touch /home/agent/.token
+# RUN chmod 666 /home/agent/.token
+
+# # Do not use start.sh in the image. Put start.sh in the job folder and run it via 'singularity run ... ./start.sh'
+# ENTRYPOINT [ "./start.sh" ]
```
BenWibking, line 35: This is odd. Do you get an error when setting this as the entrypoint?

chongchonghe:
Both moth and avatargpu agents are working. Now, we have completed the transition to fully Docker-based GPU tests. I'll refine the Dockerfiles and this PR should be ready to merge later.

.devcontainer/cuda-azp-agent-in-docker/Dockerfile
chongchonghe: There are a few problems, mostly about the way you stored the token. I find it easier to maintain if start.sh is put outside of the docker image. That way, if we want to update it, all we need to do is to restart the container, and we don't have to rebuild the image.

chongchonghe: writing .token file is the only command that requires root permission, and we can easily avoid that. So, there is really no advantage to copying a whole script into the image.

BenWibking:
Ok, great. We still need to move the `regression_tests` directory from my home directory to yours (or to a common accessible storage area), so the nightly regression tests will work.

.devcontainer/cuda-azp-agent-in-docker/Dockerfile
BenWibking: As long as the script is stored in the repo, that sounds fine.

BenWibking:
@chongchonghe Do you want me to move it to `/data/mash`?

chongchonghe:
> @chongchonghe Do you want me to move it to `/data/mash`?

Yes. I have all the containers and agents on mash . Let's keep it all together.

BenWibking:
> > @chongchonghe Do you want me to move it to `/data/mash`?
> 
> Yes. I have all the containers and agents on mash . Let's keep it all together.

I don't have access to mash for some reason. I have put it in `/data/misfit/bwibking/regression-tests` for now.

.devcontainer/cuda-azp-agent-in-docker/Dockerfile
chongchonghe: Just a note that I have to manually copy the script into the agent folder `/data/mash/cche/azp-agent-in-docker-moth/azp-agent-moth`

BenWibking:
It looks like it's failing on moth because of a Python package issue. I think we should not use Python 3.6 and should install a newer Python version inside the container, since Python 3.6 is officially EOL. I don't think most packages work with it anymore. On Frontier, for example, I always load a newer Python module (usually Python 3.11).

```
64/68 Test #64: ComputePerturbations ...............***Failed    0.50 sec
Traceback (most recent call last):
  File "/usr/lib64/python3.6/site-packages/h5py/_hl/dataset.py", line 15, in <module>
    from functools import cached_property
ImportError: cannot import name 'cached_property'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/data/mash/cche/azp-agent-in-docker-moth/azp-agent-moth/_work/1/s/src/turbulence/perturbation.py", line 1, in <module>
    import h5py
  File "/usr/lib64/python3.6/site-packages/h5py/__init__.py", line 60, in <module>
    from ._hl.files import (
  File "/usr/lib64/python3.6/site-packages/h5py/_hl/files.py", line 20, in <module>
    from .group import Group
  File "/usr/lib64/python3.6/site-packages/h5py/_hl/group.py", line 23, in <module>
    from . import dataset
  File "/usr/lib64/python3.6/site-packages/h5py/_hl/dataset.py", line 17, in <module>
    from cached_property import cached_property
ModuleNotFoundError: No module named 'cached_property'
```

BenWibking:
Since the system Python version is so old in the Frontier container (and on Frontier itself), I think it's probably best to install a newer Python with, e.g., uv: https://docs.astral.sh/uv/guides/install-python/#installing-a-specific-version.

.devcontainer/rocm-azp-agent-in-docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    useradd -m -d /home/agent -s /bin/bash -g agent agent && \
+    usermod -a -G users agent && \
+    chown -R agent:agent /azp /home/agent
+
+# Switch to agent user
+WORKDIR /home/agent
+USER agent
+# Another option is to run the agent as root.
+# ENV AGENT_ALLOW_RUNASROOT="true"
+
```
BenWibking, line 56: Install Python here, e.g.: `uv python install 3.12`. Then we can install packages with `uv pip install h5py` etc. The only downside is we have to run python scripts with `uv run my_script.py` instead of `python3 my_script.py`.

BenWibking: But we could also create a wrapper script called `python3` that actually calls `uv run` and put it first in `PATH`, so that it intercepts any shell (or CMake) commands that run python scripts.

BenWibking (changes requested):
Looks good except for the moth test failure due to the old Python version. See my recommended solution in the comments.

.devcontainer/rocm-azp-agent-in-docker/Dockerfile
chongchonghe: That would add uv dependency to the whole quokka code, because we have to change 'python' to 'uv run' everywhere. Not sure if that's an issue, though, because we don't use `python script.py` in the source code directly.

chongchonghe: I added installation of python3.11 via zypper and set /usr/bin/python3.11 as system default. That should solve the issue without changing anything else.

BenWibking: Did you push this change? I don't see it on Github.

chongchonghe: Just finished rebuilding the image and pushed the change. Now it's running again and hopefully the ROCm test will pass.

chongchonghe:
Some test failed with the following error. I set python3.11 the default python3 [here](https://github.com/quokka-astro/quokka/blob/4bb33ce4aefcb6c96bb3c5bdc5401ab05994995f/.devcontainer/rocm-azp-agent-in-docker/Dockerfile#L26), which probably has caused quokka to fail to locate the matplotlib library. 

```
time = 1.032102238e+14, Mstar = 2.005371975e+34
Mdot_exact = 6.220300533e+18
Steady state Mdot_sim = 1.324641816e+18
rel_diff = 0.7870453672
ModuleNotFoundError: No module named 'matplotlib'
  what():  Error loading module matplotlib!
SIGABRT
Abort(6) on node 0 (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 6) - process 0
```

I set

BenWibking (approved):

PR #1017, opened 2025-04-26
A novel scheme for sink particle accretion -- framework

chongchonghe:
### Description

A novel scheme for sink particle accretion. We use a buffer to store the accretion rate on cells, and apply a limiter to prevent negative mass in hydro cells before we apply the accretion rate to particles and the hydro state. This scheme eliminates the need for communication between particles and the need to find neighboring particles. We support an accretion radius of 3 dx with 4 ghost cells (since a particle can move one box size in a time step.)

We conduct two stages of accretion. In stage one, we compute the raw accretion rate on each cell from all particles and store it in a MultiFab `accretion_rate`. A `accretion_rate.SumBoundary` is called to sum the boundary values to the real cells.

In stage two, we do the following steps:
1. Apply a limiter to prevent negative mass in hydro cells by computing a `scale_factor` MultiFab. At the same time, we apply the rescaling to `accretion_rate`. Then, a `scale_down.FillBoundary` is called to fill the real values to the boundary cells.
2. Loop through all particles again to compute the accretion rate on each particle. For each particle, we loop through all cells in the accretion zone and add the accreted mass and momentum multiplied by the `scale_factor` to the particle. 
3. Subtract `accretion_rate` from hydro state.  

In between stage one and stage two, we can apply sink particle formation subject to limitation by the accretion rate. 

I've implemented a very simple threshold-based accretion method, similar to Federrath (2010). In every time step, a sink particle accretes $0.5 (\rho - \rho_{\rm th}) \Delta x$ amount of mass from each cell in its fixed accretion zone. A limiter is applied to prevent cell density from dropping by more than 90%. 

I have added a new test, `test_particle_sink`, to demonstrate the accretion scheme. The following figure shows the accretion zones of 40 sink particles randomly distributed in the box. I use the simplest threshold accretion scheme for this test, and the accretion zone is set to a cube for better visualization. The white lines are MPI box boundaries. As we can see, sink particle accretion zones can overlap through Fab boundaries or periodic domain boundaries. 

40sink-lev1-plt00001_Slice_z_n_H

The `test_particle_sink` test includes quantitative validation of both mass conservation and accretion rates. The test compares simulation results against analytical solutions. A plot of the gas density along the central horizontal line of the box after one accretion step shows that the numerical solution matches the analytical solution to machine precision, validating the accuracy of the accretion implementation.

</img></img>

### Related issues

### Checklist


chongchonghe:
@markkrumholz @BenWibking 
I plan to have two PRs for sink particle accretion. This first PR lays down the communication framework for our new sink accretion methods using a buffer state and a limiter. In the next PR, I'll implement a better accretion scheme like Krumholz et al. (2004) or [Lee et al. (2014)](https://iopscience.iop.org/article/10.1088/0004-637X/783/1/50).

This PR is well tested and ready for your review.

src/particles/particle_accretion.hpp
```diff
@@ -0,0 +1,282 @@
...
+						// use atomic operation to avoid race conditions
+						amrex::Gpu::Atomic::AddNoRet(&local_accretion_rate(ix + ii, iy + jj, iz + kk), delta_rho);
+					}
+				}
+			}
+		}
+	});
+}
+
+// Compute the scale down factor for the accretion rate. This is used to prevent accretion rates from exceeding 100% of the available mass.
```
markkrumholz, line 86: Just a note that a fixed 90% floor is fine for testing purposes, but we probably want to revisit this number, and probably make it a run-time parameter in the future. I would certainly worry that allowing a 90% reduction in mass in a single time step might lead to artificial sound waves being generated due to the sudden change in pressure.

src/particles/particle_accretion.hpp
```diff
@@ -0,0 +1,282 @@
...
+	const amrex::Long np = pti.numParticles();
+
+	// make a copy of kernel_weights for device
+	const auto kernel_weights_d = kernel_weights;
+
+	// Deposit particle data into the local buffer
+	amrex::ParallelFor(np, [=] AMREX_GPU_DEVICE(int64_t idx) {
+		auto &p = pData[idx]; // NOLINT(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+
+		// Check if this is a supernova progenitor
```
markkrumholz: I'm confused how this interacts with supernova progenitors. We might well want to use sink particles in a calculation where the sinks don't represent stars at all, and the notion of stellar evolution does not apply. Can you clarify what is going on here?

src/particles/particle_accretion.hpp
chongchonghe: Sounds good. I'll add a TODO here. For a simple test, a fixed number is convenient because it does not require computing local properties like Jeans's length, which may require access to ghost cells. This is definitely necessary and I'll add it in the next PR.

src/particles/particle_accretion.hpp
chongchonghe: We need to check if a particle accretes or not. Here I assume sink particles are star particles and they have an evolutionStage, and only LowMassStar and SNProgenitor accrete (dead stars do not). We do want to determine if a particle accretes on a particle-by-particle basis, so the particles must carry a trait field with integer component. This is convenient if all sink particles are stars and we can use StellarEvolutionStage for this. If a sink particle is not a star, we'll need another particle type with a different integer component, and we check for different integer components depending on the particle type. It helps for design consideration: can you give an example of an accreting non-star particle?

markkrumholz: Sure: a black hole particle. Or a planet.

In general I think the accretion field should be completely separated from the stellar evolution field. We want to allow star particles that do not accrete, accreting sink particles that are not stars, and arbitrary combinations of the two.

markkrumholz: I should also add: in general we will _not_ want the stars of the sort that @aditivijayan has implement to be accreting, because they are intended for a resolution such that we're not trying to follow the evolution of protostars as they accrete. In a star formation simulation on smaller scales we would not want to use that treatment, we will want a subgrid protostellar evolution model of the sort I implemented in Orion. So we certainly don't want to be forced into having accretion operations happen with the stars that @aditivijayan has implemented.

chongchonghe: A black hole is an evolutionary stage, unless we create it from files at the beginning of a simulation. 

> In general I think the accretion field should be completely separated from the stellar evolution field. We want to allow star particles that do not accrete, accreting sink particles that are not stars, and arbitrary combinations of the two.

OK. Makes sense. In that case, I'll create a second integer component to store whether or not a particle accretes, although this is only necessary for particle types that allow some particles to accrete and other particles not to accrete.

markkrumholz: I actually don't think that's the approach we want. Whether something accretes should be an aspect of the particle type, not something that can vary from particle to particle or change over the course of a simulation. I think this should be handled by a constexpr flag, the same way we declare whether a particular particle type allows creation or destruction.

chongchonghe: But why don't we allow a star to stop accretion when it is dead, like after SN explosion? We may have lots of dead stars or low-mass stars that we don't care about their accretion.

markkrumholz: I don't think we want to use accretion at all for the types of stars that are in our stochastic stellar pop treatment. The reason to follow accretion onto individual objects is when you are interested in knowing the mass distributions of those objects, for example when you're studying the IMF or black hole growth or planet formation. But in the case of the sorts of stars that are in the stochastic stellar pop treatment, we are taking the mass distribution as an input, not an output. So I would say we should declare the stochastic stellar pop stars are simply non-accreting, period.

chongchonghe: Oh. I'm not talking about stochastic stellar pop at all. I'm referring to sink particles.

chongchonghe: Although very massive stars (>8 Msun) are not very populous, there is a question of whether or not we want to disable accretion for some sink particles (like very low-mass stars whose accretion is inefficient).

markkrumholz: Can we have a quick real-time chat, either in person or by zoom? I feel like we're talking past each other.

chongchonghe: Sure. I'm working from home today, so let's meet on zoom.

chongchonghe: OK. Changes made. @markkrumholz

src/particles/particle_types.hpp
```diff
@@ -229,6 +231,38 @@ constexpr int TestParticleIntComps = 1; // stellar evolution stage
...
 template <typename problem_t> using TestParticleIterator = amrex::ParIter<TestParticleRealComps<problem_t>, TestParticleIntComps>;
 
+//-------------------- Sink particles --------------------
+
+// Indices for Sink_particles
+enum SinkParticleDataIdx {
+	SinkParticleMassIdx = 0,  // Mass of the particle
+	SinkParticleVxIdx,	  // Velocity in x direction
+	SinkParticleVyIdx,	  // Velocity in y direction
+	SinkParticleVzIdx,	  // Velocity in z direction
```
markkrumholz: I don't think sink particles should have birth or death times, or luminosities.

src/particles/particle_accretion.hpp
```diff
@@ -0,0 +1,264 @@
...
+
+	amrex::ParallelFor(state, [=] AMREX_GPU_DEVICE(int bx, int i, int j, int k) noexcept {
+		const double accretion_rate_cell = local_accretion_rate_arr[bx](i, j, k);
+		AMREX_ASSERT(accretion_rate_cell <= 0.0);
+		AMREX_ASSERT(accretion_rate_cell > -1.0);
+		const double accretion_down_factor = 1.0 + accretion_rate_cell;
+		state_arr[bx](i, j, k, HydroSystem<problem_t>::density_index) *= accretion_down_factor;
+		state_arr[bx](i, j, k, HydroSystem<problem_t>::x1Momentum_index) *= accretion_down_factor;
+		state_arr[bx](i, j, k, HydroSystem<problem_t>::x2Momentum_index) *= accretion_down_factor;
+		state_arr[bx](i, j, k, HydroSystem<problem_t>::x3Momentum_index) *= accretion_down_factor;
```
markkrumholz, line 208: You need to update the passive scalars by the same factor.

src/particles/particle_types.hpp
chongchonghe: Luminosity is necessary because we want radiation feedback from stars, right? It's set to 0 by default. If we want to update stellar luminosity based on its age, we need a birth time. Death time is unnecessary if we don't turn off luminosity or accretion over the course of a simulation.

src/particles/particle_accretion.hpp
chongchonghe: OK. I also forgot to update the energy. Now both energies and passive scalars are added.

src/particles/particle_types.hpp
markkrumholz: Sink particles themselves should not radiate. I see what you're implementing here as the basic sink particle, and we will then implement a separate class for radiating, accreting particles. I don't think it makes sense to have sink particles by themselves radiate, because there is no model attached to them to specify what the luminosity is or how it should be computed. There are generalizations of sink particles that will have this, but what I see this PR as being about is a minimal sink particle that doesn't do anything but accrete.

chongchonghe: I thought our plan was to have the luminosity field there and leave it as zeros if we don't use them since it's only one double component and doesn't cost much memory. However, I do agree that that for simplicity we should have a minimal sink particle prescription with only mass and velocity fields. I'll make that change.

chongchonghe: I've made the change.

markkrumholz (approved):

PR #1045, opened 2025-05-06
Sink particle formation: a simple Jeans model

chongchonghe:
### Description

Implemented a simple Jeans model for sink particle formation. We use the Jeans criterion (Truelove et al. 1997) for sink particle formation:

$$
\rho_{\mathrm{J}} > J^2 \frac{\pi c_s^2}{G \Delta x^2},
$$

where $J = 0.25$ is the Jeans number. 

The complete algorithm for sink particle formtion is as follows: 

1. Apply sink accretion, which returns an accretion rate at each cell. 
2. Apply sink formation. We create a sink particle with a mass of $m = (\rho - \rho_{\rm J}) \Delta x^3$ from a cell if the cell meets the following criteria 
   1. Not in an accretion zone of an existing sink particle.
   2. Density is above the local Jeans density.
   3. Density is local maximum within a sphere of radius `stencil_size * dx` . We choose `stencil_size = 3`. 

This algorithm avoids the need for merging a newborn particle with an existing particle nearby, which is done in ORION.

This PR also introduced a test problem, `ParticleSinkFormation`, to validate combined sink particle formation and accretion. This test checks:

1. There is exactly one star forming in the first step.
2. Mass is conserved to machine precision during sink formation.
3. Mass is conserved to machine precision during sink accretion. 

### Related issues

None.

### Checklist


pakshingli:
Thank you!

 

PS

 

From: ***@***.*** ***@***.***> On Behalf Of ChongChong He
Sent: Tuesday, May 6, 2025 1:21 PM
To: quokka-astro/quokka ***@***.***>
Cc: PS Li ***@***.***>; Mention ***@***.***>
Subject: [quokka-astro/quokka] [WIP] A framework for sink particle formation (PR #1045)

 


Description


Created an empty particle formation specialization for sink particles and a new test problem, ParticleSinkFormation, to validate sink formation recipe.

You can run the ParticleSinkFormation test with

ninja -j8 test_particle_sink_formation
ninja src/problems/ParticleSinkFormation/test

The test problem checks the number of stars formed at the first step and also checks for mass conservation. It will fail if no stars form or mass is not conserved.

@pakshingli <https://github.com/pakshingli>  To write your implementation of sink particle formation, update the ParticleCreationTraits<ParticleType::Sink> from line 165 of src/particles/particle_creation.hpp.


Related issues


Checklist


Before this pull request can be reviewed, all of these tasks should be completed. Denote completed tasks with an x inside the square brackets [ ] in the Markdown source below:

*	[ ] I have added a description (see above).
*	[ ] I have added a link to any related issues (if applicable; see above).
*	[ ] I have read the Contributing Guide <https://github.com/quokka-astro/quokka/blob/development/CONTRIBUTING.md> .
*	[ ] I have added tests for any new physics that this PR adds to the code.
*	[ ] (For quokka-astro org members) I have manually triggered the GPU tests with the magic comment /azp run.

  _____  


You can view, comment on, or merge this pull request online at:


  https://github.com/quokka-astro/quokka/pull/1045


Commit Summary


*	dabc201 <https://github.com/quokka-astro/quokka/pull/1045/commits/dabc2013de0c845e92e9b34846fd21f92f01a477>  merge in ParticleSink test from chong/particles/SF_test_v3_agora
*	9fcb5d2 <https://github.com/quokka-astro/quokka/pull/1045/commits/9fcb5d242d06b10975656388b80b5a82c727984f>  setup sink_formation test; passing
*	2653289 <https://github.com/quokka-astro/quokka/pull/1045/commits/265328977b3b7f9efb19ebf059a21e0961a8616d>  move createParticlesFromState
*	26a6b14 <https://github.com/quokka-astro/quokka/pull/1045/commits/26a6b145228ef1c16eb0c67bac03e9136c83ca66>  sink creation specialization
*	de59268 <https://github.com/quokka-astro/quokka/pull/1045/commits/de5926888483879d8b8a4869886005e43c7ad0bb>  add accretion_rate argument to createParticlesFromState
*	a26e899 <https://github.com/quokka-astro/quokka/pull/1045/commits/a26e8995304dd3dce7ae457c4f92f04cc99885ea>  update in
*	3eb56eb <https://github.com/quokka-astro/quokka/pull/1045/commits/3eb56eb846f3b4a2ecc42bbcb4f77a569742feee>  update sink units
*	a544211 <https://github.com/quokka-astro/quokka/pull/1045/commits/a5442111aff479eb0e717921ba45377d9ea83c76>  add print mass
*	36d6bd0 <https://github.com/quokka-astro/quokka/pull/1045/commits/36d6bd046e88e96fad3283da5ef21795e6b1d228>  particle creation routine
*	e8a93da <https://github.com/quokka-astro/quokka/pull/1045/commits/e8a93dad52620472db997330bcd7780dc6b82d1c>  update particle test
*	2c5c7c2 <https://github.com/quokka-astro/quokka/pull/1045/commits/2c5c7c21c339564250853704721efe37ec25fa3a>  minor
*	23f3404 <https://github.com/quokka-astro/quokka/pull/1045/commits/23f340434b3a3786f3a788248c0a4dc09049e38b>  turn off sink accretion
*	91b411c <https://github.com/quokka-astro/quokka/pull/1045/commits/91b411c7340b2c9ecd87350da321dacd6ddc669c>  add status check
*	ad6506f <https://github.com/quokka-astro/quokka/pull/1045/commits/ad6506f50a81aa581b45e9b62c11610cf0873d44>  check num_stars
*	615b6bf <https://github.com/quokka-astro/quokka/pull/1045/commits/615b6bf1894d13ebaee36f8ddecdf9bdf655d947>  [pre-commit.ci] auto fixes from pre-commit.com hooks


File Changes 


(9 <https://github.com/quokka-astro/quokka/pull/1045/files>  files)

*	M src/particles/PhysicsParticles.hpp <https://github.com/quokka-astro/quokka/pull/1045/files#diff-2bb8ffdf40aa2dca173b41cf3f31ff9ef81ed7398567105b93ba3e4e4ffa00ea>  (47) 
*	M src/particles/particle_creation.hpp <https://github.com/quokka-astro/quokka/pull/1045/files#diff-8dac35795c329dc145295e51af81ae9d76c4c150f89af6e6ef700443c24e0ae3>  (146) 
*	M src/particles/particle_types.hpp <https://github.com/quokka-astro/quokka/pull/1045/files#diff-0aa73f02d4907eb570b7b70bcbdfed567ca2d5023e52b9b430d1c40950e947cc>  (9) 
*	M src/problems/CMakeLists.txt <https://github.com/quokka-astro/quokka/pull/1045/files#diff-ca50142f2a13da3e9d252faa8399823b6194b41847fc8fecf2c35380c7a02d20>  (1) 
*	A src/problems/ParticleSinkFormation/CMakeLists.txt <https://github.com/quokka-astro/quokka/pull/1045/files#diff-3cb5bb9535502efa6c79d5dd73b2a649b3fc06951f102dcfb5dc07a99478b213>  (8) 
*	A src/problems/ParticleSinkFormation/test_particle_sink_formation.cpp <https://github.com/quokka-astro/quokka/pull/1045/files#diff-5bd2341fb50fd26e910c6f2b2356f8e6e35d6803f845297e7926279edf25d6ce>  (195) 
*	A src/problems/ParticleSinkFormation/test_particle_sink_formation.hpp <https://github.com/quokka-astro/quokka/pull/1045/files#diff-775a76c19915484fd1248adbb7e22a7f83153a28f4f09faa84955e91d3d0dfe4>  (12) 
*	M src/simulation.hpp <https://github.com/quokka-astro/quokka/pull/1045/files#diff-1643e903f8d7916aa711335f79a095b83a7aa844db8dd37ba137dbc62a5eafbe>  (6) 
*	A tests/ParticleSinkFormation.in <https://github.com/quokka-astro/quokka/pull/1045/files#diff-3de01b5518decf81f645c08ad5ed8cb017a520e10d23915e84257c054d4c4a4b>  (41) 


Patch Links:


*	https://github.com/quokka-astro/quokka/pull/1045.patch
*	https://github.com/quokka-astro/quokka/pull/1045.diff

chongchonghe:
@pakshingli  I have implemented a simplest model for sink particle formation: a particle is formed in a cell if the density is above a threshold, and the particle takes the mass and momentum of the gas above the threshold. You may start from there and add your implementations. You can make a copy of this branch so it becomes your branch, but you can also directly work on this branch.

chongchonghe:
@pakshingli It would be helpful if you could push your changes to this PR (or create your PR) so I can test it on my computer.

chongchonghe:
@pakshingli  @markkrumholz @BenWibking 
We need to think about how to deal with multiple cells having the same density in a formation zone. This can happen, for example, in a initial condition where the gas density is uniform and all cells are Jeans unstable. We need to come up with a algorithm that is reproducible and independent of the order you process the cells. 

I can think of a simple routine where when multiple cells have the same density, we compare the indices of the cells, i.e. largest in `i`, then largest in `j`, then largest in `k`. This choice is arbitrary and non-physical, though.

markkrumholz:
> @pakshingli @markkrumholz @BenWibking We need to think about how to deal with multiple cells having the same density in a formation zone. This can happen, for example, in a initial condition where the gas density is uniform and all cells are Jeans unstable. We need to come up with a algorithm that is reproducible and independent of the order you process the cells.
> 
> I can think of a simple routine where when multiple cells have the same density, we compare the indices of the cells, i.e. largest in `i`, then largest in `j`, then largest in `k`. This choice is arbitrary and non-physical, though.

I agree that we can, when comparing cells, break ties where the densities are exactly equal by taking the higher density to be the cell with the lower index. In practice I think this will never, ever occur outside of artificial initial conditions that are specifically set up in this way, since in a real calculation the odds of two cells having identical densities to double precision are basically zero.

pakshingli:
Yes.  It is extremely rare to have exactly the same density of two cells, probably only happens in a problem set up.  But if we want to deal with that also, we can find a mean position for the cells that have exactly the same max density and that cell is the one to create sink particle.  However, that would still depend on the sequence of looping over the grid.

chongchonghe:
> Yes. It is extremely rare to have exactly the same density of two cells, probably only happens in a problem set up. But if we want to deal with that also, we can find a mean position for the cells that have exactly the same max density and that cell is the one to create sink particle. However, that would still depend on the sequence of looping over the grid.

I think an arbitrary choice like the lowest or highest indices (in the order of i, j, k) is just fine, since it only matters in artificial initial conditions. This is probably the only simple way to have a deterministic result independent of the sequence you process the cells and avoid creating multiple stars within 3 cells of distance. 

Average position of all the cells with maximum density won't work because that will depend on the covering of the formation zone. Imagine an initial condition with uniform density and all cells are Jeans unstable. An 'average position' algorithm will still determine that all cells (except for cells that are within 3 cells away from the domain boundary) will form stars. 

 @pakshingli Can you implement this?

chongchonghe:
As I gave it more thought, I am convinced that an algorithm that avoids creating multiple stars within 4 cells in a uniform initial condition does NOT exist. I think we can simply ignore this and keep the algorithm as is, and avoid starting a test problem with homogeneous initial conditions by adding an arbitrary perturbation.

markkrumholz:
Yes, I think you're right. After all, we can come up with a pathological case of initializing a problem where an entire domain is at exactly the same Jeans-violating density, over such an extent that the cells are spread across an arbitrary number of MPI ranks. No local algorithm will do that "right".

chongchonghe:
@BenWibking @markkrumholz @pakshingli This PR is done and ready for your review.

chongchonghe:
@pakshingli I made one minor change from your original implementation: I removed the hard-coded temperature floor in the calculation of Jeans density because we use `hydro_system::EnforceLimits` (by setting `temperature_floor` at runtime) to enforce a temperature floor. See discussions in #1072 . I also merged #1092, which fixed a bug in the accretion scheme implementation that caused mass conservation to fail.

chongchonghe:
Sink particle formation and accretion are thoroughly tested and all tests passed. It's ready for your review. @markkrumholz @BenWibking

chongchonghe:
@markkrumholz Can you review this PR?

markkrumholz (approved):
LGTM. The only potential issue I see is that right now the Jeans number is hard-coded, and we probably want to change that in future, because there are applications where we will want a value other than J = 1/4 -- for example, for first star disc dynamo simulations, the standard is J = 1/32. However, that can be addressed by a future PR.

PR #1055, opened 2025-05-09
Implement the Krumholz 2004 model for sink particle accretion

chongchonghe:
### Description

We first compute the particle's Bondi-Hoyle radius,

$$
r_{\mathrm{BH}}=\frac{G M}{v_{\infty}^2+c_{\infty}^2},
$$

where $M$ is the particle's mass and $v_{\infty}$ and $c_{\infty}$ are the velocity and sound speed of the gas far from the sink particle. We compute `v_infty` as the mass-weighted mean velocity relative to the particle inside the accretion zone and `c_infty` as the mass-weighted mean sound speed inside the accretion zone. 

Then, we compute the Bondi-Hoyle-Lyttleton accretion rate: 

$$
\dot{M}=4 \pi \rho_{\infty} G^2 M^2\left[\frac{\lambda^2 c_{\infty}^2+v_{\infty}^2}{\left(c_{\infty}^2+v_{\infty}^2\right)^4}\right]^{1 / 2}=4 \pi \rho_{\infty} r_{\mathrm{BH}}^2\left(\lambda^2 c_{\infty}^2+v_{\infty}^2\right)^{1 / 2}
$$

where $\rho_{\infty}$ is the the gas density far from the sink particle. We compute `rho_infty` as the mean density inside the accretion zone. 

Last, we apply an accretion kernel on the accretion rate from each cell. We assign each cell a weight

$$
w = \exp \left(-r^2 / r_K^2\right),
$$

where $r$ is the distance from the particle to the cell center; cells outside of the accretion zone have a weight of 0. $r_K$ is the kernel radius 

$$
r_K= \begin{cases}\Delta x / 4 & r_{\mathrm{BH}}<\Delta x / 4, \\ r_{\mathrm{BH}} & \Delta x / 4 \leq r_{\mathrm{BH}} \leq r_{\mathrm{acc}} / 2, \\ r_{\mathrm{acc}} / 2 & r_{\mathrm{BH}}>r_{\mathrm{acc}} / 2\end{cases} 
$$

The accretion rate from each cell is $\dot{M} w_i / \sum_i w_i$. 

We further apply two limitations/corrections to the accretion rate in each cell, in the following order:
1. We set a cap that no more than 25% (Krumholz 2004) of the mass may be removed from a cell in any single time step.
2. If the density of a cell is above Jeans density, $\rho_{\mathrm{J}} = J^2 \frac{\pi c_s^2}{G \Delta x^2}$ where $J=0.25$, we bring down the density to $\rho_J$. 

#### Test problem

Based on the test problem introduced in #1017 . It's simple to derive an exact solution for the Bondi-Hoyle accretion from four particles in one step, if we use a uniform, cubic accretion kernel (only for testing purpose). The numerical result is in perfect agreement with the analytic one:

</img>

### Related issues

### Checklist


markkrumholz:
I think this recipe is basically fine.

Regarding what happens if you're above the Jeans threshold, I would go ahead and accrete the extra mass to keep below Jeans. Here's my reasoning: the "no more than 25%" rule is intended to avoid rapidly changing the pressure in a way that generates artificial sound waves, and this is necessary when, for example, we have a rapidly accreting massive particle moving through the grid. So the reason for the rule is to prevent launching artificial sound waves. However, the only situation where the "no more than 25%" rule and the "remove enough mass to avoid going above the Jeans density" rule can come into conflict is when there is a converging flow into a cell that is capable of increasing its density by more than 25% in a single time step, and the only way that this could be the case is if the cell in question is at the center of a highly supersonic convergence -- given the CFL condition, there's no way for a subsonic or even trans-sonic flow to increase the density in a cell that fast. A cell that is at the center of such a supersonic convergence is causally disconnected from the surrounding domain. It's not launching sound waves anywhere. Thus there is no need to limit the accretion to 25% of the mass in a cell that capable of having its mass increased by 25% in a single time step, because such cells cannot perturb those around them.

chongchonghe:
@markkrumholz @BenWibking This PR is ready for a review.

src/particles/particle_accretion.hpp
```diff
@@ -41,131 +36,279 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto compute_rho_sink(const amrex::Arra
...
+	// compute the average density, momentum, and sound speed in the accretion zone
+	double rho_infty = 0.0;
+	double px_infty = 0.0;
+	double py_infty = 0.0;
+	double pz_infty = 0.0;
+	double cs_infty = 0.0;
+	for (int ii = ix - stencil_size; ii <= ix + stencil_size; ++ii) {
+		for (int jj = iy - stencil_size; jj <= iy + stencil_size; ++jj) {
+			for (int kk = iz - stencil_size; kk <= iz + stencil_size; ++kk) {
+				const double weight = kernel_weights_normalized[std::abs(ii - ix)][std::abs(jj - iy)][std::abs(kk - iz)];
```
markkrumholz, line 53: It looks here like you're computing the weight from the difference in cell center positions, not from the position of the sink particle. I would prefer the latter, and that is what is implemented in the original 2004 paper.

src/particles/particle_accretion.hpp
```diff
@@ -41,131 +36,279 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto compute_rho_sink(const amrex::Arra
...
+					//-----------------------------------------------------------------------------------------------------
 				}
 			}
 		}
+
 		// the accretion rates are negative, so we 'subtract' them
 		p.rdata(mass_index) -= accreted_mass;
-		p.rdata(mass_index + 1) -= accreted_momentum_x;
-		p.rdata(mass_index + 2) -= accreted_momentum_y;
-		p.rdata(mass_index + 3) -= accreted_momentum_z;
```
markkrumholz, line 168: I'm confused what is is going on here. Are you tracking particle momentum or velocity? This seems consistent with neither. If it's particle momentum you're tracking then you should just be adding `accreted_momentum_x` and similarly for y and z, and if you're tracking particle velocity then the correct statement is

`p.rdata(mass_index + 1) -= accreted_momentum_x / p.rdata(mass_index)`

because that is the correct update to conserve total momentum.

markkrumholz (commented):
I left one specific comment in the code where I think there is an error, but there is also a broader design choice I'd like to discuss: in computing weights for cells, you multiply by a constant array `kernel_spherical_3_weights_normalized`, which as far as I can tell is just an array that contains (up to the normalization) the fraction of each cell's volume that lies inside the spherical kernel. However, the way this has been computed is for a spherical kernel that is centered at a cell center, not at the location of the particle, which is different for every particle. Thus I don't think that this step makes sense, and I would advocate for not multiplying by this factor, and just ignoring the fact that the kernel isn't exactly spherical, it's an approximation to a sphere formed out of cubes. Thus I would prefer that we delete all the multiplications by `kernel_spherical_3_weights_normalized`. This will also have the virtue of making the code simpler and easier to read.

Finally, I'd suggest moving the added clang_tidy stuff in this PR into a separate one, since it's not really related to sink particles.

markkrumholz:
I'd also suggest that, perhaps not as part of this PR but maybe in a subsequent one, we implement some more of the standard tests from my 2004 paper, just to prove that the sink particle accretion code works in non-trivial problems.

src/particles/particle_accretion.hpp
chongchonghe: This seems to be outdated code. Are you commenting on the most recent commit? That's an obvious mistake, but I corrected it last night.

chongchonghe:
> I'd also suggest that, perhaps not as part of this PR but maybe in a subsequent one, we implement some more of the standard tests from my 2004 paper, just to prove that the sink particle accretion code works in non-trivial problems.

Yes, I'm working on that.

src/particles/particle_accretion.hpp
chongchonghe: Oh. There is still a mistake in the recent commit. I'll fix that.

chongchonghe:
> Finally, I'd suggest moving the added clang_tidy stuff in this PR into a separate one, since it's not really related to sink particles.

I'll merge in the dev branch once the tidy PR is merged.

chongchonghe:
> I left one specific comment in the code where I think there is an error, but there is also a broader design choice I'd like to discuss: in computing weights for cells, you multiply by a constant array `kernel_spherical_3_weights_normalized`, which as far as I can tell is just an array that contains (up to the normalization) the fraction of each cell's volume that lies inside the spherical kernel. However, the way this has been computed is for a spherical kernel that is centered at a cell center, not at the location of the particle, which is different for every particle. Thus I don't think that this step makes sense, and I would advocate for not multiplying by this factor, and just ignoring the fact that the kernel isn't exactly spherical, it's an approximation to a sphere formed out of cubes. Thus I would prefer that we delete all the multiplications by `kernel_spherical_3_weights_normalized`. This will also have the virtue of making the code simpler and easier to read.

Unlike the SN prescription where we use a uniform kernel in a sphere, we are using a kernel function `~ exp(-r^2/r_K^2)` for accretion. So, I agree it's fine to use an approximate sphere formed by cubes.

src/particles/particle_accretion.hpp
markkrumholz: OK, looks good now.

markkrumholz:
OK, let me know once you have made the edits to clang_tidy and the weights and then I can re-review.

chongchonghe:
> OK, let me know once you have made the edits to clang_tidy and the weights and then I can re-review.

OK, all done.

markkrumholz (approved):
Looks good -- ready to merge.

PR #1057, opened 2025-05-10
add example SLURM script to cancel job when I/O hangs

BenWibking:
### Description
This SLURM script provides an example of how to automatically kill batch jobs that stop writing output to stdout/stderr after 5 minutes. This is useful when the filesystem is semi-broken and there are stochastic hangs when writing output files.

I have manually tested this on DeltaAI.

### Related issues
N/A

### Checklist


BenWibking:
~~On at least DeltaAI, there are some cases where the file size metadata is not updated, even though the job is still writing to the logfile. So then the job is killed spuriously.~~

Fixed

chongchonghe (approved):

chongchonghe:
@BenWibking Is there a way to ask a running Quokka process to immediately write a checkpoint? For example, a user defined kill code: `scancel --signal=10 <jobid>` . This is helpful when you have run a simulation for a long time since last checkpoint and you have to kill it. If this works, you can integrate it in your script.

BenWibking:
> @BenWibking Is there a way to ask a running Quokka process to immediately write a checkpoint? For example, a user defined kill code: `scancel --signal=10 <jobid>` . This is helpful when you have run a simulation for a long time since last checkpoint and you have to kill it. If this works, you can integrate it in your script.

If it's killed because it's stuck writing an output file (which is the most likely reason), this doesn't make sense.

chongchonghe:
> > @BenWibking Is there a way to ask a running Quokka process to immediately write a checkpoint? For example, a user defined kill code: `scancel --signal=10 <jobid>` . This is helpful when you have run a simulation for a long time since last checkpoint and you have to kill it. If this works, you can integrate it in your script.
> 
> If it's killed because it's stuck writing an output file (which is the most likely reason), this doesn't make sense.

@BenWibking  Right. But, independent of I/O failure, is there a way to trigger an checkpoint from an running process?

BenWibking:
> > > @BenWibking Is there a way to ask a running Quokka process to immediately write a checkpoint? For example, a user defined kill code: `scancel --signal=10 <jobid>` . This is helpful when you have run a simulation for a long time since last checkpoint and you have to kill it. If this works, you can integrate it in your script.
> > 
> > 
> > If it's killed because it's stuck writing an output file (which is the most likely reason), this doesn't make sense.
> 
> @BenWibking Right. But, independent of I/O failure, is there a way to trigger an checkpoint from an running process?

No. We could add it.

PR #1066, opened 2025-05-12
Change output behaviour 

chongchonghe:
### Description

As an example, let's run the ParticleCreation test, which has these time stamps: 0, 0.001, 0.002, 0.003, 0.004.  

If we set `plotfile_interval = 3` and `plottime_interval = 0.003` 
   1. Previously: `plt00003` will be written twice, because both `plotfile_interval` and `plottime_interval` checks will result in writing a plotfile. 
   2. Now: `plt00003` will be written once. 

If we set `plottime_interval = 0.003` 
   1. Previously: Will write the following outputs: 1, 3
   2. Now: Will write the following outputs: 0, 3, 4, i.e. will write the first (0) and last steps along with steps whose time is a multiple of `plottime_interval`. 

### Related issues

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -977,6 +977,8 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 
 	double next_plot_file_time = 0;
 	if (plotTimeInterval_ > 0) {
+		// We have one plotfile at the start of the simulation, so we set next_plot_file_time to plotTimeInterval_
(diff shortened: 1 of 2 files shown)


BenWibking (approved):
This is more consistent. For long running simulations that will be restarted many times, it might also be useful to have an option to disable the final plotfile, in order to avoid having to filter it out later.

PR #1072, opened 2025-05-13
Add Bondi accretion test from Krumholz 2004

chongchonghe:
### Description

Add the Bondi accretion test from Krumholz 2004. 

### Related issues

### Checklist


chongchonghe:
However, the accretion rate is not exactly right. It's ~2 times larger than the theoretical Bondi accretion rate:

particle_mass

chongchonghe:
With the correct background density, the accretion rate from the simulation agrees well with the theoretical Bondi rate.

</img>

</img></img>

</img></img>

markkrumholz:
Looks like there are some GPU issues and so the PR will need to be modified slightly, but in general this looks fine. Let me know once the build issues are resolved and I will approve.

chongchonghe:
> Looks like there are some GPU issues and so the PR will need to be modified slightly, but in general this looks fine. Let me know once the build issues are resolved and I will approve.

Some non-trivial issues. I'll fix it this afternoon and let you know once done.

chongchonghe:
Looks good now. @markkrumholz

src/particles/particle_accretion.hpp
```diff
@@ -23,12 +23,13 @@ namespace SinkAccretionUtils
 {
 
 constexpr int stencil_size = quokka::ParticleUtils::stencil_size;
+static constexpr double cs_floor_ = 1.33e4; // = 0.133 km/s = sqrt(k_B * 5 K / (2.33 m_p))
```
markkrumholz: I would prefer that we not hard wire in a separate floor here. We have a tempFloor already defined in hydro_system, and I think we should rely on that rather than setting a separate floor here. Just call hydro_system::enforceLimits before calling sink creation.

src/particles/particle_accretion.hpp
```diff
@@ -23,12 +23,13 @@ namespace SinkAccretionUtils
...
 
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto get_delta_rho(double rho, double rho_sink) -> double { return -0.5 * (rho - rho_sink) / rho; }
 
 template <typename problem_t>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto compute_Mdot_and_r_K(const amrex::Array4<const amrex::Real> &local_state, int ix, int iy, int iz, double par_mass,
-								   double par_x, double par_y, double par_z,
+								   double par_x, double par_y, double par_z, double cs_floor,
 								   const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> &plo,
```
markkrumholz: See my previous comment.

src/particles/particle_accretion.hpp
```diff
@@ -66,7 +67,11 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto compute_Mdot_and_r_K(const amrex::
 				const double px = local_state(ii, jj, kk, HydroSystem<problem_t>::x1Momentum_index);
 				const double py = local_state(ii, jj, kk, HydroSystem<problem_t>::x2Momentum_index);
 				const double pz = local_state(ii, jj, kk, HydroSystem<problem_t>::x3Momentum_index);
-				const double cs = HydroSystem<problem_t>::ComputeSoundSpeed(local_state, ii, jj, kk);
+				double cs = HydroSystem<problem_t>::ComputeSoundSpeed(local_state, ii, jj, kk);
+				if constexpr (quokka::EOS_Traits<problem_t>::gamma == 1.0) {
+					cs = quokka::EOS_Traits<problem_t>::cs_isothermal;
+				}
+				cs = std::max(cs, cs_floor);
```
markkrumholz: See previous comment about floors.

markkrumholz (changes requested):
It looks like you have added in a bunch of floors on the sound speed, but at present these are hard-wired in. I don't think this is a good idea. We already have a temperature flooring capability, which is set up to be applied in hydro_system::EnforceLimits. I don't think we should hard-wire in a different floor here, since this creates potentially hard to diagnose conflicts in the future. Instead, I think the best way to handle this is just to apply the floors, by calling EnforceLimits, before you start the sink particle creation routine.

chongchonghe:
> It looks like you have added in a bunch of floors on the sound speed, but at present these are hard-wired in. I don't think this is a good idea. We already have a temperature flooring capability, which is set up to be applied in hydro_system::EnforceLimits. I don't think we should hard-wire in a different floor here, since this creates potentially hard to diagnose conflicts in the future. Instead, I think the best way to handle this is just to apply the floors, by calling EnforceLimits, before you start the sink particle creation routine.

@markkrumholz 
OK, I've made the change and removed the hard-wired `cs_floor`. Since `EnforceLimits` is already called every time step, all I have to do is to set `sim.tempFloor_ = 10.0;` in the test problem.

markkrumholz (dismissed):

chongchonghe:
I set `sim.tempFloor_ = 10.0;` in test_particle_sink.cpp but forgot to set it in test_particle_accretion.cpp. Now it's updated, and I also renamed the test problem for consistency.  @markkrumholz Can you approve this again. I'll merge it once the GPU tests pass.

markkrumholz (approved):

PR #1075, opened 2025-05-16
Fix all clang-tidy warnings in simulation.hpp

chongchonghe:
### Description

`simulation.hpp` is tidy-free now. Any future edits to `simulation.hpp` will not trigger a clangt-tidy warning unless it's new in that change. 

Also updated `tidy.sh` to support `--fix` option to apply automatic fix. 

### Related issues

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -1974,6 +1974,7 @@ void AMRSimulation<problem_t>::fillBoundaryConditions(amrex::MultiFab &S_filled,
 		for (auto &i : coarseData) {
 			AMREX_ASSERT(!i->contains_nan(0, state.nComp()));
 			AMREX_ASSERT(!i->contains_nan()); // check ghost zones
+			amrex::ignore_unused(i);
 		}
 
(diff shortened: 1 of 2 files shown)


src/simulation.hpp
```diff
@@ -524,7 +524,7 @@ template <typename problem_t> auto AMRSimulation<problem_t>::getNewMF_fc() const
 
 template <typename problem_t> void AMRSimulation<problem_t>::initialize()
 {
-	BL_PROFILE("AMRSimulation::initialize()");
+	BL_PROFILE("AMRSimulation::initialize()"); // NOLINT(misc-const-correctness)
```
BenWibking, line 527: A better way to fix these is to just add `const` before `BL_PROFILE` like this:
```
const BL_PROFILE("AMRSimulation::initialize()");
```
This is done in some places in the code.

chongchonghe: OK. I've changed them all.

chongchonghe: However, this change caused SonarCloud Code Analysis to fail. It caused “declaration does not declare anything” in every instance of `BL_PROFILE`. I'll have to revert them. I think this is just fine. We don't have to spend time on this.

chongchonghe:
@markkrumholz Can you approve this PR?

BenWibking (approved):

PR #1100, opened 2025-05-26
Refactor restart code and add universal refinement option

BenWibking:
### Description
Refactors the entire checkpoint reading code, and on restart:
* Optionally, universally refines all levels with piecewise constant interpolation
* Chops grids on all levels in order to better load balance across all MPI ranks
* Splits particles (except for tracer particles) by a factor of  `pow(refine_factor, AMREX_SPACEDIM)`

The user can adjust `amr.n_cell` to be an integer multiple of the previous simulation. Then, the restart code will check that the new base grid size is the same integer multiple of the original grid in all dimensions, and then interpolate each level onto new refined grids.

Adds a new `CheckpointRestartRefine` test, which tests multi-level restarts with universal refinement using the HydroBlast3D problem with tracer particles.

*Warning:* **AMReX PC->Restart() does not work if there are more levels than in the checkpoint.** This is a bug in AMReX. For now, we detect this situation and abort.

### Related issues
N/A

### Checklist


BenWibking:
@chongchonghe Can you review this? This code is a bit tricky, so I would appreciate a close review.

BenWibking:
Well, this is hard to argue with:
```

⏺ PR Review: Universal Refine Feature

  Overview

  This PR implements a "universal refine" feature that allows restarting simulations with a different (higher) grid
   resolution than the original checkpoint. The feature automatically detects grid size differences and
  interpolates data from the coarser checkpoint to the finer restart grid.

  Key Changes

  Core Implementation (src/simulation.hpp):
  - Added restartRefineFactor_ member variable to track refinement ratio
  - Enhanced ReadCheckpointFile() to detect grid size mismatches between checkpoint and input
  - Implemented grid rescaling logic with validation that new grid is integer multiple of original
  - Added interpolation for both cell-centered and face-centered data using AMReX's InterpFromCoarseLevel

  Strengths

  1. Robust Error Checking: The code validates that the new grid dimensions are exact integer multiples of the
  checkpoint grid, preventing invalid configurations.
  2. Comprehensive Data Handling: Supports interpolation for both cell-centered and face-centered variables,
  maintaining physical consistency.
  3. Proper Geometry Handling: Creates appropriate coarse geometry objects for interpolation and handles boundary
  conditions correctly.
  4. Conservative Approach: Only applies interpolation when needed (restartRefineFactor_ > 1), using efficient
  ParallelCopy for same-resolution restarts.

  Technical Issues

  1. Missing Particle Support: The TODO comment at line 3210 indicates particle data splitting is not yet
  implemented, which could be problematic for simulations with particles.
  2. Limited Documentation: The commit messages are brief and don't explain the use case or testing approach.
  3. Hardcoded Assumptions: The code assumes uniform refinement in all spatial dimensions, which may not always be
  desired.

  Code Quality

  Positive:
  - Follows project conventions (camelCase with trailing underscore)
  - Uses appropriate AMReX interpolation functions
  - Includes boundary condition handling

  Areas for Improvement:
  - The logic is complex and could benefit from factoring into separate helper functions
  - Some variable names could be clearer (e.g., reDom vs restartDomain)

  Testing

  Concerns:
  - Only minor test file changes (adding checkpoint interval)
  - No apparent comprehensive tests for the universal refine functionality
  - Need verification that interpolated solutions maintain accuracy

  Recommendations

  1. Add comprehensive tests that verify:
    - Correct interpolation behavior
    - Conservation properties
    - Boundary condition handling
  2. Implement particle support or document limitations clearly
  3. Consider refactoring the restart logic into separate functions for readability
  4. Add documentation explaining when and how to use this feature

  Overall Assessment

  This is a valuable feature that addresses a real need for resolution studies and simulation continuation at
  higher resolution. The implementation is technically sound but needs better testing and documentation. The code
  quality is good but could benefit from some refactoring for maintainability.

  Recommendation: Approve with required changes for particle support and comprehensive testing.
```

chongchonghe:
I got this error while testing it with BinaryOrbit which has particles:

```bash
-> % ../build/build-3D-MPI/src/problems/BinaryOrbitCIC/binary_orbit ./BinaryOrbit.in max_timesteps=30 
Initializing AMReX (25.04-26-gd49d22968b40)...
AMReX (25.04-26-gd49d22968b40) initialized
Successfully read inputs file ... 
Using derived variables:
        gpot

Restart from checkpoint last_chk
Input grid dimensions on level 0: (64,64,64)
Restart file grid dimensions on level 0: (32,32,32)
Rescaling MultiFabs in restart file by a factor of 2...
Reading last_chk/metadata.yaml...
        quokka_version = 25.03
        git_hash_quokka = 958425d658d57a70ce3cae885729ade4f3fb3191
        git_hash_amrex = d49d22968b403bc978749b5e2feb7c94a7710158
        units has unknown type! skipping this entry.
        constants has unknown type! skipping this entry.
amrex::Abort::0::ParticleContainer::locateParticle(): invalid particle. !!!
SIGABRT
```

chongchonghe:
We need to properly deal with particles. I would recommend to deal with particle splitting independently from re-meshing. We could set the default particle splitting factor to be `(n_cell/n_cell_old)^3`. 

I created a new PR for particle splitting: #1104

BenWibking:
> I got this error while testing it with BinaryOrbit which has particles:
> 
> ```shell
> -> % ../build/build-3D-MPI/src/problems/BinaryOrbitCIC/binary_orbit ./BinaryOrbit.in max_timesteps=30 
> Initializing AMReX (25.04-26-gd49d22968b40)...
> MPI initialized with 1 MPI processes
> MPI initialized with thread support level 0
> AMReX (25.04-26-gd49d22968b40) initialized
> Successfully read inputs file ... 
> Using derived variables:
>         gpot
> 
> Restart from checkpoint last_chk
> Input grid dimensions on level 0: (64,64,64)
> Restart file grid dimensions on level 0: (32,32,32)
> Rescaling MultiFabs in restart file by a factor of 2...
> Reading last_chk/metadata.yaml...
>         quokka_version = 25.03
>         git_hash_quokka = 958425d658d57a70ce3cae885729ade4f3fb3191
>         git_hash_amrex = d49d22968b403bc978749b5e2feb7c94a7710158
>         units has unknown type! skipping this entry.
>         constants has unknown type! skipping this entry.
> amrex::Abort::0::ParticleContainer::locateParticle(): invalid particle. !!!
> SIGABRT
> ```

I'm looking into this bug. The `ParticleContainer->Restart()` function appears to assume the grids are unchanged after restart, which is not true. I will see if I can add a workaround.

BenWibking:
@chongchonghe Can you test this again manually? It should work now, but the refactoring is complex and I might have missed something.

BenWibking:
@chongchonghe I am getting crashes when restarting and re-reading the StochasticStellarPop container. Have you been able to restart a particle container that has zero particles? I think this may be an AMReX bug. However, we need to fix it ourselves, since I cannot run the weak scaling tests without a fix.

BenWibking:
> @chongchonghe I am getting crashes when restarting and re-reading the StochasticStellarPop container. Have you been able to restart a particle container that has zero particles? I think this may be an AMReX bug. However, we need to fix it ourselves, since I cannot run the weak scaling tests without a fix.

Ah, nevermind. I can just check whether there are any particles, and avoid calling `->Restart()` if there are zero particles.

chongchonghe:
Just tested. Looks pretty good.

chongchonghe:
Tested restarting the binary_orbit test with a factor of 2 refactoring. Looks good. Before and after super-resolution:

</img></img>

chongchonghe:
@BenWibking  However, the particles are NOT splitted. This line seems to suggest that you let the particles split by a factor of `split_factor^3`. However, in the plotfile, there are only 2 particles in the binary_orbit test after restarting with refactoring.
https://github.com/quokka-astro/quokka/blob/d0ada35abec2d453e3658e8fcf7983a65ed2ad27/src/simulation.hpp#L3427

chongchonghe:
I think we don't always want to split the particles by a factor of `factor^3`, because the number of particles/stars in a simulation usually does not scale with the number of hydro cells. So, I would recommend we do the following: when restarting with refactoring, split particles according to your previously defined parameter `particles.split_factor`.

BenWibking:
Hmm, this is very bizarre. When I test it with the Agora problem, I see that some particles are split, but not the right number:
```
In [1]: import yt

In [2]: ds = yt.load("plt00000/")
yt : [INFO     ] 2025-05-30 23:20:01,245 Parameters: current_time              = 0.0
yt : [INFO     ] 2025-05-30 23:20:01,245 Parameters: domain_dimensions         = [256 256 256]
yt : [INFO     ] 2025-05-30 23:20:01,245 Parameters: domain_left_edge          = [-1.85140655e+24 -1.85140655e+24 -1.85140655e+24]
yt : [INFO     ] 2025-05-30 23:20:01,245 Parameters: domain_right_edge         = [1.85140655e+24 1.85140655e+24 1.85140655e+24]

In [3]: ds.particle_type_counts
Out[3]: {'StochasticStellarPop_particles': 0, 'CIC_particles': 848247}

In [4]: ds = yt.load("plt00000.old.7664667/")
yt : [INFO     ] 2025-05-30 23:20:15,142 Parameters: current_time              = 0.0
yt : [INFO     ] 2025-05-30 23:20:15,142 Parameters: domain_dimensions         = [128 128 128]
yt : [INFO     ] 2025-05-30 23:20:15,142 Parameters: domain_left_edge          = [-1.85140655e+24 -1.85140655e+24 -1.85140655e+24]
yt : [INFO     ] 2025-05-30 23:20:15,142 Parameters: domain_right_edge         = [1.85140655e+24 1.85140655e+24 1.85140655e+24]

In [5]: ds.particle_type_counts
Out[5]: {'StochasticStellarPop_particles': 0, 'CIC_particles': 212500}
```

chongchonghe:
Probably easier to test with a simple test problem like BinaryOrbit. I tested it and the particles are not splitted at all. I can run a debugger and see what's wrong.

chongchonghe:
I found a bug. Here, when restarting from a sim with `max_level=0`, `finestLevel() = 0`, so this loop is never reached.

```cpp
		for (int lev = 0; lev < finestLevel(); ++lev) {
			descriptor->splitParticles(lev, split_factor);
		}
```

It should be `lev <= ...`

chongchonghe:
So, I guess what happens in your Agora sims is that all particles are splitted except those in the finest level.

BenWibking:
> So, I guess what happens in your Agora sims is that all particles are splitted except those in the finest level.

That fixed it:
```

In [3]: ds = yt.load("plt00000.old.7220084/")
yt : [INFO     ] 2025-05-30 23:43:15,533 Parameters: current_time              = 0.0
yt : [INFO     ] 2025-05-30 23:43:15,533 Parameters: domain_dimensions         = [128 128 128]
yt : [INFO     ] 2025-05-30 23:43:15,533 Parameters: domain_left_edge          = [-1.85140655e+24 -1.85140655e+24 -1.85140655e+24]
yt : [INFO     ] 2025-05-30 23:43:15,533 Parameters: domain_right_edge         = [1.85140655e+24 1.85140655e+24 1.85140655e+24]

In [4]: ds.particle_type_counts
Out[4]: {'StochasticStellarPop_particles': 0, 'CIC_particles': 212500}

In [5]: ds = yt.load("plt00000")
yt : [INFO     ] 2025-05-30 23:43:25,883 Parameters: current_time              = 0.0
yt : [INFO     ] 2025-05-30 23:43:25,883 Parameters: domain_dimensions         = [256 256 256]
yt : [INFO     ] 2025-05-30 23:43:25,884 Parameters: domain_left_edge          = [-1.85140655e+24 -1.85140655e+24 -1.85140655e+24]
yt : [INFO     ] 2025-05-30 23:43:25,884 Parameters: domain_right_edge         = [1.85140655e+24 1.85140655e+24 1.85140655e+24]

In [6]: ds.particle_type_counts
Out[6]: {'StochasticStellarPop_particles': 0, 'CIC_particles': 1700000}
```

chongchonghe:
I have set up the BinaryOrbit test to validate this. Do you want me to push it? @BenWibking

chongchonghe:
It seems like the problem is fixed, but a simple test problem is always helpful to debug any future problems.

BenWibking:
Now the restart fails on Frontier due to a NaN issue. Unfortunately, I don't see this at any smaller scale:
```
Splitting CIC_particles using split_factor = 8
WARNING: No particle data found in checkpoint (no Level_* directories) in ../ncell_128/chk00000/StochasticStellarPop_particles
Splitting StochasticStellarPop_particles using split_factor = 8
Doing Poisson solve...

0::Assertion `!rhs[lev].contains_nan()' failed, file "/ccs/home/wibking/quokka/src/simulation.hpp", line 1318 !!!
SIGABRT
```

chongchonghe:
> No particle data found in checkpoint

So, the particles are not written to the disk.

BenWibking:
> > No particle data found in checkpoint
> 
> So, the particles are not written to the disk.

No, there are just no star particles created yet.

BenWibking:
> I have set up the BinaryOrbit test to validate this. Do you want me to push it? @BenWibking

Yes, it would be useful to have another test problem. I didn't check the particle splitting in my test, so we should add that.

BenWibking:
@chongchonghe Can you try running (on the `BenWibking/agora-galaxy` branch):
```
mpirun -np 1 ../build/src/problems/AgoraGalaxy/agora_galaxy AgoraGalaxy.in max_timesteps=0
```
and then restart with:
```
mpirun -np 8 ../build/src/problems/AgoraGalaxy/agora_galaxy AgoraGalaxy_256.in restartfile=chk00000
```
I think I've made an obvious mistake somewhere that is creating a NaN, but I don't see where it is.

I get this:
```
Restart from checkpoint chk00000
Input grid dimensions on level 0: (256,256,256)
Restart file grid dimensions on level 0: (128,128,128)
Rescaling MultiFabs in restart file by a factor of 2...
Reading chk00000/metadata.yaml...
	quokka_version = 25.03
	git_hash_quokka = e7ea9433d74b2bb2b8a6de0c9d3bd144a72e6411
	git_hash_amrex = d4b4bf7a4134782eaae78f3140801950709132f8
	units has unknown type! skipping this entry.
	constants has unknown type! skipping this entry.
1::Assertion `!state_new_cc_[lev].contains_nan()' failed, file "/Users/benwibking/amrex_codes/quokka/src/simulation.hpp", line 3218 !!!
SIGABRT
```

chongchonghe:
@BenWibking 
This is the new CMakeLists.txt in BinaryOrbit. This test suite requires the `BinaryOrbitCICRefactorInit` test to run before the `BinaryOrbitCICRefactor` test, as the former dumps a checkpoint for the latter to restart from. Can we rely on that? I don't know what happens when we run ctest in parallel.

```
    add_test(NAME BinaryOrbitCICSplit COMMAND binary_orbit BinaryOrbit_split.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
    add_test(NAME BinaryOrbitCICRefactorInit COMMAND binary_orbit BinaryOrbit_refactor_init.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
    add_test(NAME BinaryOrbitCICRefactor COMMAND binary_orbit BinaryOrbit_refactor.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
    add_test(NAME BinaryOrbitCIC COMMAND binary_orbit BinaryOrbit.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
    add_test(NAME BinaryOrbitCICAMR COMMAND binary_orbit BinaryOrbitAMR.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
```

chongchonghe:
OK, I'll try your `agora-galaxy` branch.

BenWibking:
> @BenWibking This is the new CMakeLists.txt in BinaryOrbit. This test suite requires the `BinaryOrbitCICRefactorInit` test to run before the `BinaryOrbitCICRefactor` test, as the former dumps a checkpoint for the latter to restart from. Can we rely on that? I don't know what happens when we run ctest in parallel.
> 
> ```
>     add_test(NAME BinaryOrbitCICSplit COMMAND binary_orbit BinaryOrbit_split.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
>     add_test(NAME BinaryOrbitCICRefactorInit COMMAND binary_orbit BinaryOrbit_refactor_init.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
>     add_test(NAME BinaryOrbitCICRefactor COMMAND binary_orbit BinaryOrbit_refactor.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
>     add_test(NAME BinaryOrbitCIC COMMAND binary_orbit BinaryOrbit.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
>     add_test(NAME BinaryOrbitCICAMR COMMAND binary_orbit BinaryOrbitAMR.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
> ```

https://cmake.org/cmake/help/latest/prop_test/DEPENDS.html

chongchonghe:
This failed:

```
../build/build-3D-MPI-agora/src/problems/AgoraGalaxy/agora_galaxy AgoraGalaxy.in max_timesteps=0
mpirun -np 8 ../build/build-3D-MPI-agora/src/problems/AgoraGalaxy/agora_galaxy AgoraGalaxy_256.in restartfile=chk00000 max_timesteps=1 
```

This worked:

```
mpirun -np 8 ../build/build-3D-MPI-agora/src/problems/AgoraGalaxy/agora_galaxy AgoraGalaxy.in max_timesteps=0
mpirun -np 8 ../build/build-3D-MPI-agora/src/problems/AgoraGalaxy/agora_galaxy AgoraGalaxy_256.in restartfile=chk00000 max_timesteps=1 
```

chongchonghe:
@BenWibking  I was able to reproduce the error you came across. The solution is, at least it works on my computer, to run the first run with `mpirun -np 8`, matching the second run. Then, the problem disappeared.

chongchonghe:
Alright. That's not "the solution", because obviously in your test you want to keep the right weak scaling. However, that means there is something wrong when we restart with different number of processors.

chongchonghe:
However, this also worked:
```
../build/build-3D-MPI-agora/src/problems/AgoraGalaxy/agora_galaxy AgoraGalaxy.in max_timesteps=0
mpirun -np 8 ../build/build-3D-MPI-agora/src/problems/AgoraGalaxy/agora_galaxy AgoraGalaxy.in max_timesteps=1 restartfile=chk00000
```

The original code works fine when restarting a sim with different processes (obviously, because we don't see such problems before). That means it's a bug introduced by the refactoring in this PR.

chongchonghe:
@BenWibking OK. I was able to reproduce the exact same behaviour with the BinaryOrbit test. The error occurs when you have nested refinement grids. I made the BinaryOrbit test like that by enforcing the particles to live in the finestest grids. Then, restarting it causes the same NAN error.

chongchonghe:
I pushed my changes. Now, I can reproduce the same error (or similar) on my MacBook with a single process:

```bash
../build/build-3D-MPI/src/problems/BinaryOrbitCIC/binary_orbit ./BinaryOrbit_refactor_init.in 
../build/build-3D-MPI/src/problems/BinaryOrbitCIC/binary_orbit ./BinaryOrbit_refactor.in 
```

chongchonghe:
By running a debugger, I found that one bug comes from line 3107 of simulation.hpp:
```
			header.istep[i++] = std::stoi(word);
```
error:
```
0::Assertion `i < (this->std::vector<T, Allocator>::size())' failed, file "/Users/cche/softwares/quokka/quokkawt1/extern/amrex/Src/Base/AMReX_Vector.H", line 35 !!!
```

when restarting from a checkpoint, `finest_level = 0`, so istep size is 1, but there are two words in that line, causing an out-of-bound error. I don't immediately know how to fix it, but I'm sure you know how.

BenWibking:
Ok, I fixed that bug. Now it fails on BinaryOrbitCIC when it calls Restart on the particle container:
```

 5: amrex::ParticleContainer_impl<amrex::Particle<4, 0>, 0, 0, amrex::DefaultAllocator, amrex::DefaultAssignor>::Restart(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) (in binary_orbit) (AMReX_ParticleIO.H:773)

 6: void AMRSimulation<BinaryOrbit>::restartParticleContainerWithRefinement<amrex::AmrParticleContainer_impl<amrex::Particle<4, 0>, 0, 0, amrex::DefaultAllocator, amrex::DefaultAssignor>>(std::__1::unique_ptr<amrex::AmrParticleContainer_impl<amrex::Particle<4, 0>, 0, 0, amrex::DefaultAllocator, amrex::DefaultAssignor>, std::__1::default_delete<amrex::AmrParticleContainer_impl<amrex::Particle<4, 0>, 0, 0, amrex::DefaultAllocator, amrex::DefaultAssignor>>>&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) (in binary_orbit) (simulation.hpp:3395)

 7: void AMRSimulation<BinaryOrbit>::initializeParticleContainerFromCheckpoint<amrex::AmrParticleContainer_impl<amrex::Particle<4, 0>, 0, 0, amrex::DefaultAllocator, amrex::DefaultAssignor>>(std::__1::unique_ptr<amrex::AmrParticleContainer_impl<amrex::Particle<4, 0>, 0, 0, amrex::DefaultAllocator, amrex::DefaultAssignor>, std::__1::default_delete<amrex::AmrParticleContainer_impl<amrex::Particle<4, 0>, 0, 0, amrex::DefaultAllocator, amrex::DefaultAssignor>>>&, quokka::ParticleType, bool) (in binary_orbit) (simulation.hpp:3432)

 8: AMRSimulation<BinaryOrbit>::ReadCheckpointFile() (in binary_orbit) (simulation.hpp:3299)

 9: AMRSimulation<BinaryOrbit>::setInitialConditions() (in binary_orbit) (simulation.hpp:816)

10: problem_main() (in binary_orbit) (binary_orbit.cpp:201)

11: main (in binary_orbit) (main.cpp:54)

```

BenWibking:
Ok, my bug turned out to due to the fact that, unlike the standard AMR use case for interpolation, the new grids are not properly nested with respect to the old ones, so the restart refinement cannot use linear interpolation. It *must* use piecewise constant interpolation to fill the new multifabs.

I still don't know what is causing the new bug in BinaryOrbitCIC.

BenWibking:
Ok, now I think I am seeing a similar (maybe the same?) bug in AgoraGalaxy as in BinaryOrbitCIC:
```
...[level 4] estimated hydro timestep: 9.568797e+12
...[level 4] 	hydro timestep limited at cell [2040, 2012, 2059] with signal speed = 2.834234e+07
...[level 4] estimated particle timestep: 1.797693e+308
...[level 4] 	max particle velocity: 0.000000e+00
...[level 4] 	particle timestep limited at position [nan, nan, nan]
```
There is something messed up with the particle restarts when refining.

BenWibking:
For your bug, the issue is that max_level=1, but only level 0 has any particles. So it crashes when restarting:
```
Restart from checkpoint chk00010
Input grid dimensions on level 0: (64,64,64)
Restart file grid dimensions on level 0: (32,32,32)
Rescaling MultiFabs in restart file by a factor of 2...
Reading chk00010/metadata.yaml...
	quokka_version = 25.03
	git_hash_quokka = 6b6d95197ef43a57b888445652d608f2a60927ab-dirty
	git_hash_amrex = d49d22968b403bc978749b5e2feb7c94a7710158
	units has unknown type! skipping this entry.
	constants has unknown type! skipping this entry.
Before resizeData: particles->numLevels() = 2
After resizeData: particles->numLevels() = 2
About to call Restart with finest_level = 0
0::Assertion `i < (this->std::vector<T, Allocator>::size())' failed, file "/Users/benwibking/amrex_codes/quokka-restart-refactor/extern/amrex/Src/Base/AMReX_Vector.H", line 35 !!!
SIGABRT
```
I think this is a bug in AMReX.

BenWibking:
Claude Code actually diagnosed it:

```
⏺ Good, we can see the warning is printed, confirming the level mismatch. But the error still occurs. The problem is that even
  though we're setting the geometry for the extra levels, the internal vectors in the Restart() function are still sized based on
  the checkpoint file's number of levels.

  Looking at the AMReX code, the issue is in Restart() at line 773 where it's accessing old_dms[lev] which is sized based on
  finest_level_in_file, but the loop uses particles->finestLevel(). This is a bug in AMReX.
```

BenWibking:
@chongchonghe The crash in BinaryOrbitCIC is due to a bug in AMReX. I've modified the code to abort when it detects this situation until we can create a full PR to fix AMReX.

I realized I actually started to fix this bug earlier in this PR (https://github.com/AMReX-Codes/amrex/pull/4424), but I thought it was only necessary if max_level increases on restart. It turns out it is necessary even when max_level is the same, because some levels can have zero particles and therefore don't appear at all in the checkpoint.

BenWibking:
Unfortunately, neither of the crashes seen with BinaryOrbitCIC appear to be related to the crashes I am seeing in AgoraGalaxy.

BenWibking:
Ok, AgoraGalaxy works now on Frontier with the latest version 🥳

chongchonghe:
I don't know why BinaryOrbit_refactor_init.in produces `max_level=0` when running to step 10. It has `max_level=1` at step 3 or step 20. I have changed it to `max_timesteps = 20` in BinaryOrbit_refactor_init.in and restart from 20 in the BinaryOrbit_refactor.in. Now, the BinaryOrbit test problem should pass. 

@BenWibking Is this PR ready to merge, assuming all tests will pass?

BenWibking:
> @BenWibking Is this PR ready to merge, assuming all tests will pass?

Yes!

BenWibking:
Ah, actually, I forgot that I didn't implement the custom parameter for the particle splitting factor. It always refines based on the (grid refinement factor)**3 in 3D.

chongchonghe:
> Ah, actually, I forgot that I didn't implement the custom parameter for the particle splitting factor. It always refines based on the (grid refinement factor)**3 in 3D.

This new `split_factor` parameter can be shared with the function you introduced in #1104 .

chongchonghe:
@BenWibking Do you want to add a `split_factor` parameter in this PR? Otherwise, I can approve this PR and merge it. I think it's fine to leave it as is for this PR.

chongchonghe (dismissed):

src/problems/BinaryOrbitCIC/binary_orbit.cpp
```diff
@@ -151,12 +150,12 @@ template <> void QuokkaSimulation<BinaryOrbit>::computeAfterTimestep()
 
 template <> void QuokkaSimulation<BinaryOrbit>::refineGrid(int lev, amrex::TagBoxArray &tags, amrex::Real /*time*/, int /*ngrow*/)
 {
-	for (amrex::MFIter mfi(state_new_cc_[lev]); mfi.isValid(); ++mfi) {
-		const amrex::Box &box = mfi.validbox();
-		const auto tag = tags.array(mfi);
+	// for (amrex::MFIter mfi(state_new_cc_[lev]); mfi.isValid(); ++mfi) {
+	// 	const amrex::Box &box = mfi.validbox();
+	// 	const auto tag = tags.array(mfi);
```
BenWibking: @chongchonghe Did you change this just for testing purposes?

src/problems/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+add_subdirectory(BinaryOrbitCIC)
+
```
BenWibking: @chongchonghe Is this change necessary?

src/problems/BinaryOrbitCIC/binary_orbit.cpp
chongchonghe: Yes, I changed this to reproduce the nested grids in the Agora sim in order to reproduce the error. Without the change, the whole domain is always refined to max level. I was surprised that the 3D GPU test passed. In the BinaryOrbitAMR test, the block_factor is 32 and the max grid is 64^3, which means only 2 of the 8 quadrants the particles live in are refined to the max level. I think this is fine. It's still a valid and more stringent test.

src/problems/CMakeLists.txt
chongchonghe: I changed it in order to see the outcome of the GPU test earlier. Not necessary but won't hurt.

chongchonghe:
@BenWibking Please make any change you want. Let me know when you are done and I'll approve it. My approval will probably still count if you make the last push, although I also pushed to this PR.

src/problems/BinaryOrbitCIC/binary_orbit.cpp
BenWibking: Ok, I just cleaned this up by removing the commented-out code.

src/problems/CMakeLists.txt
BenWibking: Ok, I've reverted this change.

BenWibking:
> @BenWibking Please make any change you want. Let me know when you are done and I'll approve it. My approval will probably still count if you make the last push, although I also pushed to this PR.

@chongchonghe I've pushed a clean up change, nothing functional. It needs your approval again to merge.

chongchonghe (approved):

chongchonghe:
@claude Create a new PR to add a new feature: add a runtime parameter split_particle_when_restart_with_refine to toggle splitting particles when doing refinement at restart.

PR #1108, opened 2025-05-29
Update simulation.hpp to fix plotfileMF_fc bug

lizmcole:
Fix bug plotfileMF_fc 

### Description
One line change to plotMF_fc so that the grid sizes are for the faces and not cell centers

### Related issues

### Checklist


chongchonghe:
@lizmcole Looks good. Is this PR ready to merge? This is currently at draft mode. You can turn it into "Ready for review" mode when it's ready and I will approve.

lizmcole:
@chongchonghe It's ready any time for review.

chongchonghe (approved):

chongchonghe (changes requested):
Oh, I missed this. I recommend some changes.

chongchonghe:
I missed this previously. Can you make these changes? @lizmcole 
1. The HydroWaveFC test has the same name as the HydroWave test. This could be confusing, as it appears in the test log. The test 'name' is defined in CMakeLists.txt `add_test(NAME HydroWave ...`. We should change that. 
2. It seems like the HydroWaveFC is almost the same as HydroWave with a minor difference: `do_tracers = 1`. Is there any reason you would keep both? If we want to keep both, we can combine them into one folder with shared cpp file but different .in file. See an example in problems/ParticleCreation. Otherwise, we can set `do_tracers = 1` in the original HydroWave test and delete HydroWaveFC.

lizmcole:
@chongchonghe 

Re: 1, I could rename it to something different, suggestions? I have now added the FC to the end of HydroWave in the CMakeLists file to fix that bug

Re: 2, As per the debugging session where it was discussed, the original HydroWave should remain untouched because it is part of the test suites and has specific output. I also removed the functions that generated the python files in the original HydroWave.

chongchonghe:
`HydroWaveFC` is an obvious choice. 

In that case, I would combine the two tests into one folder. This makes the codebase cleaner. 

```
add_test(NAME HydroWave COMMAND test_hydro_wave hydro_wave.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
add_test(NAME HydroWaveFC COMMAND test_hydro_wave hydro_wave_fc.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
```

lizmcole:
The two tests have been moved to one folder.

src/problems/HydroWave/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
 add_executable(test_hydro_wave test_hydro_wave.cpp ${QuokkaObjSources})
+add_executable(test_hydro_wave_fc test_hydro_wave_fc.cpp ${QuokkaObjSources})
```
chongchonghe: You don't need two executables. One is enough. The idea is that we use the same executable from the same cpp file. The only difference between the two tests is that they have a different runtime parameter `do_tracer`, which is distinguished by hydro_wave.in and hydro_wave_fc.in.

chongchonghe:
You don't need two executables. One is enough. The idea is that we use the same executable from the same cpp file. The only difference between the two tests is that they have a different runtime parameter `do_tracer`, which is distinguished by hydro_wave.in and hydro_wave_fc.in.

lizmcole:
test_hydro_wave.cpp original has this line at 118:
`	sim.plotfileInterval_ = -1;
`
Can this be overridden in the .in files? Otherwise, it won't make any pltNNNNN files, which are the ones that contain the fc quantities.

BenWibking:
> test_hydro_wave.cpp original has this line at 118: ` sim.plotfileInterval_ = -1; ` Can this be overridden in the .in files? Otherwise, it won't make any pltNNNNN files, which are the ones that contain the fc quantities.

No, but https://github.com/quokka-astro/quokka/pull/1114 will fix this.

lizmcole:
So I should merge 1114 into this one? Or just focus on the initial issue (the bug) and make a different pull request for changing the directory structure so that it is a single file with 2 .in files?

chongchonghe:
You don't have to wait to merge that branch. Let's focus on what this PR does: fix the bug, and also merge the two test that are the same.  However, we do want to make sure that we don't have unused parameters in the .in files.

chongchonghe:
Please take a look at my other comment above. We don't need to compile the problem twice. You can see an example in problems/ParticleCreation/CMakeLists.txt

chongchonghe:
> test_hydro_wave.cpp original has this line at 118: ` sim.plotfileInterval_ = -1; ` Can this be overridden in the .in files? Otherwise, it won't make any pltNNNNN files, which are the ones that contain the fc quantities.

You could change that to `= 1` in the cpp file and it will write plotfiles in this PR.

lizmcole:
The initial discussion was to make no changes to the original .cpp file. I had originally done that, but was asked to revert it. I can make the changes just with the assumption that when PR 1114 comes into effect, the .in file will override that line and output pltNNNNN.

chongchonghe:
> The initial discussion was to make no changes to the original .cpp file. I had originally done that, but was asked to revert it. I can make the changes just with the assumption that when PR 1114 comes into effect, the .in file will override that line and output pltNNNNN.

I have no opinion on that choice. Buy maybe I was not clear. I meant to remove the `sim.plotfileInterval_ = -1` line in test_hydro_wave.cpp, and set `plotfile_intervel = -1` in hydro_wave.in and `= 1` in hydro_wave_fc.in.

lizmcole:
Okay, two commits, one without the changes to Ben's original .cpp, and one commenting out that line and moving it to the .in file (most recent).

chongchonghe (approved):
Looks good.

PR #1111, opened 2025-05-29
Allow specifying which variables to include in plotfiles

### LINKED ISSUE #735: User can choose which variables to store in pltfiles
**Describe the proposal**
We currently don't have the capability to let the user choose which variables they want stored in pltfiles. We should add that in.

Alternatively, we have to load in large pltfiles while analyzing data which will become problematic for users wanting to do analysis on local machines.

**Additional context**
This will be crucial for simulations including chemistry with many species, because we don't want to store all the species otherwise the pltfiles will become unnecessarily large. Even for primordial chemistry (with 14 species), the size of a single pltfile with 20 AMR levels is 16 GB. This could be reduced by more than half if we only stored the species we really care about.

BenWibking:
### Description
Allows the user to specify which variables are included as cell-centered outputs in plotfiles. By default, all variables are included except for RiemannVelocity variables (which are only useful in their raw, face-centered form).

### Related issues
Fixes https://github.com/quokka-astro/quokka/issues/735.

### Checklist


chongchonghe:
How do you set the variables to write? I don't see a ParmParser to set `plotfileVarsToInclude_cc_`. 

Also, can you modify a test problem to showcase how to use it? That will also test this new functionality.

markkrumholz (approved):

PR #1114, opened 2025-05-30
Ensure runtime parameters override compile-time settings

BenWibking:
### Description
This pull request ensures that runtime parameters specified in input files always take precedence over compile-time values set in `problem_main()` functions. Previously, parameters directly assigned in problem code (before calling `sim.evolve()`) would override user input, which was counterintuitive and limited flexibility. The fix adds a `rereadRuntimeParameters()` method that is called at the beginning of evolve() to re-read all runtime parameters, ensuring that user-specified values take precedence over any compile-time assignments made earlier in `problem_main()`.

### Related issues
No related issues.

### Checklist
_Before this pull request can be reviewed, all of these tasks should be completed. Denote completed tasks with an `x` inside the
square brackets `[ ]` in the Markdown source below:_

### DIFF
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -186,6 +186,7 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 	[[nodiscard]] static auto getScalarVariableNames() -> std::vector<std::string>;
 	void defineComponentNames();
 	void readParmParse();
+	void rereadRuntimeParameters(); // Re-read parameters to ensure runtime values override compile-time settings
 
 	void checkHydroStates(amrex::MultiFab &mf, char const *file, int line);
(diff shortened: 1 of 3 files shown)


BenWibking:
Sonarcloud is complaining because we have 36 (!) class member functions in AMRSimulation now. That is not ideal, but I don't think we can fix that in this PR.

chongchonghe:
One HydroWave test problem failed. I speculate this is because the runtime parameter file in that test overwrites some parameters and causes the test to fail, but it was ignored previously. 

This is also a good chance to clean up all the .in files and remove unnecessary/unused parameters.

chongchonghe:
Probably `cfl = 0.1` in hydro_wave_fc.in will fix it.

chongchonghe (approved):

PR #1115, opened 2025-05-31
Update Bondi accretion test

chongchonghe:
### Description

The initial condition in Bondi accretion test was wrong. This PR fixed it. Now, this test produces results with errors comparable to those from the Krumholz2004 paper, for both the Bondi test and the Bondi-Hoyle test. 

Changes:
1. make the simulation box 4 times bigger, and use two refinement levels each with the same dimensions. 
2. Use reflective B.C. instead of foextrap 
3. Fixed a bug in the accretion routine: replace `ii * dx[0]` with `(ii + static_cast<amrex::Real>(0.5)) * dx[0]`. 

### Related issues
None.

### Checklist


chongchonghe:
The Bondi accretion problem does not produce the right result. 

I run the Bondi Accretion test with M_sink = 3.16 M_sun, implying r_BH = 3.16 dx. The initial condition is an exact solution to the Bondi problem; therefore, the system is expected to evolve and settle into a state similar to the initial state. 

However, the system does not evolve as expected. Here is the density slices at t = 0, t_B, 10 t_B, and 20 t_B : 

</img></img></img></img>

Here is the accretion rate over time: 
mdot_run-M3 16 
> The red line is the numerical accretion rate. The horizontal line is the expected Bondi accretion rate at the initial mass of the particle, and the green curve is the expected Bondi accretion rate at real-time particle mass, proportional to the particle mass squared. The Bondi accretion time is ~2 Myr. 

The gas structure is not stable at all, and the accretion rate keeps increasing. There must be something wrong with the initial condition or the boundary condition in this problem. Gas are falling into the star too fast, suggesting the particle's gravity is stronger than expected. The particles should affect a region of the size of the accretion radius (~3 cell size), but apparently its gravity has influenced the whole box at ~20 t_BH = 40 Myr, at which time the particle mass has not increased much. 

@BenWibking @markkrumholz  Can you help check if the B.C. implemented in the test problem looks right? I use `foextrap` for all variables in all dimensions.

BenWibking:
> @BenWibking @markkrumholz Can you help check if the B.C. implemented in the test problem looks right? I use `foextrap` for all variables in all dimensions.

`foextrap` will give a poor solution if the flow is subsonic anywhere along the boundary. The is true for all MOL hydro methods, not just Quokka. (ORION does not have this issue because it uses characteristic tracing for hydro.) I would recommend either using NSCBC, or just making the box big enough that you can use a periodic boundary condition and it doesn't matter.

markkrumholz:
Yes, agree that FOExtrap is the wrong BC for this problem. The issue is that it reinforces whatever you give it, so if there are some radially inward momentum at the boundary, FOEextrap will add more radial inward momentum, and this becomes a positive reinforcement loop, leading to exactly the runaway you're seeing. The solution is just to use a problem domain big enough that you can use something like reflecting BCs, and it won't matter because you don't run the problem long enough for a sound wave to propagate from the center to the boundary. You can do this with AMR if necessary, so that you can have a really, really big domain and yet still have the cost be reasonable because you're only using high resolution out to ~10 r_B or something like that.

chongchonghe:
@markkrumholz @drannarosen 

I've finished the Bondi test for various kernel sizes. The results seem to suggest that the kernel size doesn't really matter. At the two limits (r_B >> dx or r_B << dx), the numerical results are pretty good, no matter how many grids (2, 3, or 4) we use for the accretion zone. At the marginal regime (r_B ~ dx), none of them works well. In a particular case where r_B = 3 dx, a kernel size of 4 gives worse results than from sizes of 2 or 3. I speculate this is because with a kernel size of 4, the particle inevitably accretes from outside of the Bondi radius. However, this error can be reduced by optimizing the kernel function ( #1055 , Eq. 14 of Krumholz+2004):

$$
w = \exp \left(-r^2 / r_K^2\right). 
$$

So, my suggestion is as follows. I'll make the kernel size a compile-time parameter, and we can choose to use 2, 3, or 4 depending on what other feedback mechanisms we include. We could prioritize the correctness of other feedback like winds or outflows, and we don't have to worry too much about accretion as long as we have 2 or more cells. A caveat is that, in these tests, the particle is placed at a cell center, the most ideal scenario. I tend to believe a kernel size of 3 balances good accuracy and performance. 

> Fig. Accretion rate over time for various $r_B / \Delta x$ (rows) and kernel size (columns). x axis is time in units of Bondi time $t_B = r_B / c_s$. 

CleanShot 2025-06-23 at 13 23 03@2x

> Fig. The relative error at $t = 14 t_{\rm B}$ as a function of resolution for various kernel sizes:

mdot_last_rel_error_kernel_vs_mass

markkrumholz:
I agree that this is fine. I don't think there's a lot of advantage to optimizing the kernel shape to get better results as I did in 2004 -- in the time since then, it has become clear that these fine details matter a lot less than things like how treat the equation of state / cooling near the grid resolution, so there's not anywhere near the same level of impulse to make sure things are exactly right. The important thing is that we're getting down to ~1% errors once r_B/dx is a factor of 10 larger or smaller than unity.

chongchonghe:
and even a kernel size of 1 cell size is decent: 

mdot_last_rel_error_kernel_vs_mass

markkrumholz:
Yes, but a kernel size of 1 is bad for other reasons: if the particle advects from cell to cell, with a kernel size of 1 I at least found in Orion that this caused artificial ringing, because the accretion would instantaneously jump from all happening in one cell to all happening in another. You could see sound waves being launched whenever the particle crossed a cell boundary.

chongchonghe:
> Yes, but a kernel size of 1 is bad for other reasons: if the particle advects from cell to cell, with a kernel size of 1 I at least found in Orion that this caused artificial ringing, because the accretion would instantaneously jump from all happening in one cell to all happening in another. You could see sound waves being launched whenever the particle crossed a cell boundary.

You are probably talking about the kernel = 0.5 dx (or similar) case in my definition. In my definition, all cells whose center is within a distance of `kernel * dx` from the particle are counted as an accretion zone. That means, in the case of kernel = 1 dx, most of the time, a particle accretes from two cells in the x dimension. 

But I agree, this is still such a small accretion zone that would cause the jumping effect you described.

chongchonghe:
It also just occurred to me that there is no big advantage to implementing a complex formula for overlapping volumes, as we discussed in the Quokka meeting, even if we use a kernel size of 2 dx, because the weighting kernel function mentioned above naturally accounts for the same effect.

markkrumholz:
What's going on with the changes to this PR? Should I review it again, or is it still in progress?

chongchonghe:
@markkrumholz It's ready for review. I just fixed some trivial errors.

markkrumholz (dismissed):

chongchonghe:
I fixed a warning. Can you approve it again? @markkrumholz

markkrumholz (approved):

Issue #740, opened 2024-09-14
convert cooling tables to ASDF, drop HDF5 dependency


BenWibking:
**Describe the proposal**
Users have a lot of trouble compiling against a working version of HDF5. However, we only use HDF5 to read the cooling tables. It would be good to avoid it altogether. Some HDF5 versions are also buggy, and we've wasted time working around that.

* The upside: we could use ASDF instead. It supports self-describing arrays and unstructured metadata, and there is both a C++ (https://github.com/eschnett/asdf-cxx) and Python implementation (https://github.com/asdf-format/asdf). We can include the C++ implementation as a submodule, like we do already for AMReX, OpenPMD, and yaml-cpp.
* The downside: we would have to convert the Grackle HDF5 files to ASDF using a Python script, and then save those tables in our repository.

**Describe alternatives you've considered**
We could stick with HDF5, or we could use a raw binary format that we implement ourselves (but why do that?)

**Additional context**
ASDF is used in the astronomical community, and is currently supported by STSci.

BenWibking:
This will also require changing `perturbations.py` to save ASDF files (should be trivial), and then change the code to read in the perturbations in the initial conditions.

BenWibking:
Superceded by https://github.com/quokka-astro/quokka/issues/835.

Issue #57, opened 2022-06-23
use not-quite-transcendental functions for scaled interpolation of cooling table


BenWibking:
See code: https://github.com/lanl/not-quite-transcendental and paper: https://arxiv.org/abs/2206.08957.

This should make the cooling solve (depends on performing 10^x multiple times for each substep in each cell!) significantly faster.

BenWibking:
For reference, reproduction of Table 1[1]:
![nqt_timing](https://user-images.githubusercontent.com/1781919/175184355-f65eadad-e859-4fe7-b564-7a5aa42524b5.png)

[1] https://arxiv.org/abs/2206.08957

PR #1124, opened 2025-06-03
Resample cooling tables to FastLog-spaced rho, e

BenWibking:
### Description
This Python script reads Grackle HDF5 tables and resamples them into a single lookup table for the net cooling rate in terms of mass density $\rho$ and specific internal energy $e$ (rather than Grackle's sampling in $n_H$ and temperature $T$).

Generated lookup tables:
* net cooling rate
* temperature
* sound speed

This means that it is not necessary to compute the temperature at all during the cooling solve, which should reduce the cost of cooling by at least an order of magnitude. We also use `FastMath::log2` to construct the sampling points in $\rho$ and $e$, so calculating transcendental functions as part of the interpolation during the cooling solve is eliminated.

The next PR will handle reading this table into Quokka and using it within the cooling solver.

Example usage:
```console
$ python3 resample_cooling_tables.py ../grackle_data_files/input/CloudyData_UVB=HM2012.h5 --output HM2012_resampled.hdf5
Table properties:
	nH min: 1.000000e-10
	nH max: 1.000000e+04
	T min: 1.000000e+01
	T max: 1.000000e+09

Mean molecular weight range: 0.612 to 1.280
Specific internal energy range: 2.42e+09 to 2.02e+17 erg/g
Verifying grid boundaries...
  rho_grid[0] = 2.336821593300006e-34, rho_min = 2.336821593300000e-34
  rho_grid[-1] = 2.336821593299999e-20, rho_max = 2.336821593300000e-20
  eint_grid[0] = 2.418246265945187e+09, eint_min = 2.418246265945190e+09
  eint_grid[-1] = 2.022491441386060e+17, eint_max = 2.022491441386058e+17
✓ Grid boundaries are correct!

Resampling cooling tables on 100 x 100 grid using not-quite-logarithmic spacing...
Density range: 2.34e-34 to 2.34e-20 g/cm^3
Specific energy range: 2.42e+09 to 2.02e+17 erg/g

Processing density point 1/100
Processing density point 11/100
Processing density point 21/100
Processing density point 31/100
Processing density point 41/100
Processing density point 51/100
Processing density point 61/100
Processing density point 71/100
Processing density point 81/100
Processing density point 91/100

Saving resampled tables to HM2012_resampled.hdf5
Done!

Statistics:
Valid cooling rates: 10000 / 10000
Temperature range: 1.20e+01 to 1.00e+09 K
Cooling rate range: -5.48e-14 to 1.60e-21 erg/cm^3/s
Sound speed range: 5.18e+04 to 4.74e+08 cm/s
```

### Related issues
https://github.com/quokka-astro/quokka/issues/57
https://github.com/quokka-astro/quokka/issues/740

### Checklist


markkrumholz (commented):
Design in general looks good. I also suggest adding fields to report sound speed and thermal pressure, since we need to evaluate these as part of the hydro solve. They're not as important as temperature during cooling, since they just need to be done once at the start of the step rather than iterative, but might as well speed that up too.

chongchonghe:
Here are my notes from a previous discussion with Mark. @markkrumholz Please correct me if I miss anything.

We want
Inputs: $\rho$ and $e$. 
Outputs: Cooling rate $\Lambda$, temperature $T$, sound speed $c_s$, pressure $p$, mean molecular weight $\mu$. 

Note that the sound speed is $\sqrt{\partial p/ \partial \rho}$, so it's not directly derivable from other quantities. These four quantities, $e$, $\rho$, $\mu$, and $T$, are correlated via $e = (3/2) \rho k_B T / \mu$, so we can eliminate one of them. 

In summary, we want four 2D tables outputting $\Lambda$, $T$, $c_s$, and $p$.

BenWibking:
I don't think we can eliminate one of them, because that equation is only true when the adiabatic index is 5/3, since the prefactor is 1/(gamma - 1). On the other hand, I don't think $\mu$ is ever needed for anything.

We need $c_s$, $P$, $T$, $\Lambda$, and the fundamental derivative $\mathcal{G}$ (for the shock speed estimate inside the Riemann solver).

When using the cooling tables, we currently assume within the hydro update that $\gamma = 5/3$ and assume the pressure is given by the ideal gas law with that adiabatic index. I don't see an urgent need to change this, since this is what every other galaxy simulation code does as well. Changing it for hydro is easy, but changing it for MHD will require thinking carefully about how to modify the HLLD solver.

BenWibking:
Ok, I guess we also need the specific heat $c_v$ for the radiation-matter exchange.

BenWibking:
If we interpolate everything separately (rather than taking derivatives of an interpolated free energy), we won't get an exactly thermodynamically consistent state, but maybe that is okay since we aren't simulating stellar interiors.

markkrumholz:
I don't think we have to worry about not having an exactly consistent thermodynamic state, since as you point out we are already ignoring the fact that gamma != 5/3 when doing the hydro update. The penalty for this at worst is a drop in temporal accuracy to first order, since this means we're effectively operator-splitting the thermodynamics from the hydro.

chongchonghe:
Specific heat can be derived from other quantities -- we can assume $c_v = e / (\rho T)$. 

> On the other hand, I don't think μ is ever needed for anything.

You do use `mu` to compute electron density, which is used for the photoelectric heating term. We still need this, right?

BenWibking:
> Specific heat can be derived from other quantities -- we can assume c v = e / ( ρ T ) .

Ah, that makes sense.

> 
> > On the other hand, I don't think μ is ever needed for anything.
> 
> You do use `mu` to compute electron density, which is used for the photoelectric heating term. We still need this, right?

Yes, but it's pre-calculated and included in the table now: https://github.com/quokka-astro/quokka/blob/088dae9e6ca1f586a6b2821022ea9e6b430159cb/extern/cooling/grackle_tables.py#L134

BenWibking:
So to replace the existing functionality, I think we only need a table for temperature $T$ and net cooling rate $\Lambda_{\text{net}}$.

chongchonghe:
> Yes, but it's pre-calculated and included in the table now:

Oh. OK, that's nice and convenient. 

> So to replace the existing functionality, I think we only need a table for temperature T and net cooling rate Λ n e t .

Do you already have `c_s` and `P`?

BenWibking:
> Do you already have `c_s` and `P`?

Those are computed assuming $\gamma = 5/3$ currently, so we don't need a table.

To be specific, we are computing in the hydro solver:
$P = (\gamma - 1) e$
$c_s = \sqrt{\gamma P / \rho}$

BenWibking:
> We also need `mu` to convert `T` and `rhot` to `P`. Or, are we assuming `mu` = whatever is given in EOS_traits as a constexpr?

We *never* compute P from temperature in the hydro solver. We always compute $P = (\gamma - 1) e$ in the hydro solver. I wrote the hydro solver so that a complete EOS is not required -- it never needs to know the temperature to do the hydro update.

chongchonghe:
`P = (gamma - 1) e` makes sense, but `c_s` could be tricky. Since the Grackle table has that, we could just interpolate it to get a more accurate result. However, we don't have to do it now. It's good enough to assume `c_s = sqrt(gamma P / rho)`

BenWibking:
> `P = (gamma - 1) e` makes sense, but `c_s` could be tricky. Since the Grackle table has that, we could just interpolate it to get a more accurate result. However, we don't have to do it now. It's good enough to assume `c_s = sqrt(gamma P / rho)`

It's what every other code does, including GIZMO, Enzo, and (AFAIK) RAMSES.

If we drop this approximation, I think we should drop $\gamma = 5/3$ everywhere and compute all the thermodynamic quantities from tables (or from derivatives of the tabulated free energy).

BenWibking:
Actually, we *cannot* make a different thermodynamic assumption for $P$ than for $c_s$ in the Riemann solver. It will break the Riemann solver, because the wavespeeds will be inconsistent with the flux it produces, so it will not produce enough entropy in shocks.

chongchonghe:
Sounds good. Let's stick with the simplified version for now.

BenWibking:
@chongchonghe @markkrumholz Ok, I've fixed a subtle floating-point error in computing the inverse of the FastLog function. While the inverse exists and is continuous, the naive algebraic inverse suffers from a fatal floating-point rounding error. I instead used Brent's method to find the inverse. This is only needed for the abcissas of the grid sampling points when constructing the table. The inverse of FastLog is never needed when actually using the table to interpolate.

I've also pre-computed the resampled table for the Grackle table with the Haardt & Madau (2012) UV background, since that's the only table we've actually used.

Should be ready to be reviewed in detail or approved.

BenWibking:
Ah, wait. I need to add a table for the sound speed. I used that in the ShockCloud problem to compute the cooling length.

BenWibking:
Ok, I probably need to write a script to compute the interpolation accuracy across a very grid of density and specific internal energy (or temperature), and do a quantitative measurement of how accurate this is.

BenWibking:
I may need to split this into a cooling table and a separate heating rate table, rather than the net cooling rate.

Or maybe I need to use the second-order version of this method: https://iopscience.iop.org/article/10.3847/1538-4365/adbbd4

BenWibking:
May also need to interpolate (Edot/rho**2) instead of just Edot.

BenWibking:
Ok, that didn't help. Something is wrong here. The cooling is not close to being right in the ShockCloud problem. Need to check this step by step.

chongchonghe:
In the SN test, the ODE solver for cooling failed at ~1e4 yr when the cooling time becomes comparable to the timestep.

Here is the last error message:

```
Coarse STEP 147 at t = 3.486868805e+11 (3.683073496%) starts ...
...[level 0] estimated hydro timestep: 1.039700e+11
...[level 0] 	hydro timestep limited at cell [17, 18, 17] with signal speed = 7.122847e+07
...[level 0] 	cell density = 1.923275e-25, |v| = 2.760013e+07, cs = 4.362834e+07
...[level 0] estimated particle timestep: 1.797693e+308
...[level 0] 	max particle velocity: 0.000000e+00
...[level 0] 	particle timestep limited at position [nan, nan, nan]
...coarse timestep set by level 0
[Level 0 step 147] ADVANCE with time = 3.4868688048e+11 dt = 3.4868719605e+10
max substeps exceeded! rho = 2.31993912994686001e-25, Eint = nan, cooling time = nan, dt = 1.74343598027085876e+10
max substeps exceeded! rho = 2.31993912994686001e-25, Eint = nan, cooling time = nan, dt = 1.74343598027085876e+10
```

chongchonghe:
I think we don't have to worry too much about the accuracy of the new table for now, as long as it gives acceptable results for the test problems (ShockCloud and SN). We can refine that after the INCITE proposal is due.

chongchonghe:
The code ran and produced the new table as expected, but the SN test didn't pass. See comments in #1126 . ~~It might be a problem from the table, so I'll wait till the test problems pass.~~ It's unlikely this is due to an error in the table, but I'll hold on approving this PR till the test problems pass.

chongchonghe:
`resample_grackle_cooling_tables.py` works fine, but `resample_cloudy_cooling_tables.py` does not:

```
-> % python3 ./extern/cooling/resample_cloudy_cooling_tables.py extern/grackle_data_files/input/CloudyData_UVB=HM2012.h5 --output a.h5 
Traceback (most recent call last):
  File "/Users/cche/softwares/quokka/quokkawt1/./extern/cooling/resample_cloudy_cooling_tables.py", line 505, in <module>
    main()
  File "/Users/cche/softwares/quokka/quokkawt1/./extern/cooling/resample_cloudy_cooling_tables.py", line 496, in main
    resample_cloudy_cooling_tables(
  File "/Users/cche/softwares/quokka/quokkawt1/./extern/cooling/resample_cloudy_cooling_tables.py", line 272, in resample_cloudy_cooling_tables
    tables = read_tables(cloudy_file)
  File "/Users/cche/softwares/quokka/quokkawt1/extern/cooling/cloudy_cooling_tools_tables.py", line 29, in read_tables
    tables.log_nH = f['Parameter1']
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/Users/cche/anaconda3/envs/quokka-base/lib/python3.10/site-packages/h5py/_hl/group.py", line 357, in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5o.pyx", line 189, in h5py.h5o.open
KeyError: "Unable to open object (object 'Parameter1' doesn't exist)"
```

BenWibking:
> `resample_grackle_cooling_tables.py` works fine, but `resample_cloudy_cooling_tables.py` does not:
> 
> ```
> -> % python3 ./extern/cooling/resample_cloudy_cooling_tables.py extern/grackle_data_files/input/CloudyData_UVB=HM2012.h5 --output a.h5 
> Traceback (most recent call last):
>   File "/Users/cche/softwares/quokka/quokkawt1/./extern/cooling/resample_cloudy_cooling_tables.py", line 505, in <module>
>     main()
>   File "/Users/cche/softwares/quokka/quokkawt1/./extern/cooling/resample_cloudy_cooling_tables.py", line 496, in main
>     resample_cloudy_cooling_tables(
>   File "/Users/cche/softwares/quokka/quokkawt1/./extern/cooling/resample_cloudy_cooling_tables.py", line 272, in resample_cloudy_cooling_tables
>     tables = read_tables(cloudy_file)
>   File "/Users/cche/softwares/quokka/quokkawt1/extern/cooling/cloudy_cooling_tools_tables.py", line 29, in read_tables
>     tables.log_nH = f['Parameter1']
>   File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
>   File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
>   File "/Users/cche/anaconda3/envs/quokka-base/lib/python3.10/site-packages/h5py/_hl/group.py", line 357, in __getitem__
>     oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
>   File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
>   File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
>   File "h5py/h5o.pyx", line 189, in h5py.h5o.open
> KeyError: "Unable to open object (object 'Parameter1' doesn't exist)"
> ```

This script does not work with Grackle tables. It works with tables output directly from cloudy-cooling-tools. You can use it on the `isrf_1000Go_grains.h5` file.

BenWibking:
I think I have made a units error in the cooling table. Need to investigate more.

BenWibking:
Ok, there is just a bug in the number of powers of rho that undos the scaling applied to the cooling rates in the Grackle tables. The underlying interpolation method works fine.

BenWibking:
Here are the Python-only test problems. I will re-implement these in C++ as an actual Quokka problem tomorrow.

cooling_comparison_cool_dense
cooling_comparison_warm_medium
cooling_comparison_hot_diffuse
cooling_comparison_shocked_gas

chongchonghe:
Tomorrow (Friday), I'll implement a Backward Euler cooling solver using the new DataTable introduced in #1128 . Once you have the one-zone test problem ready, I'll use that for testing.

BenWibking:
Ok, this requires too much back and forth between the python scripts and the C++ code, so I'm closing this PR and will continue everything in https://github.com/quokka-astro/quokka/pull/1126.

PR #1134, opened 2025-06-07
Disable FPE on AppleClang

chongchonghe:
### Description

FPE behaves spuriously on macOS with Apple chips, so we'd better turn it off for all compilers on macOS.

### Related issues
None. 

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 endif()
 
 # this is necessary to avoid spurious FPEs due to vectorization
-if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" OR CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" OR (CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
 	option(ENABLE_TESTS_FPE "Enable floating-point exceptions when running tests" OFF)
 else()
 	option(ENABLE_TESTS_FPE "Enable floating-point exceptions when running tests" ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,6 @@ option(QUOKKA_PYTHON "Compile with Python support (on/off)" ON)
 option(DISABLE_FMAD "Disable fused multiply-add instructions on GPU (on/off)" ON)
 option(ENABLE_ASAN "Enable AddressSanitizer and UndefinedBehaviorSanitizer" OFF)
 option(ENABLE_HWASAN "Enable HWAddressSanitizer" OFF)
-option(ENABLE_TESTS_FPE "Enable floating-point exceptions when running tests" ON)
 option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" OFF)
 option(QUOKKA_OPENPMD "Enable OpenPMD output (on/off)" OFF)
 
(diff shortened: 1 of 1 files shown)


BenWibking:
Did you test gcc on your Mac? I haven't seen this happen with gcc.

chongchonghe:
> Did you test gcc on your Mac? I haven't seen this happen with gcc.

No, I haven't tested gcc. I don't use gcc on my mac. 

I just checked and found `gcc` on my mac is Apple clang as well:

```
-> % gcc -v
Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: arm64-apple-darwin24.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

chongchonghe:
Here is the updated version. It checks AppleClang or Darwin+Clang and should avoid gnu.

```
if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" OR (CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
	option(ENABLE_TESTS_FPE "Enable floating-point exceptions when running tests" OFF)
else()
	option(ENABLE_TESTS_FPE "Enable floating-point exceptions when running tests" ON)
endif()
```

BenWibking:
> > Did you test gcc on your Mac? I haven't seen this happen with gcc.
> 
> No, I haven't tested gcc. I don't use gcc on my mac.
> 
> I just checked and found `gcc` on my mac is Apple clang as well:
> 
> ```
> -> % gcc -v
> Apple clang version 17.0.0 (clang-1700.0.13.5)
> Target: arm64-apple-darwin24.5.0
> Thread model: posix
> InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
> ```

Right, you have to install gcc via Homebrew, and then use, e.g. `gcc-14`: https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX.html#gcc-on-macos

chongchonghe:
@BenWibking OK. I tested it. `gcc-15` installed via homebrew works just fine with FPE on. This PR is ready now. Can you review and approve it?

CMakeLists.txt
```diff
@@ -78,8 +77,10 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 endif()
 
 # this is necessary to avoid spurious FPEs due to vectorization
-if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
-  add_compile_options(-ffp-exception-behavior=strict)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" OR (CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
```
BenWibking, line 80: Is the Darwin+Clang combination necessary? Did you test non-Apple Clang on macOS?

BenWibking (changes requested):

CMakeLists.txt
chongchonghe: Yes. I have `mpicc` installed via conda on my Mac. That's the only way to get MPI work (the brew version didn't). `mpicc` appears as `CMAKE_CXX_COMPILER_ID = Clang` and not `AppleClang`. So, `(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang")` is necessary.

BenWibking: Ah, ok. That's very weird. When openmpi in Homebrew broke a few weeks ago, I installed openmpi via Spack on my laptop. That works much better than conda for me.

BenWibking (approved):

CMakeLists.txt
BenWibking: But some users might have Conda installed already. So we should make that work too. I'll approve.

PR #1135, opened 2025-06-07
[WIP] Stable particle redistribute on GPUs

### LINKED ISSUE #1125: reproducible particle operations
**Describe the bug**
Star formation is not reproducible, because:
 i) we don't set the random seed to a different (but reproducible) value on each MPI rank, and
ii) we use the non-deterministic particle redistribution in AMReX.

We need to fix both of these things to ensure galaxy simulations are reproducible.

**To Reproduce**
1. Compile the DiskGalaxy problem.
2. Run it multiple times until it forms stars (about 200 timesteps).
3. See the output is different each time.

**Additional context**
AMReX StableRedistribute: https://github.com/AMReX-Codes/amrex/pull/4200

chongchonghe:
### Description

This PR attempts to add a new parameter to make particle redistribute stable on GPU. 

Related: https://github.com/AMReX-Codes/amrex/pull/4200 
Following implementation from: https://github.com/erf-model/ERF/pull/2021 

This is not tested. It compiles and runs, but I haven't tested it on GPUs. I couldn't reproduce different results on GPU runs (as reported in #1125). 

### Related issues
Closes #1125 

### Checklist

### DIFF
--- a/src/particles/particle_types.hpp
+++ b/src/particles/particle_types.hpp
@@ -315,7 +315,7 @@ inline bool sink_particle_use_uniform_kernel = false; // NOLINT. If true, use un
 inline int particle_verbose = 0; // NOLINT print particle logistics
 
 // Random seed for particle creation. Default is 0, which means the random seed is not fixed.
-inline int particle_random_seed = 0; // NOLINT
+inline int particle_random_seed = 1024; // NOLINT
 
 // Stable redistribute for particle creation. Default is 0 (false).
 inline int particle_stable_redistribute = 0; // NOLINT
--- a/src/particles/particle_types.hpp
+++ b/src/particles/particle_types.hpp
@@ -314,6 +314,12 @@ inline bool sink_particle_use_uniform_kernel = false; // NOLINT. If true, use un
 // Verbosity for particle operations
 inline int particle_verbose = 0; // NOLINT print particle logistics
 
+// Random seed for particle creation. Default is 0, which means the random seed is not fixed.
+inline int particle_random_seed = 0; // NOLINT
+
+// Stable redistribute for particle creation. Default is 0 (false).
+inline int particle_stable_redistribute = 0; // NOLINT
+
 // Function to parse particle parameters from input file
 // The 'inline' keyword allows this function to be defined in a header file without
 // causing multiple definition errors when the header is included in multiple source files.
@@ -339,6 +345,12 @@ inline void particleParmParse()
 	// Placeholder parameters for particles
 	pp.query("param1", particle_param1);
 	pp.query("param2", particle_param2);
+
+	// Handle random seed
+	pp.query("random_seed", particle_random_seed);
+
+	// Handle stable redistribute
+	pp.query("stable_redistribute", particle_stable_redistribute);
 }
 
 } // namespace quokka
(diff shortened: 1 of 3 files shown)


src/simulation.hpp
```diff
@@ -2293,13 +2293,21 @@ template <typename problem_t> void AMRSimulation<problem_t>::InitPhyParticles()
 	// Read particle parameters from input file
 	quokka::particleParmParse();
 
+	if (quokka::particle_random_seed != 0) {
+		amrex::InitRandom(static_cast<unsigned long>(quokka::particle_random_seed)); // NOLINT(google-runtime-int)
```
BenWibking, line 2297: You need to set this to a different value on each MPI rank. You also need to set the GPU seed separately: https://github.com/AMReX-Codes/amrex/blob/6ab61fb6d8a752a5b586cc98619c6cc13e417394/Src/Base/AMReX_Random.cpp#L89C1-L89C11

src/particles/particle_types.hpp
```diff
@@ -314,6 +314,12 @@ inline bool sink_particle_use_uniform_kernel = false; // NOLINT. If true, use un
 // Verbosity for particle operations
 inline int particle_verbose = 0; // NOLINT print particle logistics
 
+// Random seed for particle creation. Default is 0, which means the random seed is not fixed.
+inline int particle_random_seed = 0; // NOLINT
```
BenWibking: I think this should be fixed to some non-zero value by default.

src/particles/particle_types.hpp
```diff
@@ -314,6 +314,12 @@ inline bool sink_particle_use_uniform_kernel = false; // NOLINT. If true, use un
...
+
+// Stable redistribute for particle creation. Default is 0 (false).
+inline int particle_stable_redistribute = 0; // NOLINT
```
BenWibking, line 321: I think we should turn on StableRedistribute by default.

BenWibking (changes requested):

src/particles/particle_types.hpp
chongchonghe: OK. Done.

src/particles/particle_types.hpp
chongchonghe: This makes the code slower, as Andrew Myers said. Do you really want turn it on by default?

BenWibking: I think the question is: how much slower does it make it on a production simulation? If the answer is "a few percent", then I think yes, being reproducible is worth it.

BenWibking: For a code that has 100s of particles in every cell (like WarpX) it may not be worth it. For our case, there is << 1 particle per cell, so I think it will have a very small performance impact.

BenWibking: The particle container Redistribute() function does not even appear in the profiling output, so it is currently negligible. That is why I am almost certain it will not matter for our simulations.

chongchonghe: OK, that makes sense.

BenWibking:
@chongchonghe Do we still need this PR?

chongchonghe:
Probably not needed. I'll close it. We can revisit this if we find this is necessary.

PR #1163, opened 2025-06-26
update Microphysics to make EOS names constexpr

BenWibking:
### Description
This updates the Microphysics submodule so that the EOS names are now constexpr. This is needed so we can check whether a non-gamma-law EOS is used with MHD and produce an error at compile time.

### Related issues
N/A

### Checklist


BenWibking:
Hmm, somehow this broke several of our hydro tests.

chongchonghe:
I think you have to update the `EOS_Traits` in EOS.hpp for this PR, right?

chongchonghe:
BTW, do we really need Microphysics for a simple gamma-law EOS? You mentioned a long time ago about getting rid of reliance on Microphysics when using gamma-law EOS. Are you still planning to do that?

BenWibking:
> BTW, do we really need Microphysics for a simple gamma-law EOS? You mentioned a long time ago about getting rid of reliance on Microphysics when using gamma-law EOS. Are you still planning to do that?

I've thought about this more and I think, in terms of the EOS, we should either use Microphysics for all EOSes or for none of them. Having two different code paths for the same physics is the best way to create bugs and was something I wanted to avoid in Quokka since the beginning. So as long as we need Microphysics for the more complicated EOSes, I think we should stick to using Microphysics for gamma law as well.

PR #1175, opened 2025-07-01
Add a log message of maximum supernova remnant velocity 

chongchonghe:
### Description

Print the maximum net velocity (speed + c_s) in supernova remnant regions at each step. Print a WARNING when net velocity > 0.03 c (equivalent to T = 1e10 K). 

### Related issues
None. 

### Checklist

### DIFF
--- a/src/particles/particle_deposition.hpp
+++ b/src/particles/particle_deposition.hpp
@@ -570,7 +570,7 @@ void SNDeposition(ContainerType *container, amrex::MultiFab &state, amrex::Multi
 	SNFeedbackUtils::addBufferToState<problem_t>(state, state_buffer, SN_scheme_d, p_max_velocity);
 
 	// Step 4: Check maximum velocity and print warning if needed
-	amrex::Real max_velocity = max_velocity_buffer.data()[0];
+	const amrex::Real max_velocity = max_velocity_buffer.data()[0];
 	constexpr amrex::Real c_light = C::c_light;
 	constexpr amrex::Real velocity_threshold = 0.03 * c_light;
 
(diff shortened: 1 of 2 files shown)


src/particles/particle_deposition.hpp
```diff
@@ -331,6 +331,12 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE void addCompositeBufferToState(amrex::Array4
 	const double e_tot = local_state(i, j, k, HydroSystem<problem_t>::energy_index);
 
 	const double d_rho = local_buffer(i, j, k, HydroSystem<problem_t>::density_index);
+
+	// Skip if there is no SN feedback
+	if (d_rho < 10. * std::numeric_limits<Real>::min()) {
```
BenWibking: I'm not sure if this is a good idea to do the comparison this way. If you take 10 times smaller than the minimum value, I think you get a denormal floating-point value (https://en.wikipedia.org/wiki/Subnormal_number), which is either very slow or rounds to zero.

chongchonghe: OK. I was always nervous at using something like 'x == 0.0', but it's probably fine, especially when d_rho is original set to 0.0 and not changed later. 
Another approach would be `amrex::almostEqual(d_rho, 0.0_rt)`, or `d_rho == 0.0_rt` is equally fine? We can make this change later when we want to support single-precision.

BenWibking: If it's set to 0.0 (exactly) and never changed, then that is one case where testing for exact equality is the correct thing to do.

BenWibking: This is true regardless of the floating-point precision.

chongchonghe: OK, that makes sense. good to know.

chongchonghe: I made that change.

BenWibking (approved):
LGTM

chongchonghe:
With this change, quokka will print out the maximum net velocity (|v| + cs) at SNR regions. When there is no SN explosion, it will print 0. If you see a large number (>0.01 c), that means SN is doing strange things. If not, that means the extremely high velocity comes form hydrodynamics.

PR #1181, opened 2025-07-05
Extrema-preserving PPM

BenWibking:
### Description
This implements the 5th-order, extrema-preserving, hybrid PPM-WENO reconstruction method of [Rider, Greenough, and Kamm (2007)](https://ui.adsabs.harvard.edu/abs/2007JCoPh.225.1827R/abstract).

The number of ghost zones required is the same as classic PPM.

The default reconstruction method is unchanged (classic PPM). The advection tests and `RadhydroShockCGS` are modified to use this method by default, so this new method is tested in the test suite. (While testing this, I also discovered the RadhydroShock tests were slightly too under-resolved to get the Zeldovich spike right, so those are all increased to 256 cells.)

Caveats:
* Whether this method perfectly maintains an isolated contact wave is compiler-dependent. There appears to be sensitivity to how the floating-point operations are grouped. This causes the `HydroContact` test to fail for one CI configuration when xPPM is enabled for this problem.
* The `RadhydroShell` problem crashes with this method. It might need to be run at a lower CFL number.
* Note that only the dimension-by-dimension reconstruction is 5th order: the overall order of accuracy in space in still 2nd order, since the conservative-to-primitive conversion assumes that cell averages and cell centers are identical (also the face center and face average fluxes are assumed to be identical).

### Related issues
N/A

### Checklist


BenWibking:
The one concern is that the temperature spike in the subcritical radiative shocks is... kind of weird:
[radshock_multigroup_temperature.pdf](https://github.com/user-attachments/files/21083917/radshock_multigroup_temperature.pdf)
[radshock_cgs_temperature.pdf](https://github.com/user-attachments/files/21083918/radshock_cgs_temperature.pdf)

src/hyperbolic_system.hpp
```diff
@@ -432,6 +414,211 @@ AMREX_GPU_HOST_DEVICE void HyperbolicSystem<problem_t>::ReconstructStatesPPM(quo
...
+    -> std::pair<double, double>
+{
+	// compute monotone edge values
+	const double qL_star = median(q, qL_in, qminus);
+	const double qR_star = median(q, qR_in, qplus);
+
+	// this does something weird to the left side of the sawtooth advection
+	// problem, but is absolutely essential for stability in other problems
+	const double qL = median(q, qL_star, 3. * q - 2. * qR_star);
+	const double qR = median(q, qR_star, 3. * q - 2. * qL_star);
```
BenWibking, line 428: I think this there is a sign error on line 427 or 428. Need to debug.

BenWibking: This appears to be right. Also, if I remove this step, the HydroHighMach problem is completely wrong. Somehow this is the right thing to do for this method, but I have absolutely no idea why.

BenWibking:
These tests fail on GPU:
```
The following tests FAILED:
	 14 - HydroContact (Failed)
	 44 - RadhydroShell (Failed)
```

These fail with FPE enabled on CPU:
```
Short test summary info
✗ FAIL 2ebbcec HydroContact (00:00:03): a non-zero exit status was returned
✗ FAIL e62eacd HydroLeblanc (00:00:03): a non-zero exit status was returned
```

In both cases, HydroContact fails because the error is not exactly zero. So whether there is exact cancellation of the flux at the contact interface appears to be compiler-dependent with this method.

HydroLeblanc causes an FPE.

RadhydroShell has a stability problem.

BenWibking:
The HydroLeblanc FPE appears in the Riemann solver, but the variables in the debugger are all non-NAN, so I don't know how this FPE is happening.

src/linear_advection/AdvectionSimulation.hpp
```diff
@@ -116,7 +116,7 @@ template <typename problem_t> class AdvectionSimulation : public AMRSimulation<p
 
 	amrex::Real errorNorm_ = NAN;
 
-	static constexpr int reconstructOrder_ = 3; // PPM = 3 ['third order'], piecewise constant == 1
+	static constexpr int reconstructOrder_ = 5; // xPPM = 5 (fifth order), piecewise constant == 1
```
chongchonghe: A search of the repo indicates `reconstructOrder_` is never used anywhere. Is this a legacy constexpr parameter you forgot to delete? All the test problems use runtime parameter `sim.reconstructionOrder_`.

chongchonghe: I see you manually set to use the EP reconstruction at line 499, so this test is valid. You can simply delete this line to avoid confusion.

chongchonghe: A search of the repo indicates reconstructOrder_ is never used anywhere. Is this a legacy constexpr parameter you forgot to delete? All the test problems use runtime parameter sim.reconstructionOrder_.

I see you manually set to use the EP reconstruction at line 499, so this test is valid. You can simply delete this line to avoid confusion.

chongchonghe:
"Pending" bug again, so I'm gonna paste my comments here (this is the line 119 of AdvectionSimulation.hpp):

A search of the repo indicates `reconstructOrder_` is never used anywhere. Is this a legacy constexpr parameter you forgot to delete? All the test problems use runtime parameter `sim.reconstructionOrder_`.
I see you manually set to use the EP reconstruction at line 499, so this test is valid. You can simply delete this line to avoid confusion.

chongchonghe:
The overall implementation looks good to me, although I don't have the capacity to review the details of the 5th reconstruction scheme, so I would request @markkrumholz to take a review. 

One minor comment: I suggest we modify one of the all-around tests (like RadhydroShock) to validate this new scheme, instead of only replying on the Advection test, which is isolated from the rest of the code. In this PR, the 'ReconstructStatesPPM_EP' function in QuokkaSimulation.hpp is never touched and that is a concern to me.

BenWibking:
> "Pending" bug again, so I'm gonna paste my comments here (this is the line 119 of AdvectionSimulation.hpp):
> 
> A search of the repo indicates `reconstructOrder_` is never used anywhere. Is this a legacy constexpr parameter you forgot to delete? All the test problems use runtime parameter `sim.reconstructionOrder_`. I see you manually set to use the EP reconstruction at line 499, so this test is valid. You can simply delete this line to avoid confusion.

I've deleted this line in AdvectionSimulation.hpp.

BenWibking:
> The overall implementation looks good to me, although I don't have the capacity to review the details of the 5th reconstruction scheme, so I would request @markkrumholz to take a review.
> 
> One minor comment: I suggest we modify one of the all-around tests (like RadhydroShock) to validate this new scheme, instead of only replying on the Advection test, which is isolated from the rest of the code. In this PR, the 'ReconstructStatesPPM_EP' function in QuokkaSimulation.hpp is never touched and that is a concern to me.

I've updated `inputs/radshock.in` to use xPPM reconstruction.

BenWibking:
@chongchonghe I've made the changes you suggested here.

chongchonghe (approved):
LGTM.

PR #1183, opened 2025-07-07
Don't store face velocities in state_fc_

### LINKED ISSUE #902: slight face-centered velocity inconsistency when AMR subcycling
**Describe the bug**
The face-centered velocity is inconsistent by a half-timestep at the coarse-fine boundary when doing AMR subcycling. (This is because the face-centered velocity values are averaged over the timestep, but the AMR interpolator assumes they are at the end of the timestep, and they are interpolated from the coarser level onto the finer level at a different time.)

This causes a very slight inconsistency in the trajectories of tracer particles at the coarse-fine AMR boundary. If the face-centered velocities are also used for the MHD emf update, then there is also an inconsistency there.

Two possible fixes:
* advect the tracer particles using the velocities at the beginning of the timestep and then advect them again at the end of the timestep (e.g., the [RK2 method used in FLASH for passive tracers](https://flash.rochester.edu/site/flashcode/user_support/flash_ug_devel/node136.html#SECTION06813000000000000000)). This also has the advantage that the time integration can be made exactly consistent with that used for hydro. Then we never need to deal with time-centered velocity fields. However, this can't be done with the built-in AMReX tracer particle advection code.
* change the AMR interpolation code so that it knows that the coarse-level face-centered velocities should be treated as time-centered. (This should be easy once we have https://github.com/quokka-astro/quokka/issues/748 implemented.)

**To Reproduce**
1. Run a problem with tracer particles and AMR subcycling
2. Examine face-centered velocity field at coarse-fine interface

**Additional context**
This problem arises because we use the face-centered marker-and-cell (MAC) advection method from AMReX, which requires velocities that are averaged over the timestep. Most other AMReX codes use cell-centered velocities to advect the tracer particles, where this problem doesn't arise in the same way.

BenWibking:
### Description
Since we now have 6 ghost cells, we don't need to store the face velocities as part of the global simulation state. We only need to store them temporarily during the hydro advance on a given level. This removes them to save memory and simplify the code.

### Related issues
Fixes https://github.com/quokka-astro/quokka/issues/902.

### Checklist

### DIFF
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -1546,31 +1546,9 @@ auto QuokkaSimulation<problem_t>::advanceHydroAtLevel(amrex::MultiFab &state_old
 	amrex::Gpu::streamSynchronizeAll();
 
 	// advect tracer particles using avgFaceVel
-#ifdef AMREX_PARTICLES
 	if (do_tracers != 0) {
-		// copy avgFaceVel to state_new_fc_[lev]
-		for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
-			amrex::Copy(state_new_fc_[lev][idim], avgFaceVel[idim], Physics_Indices<problem_t>::velFirstIndex, 0,
-				    Physics_NumVars::numVelVars_per_dim, nghost_vel);
-		}
-
-		// fill ghost faces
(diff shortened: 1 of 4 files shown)


chongchonghe (approved):
LGTM. My only minor concern is this will eliminate the possibility of using fewer ghost cells for some problems. However, since we are going to include magnetic fields in most production runs once we have MHD working, this is not a concern.

PR #1188, opened 2025-07-09
Remove all AMREX_PARTICLES ifdef blocks

### LINKED ISSUE #1099: remove ifdef AMREX_PARTICLES
**Describe the proposal**
I thought this would be useful to have the ability to turn off particles to reduce the compile time for problems that don't need them, but we need particles for all of our production simulations. Also the compile time difference is very small.

N/A

**Additional context**

BenWibking:
Remove all conditional compilation blocks for AMREX_PARTICLES while preserving the particle functionality code inside them. This makes particle support always available instead of being conditionally compiled.

Changes:
- Remove all #ifdef AMREX_PARTICLES and corresponding #endif blocks
- Keep all particle-related code that was previously conditional
- Affects 13 locations across simulation.hpp and QuokkaSimulation.hpp

Resolves #1099

### DIFF
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -1157,14 +1157,12 @@ void QuokkaSimulation<problem_t>::advanceHydroAtLevelWithRetries(int lev, amrex:
 		amrex::Copy(originalFineData, fineData, 0, 0, fineData.nComp(), 0);
 	}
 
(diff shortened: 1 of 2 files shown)


BenWibking:
@chongchonghe Can you review this PR?

chongchonghe (approved):

PR #1194, opened 2025-07-11
Fix fextract function for MPI-parallel runs

### LINKED ISSUE #1101: `fextract` with MPI causes crash
**Describe the proposal**
The following code caused `SIGABRT` when running with more than 1 process. Does `fextract` not work with MPI? 

```cpp
auto [position1, values1] = fextract(sim.state_new_cc_[0], sim.Geom(0), 0, 0.0, true);
```

**Additional context**
None.

BenWibking:
The fextract function was causing SIGABRT when running with multiple MPI processes due to null pointer issues in ParallelDescriptor::Gatherv operations. When some processes had no data for the slice, their vectors were empty, causing data() to return null pointers.

This fix uses dummy Real values when vectors are empty to ensure we never pass null pointers to MPI gather operations, while the numpts parameter correctly indicates the actual number of elements to gather.

Fixes #1101

### DIFF
--- a/src/util/fextract.cpp
+++ b/src/util/fextract.cpp
@@ -155,13 +155,13 @@ auto fextract(MultiFab &mf, Geometry &geom, const int idir, const Real slice_coo
 		// Use a dummy Real value when vectors are empty to avoid passing null pointers
 		Real dummy_pos = 0.0;
 		Real dummy_data = 0.0;
-		const Real* pos_ptr = pos.empty() ? &dummy_pos : pos.data();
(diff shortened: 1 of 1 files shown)


BenWibking:
@claude This fix introduces spurious zero values to the extracted profile. Can you fix the root issue by passing a zero-size array (with a non-nullptr value) to ParallelDescriptor::Gatherv?

BenWibking:
@chongchonghe This might fix the issue you were seeing earlier.

markkrumholz (approved):

PR #1215, opened 2025-07-13
Implement user-specified particle radiation

chongchonghe:
### Description
**Implement user-specified particle radiation**

This pull request adds support for setting particle luminosity based on particle properties. The implementation provides a  new mechanisms for updating the luminosity field at each hydro step.

**Key Features:**
- **New particle radiation framework** (`src/particles/particle_radiation.hpp`): Introduces template traits system allowing problems to specify custom luminosity functions via `LuminosityTraits<problem_t>::stellarLuminosity()`

**Core Components:**
- `updateProperties` traits class for computing luminosity values on-the-fly from particle data. 
- New test problem `ParticleRadiation` demonstrating and validating the functionality with simple linear mass-luminosity relation

### Related issues
None.

### Checklist


chongchonghe:
@claude help me write a description for this PR

chongchonghe:
@markkrumholz can you review this PR?

markkrumholz (changes requested):
Overall this mostly looks fine, but I think mass-based luminosity is too specific. First of all, the luminosity isn't just based on mass, it is based on mass and age, so the name is confusing. But second and more importantly, I'm not sure it makes sense to have this option exist at all, because then there still needs to be a particle physics-dependent implementation of the function to look up the luminosity given the mass and age. So I think I would prefer a more general solution of just using the luminosity index, and then sticking all the customization in to the function responsible for setting the luminosity. This will be an example of a generic particle updateState functionality -- that is, there is an updateState routine that is called on every particle at every time step, and which can change particle properties. The implementation of this function will then depend on the particle type, so for stochastic stellar particles it will just set the particle luminosity, while for something like Orion protostar particles it will do a more sophisticated calculation where it updates a bunch of other variables that are part of the model (e.g., stellar radius, current evolutionary stage, mass of unburned deuterium, etc.). But that way we put all the customization that is specific to the particular type of particle in one place, and leave the radiation deposition routine generic.

chongchonghe:
OK, this is a good design. By always keeping the luminosity as a stored property of the particles, it allows us to write the luminosity into the disk.

chongchonghe:
However, I still think it's helpful to provide a function `stellarLuminosity` for the user to specify a stellar property-luminosity relation. There are cases where the user want to specify their own customized relation or luminosity table, or a simple analytical mass-luminosity relation (like I did with RAMSES-RT in my 2020 paper).

markkrumholz:
I guess I have no objection to an implementation that lets the user provide either a stellar luminosity function or tells the code to get the stellar luminosity from a particle field. But then the stellar luminosity function should get all of the particle data as an argument, rather than just some fields.

chongchonghe:
@markkrumholz Ready for your review again. 
I've changed the design. In this PR, I use a simple analytical formula for luminosity-mass-age relation. In the next PR, I'll implement table interpolation.

chongchonghe:
@markkrumholz I have made the change as we discussed -- removing unnecessary parameters passed to `updateProperties`. It's ready for your review.

src/particles/PhysicsParticles.hpp
```diff
@@ -767,6 +772,27 @@ class StarParticleDescriptor : public PhysicsParticleDescriptor<ContainerType, p
 	}
 
 #if AMREX_SPACEDIM == 3
```
markkrumholz, line 774: I suggest that the code block below be moved into the generic update particle properties function above, so avoid duplicating the boilerplate code that loops over particles, grabs their data, and then passes them to the type-specific updateProperties function. This code will be the same for any particle type that has an update, so I suggest making this generic, and leaving all the type-specific specialization just in the updateProperties<problem_t> function.

src/simulation.hpp
```diff
@@ -1098,29 +1100,38 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
...
 #if AMREX_SPACEDIM == 3
-		if constexpr (Physics_Traits<problem_t>::is_self_gravity_enabled) {
-			kickParticlesAllLevels(dt_[0]);
-		}
+		if constexpr (Particle_Traits<problem_t>::particle_switch != ParticleSwitch::None) {
+			if constexpr (Physics_Traits<problem_t>::is_self_gravity_enabled) {
+				kickParticlesAllLevels(dt_[0]);
+			}
+
+			// Stellar evolution and SN deposition; only apply to star particles
```
markkrumholz, line 1122: As written right now this only calls the update on star particles, not on any other particle that has an update. I think that we either need a flag to register that a particle type provides an update function, or we want to call update particle properties on all particle types, and just let it be a no-op for any particle type that doesn't override the default. The compiler should then be smart enough to optimize that out, so there won't be any performance hit.

markkrumholz (changes requested):
I left some comments suggesting tweaks to the design to support future functionality where we will have multiple particle types that have internal states that will need updating.

src/particles/PhysicsParticles.hpp
chongchonghe: If I understand what you mean correctly, I'm already doing that. If you look at the implementation of updateProperties in particle_radiation.cpp, you'll see the amount of duplication is already minimized. The user specification does not have to repeat any code that's already here.

markkrumholz: Maybe I'm misunderstanding. The issue I was raising here is that you have chosen to hang this function off StarParticleDescriptor, rather than off the higher-level ParticleDescriptorBase. My concern is that, in a few months, people are going to be implementing protostar particles, which will not be derived from StarParticleDescriptor and will instead be an entirely new particle type -- but which will still need to have exactly the same boilerplate code to loop over the particles in the container and call the update on each of them. So I am proposing moving this code up a level in the class hierarchy, so that it is accessible to any particle type, not just particle types derived from StarParticleDescriptor.

src/simulation.hpp
chongchonghe: Currently, `updateParticleProperties` applies to all star particles types, and only to star particle types, and it's default to no-op for base particle types (even if you specialize a `updateProperties` for that type). This is intensional because I assumed only star particles are allowed to chnage its state. The requirement to avoid confusion is to keep in mind that we only define `updateProperties` for star particle types. 

Are you suggesting remove the `if (particleRegister_.HasStarParticles())` clause?

markkrumholz: Yes, in conjunction with my previous change: I don't think that updateParticleProperties should be limited to star particle types, because there are going to be lots of other particle types coming that also have internal states that need to be updated.

src/particles/PhysicsParticles.hpp
chongchonghe: I think you are misunderstanding the particle architecture I have implemented here. We don't have a particle type called "StarParticle". 

Here is a big picture of the particle types:

PhysicsParticleRegister registers base particle types: Rad, CIC, and CICRad. StarParticleRegister registers more feature-rich particle types, currently including StochasticStellarPop, Sink, and Test, and will extend to StellarParticle, ProtostellarParticle, etc. 

So, this `updateParticleProperties` function will be shared among all feature-rich particle types, including Protostars.

markkrumholz: I see. OK, I was confused by the naming. So in your design protostellar particles would sit under StarParticleRegister. That's fine, but then I don't quite understand why it's called StarParticleRegister. What determines whether a particle type sits under this versus under the generic PhysicParticleRegister? What is the distinguishing functionality for StarParticleRegister?

chongchonghe: The only reason I made this two "super classes" of particle types is that many of the complex features like star formation, SN, particle-mesh interaction are irrelevent to CIC (and Rad, CICRad) particles, and you, in your previous PR reviewes, recommended we don't want to provide a "particleCreation" function for CIC particle types. I think if this confuses you, it will confuse everyone. So, perhaps, I should simply remove this 'StarParticleRegister' and use 'PhysicsParticleRegister' for all particle types. This wont' make much difference because CIC particle is the only non-star particle type that will actually be used in production (Rad and CICRad are useless).

markkrumholz: OK, I understand. Yes, this is confusing. So let's have one particle class, and in cases like CIC particles where creation is not really a defined function, we can just have it default to the empty implementation, the same way we will for state updating.

chongchonghe: and the reason we can't do template specialization for the specific particle type that a function applies to is: we are using this visitor pattern instead of class inheritance to deal with all particle types. If we do class ineritance, our "DepositeMass" function would look like this: 

```cpp
DepositeMass(CICParticleType &p);
DepositeMass(RadParticleType &p);
DepositeMass(SinkParticleType &p);
...
```

this is exactly the thing we want to avoid when you and Ben first desighed this architecture.

chongchonghe: OK, I'll make that change in another PR. I think we can keep this PR as is because it's consistent with the current implementation.

src/simulation.hpp
chongchonghe: yes, as I explained in the other comments, this is already the case. it applies to all particle types that belong to the group of 'StarParticle' types.

src/particles/PhysicsParticles.hpp
markkrumholz: OK, that's fine.

markkrumholz (approved):

PR #1219, opened 2025-07-14
[WIP] move particleMeshInteraction in front of kick-drift-kick

chongchonghe:
### Description
Move `particleMeshInteraction` in front of particle movement. As discussed earlier, this would allow us to use all the ghost cells (4) for particle-mesh interaction. Additionally, this makes more sense because particles should deposit radiation or other feedback into hydro cells before hydro/radiation update. 

### Related issues

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -1084,6 +1084,15 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 		computeBeforeTimestep();
 
 #if AMREX_SPACEDIM == 3
+		// TODO(cch): Need to take care of AMR subcycling
+		particleRegister_.destroyParticles(0, cur_time, dt_[0]);
+
+		// Stellar evolution and SN deposition; only apply to star particles
+		if (particleRegister_.HasStarParticles()) {
+			// TODO(cch): Need to take care of AMR subcycling
+			particleMeshInteraction(cur_time, dt_[0]);
+		}
+
 		// do particle leapfrog (first kick at time t)
 		if constexpr (Physics_Traits<problem_t>::is_self_gravity_enabled) {
 			kickParticlesAllLevels(dt_[0]);
@@ -1111,16 +1120,6 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 		if constexpr (Physics_Traits<problem_t>::is_self_gravity_enabled) {
 			kickParticlesAllLevels(dt_[0]);
 		}
-
-		// Stellar evolution and SN deposition; only apply to star particles
-		if (particleRegister_.HasStarParticles()) {
-			// TODO(cch): Need to take care of AMR subcycling
-			particleMeshInteraction(cur_time, dt_[0]);
-		}
-
-		// Use the new type-aware particle destruction method
-		// TODO(cch): Need to take care of AMR subcycling
-		particleRegister_.destroyParticles(0, cur_time, dt_[0]);
 #endif
 
 		cur_time += dt_[0];
--- a/src/problems/ParticleSink/test_particle_sink.cpp
+++ b/src/problems/ParticleSink/test_particle_sink.cpp
@@ -182,7 +182,7 @@ auto problem_main() -> int
 	const double total_total_mass_init = total_mass_init + total_particle_mass;
 
 	// evolve
-	sim.maxTimesteps_ = 1;
+	sim.maxTimesteps_ = 0;
 	sim.evolve();
 
 	// get total gas mass in the final state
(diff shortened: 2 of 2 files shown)


chongchonghe:
@markkrumholz Ready for your review. I would like to merge this before #1215

BenWibking:
I don't think this is a good idea for SN feedback, because it will change the CFL-stable timestep.

In principle, the hydro retry mechanism can save the simulation from crashing, but it may change the timestep by a large enough factor that it still crashes. If the simulation doesn't crash, there is still an accuracy issue since each level retries the timestep independently of the other levels.

I think it would be better if we either i) deposited the SN feedback after hydro (as we currently do), or ii) computed the timestep *after* doing the particle deposition. However, we can't easily generalize the latter to work with AMR subcycling.

src/simulation.hpp
```diff
@@ -1084,6 +1084,15 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 		computeBeforeTimestep();
 
 #if AMREX_SPACEDIM == 3
+		// TODO(cch): Need to take care of AMR subcycling
+		particleRegister_.destroyParticles(0, cur_time, dt_[0]);
+
+		// Stellar evolution and SN deposition; only apply to star particles
+		if (particleRegister_.HasStarParticles()) {
+			// TODO(cch): Need to take care of AMR subcycling
+			particleMeshInteraction(cur_time, dt_[0]);
```
BenWibking, line 1093: This will change the CFL-stable timestep. See my longer comment. I think this change needs very careful consideration.

BenWibking (changes requested):

BenWibking:
Unfortunately, I think this PR will really break things, so I've added a "do not merge" label for now.

chongchonghe:
> I don't think this is a good idea for SN feedback, because it will change the CFL-stable timestep.
> 
> In principle, the hydro retry mechanism can save the simulation from crashing, but it may change the timestep by a large enough factor that it still crashes. If the simulation doesn't crash, there is still an accuracy issue since each level retries the timestep independently of the other levels.
> 
> I think it would be better if we either i) deposited the SN feedback after hydro (as we currently do), or ii) computed the timestep _after_ doing the particle deposition. However, we can't easily generalize the latter to work with AMR subcycling.

OK, I didn't think about that. We would definitely have to compute the CFL condition after feedback, or after any particle-mesh interaction in general. I'm not sure why there is a problem with your option (ii), but I'll set this PR to draft and come back to it later.

chongchonghe:
Strange. "Turn into draft" button is missing.

BenWibking:
> > I don't think this is a good idea for SN feedback, because it will change the CFL-stable timestep.
> > In principle, the hydro retry mechanism can save the simulation from crashing, but it may change the timestep by a large enough factor that it still crashes. If the simulation doesn't crash, there is still an accuracy issue since each level retries the timestep independently of the other levels.
> > I think it would be better if we either i) deposited the SN feedback after hydro (as we currently do), or ii) computed the timestep _after_ doing the particle deposition. However, we can't easily generalize the latter to work with AMR subcycling.
> 
> OK, I didn't think about that. We would definitely have to compute the CFL condition after feedback, or after any particle-mesh interaction in general. I'm not sure why there is a problem with your option (ii), but I'll set this PR to draft and come back to it later.

The timestep computation is global over all levels and happens before calling the advanceWithSubcycling function, but the feedback needs to be deposited at each level during each subcycle. So we can't deposit all the feedback before the calling ComputeTimestep() if we are doing AMR subcycling.

chongchonghe:
A primary motivation for doing particle-mesh interaction before `ellipticSolveAllLevels` is we can use all the ghost cells for particle-mesh interaction. This gives us 4 zones for current RHD solver and 6 zones for RMHD. This is great because sink particle + stellar wind would require more cells. With the current implementation of particle-mesh interaction after redistribution in ellipticSolveAllLevels, we have to use one fewer cells. Then, the best we can do is 5 cells (after we merge MHD): 3 cells for accretion and 2 cells for winds. With this change, we can do 3+3.

BenWibking:
> A primary motivation for doing particle-mesh interaction before `ellipticSolveAllLevels` is we can use all the ghost cells for particle-mesh interaction. This gives us 4 zones for current RHD solver and 6 zones for RMHD. This is great because sink particle + stellar wind would require more cells. With the current implementation of particle-mesh interaction after redistribution in ellipticSolveAllLevels, we have to use one fewer cells. Then, the best we can do is 5 cells (after we merge MHD): 3 cells for accretion and 2 cells for winds. With this change, we can do 3+3.

Yes, I understand the motivation. The problem is that it just won't work with AMR level subcycling in time. Since we want this to work in the future, I don't think it makes sense to go down this path, since that would require a separate code path for global timestep vs subcycling.

I think the way to get around the limited number of ghost cells is to just do another `fillBoundaryConditions` call before doing particle deposition and feedback. Then we can fill as many ghost cells as needed for whatever particle deposition stencil we are using. It only requires nearest neighbors communication, so there is no scaling issue with doing so.

BenWibking:
For the reasons discussed above, I think this is not the optimal approach to solving this problem. @chongchonghe, if it's okay with you, can we close this PR?

PR #1222, opened 2025-07-15
Add bounds_error parameter to interpolate_value function

### LINKED ISSUE #1221: add tramp paremeter to `interpolate_value`
**Describe the proposal**
We should add a parameter `bool bounds_error = true` to `interpolate_value` in interpolate.hpp and interpolate.cpp. If set to false, the interpolation function would return the first and last element of arr_y if x is out of bounds of arr_x. Keep a default value bounds_error = true to be consistent with the rest of the code. We should also search the codebase and find where interpolate_value is used with this clamping behaviour and replace it with `bounds_error=false` if possible. 

@claude Please help me create a PR to implement this. 

A clear and concise description of any alternative solutions you've considered.

**Additional context**

chongchonghe:
Updated `interpolate_value` function to a template function with a default boundary policy parameter (`BoundaryPolicy::ErrorOnBounds`). When called without explicitly specifying the template parameter, it defaults to the `ErrorOnBounds` policy, which returns `NAN` and asserts if the input value `R` falls outside the bounds of the `R_table` array. This is consistent with previous implementation, so no change is required if `ErrorOnBounds` is the desired behaviour. 

Fixes #1221

PR created by [Claude Code](https://claude.ai/code) and @chongchonghe .


chongchonghe:
This would partially fix the bugs in #712 (although you can always do the clamping outside of `bounds_error`. I understand that we have plans to have a better and more comprehensive handle of interpolation bounds using templates, but the change introduced in this PR is immediately usable and fix some problems. So, I think we can merge this.

src/math/interpolate.cpp
```diff
@@ -133,7 +131,7 @@ void interpolate_arrays(double *x, double *y, int len, double *arr_x, double *ar
 }
 
 AMREX_GPU_HOST_DEVICE
-double interpolate_value(double x, double const *arr_x, double const *arr_y, int arr_len)
+auto interpolate_value(double x, double const *arr_x, double const *arr_y, int arr_len, bool bounds_error) -> double
```
BenWibking: Two comments:
1. Can you name this something more directly related to the effect on the interpolation? e.g., `constant_extrapolation`? That seems more intuitive to understand its effect. 
2. I think it would be easy to change this to a template parameter. Can you do that? That would avoid the negative performance impact of having an additional local variable due to increased register usage.

BenWibking (changes requested):

BenWibking:
@claude Can you suggest how to modify this PR so the new function argument to interpolate_value is instead a template parameter that passes an opaque type to change the behavior of the function?

chongchonghe:
@claude  Make the following changes and push to this PR:
1. define a enum class for BoundaryPolicy with these options: ErrorOnBounds, Clamp, Extrapolation.
2. template interpolate_value with BoundaryPolicy, and deal with out-of-bounds as the name suggests.

src/math/interpolate.cpp
chongchonghe: OK. I'll ask Claude to make it a enum template.

chongchonghe:
@claude Your design looks good. Please implement it.

src/math/interpolate.cpp
```diff
@@ -132,8 +130,8 @@ void interpolate_arrays(double *x, double *y, int len, double *arr_x, double *ar
 	}
 }
 
-AMREX_GPU_HOST_DEVICE
-double interpolate_value(double x, double const *arr_x, double const *arr_y, int arr_len)
+template<BoundaryPolicy Policy>
+AMREX_GPU_HOST_DEVICE auto interpolate_value(double x, double const *arr_x, double const *arr_y, int arr_len) -> double
```
BenWibking: I think this has to go in `interpolate.hpp` now since it's a template function.

BenWibking (changes requested):
Looks good, but I think template functions should go in headers rather than in .cpp files.

src/math/interpolate.cpp
chongchonghe: can we get rid of the cpp file? is it necessary at all?

BenWibking: If everything is a template function and/or marked `inline`, then it's not needed.

chongchonghe: OK, I see. But it's not, so we have to keep it.

BenWibking: Hmm. Where is the non-template function?

src/math/interpolate.cpp
```diff
@@ -26,15 +27,17 @@
  * @return index
  */
 AMREX_GPU_HOST_DEVICE
-int64_t binary_search_with_guess(const double key, const double *arr, int64_t len, int64_t guess)
+auto binary_search_with_guess(const double key, const double *arr, int64_t len, int64_t guess) -> int64_t
```
BenWibking: If we mark this as `inline`, then we can get rid of the .cpp file.

src/math/interpolate.cpp
chongchonghe: binary_search_with_guess, but we can make it inline as you suggested.

src/math/interpolate.cpp
chongchonghe: sure. I'll do that.

chongchonghe:
@BenWibking Can you review this?

chongchonghe:
This PR is ready to merge unless you recommend more changes. @BenWibking

BenWibking (approved):

PR #1243, opened 2025-07-30
Update DataTable to read HDF5 files and do table interpolation up to four dimensions

chongchonghe:
This new DataTable class provides reading and interpolation of cooling rates and thermodynamic quantities using resampled lookup tables. The implementation features a generic n-dimensional (1 <= n <= 4) `DataTable` class with HDF5 integration.

### DataTable Class

The `DataTable<Ndim, Nout>` template class provides:

- **N-dimensional support**: Handle 1D-4D lookup tables with compile-time dimension checking
- **Multiple outputs**: Single coordinate system can map to multiple physical quantities
- **High-performance interpolation**: N-linear interpolation (linear, bilinear, trilinear, quadrilinear)
- **GPU compatibility**: GPU-friendly const table access for device kernels

#### Template Parameters

```cpp
template <int Ndim, int Nout = 1>
class DataTable
```

- `Ndim`: Number of coordinate dimensions (1-4)
- `Nout`: Number of output quantities sharing the same coordinate system

#### Supported Interpolation Methods

- **1D**: Linear interpolation
- **2D**: Bilinear interpolation  
- **3D**: Trilinear interpolation
- **4D**: Quadrilinear interpolation

### H5Reader Integration

The `H5Reader` static method provides seamless HDF5 integration for reading cooling tables:

```cpp
auto table = DataTable<2, 1>::H5Reader(file_id, "/data/cooling_rates", 
                                        {"rho", "eint"}, 1, &coord_bounds);
```

Cooling table files use a standardized HDF5 structure:

```
cooling_table.h5
├── /metadata          # Grid metadata and bounds
├── /grids             # Coordinate arrays
└── /data              # Physical quantity datasets
```

#### Metadata Group (`/metadata`)

Contains attributes describing the grid structure and physical bounds:

```
/metadata
├── n_rho              # Number of density grid points (int)
├── n_eint             # Number of specific energy grid points (int)  
├── rho_min            # Minimum density [g/cm³] (double)
├── rho_max            # Maximum density [g/cm³] (double)
├── eint_min           # Minimum specific energy [erg/g] (double)
├── eint_max           # Maximum specific energy [erg/g] (double)
└── cloudy_H_mass_fraction  # Hydrogen mass fraction used in Cloudy (double)
```

**Generic Pattern**: For coordinate `<coord>`, expect attributes:
- `n_<coord>`: Grid size
- `<coord>_min`: Minimum value  
- `<coord>_max`: Maximum value

#### Coordinate Grids (`/grids`)

Contains 1D arrays defining the coordinate grid points:

```
/grids
├── fast_log_rho       # log₁₀(ρ) grid points [dimensionless] (1D double array)
└── fast_log_eint      # log₁₀(e_int) grid points [dimensionless] (1D double array)
```

**Generic Pattern**: For coordinate `<coord>` with `is_fast_log=1`:
- Dataset name: `fast_log_<coord>`
- Without fast_log: `<coord>`

#### Data Group (`/data`)

Contains 2D arrays with physical quantities on the coordinate grid:

```
/data
├── cooling_rates      # Λ/ρ² [erg⋅cm³⋅g⁻²⋅s⁻¹] (2D double array: n_rho × n_eint)
├── temperatures       # Temperature [K] (2D double array: n_rho × n_eint)  
├── sound_speeds       # Sound speed [cm/s] (2D double array: n_rho × n_eint)
├── pressures          # Pressure [dyne/cm²] (2D double array: n_rho × n_eint)
└── entropies          # Entropy [erg⋅g⁻¹⋅K⁻¹] (2D double array: n_rho × n_eint)
```

#### Data Layout

All N-D datasets use **C-order** (row-major) layout:
- First dimension: output index
- Second dimension: first coordinate index
- Third dimension: second coordinate index
- ...
- Last dimension: N-th coordinate index
- Array element `[i,j,k,...]` corresponds to `(output[i], coord[0][j], coord[1][k], ...)`

#### Single-Output Tables

```cpp
#include "util/DataTable.hpp"

// Read cooling table from HDF5 file
hid_t file_id = H5Fopen("cooling_table.h5", H5F_ACC_RDONLY, H5P_DEFAULT);

std::vector<std::string> coords = {"rho", "eint"};
std::array<std::pair<amrex::Real, amrex::Real>, 2> bounds;

auto cooling_table = quokka::DataTable<2, 1>::H5Reader(
    file_id, "/data/cooling_rates", coords, 1, &bounds);

H5Fclose(file_id);

// Interpolate cooling rate at specific (ρ, e_int)
double rho = 1e-24;     // g/cm³
double eint = 1e12;     // erg/g  
std::array<amrex::Real, 2> point = {FastMath::fastlg(rho), FastMath::fastlg(eint)};

double cooling_rate = cooling_table.const_tables().interpolate_single(point);
```

#### Multi-Output Tables

Suppose a H5 file contains a dataset called `/data/data` with three dimensions, the first dimension is the outputs: (e.g. temperature, pressure, entropy), and the rest are the coordinate grids. The following code reads the table and interpolates all outputs simultaneously.

```cpp
// Table with multiple outputs (temperature, pressure, entropy)
auto thermo_table = DataTable<2, 3>::H5Reader(file_id, "/data/data", coords, 1);

// Interpolate all outputs simultaneously  
std::array<amrex::Real, 3> results = thermo_table.const_tables().interpolate(point);
// results[0] = temperature, results[1] = pressure, results[2] = entropy
```

### Next

Need @BenWibking 's help to regenerate the H5 file so that we can use the following generic format (align with JAFF):

### Related issues
None.

### Checklist


chongchonghe:
I'll implement reading tables of arbitrary dimensions next.

BenWibking:
The HDF5 API is so ugly that I think it's worth hiding the H5Open/H5Close calls inside the DataTable constructor:
```
// Read cooling table from HDF5 file
std::vector<std::string> coords = {"rho", "eint"};
std::array<std::pair<amrex::Real, amrex::Real>, 2> bounds;

auto cooling_table = quokka::DataTable<2, 1>::H5Reader(
    "cooling_table.h5", "/data/cooling_rates", coords, 1, &bounds);
```
It will be slower, but it doesn't matter.

BenWibking:
@claude Provide a review of this PR.

BenWibking:
Ok, I agree with the AI about its suggestions for improvement.

Unit tests would also be nice, but since we don't have any of those for the rest of code, I'm okay with not including those.

chongchonghe:
@BenWibking I addressed all the comments and updated the description.

chongchonghe:
There are a little work to do in order to make this consistent with the JAFF format. Once we agree on a final format, I'll update the code to use the new format, and I'll need @BenWibking 's help to regenerate the HDF5 file with the corresponding format.

BenWibking (approved):

markkrumholz:
Looking at this design, I think there are a couple of issues:

1. The format is not fully self-documenting because nothing in the HDF5 file tells you how the inputs dimensions are ordered. That is, in the given example, someone who was handed the file would have no way of knowing whether rho is coordinate 1 and eint is coordinate 2 or vice-versa. I therefore think that there needs to be a metadata entry giving the names of the coordinates in order.
2. The description of how multiple outputs are handled is contradictory. In the example given it looks like if there are N_out outputs, they are stored in N_out distinct datasets, each of size `Nin1 x Nin2 x ...` -- but then the description of the data layout seems to imply that multiple outputs are stored in a single data set of size `Nout x Nin1 x Nin2 x ....` We need to pick one or the other and be clear about which we're picking. There are advantages and disadvantages to both options -- packing the data into a single array is probably more efficient computationally if the typical use is always to interpolate all quantities at once, while keeping them separate is probably preferable if the typical use case is to interrogate one quantity at a time, but we should probably pick one.
3. Related to the previous point, if we pack the data into a single array, we need a standardized way of specifying the names of the output quantities. Right now the information that, for example, cooling_rates are the first output, temperatures are the second output, etc., are embedded in the names of the data sets, but if we're going to pack the data in to a single array we can't rely on that strategy and instead need to explicitly store the names of the output quantities. Related to this, for multiple output tables it would probably be useful to let users specify that they want to interpolate a single quantity and return a scalar, and describe that single quantity using either a string or the quantity number, in addition to interpolating all quantities.
4. The treatment of fast_log spacing is inconsistent. First, if we store the positions of the data points rather than their logarithms for log spacing, why don't we do the same for fast_log? This is confusing and likely to cause errors. I propose that we always store the value of the variable itself, no matter what spacing we are using. Second, the `is_fast_log` variable is not easily extensible if, as @BenWibking suggests, we might at some point want to implement the alternative v2 fast_log, and support that as well as the current fast_log. I therefore propose that we get rid of `is_fast_log` and instead have an attribute that is a string that gives the spacing, and which can be "linear", "log", and "fast_log", and which in the future we can extend to "fast_logv2" or something like that. Depending on how ambitious we want to be, this can be either uniform for all inputs, or can be different for different dimensions.

markkrumholz (changes requested):
See my comments above on the data format design.

chongchonghe:
@markkrumholz  I forgot to explain that some of the design consideration in this PR is for backward compatibility and they are temporary. Particularly, note that this cooling module is a *one-output* interpolation problem, because only the cooling rate is ever used in `resampled_cooling_function`. The other quantities (temperatures, sound_speeds, etc) are only used in some test problems (cloud.cpp) for reporting. So, what I am doing here is interpolating to one output, although the code support multiple outputs.

markkrumholz:
Doesn't the code also need the pressure or sound speed at some point?

chongchonghe:
Since now we have agreed on the final format of the HDF5 file, let me update this PR to it's final version and include all the features we have discussed. @BenWibking can you help to regenerate the HDF5 with the following format?
1. The main data should be stored in '/data/data', so we should rename '/data/cooling_rates' to '/data/data', and keep the other outputs unchanged. For future applications where there are multiple outputs that share the same coordinates and *we use all of them in the application*, we should store them in a single data object '/data/data' with the first dimension being Nout. 
2. In the metadata, we need
   - `Ndim = 2`
   - `Nout = 1`
   - `input_names = {'rho', 'eint'}` (suppose HDF5 supports list of strings)
   - `output_names = {'cooling_rates'}`
   - `input_sizes = {n_rho, n_eint}` (replace with actual values)
   - `input_units = {'g/cm3', 'erg/g'}`
   - `output_units = {'erg cm3 g-2 s-1'}` (please check if the units are correct)
   - `xlo = {rho_min, eint_min}` (replace with actual values)
   - `xhi = {rho_max, eint_max}` (replace with actual values)
   - `spacing = {'fast_log', 'fast_log'}`
3. For this HDF5 file, we don't need '/grids' since the grids are regular and defined in metadata

chongchonghe:
> Doesn't the code also need the pressure or sound speed at some point?

In the cooling calculation, no, only cooling rate is used, because that's all you need to evolve the internal energy.

markkrumholz:
OK, I think we're agreed on the format. Just to put it all in one place for the record, and to make sure we agree, here's what I think the final converged format is:

1. There will be one group called `data`.
2. The `data` group will contain the following attributes:

- `Ndim` (integer): number of dimensions of input
- `Nout` (integer): number of dimensions of output
- `input_names` (array of string with Ndim entries): names of all input quantities
- `output_names` (array of string with Nout entries): names of all output quantities
- `input_units` (array of string with Ndim entries): units of all input quantities (unitless quantities are empty strings)
- `output_units` (array of string with Nout entries): units of all output quantities (unitless quantities are empty strings)
- `xlo` (array of real with Ndim entries): lower limits on the range input quantities; even if quantities use logarithmic spacing, we these values are the quantity itself, not its logarithm; in the future if we expand to support irregular grids, some entries may be left as NaN or unused
- `xhi` (array of real with Ndim entries): upper limits on the range of input quantities; same notes as `xlo`
- `spacing` (array of string with Ndim entries): array specifying how each dimension is spaced; allowed entries are `linear`, `log`, `fast_log`, and in the future we may also support `irregular` and `fast_log2`.

3. The `data` group will contain one dataset, called `data`, which will be of shape `[Nout, Nin1, Nin2, ... NinNdim]`. The data stored in this dataset are the values themselves, not their logarithms, even if we are using logarithmic interpolation.
4. There can also be a second group called `grids` which can contain `Ndim` datasets, with names matching those in `input_names`, each of which is a size `NinN` array of real containing the locations of each data point in each dimension. This will be mandatory if and when we have irregularly-sampled dimensions, but for now is optional.

Assuming we are agreed on that, I'm happy to approve and merge.

markkrumholz:
Notes for possible future expansions:

1. As mentioned above, we may at some point support irregularly-sampled data and data sampled in 2nd generation fast logarithm.
2. For cases where there is more than one output, we should support both interpolation of all quantities (returning an array of `Nout` entries and interpolation of a single quantity, which can be specified either by index number of by name.
3. In principle we could also specify the interpolation type separately from the spacing, in that we could interpolate on either the log of the stored quantity or on the quantity itself. This would allow, for example, log-linear or linear-log interpolation.

Again, just recording these ideas here for future reference. We certainly do not need to implement these now.

chongchonghe:
All looks good; just one minor comment: my original design (which is also the current implementation) has `/data/data`, `/metadata/xxx`, and `/grids/xxx`, but your idea is to put both `data` and `metadata` entries inside `/data/`. Just to point this out, but I don't have opinion on this and either one is okay.

markkrumholz:
Since attributes are generally attached to the same group as the dataset, I think it makes more sense to put them into the same place as the data themselves. I therefore favor there being just a `data` group, with the metadata as attributes within it.

BenWibking:
Ok, so this format does not support having multiple tables in the same file? For tables that don't need to be output as vectors, or that have different axes, that would be useful to have.

BenWibking:
@chongchonghe You can modify the Python code that outputs the HDF5 files in these two places:
* https://github.com/quokka-astro/quokka/blob/195ee995a4e3ef8ff49e0a3ac17b6299c7b18b5f/extern/cooling/resample_cloudy_cooling_tables.py#L359
* https://github.com/quokka-astro/quokka/blob/195ee995a4e3ef8ff49e0a3ac17b6299c7b18b5f/extern/cooling/resample_grackle_cooling_tables.py#L289

markkrumholz:
> Ok, so this format does not support having multiple tables in the same file? For tables that don't need to be output as vectors, or that have different axes, that would be useful to have.

I suppose we could generalize this by having multiple data sets, which we could label data1, data2, etc. I think for tables that share axes there is no benefit to separating tables into different data sets -- we can handle the case of extracting just one quantity from a vector of quantities by providing a method to extract a single scalar quantity by name or number, as I suggested -- but I agree that one might want to put data sets with different axes into a single hdf5 file.

BenWibking:
Right. I think the way to handle the case of a single scalar quantity from a table that includes multiple quantities is to provide a helper object that precomputes the offset in the table appropriate for that component. That way, the component can be specified outside the `amrex::ParallelFor` and we reduce the number of variables needed  to compute the index inside the ParallelFor by one (which is significant on the GPU).

chongchonghe:
> Ok, so this format does not support having multiple tables in the same file? For tables that don't need to be output as vectors, or that have different axes, that would be useful to have.

It does support this. Simply store the data under whatever name you like, e.g. '/data/temperature', or for better readability, '/data/data_temperature'

chongchonghe:
nothing has to be changed. i'll keep the api as the current one: the default data name is '/data/data', which could be one-output or multi-output. you can optionally pass a data_name variable to the H5Reader function to select the data you want.

markkrumholz (approved):
OK, I'm happy to consider this converged. I will approve now.

markkrumholz:
I'm holding off on merging just to make sure that there are no further changes that @chongchonghe wants to make, but I'm satisfied now.

chongchonghe:
This PR is fully functional and coherent itself, so i'm okay with merging it right now or implementing what we discussed above before merging. Since both of you have approved, i'll merge it now and make more updates in another PR.

markkrumholz:
Just a note as I am coding up the jaff implementation of this standard that the method of handling multiple data sets in the same file has to be slightly different than we were saying. The format has to be that each group is a separate data set, and groups can be called anything. That is, if we want a file to contain two lookup tables `tab1` and `tab2` with different grids, the structure needs to be that there are two groups called `tab1` and `tab2`, and that within each group you have the set of attributes listed above describing the grid layout and interpolation method, and then the data within that group are stored in a dataset called `data`. So for example you could have two data sets, `tab1/data` and `tab2/data`. It doesn't work to have `data/tab1` and `data/tab2` as @chongchonghe was suggesting, because then you don't know which of the attributes in the `data` group apply to which table.

PR #1263, opened 2025-08-11
Fix race conditions in ComputeEMF

### LINKED ISSUE #1258: GPU race condition in MHD code
**Describe the bug**
There is a GPU race condition in the MHD code that causes incorrect results. I think this is because ComputeEMF loops over all the boxes with an MFIter, but the rest of the hydro update uses `ParallelFor`s over Multifabs.

**As a temporary workaround, if you are running MHD problems on GPU, please add `amrex.max_gpu_streams=1` to your command-line options.**

**To Reproduce**
See https://github.com/quokka-astro/quokka/pull/1142#issuecomment-3172857993.

BenWibking:
### Description
Eliminates the race condition present in the ComputeEMF function on GPU by only launching into the default AMReX stream. This only fixes non-AMR MHD runs.

The cause of the race condition is as follows: when there are multiple boxes on a given GPU, each box in an MFIter loop is normally launched into a separate stream. These extra streams (beyond the first stream) may start executing their kernels before the previous kernels in the hydro update have finished (which all use the first stream, since they execute over the whole MultiFab, rather than individual boxes).

This has a potential negative performance impact. A performance-optimal fix is deferred to a later PR.

### Related issues
Fixes https://github.com/quokka-astro/quokka/issues/1258.

### Checklist

### DIFF
--- a/docs/markdown/debugging.md
+++ b/docs/markdown/debugging.md
@@ -57,6 +57,35 @@ This will cause the CPU to wait until the GPU kernel execution is complete befor
 
 For more details, refer to the [AMReX GPU debugging guide](https://amrex-codes.github.io/amrex/docs_html/Debugging.html#basic-gpu-debugging).
 
+## Testing for GPU race conditions
+
+GPU race conditions can cause non-deterministic behavior where the same simulation produces different results on different runs, or crashes only intermittently. To systematically test for race conditions, use the provided test script:
+
+```bash
+./scripts/bash/test_gpu_race.sh -b <binary> -i <input_file> -n <max_timesteps>
+```
+
+**Parameters:**
+- `-b <binary>`: Path to the test binary (e.g., `./build/src/FieldLoop/test_field_loop`)
+- `-i <input_file>`: Path to the input file (e.g., `inputs/field_loop.in`)
+- `-n <max_timesteps>`: Maximum number of timesteps to run
+
+**Example:**
+```bash
+./scripts/bash/test_gpu_race.sh -b ./build/src/FieldLoop/test_field_loop -i inputs/field_loop.in -n 10
+```
+
+**How it works:**
+1. Runs the same simulation twice - once with `CUDA_LAUNCH_BLOCKING=1` (synchronous) and once with `CUDA_LAUNCH_BLOCKING=0` (asynchronous)
+2. Compares the final plotfiles using AMReX's `fcompare` tool with zero tolerance
+3. If the results differ, this indicates a race condition where kernel execution order affects the outcome
+
+**Interpreting results:**
+- **SUCCESS**: Plotfiles are identical - no race condition detected
+- **FAILURE**: Plotfiles differ or non-blocking run crashes - race condition detected
+
+The script preserves temporary files for analysis when a race condition is found. These can be examined to understand the nature of the differences between the two runs.
+
 ## When all else fails: Debugging with `printf`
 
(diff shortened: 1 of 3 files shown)


BenWibking:
The root cause of the race condition is explained in the [AMReX documentation](https://amrex-codes.github.io/amrex/docs_html/GPU.html#fig-gpu-streams):

> AMReX further parallelizes GPU applications by utilizing streams. Streams guarantee execution order of kernels within the same stream, while allowing different streams to run simultaneously. AMReX places each iteration of MFIter loops on separate streams, allowing each independent iteration to be run simultaneously and sequentially, while maximizing GPU usage.
> 
> The AMReX implementation of streams is illustrated in [Fig. 13](https://amrex-codes.github.io/amrex/docs_html/GPU.html#fig-gpu-streams). The CPU runs the first iteration of the MFIter loop (blue), which contains three GPU kernels. The kernels begin immediately in GPU Stream 1 and run in the same order they were added. The second (red) and third (green) iterations are similarly launched in Streams 2 and 3. The fourth (orange) and fifth (purple) iterations require more GPU resources than remain, so they have to wait until resources are freed before beginning. Meanwhile, after all the loop iterations are launched, the CPU reaches a synchronize in the MFIter’s destructor and waits for all GPU launches to complete before continuing.

cc @markkrumholz @AstroKriel

BenWibking:
@AstroKriel @markkrumholz Weiqun has pointed out there is a more fundamental issue with the ComputeEMF function, which is that the temporary FABs are not async-safe. I forgot about this issue, since it last came up when I originally wrote the hydro code for Quokka in 2021.

The issue (and solution) is to use `amrex::The_Async_Arena()` when allocated temporary FABs: https://amrex-codes.github.io/amrex/docs_html/GPU.html#async-arena

@AstroKriel will need to modify the code to use this instead of calling `.resize()` for the temporary FABs. You can look at the radiation code to see an example of how to do this.

BenWibking:
> @AstroKriel @markkrumholz Weiqun has pointed out there is a more fundamental issue with the ComputeEMF function, which is that the temporary FABs are not async-safe. I forgot about this issue, since it last came up when I originally wrote the hydro code for Quokka in 2021.
> 
> The issue (and solution) is to use `amrex::The_Async_Arena()` when allocated temporary FABs: https://amrex-codes.github.io/amrex/docs_html/GPU.html#async-arena
> 
> @AstroKriel will need to modify the code to use this instead of calling `.resize()` for the temporary FABs. You can look at the radiation code to see an example of how to do this.

BenWibking:
@claude Modify the `ComputeEMF` function so that all of the temporary `amrex::FArrayBox`es are allocated with amrex::The_Async_Arena(), as described in the AMReX documentation (https://amrex-codes.github.io/amrex/docs_html/GPU.html#async-arena). Also, make sure there are no calls to `.resize()` for any of the FABs.

This is needed to avoid the FArrayBoxes going out of scope before the GPU kernel runs.

BenWibking:
@markkrumholz Can you review this PR? This should be the last (I think) fix needed for uniform grid MHD simulations.

markkrumholz:
Something seems to have failed in the GPU build on avatargpu. I'll try running again, since it looks like a failure in the agent rather than the build itself.

markkrumholz (approved):

markkrumholz:
I removed from the merge queue while waiting for the GPU build tests to run. This can be merged once they pass.

BenWibking:
GPU pass!

Issue #1266, opened 2025-08-16
Multifab containers with strongly-typed components


BenWibking:
**Describe the proposal**
We can solve a lot of component indexing problems by using C++ template magic. Namely, we can use [variadic template parameter packs](https://en.cppreference.com/w/cpp/language/parameter_pack.html) to encode the types of each components.

Then variables could be declared like this:
```
#define VARIABLE(ns, varname)                                                            \
  struct varname : public variable_names::base_t<false> {                     \
    template <class... Ts>                                                               \
    AMREX_INLINE varname(Ts &&...args)                                         \
        : variable_names::base_t<false>(std::forward<Ts>(args)...) {}         \
    static std::string name() { return #ns "." #varname; }                               \
  }

namespace Conserved {
VARIABLE(advection, phi);
VARIABLE(advection, C);
VARIABLE(advection, D);
} // namespace Conserved
```
And we could return `quokka::TypedMultifab` objects:
```
using HydroTypeList = TypeList<Conserved::phi_fine, Conserved::C, Conserved::D>;
auto mf = quokka::TypedMultifab<HydroTypeList>(ba, dmap, ng);
mf.arrays[box](i, j, k, Conserved::C()) = 0.;
```

This means we can ensure that physics functions (e.g., in HydroSystem, RadiationSystem, MHDSystem, etc.) are passed Multifabs with the exact set of components that are desired, and that is impossible to pass a typed Multifab with the wrong set of components.

Also, the underlying indexing of the components can be changed "behind-the-scenes" without any effect on the rest of the code.

**Describe alternatives you've considered**
Continue using Physics_Indices and the enums defined in HydroSystem, RadSystem, etc. for component indexing.

**Additional context**
In another AMR library, this is implemented as follows:
```
// Convenience struct for holding a variadic pack of types
// and providing compile time indexing into that pack as
// well as the ability to get the index of a given type within
// the pack. Functions are available below for compile time
// concatenation of TypeLists
template <class... Args>
struct TypeList {
  using types = std::tuple<Args...>;

  static constexpr std::size_t n_types{sizeof...(Args)};

  template <std::size_t I>
  using type = typename std::tuple_element<I, types>::type;

  template <std::size_t... Idxs>
  using sublist = TypeList<type<Idxs>...>;

  template <class T, std::size_t I = 0>
  static constexpr std::size_t GetIdx() {
    static_assert(I < n_types, "Type is not present in TypeList.");
[...]

PR #1268, opened 2025-08-18
Balsara Method for ComputeEMF

lizmcole:
### Description
Uses the method from Balsara's 2025 paper to compute the emf by getting v x B at cell center, then reconstructing face-centered B-fields to the edge, and then using equations 3.3, 3.5, 3.9, and 3.10 to get the EMF at the edge.

### Related issues
This still has the GPU problem with MHD. e.g.,
https://github.com/quokka-astro/quokka/issues/1258
https://github.com/quokka-astro/quokka/issues/1266

### Checklist
- [x ] I have added a link to any related issues (if applicable; see above).


src/hydro/mhd_system.hpp
```diff
@@ -158,7 +163,7 @@ void MHDSystem<problem_t>::ComputeEMF(std::array<amrex::MultiFab, AMREX_SPACEDIM
 				    amrex::grow(box_cc, (nghost_cc - 1) * vec_fc2ec); // note, the reconstruct function will uniformly grow the bounds by 1
 				const amrex::Box box_fc_U = amrex::grow(box_fc, (nghost_cc - 1) * vec_fc2ec + 1);
 
-				// extrapolate both required cell-centered velocity fields to the cell-edge
+				// extrapolate cell-centered electic fields to the cell-edge
```
markkrumholz: Typo -- "electic"

src/hydro/mhd_system.hpp
```diff
@@ -329,6 +334,209 @@ void MHDSystem<problem_t>::ComputeEMF(std::array<amrex::MultiFab, AMREX_SPACEDIM
...
+	// significantly reduces the total memory used, which is a much bigger bottleneck.
+
+	// loop over each box-array on this level
+	constexpr int nstreams = 1; // only run on 1 GPU stream to avoid race conditions
+	for (amrex::MFIter mfi(cc_mf_cVars, amrex::MFItInfo().SetNumStreams(nstreams)); mfi.isValid(); ++mfi) { // keep
+		const amrex::Box &box_cc = mfi.validbox();
+		std::array<amrex::FArrayBox, 3> cc_fabs_EMF;
+		{
+			const amrex::Box &box_cc_EMF = amrex::grow(box_cc, nghost_cc); // only need cc ghost zones for Balsara method
+			cc_fabs_EMF[0].resize(box_cc_EMF, 1);
```
markkrumholz: See the discussion of the problems of using resize as part of #1263. This should be rewritten using the fix proposed there.

src/hydro/mhd_system.hpp
```diff
@@ -329,6 +334,209 @@ void MHDSystem<problem_t>::ComputeEMF(std::array<amrex::MultiFab, AMREX_SPACEDIM
...
+				const auto &emf_cc = fcx_mf_fspds[w1_comp][mfi].const_array();
+
+				// LLF variant
+				amrex::ParallelFor(box_ec, [=] AMREX_GPU_DEVICE(int i, int j, int k) {
+					const double S0_m = std::max(fspd_x0(i, j, k, 0), fspd_x0(i + delta_w0[0], j + delta_w0[1], k + delta_w0[2], 0));
+					const double S0_p = std::max(fspd_x0(i, j, k, 1), fspd_x0(i + delta_w0[0], j + delta_w0[1], k + delta_w0[2], 1));
+					const double S1_m = std::max(fspd_x1(i, j, k, 0), fspd_x1(i + delta_w1[0], j + delta_w1[1], k + delta_w1[2], 0));
+					const double S1_p = std::max(fspd_x1(i, j, k, 1), fspd_x1(i + delta_w1[0], j + delta_w1[1], k + delta_w1[2], 1));
+
+					if (i == 0 && j > 7 && k == 0) {
```
markkrumholz: Debugging prints should be removed before merging.

src/hydro/mhd_system.hpp
```diff
@@ -329,6 +334,209 @@ void MHDSystem<problem_t>::ComputeEMF(std::array<amrex::MultiFab, AMREX_SPACEDIM
...
+					}
+
+					const double S = std::max({std::abs(S0_p), std::abs(S0_m), std::abs(S1_p), std::abs(S1_m)}); // maximum wavespeed
+					// 		 |
+					//   LU  |  RU
+					// ------+------
+					//   LD  |  RD
+					// 		 |
+
+					const auto E2_LD = cc_E2_array(i, j, k);
```
markkrumholz: It seems like a lot of the quantities you define here never get used. For example, `B0_dstar` is defined, but then never referenced. Can you go through here and check that you are only defining variables that you are actually going to use? You can also increase efficiency by moving a bunch of these calculations inside the final if block, so for example you don't compute `E2_L_star` in cases where the wavespeeds imply that you don't need it.

markkrumholz (changes requested):
I left comments inline in the code.

src/hydro/mhd_system.hpp
lizmcole: Okay, I copied the original compute-emf function from this and changed the parts of the computeEMF_Balsara that seemed to reflect what needed changed.

src/hydro/mhd_system.hpp
lizmcole: removed

src/hydro/mhd_system.hpp
lizmcole: Still todo; it was written exactly as it was in the paper so I can double check equations, but yes, some of the star quantities are written but not used and there is definitely a more efficient way to do it.

markkrumholz: Understood. Just ping me when this is ready for re-review.

src/hydro/mhd_system.hpp
markkrumholz: Understood. We are in the process of fixing that, and we will need to apply the same fixes here for this to run stably on GPU.

src/hydro/mhd_system.hpp
markkrumholz: I still see it. Did you perhaps forget to push the update?

BenWibking (commented):
I don't see any hooks to use the new method - is that the case? I think it would be useful to have one of the test problems use the new method when running in the test suite.

src/hydro/mhd_system.hpp
lizmcole: I pushed to the wrong branch, should be there now.

lizmcole:
The indexing is not correct for E2_LD, LU, RD, and RU. I have to work out how to shift the 4 values thing one box down and to the left relative to the direction for a particular edge.

src/hydro/mhd_system.hpp
```diff
@@ -329,6 +330,204 @@ void MHDSystem<problem_t>::ComputeEMF(std::array<amrex::MultiFab, AMREX_SPACEDIM
...
+				std::array<int, 3> delta_w_both = {0, 0, 0};
+				delta_w0[w0_comp] = 1;
+				delta_w1[w1_comp] = 1;
+				delta_w_both[w0_comp] = 1;
+				delta_w_both[w1_comp] = 1;
+				const auto &fspd_x0 = fcx_mf_fspds[w0_comp][mfi].const_array();
+				const auto &fspd_x1 = fcx_mf_fspds[w1_comp][mfi].const_array();
+				const auto &emf_cc = fcx_mf_fspds[w1_comp][mfi].const_array();
+
+				// LLF variant
```
BenWibking: I'm confused by this comment, since I don't think this is actually the LLF variant. From reading Balsara et al. (2025), that appears to be given by Equation 3.9, where there are no special cases for supersonic states and there is one one case to handle. However, 3.10 handles all the varying cases of supersonic states in all the combinations of directions, but that is not an LLF solver and I don't think uses 3.9 at all.

lizmcole: I switched it all over to handle supersonic states now.

BenWibking (changes requested):
It looks like there are a couple of unused variable warnings left:
```
  425 |                                 const auto cc_E2_array = cc_a4_EMF_array[iedge];
      |                                            ^~~~~~~~~~~
/home/runner/work/quokka/quokka/src/hydro/mhd_system.hpp:460:43: error: unused variable ‘w2_comp’ [-Werror=unused-variable]
  460 |                                 int const w2_comp = iedge;
      |                                           ^~~~~~~
```

The GPU failures look like they're unrelated to this PR, but I think Chongchong should figure out what's going on before we merge any new PRs, just in case there's a subtle failure hiding underneath.

Can you point out where in Balsara's paper the reconstruction is explained? Does Balsara do the same reconstruction in both ways to the edge?

BenWibking:
Also: is this method stable with the non-EP reconstruction schemes? Whichever way it turns out, that would be extremely helpful to document in this PR.

lizmcole:
@BenWibking 
Balsara just said to do the EMF reconstruction the same way we already do our other reconstructions when I asked.

BenWibking:
> @BenWibking Balsara just said to do the EMF reconstruction the same way we already do our other reconstructions when I asked.

Ok, that's interesting. I think this means there are 4 possible MHD schemes to consider:
1. Felker & Stone reconstruction, LD04 EMF Riemann solver
2. Felker & Stone reconstruction, Balsara EMF Riemann solver
3. Face velocity reconstruction, LD04 EMF Riemann solver
4. Face velocity reconstruction, Balsara EMF Riemann solver (would have to compute EMF in quadrants after velocity reconstruction for this case, rather than cell-centered EMF)

Would it make sense to split the reconstruction and EMF Riemann solver into separate functions in MHDSystem to test these four combinations?

BenWibking:
Although the first three are already implemented, so maybe that doesn't actually save any effort. Case 4 seems interesting, and could just be copy-and-pasted into its own ComputeEMF_* function.

Edit: I think the reason case 4 is interesting is that it will let us test the non-EP reconstructions with the Balsara EMF Riemann solver.

markkrumholz:
I don't think case 4 exists, because if you're using the Balsara scheme you don't need the velocities at the faces or corners at all. The whole point of the scheme is that you compute the v x B terms at cell centers and then extrapolate that quantity to the corners. When you do that, the velocity by itself never appears at the corners.

BenWibking:
> I don't think case 4 exists, because if you're using the Balsara scheme you don't need the velocities at the faces or corners at all. The whole point of the scheme is that you compute the v x B terms at cell centers and then extrapolate that quantity to the corners. When you do that, the velocity by itself never appears at the corners.

I agree that is a point of the scheme, but to me the more salient feature is that Balsara's Riemann solver treats supersonic states whereas the LD04 Riemann solver assumes you are always in the subsonic strongly interacting state. So for low order reconstructions, the Balsara Riemann solver should be significantly less numerically resistive.

markkrumholz:
> > I don't think case 4 exists, because if you're using the Balsara scheme you don't need the velocities at the faces or corners at all. The whole point of the scheme is that you compute the v x B terms at cell centers and then extrapolate that quantity to the corners. When you do that, the velocity by itself never appears at the corners.
> 
> I agree that is a point of the scheme, but to me the more salient feature is that Balsara's Riemann solver treats supersonic states whereas the LD04 Riemann solver assumes you are always in the subsonic strongly interacting state. So for low order reconstructions, the Balsara Riemann solver should be significantly less numerically resistive.

OK, that's fine, but in that case I think your taxonomy of possible combinations is a bit confusing. It seems like there are two independent axes that could be varied. One is how you reconstruct the corner EMFs, for which we have three possibilities:

1. Felker & Stone scheme as originally published
2. Our variant of Felker & Stone with velocities coming from the Riemann solver.
3. The Balsara scheme where you compute v x B at cell centers and then extrapolate that to corners.

Then we have two distinct Riemann solvers: the LD04 one and the Balsara one. So I guess in principle there are 6 distinct combinations. I don't know that it makes sense to test all of them, but from a code legibility standpoint we should probably separate out the flags that control the choice of reconstruction to the corner scheme and those that control the corner Riemann solver, since in principle you can (I think) mix and match any of these, though of course not all are guaranteed to work.

BenWibking:
> > > I don't think case 4 exists, because if you're using the Balsara scheme you don't need the velocities at the faces or corners at all. The whole point of the scheme is that you compute the v x B terms at cell centers and then extrapolate that quantity to the corners. When you do that, the velocity by itself never appears at the corners.
> > 
> > 
> > I agree that is a point of the scheme, but to me the more salient feature is that Balsara's Riemann solver treats supersonic states whereas the LD04 Riemann solver assumes you are always in the subsonic strongly interacting state. So for low order reconstructions, the Balsara Riemann solver should be significantly less numerically resistive.
> 
> OK, that's fine, but in that case I think your taxonomy of possible combinations is a bit confusing. It seems like there are two independent axes that could be varied. One is how you reconstruct the corner EMFs, for which we have three possibilities:
> 
> 1. Felker & Stone scheme as originally published
> 2. Our variant of Felker & Stone with velocities coming from the Riemann solver.
> 3. The Balsara scheme where you compute v x B at cell centers and then extrapolate that to corners.
> 
> Then we have two distinct Riemann solvers: the LD04 one and the Balsara one. So I guess in principle there are 6 distinct combinations. I don't know that it makes sense to test all of them, but from a code legibility standpoint we should probably separate out the flags that control the choice of reconstruction to the corner scheme and those that control the corner Riemann solver, since in principle you can (I think) mix and match any of these, though of course not all are guaranteed to work.

That makes sense. That's a better taxonomy.

lizmcole:
@BenWibking I tested with reconstruction method 3, it was less stable than the regular method. 

I removed the extra variables.

src/QuokkaSimulation.hpp
```diff
@@ -169,7 +169,9 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 	int nghost_vel_ = Physics_Traits<problem_t>::is_mhd_enabled ? 3 : 2;
 
 	EMFAvgType emfAveragingType_ = EMFAvgType::LD04; // method to use to average EMF at edges
-	int emf_scheme_ = 1; // 0 == Felker and Stone scheme state; 1 == Felker and Stone scheme using the FC velocity from Riemann solver (default)
+	int emf_scheme_ = 1;				 // 0 == Felker and Stone scheme state;
```
AstroKriel: Are the test problems actually running on this new scheme in the GitHub workflows (in this PR)? I think not (correct me if I'm missing something), since the default emf_scheme is set to 1 (our amended Felker & Stone scheme), unless the input files explicitly set it to 2 (the number for this scheme). Could I suggest temporarily setting this to 2, by default, so we make sure the scheme you’ve implemented actually gets run in this PR?

AstroKriel: Mhmm, just requesting this change again, now that I've added myself as a reviewer. I do think this should be done to make sure all of the automated tests work.

AstroKriel (changes requested):

src/QuokkaSimulation.hpp
lizmcole: Done

BenWibking:
@claude Update the documentation with the new MHD runtime parameters.

BenWibking:
@claude Modify the `mhd.emf_scheme` runtime parameter so it accepts a string value using `AMREX_ENUM`, rather than using magic numbers to specify the three allowable options. Also, update the documentation accordingly.

markkrumholz:
@BenWibking, I'd suggest holding off on this PR, because we're going to change these parameters shortly when we refactor to separate the parameter controlling the extrapolation scheme and the ones controlling the 2D Riemann solver.

markkrumholz:
Sorry, just to clarify: I meant that we shouldn't bother updating the input file to select the scheme, not the whole PR.

AstroKriel (approved):

BenWibking:
Code looks good.

What is the longer term plan for testing all of the combinations of schemes in the CI?

markkrumholz:
> Code looks good.
> 
> What is the longer term plan for testing all of the combinations of schemes in the CI?

I don't know that we necessarily want to test all, because not all of them are necessarily going to play well together -- we don't know yet. I would therefore advocate that we first get in the PR where we refactor the code to let us control the extrapolation and Riemann solver independently, do some preliminary testing to see which combinations are viable, and then set up the Alfven wave and fast wave tests for the combinations that we decide are viable. We might also then do a couple of the larger test problems for one or two recommended combinations on top of that.

BenWibking:
> > Code looks good.
> > What is the longer term plan for testing all of the combinations of schemes in the CI?
> 
> I don't know that we necessarily want to test all, because not all of them are necessarily going to play well together -- we don't know yet. I would therefore advocate that we first get in the PR where we refactor the code to let us control the extrapolation and Riemann solver independently, do some preliminary testing to see which combinations are viable, and then set up the Alfven wave and fast wave tests for the combinations that we decide are viable. We might also then do a couple of the larger test problems for one or two recommended combinations on top of that.

Ok. My only comment is that it would be useful to disable or remove the combinations that we find don't really work. That way there's no risk of users accidentally turning on something that they might be expecting to work.

markkrumholz:
> > > Code looks good.
> > > What is the longer term plan for testing all of the combinations of schemes in the CI?
> > 
> > 
> > I don't know that we necessarily want to test all, because not all of them are necessarily going to play well together -- we don't know yet. I would therefore advocate that we first get in the PR where we refactor the code to let us control the extrapolation and Riemann solver independently, do some preliminary testing to see which combinations are viable, and then set up the Alfven wave and fast wave tests for the combinations that we decide are viable. We might also then do a couple of the larger test problems for one or two recommended combinations on top of that.
> 
> Ok. My only comment is that it would be useful to disable or remove the combinations that we find don't really work. That way there's no risk of users accidentally turning on something that they might be expecting to work.

Agree. I think we won't know until we've done the testing.

BenWibking (approved):

markkrumholz (approved):

lizmcole:
Any other changes I have forgotten to make? Everything seems to be working.

markkrumholz:
Looks good to me. If you're happy, you can merge as soon as the automated test suite passes.

chongchonghe:
The AlfvenWaveCircular test failed due to a nan error:

```
[FOFC-1] flux correcting 80 cells on level 0
Coordinates: (0.2734375, 0.5625, 0.0625):  At cell (17,4,0) in Box ((0,0,0) (63,7,7) (0,0,0)): -nan, nan, nan, nan, -nan, -nan
[FOFC-1] Flux correction failed:
Coordinates: (0.2734375, 0.5625, 0.0625):  At cell (17,4,0) in Box ((0,0,0) (63,7,7) (0,0,0)): -nan, -nan, -nan, -nan, -nan, nan
[FOFC-1] failed for 80 cells on level 0
	>> WARNING: Hydro advance failed on level 0

QUOKKA FATAL ERROR
Hydro update exceeded max_retries on level 0. Cannot continue, crashing...
```

@lizmcole can you take a look?

PR #1274, opened 2025-08-18
update exit behavior when max_walltime is set

BenWibking:
### Description
Exiting at 90% of `max_walltime` is not very efficient when the job walltime is large. Instead, we stop the simulation 5 minutes before `max_walltime` (unless the walltime is less than 50 minutes, in which case we fall back to stopping at 90% of the limit).

### Related issues
N/A

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -1224,7 +1224,7 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 			break;
 		}
 
-		if (maxWalltime_ > 0 && getWalltime() > std::max(0.9 * maxWalltime_, maxWalltime_ - 300)) {
+		if (maxWalltime_ > 0 && getWalltime() > std::max(0.9 * maxWalltime_, static_cast<double>(maxWalltime_ - 300))) {
 			// we have exceeded the walltime limit
 			break;
 		}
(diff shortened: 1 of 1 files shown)


BenWibking:
Ok, I need to add a special case for when max_walltime itself is less than 300 seconds...

markkrumholz:
Will wait to review until you have added the special case, unless you'd prefer that I do so now.

src/simulation.hpp
```diff
@@ -1224,8 +1224,8 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 			break;
 		}
 
-		if (maxWalltime_ > 0 && getWalltime() > 0.9 * maxWalltime_) {
-			// we have exceeded 90% of maxWalltime_
+		if (maxWalltime_ > 0 && getWalltime() > std::max(0.9 * maxWalltime_, maxWalltime_ - 300)) {
```
BenWibking: This should now sensibly handle the case where the walltime is less than 5 minutes.

BenWibking:
> Will wait to review until you have added the special case, unless you'd prefer that I do so now.

@markkrumholz I've added the special case. Can you re-review?

markkrumholz (approved):

PR #1292, opened 2025-09-01
Add option to save particles to csv files

chongchonghe:
### Description
Add a function `saveParticleDataToFile` to save particle data to CSV files. This is helpful for quick debugging, or visualizing particles with any tools. An example particle folder looks like this:

```
part00000/StochasticStellarPop_particles.csv
```

The CSV file will contain the following columns:
- ID: Particle ID
- x, y, z: Particle positions
- real_0, real_1, ...: Real components (e.g., mass, velocities, etc.)
- int_0, int_1, ...: Integer components (if any)

This is controlled by the runtime parameter `particle_csv_interval`, which works the same way as `plotfile_interval`. CSV data writing is turned off by default, and if turned on, it will write all particle types with fewer than 1000 particles. 

Also changed:
- moved the IO functions (getParticleDataAtLevelZero getParticleDataAtLevel writeUnitsFile printParticleStatistics) into a new file particle_IO.hpp 

### Related issues

### Checklist

### DIFF
--- a/inputs/ParticleRadiation.in
+++ b/inputs/ParticleRadiation.in
@@ -29,7 +29,8 @@ particles.disable_SN_feedback = true
 
 stop_time = 8.0e20
 max_timesteps = 4
-plotfile_interval = -1
+plotfile_interval = 4
+particle_interval = 4
 checkpoint_interval = -1
 
 tiny_profiler.enabled = 0
--- a/inputs/ParticleRadiation.in
+++ b/inputs/ParticleRadiation.in
@@ -30,7 +30,7 @@ particles.disable_SN_feedback = true
 stop_time = 8.0e20
 max_timesteps = 4
 plotfile_interval = 4
-particle_interval = 4
+particle_csv_interval = 4
 checkpoint_interval = -1
 
 tiny_profiler.enabled = 0
--- a/inputs/ParticleSink.in
+++ b/inputs/ParticleSink.in
@@ -30,6 +30,6 @@ particles.verbose = 1
 particles.sink_particle_use_uniform_kernel = true
 
 max_timesteps = 1
-plotfile_interval = -1
+plotfile_interval = 1
 
 tiny_profiler.enabled = 0
--- a/inputs/ParticleSink.in
+++ b/inputs/ParticleSink.in
@@ -31,5 +31,6 @@ particles.sink_particle_use_uniform_kernel = true
 
 max_timesteps = 1
 plotfile_interval = 1
+particle_interval = 1
 
 tiny_profiler.enabled = 0
--- a/src/particles/PhysicsParticles.hpp
+++ b/src/particles/PhysicsParticles.hpp
@@ -186,7 +186,7 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 	// @return: tuple of vectors of particle data on rank 0, empty vectors on other ranks
 	[[nodiscard]] auto getParticleDataAtLevelZero() const -> std::pair<std::vector<std::vector<double>>, std::vector<std::vector<int>>> override
 	{
-		return particle_io::getParticleDataAtLevelZero(container_);
+		return particle_io::getAllParticleData(container_);
 	}
 
 	[[nodiscard]] auto getParticleDataAtLevel(int lev) const -> std::pair<std::vector<std::vector<double>>, std::vector<std::vector<int>>> override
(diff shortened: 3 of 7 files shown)


BenWibking:
This would be okay for debugging, but I am concerned that people might actually end up using this for production simulations/analysis.

Are there any issues with the current tools that prevent you from being able to use them to read the particles for testing/debugging purposes?

chongchonghe:
> This would be okay for debugging, but I am concerned that people might actually end up using this for production simulations/analysis.

Anything wrong with this? If disk space and I/O speed is not an issue, CSV files have all the advantages. Particularly, you can use any tools you like. 

> Are there any issues with the current tools that prevent you from being able to use them to read the particles for testing/debugging purposes?

My inability to read binary files with my eyes prevents me from debugging efficiently 😂

chongchonghe:
> This would be okay for debugging, but I am concerned that people might actually end up using this for production simulations/analysis.

This is a good point, though. I should rename the parameter `particle_interval` to `ASCII_particle_interval` or `particle_csv_interval` to indicate that this is for writing ASCII data and should not be used when the number of particles is huge.

BenWibking:
> > This would be okay for debugging, but I am concerned that people might actually end up using this for production simulations/analysis.
> 
> Anything wrong with this? If disk space and I/O speed is not an issue, CSV files have all the advantages. Particularly, you can use any tools you like.
> 
> > Are there any issues with the current tools that prevent you from being able to use them to read the particles for testing/debugging purposes?
> 
> My inability to read binary files with my eyes prevents me from debugging efficiently 😂

There's nothing "wrong" with this... just adding more ways to do things means there is potentially a larger maintenance burden down the road, and it may prevent the tools that read the binary particles from actually getting fixed if it's not in the critical path for all users. Unfortunately, I have seen this happen many times with other codes/tools.

So I am not totally opposed to this, but I think there needs to be a very good reason to add another way to do it.

BenWibking:
Maybe it could refuse to write an ASCII file if there are more than, say, 1000 particles.

chongchonghe:
> Maybe it could refuse to write an ASCII file if there are more than, say, 1000 particles.

Sounds good. I'll do that. We can also make this a runtime parameter.

chongchonghe:
Actually, we should make the maximum number of particles allowed to write to CSV files a runtime parameter, because later I will add a new parameter called "particles_to_save_to_csv" to selectively save some particle types to save to CSV files. This way we can save sink particles or star particles at high cadence but filtering out CIC_particles.

chongchonghe:
This brings back what we previously discussed about selectively saving particle types at high cadence. We can choose to do high-cadence particles with CSV files. One advantage is that this will not mess up with the Quokka folder structure: if a folder plt00004 exists but it doesn't have a subfolder Level_0, the yt reader will crash. However, sometimes we do want to save high-cadence particle data to binary boxlib files, so we also want to have another parameter to do this. 

In any case, I think having more options is (almost) always better than fewer options. You can always turn off those extra options, and we should leave them off by default, but it's nice when these extra features help.

BenWibking:
> This brings back what we previously discussed about selectively saving particle types at high cadence. We can choose to do high-cadence particles with CSV files. One advantage is that this will not mess up with the Quokka folder structure: if a folder plt00004 exists but it doesn't have a subfolder Level_0, the yt reader will crash. However, sometimes we do want to save high-cadence particle data to binary boxlib files, so we also want to have another parameter to do this.
> 
Ok, this is a case of exactly what I meant. We should just fix this in the yt reader. I think having high-cadence ASCII files is actually a bad idea, because it is easy to accidentally waste a huge amount of disk space this way *and* it means that the yt reader might never get fixed to be able to read binary particle plotfiles that don't have multifabs.

> In any case, I think having more options is (almost) always better than fewer options. You can always turn off those extra options, and we should leave them off by default, but it's nice when these extra features help.

I don't agree with this. My design philosophy for Quokka from the beginning has been, as much as is practically possible, to have one way to do things. Codes that have done the opposite have ended up having huge problems (or actual bugs) due to this.

chongchonghe:
I agree we should prefer to use amrex to write high-cadence particle data. In this case, I won't add complex control to the CSV data writer. Currently, CSV data is turned off by default, and if turned on, it will write all particle types that have fewer than 1000 particles. For a particle type with more than 1000 particles, it won't do MPI communication, so there is no performance issue. 

I think this is very helpful for debugging and for star/protostar formation simulations, so it's worth merging it as it currently stands. 

@BenWibking

markkrumholz:
I agree that this is useful, but I will hold off on approving the merger right now since it is in the middle of the night for @BenWibking -- I want to give him a chance to weigh in.

chongchonghe:
No worry. We can hold off on this. I am working on a particle reproducibility branch which is based on this branch.

chongchonghe:
I immediately find a potential cause of non-reproducibility in the stochastic SF routine, using this CSV debugging tool.

BenWibking (approved):

PR #1297, opened 2025-09-03
Fix bug in getParticleDataAtAllLevels()

chongchonghe:
### Description
The current code uses `analysisPC.copyParticles(*container);` to copy all particles from the particle container to a buffer container `analysisPC`. However, `copyParticles` expects the destination container to have the same number of levels as the source. When amr.max_level > 0, the source container has multiple levels but the analysis container only has level 0, causing a segfault.

In this update, I use the paradigm of `getParticleDataAtLevel` in `getParticleDataAtAllLevels` and loop over all levels to collect all particles from all levels and all multifabls into a single vector. This is robust for arbitrary `max_level` on multiple ranks. 

Set `particle_csv_interval = 200` in SphericalCollpase.in to test this update. Before this update, this test will fail with a segfault. 

### Related issues

### Checklist

### DIFF
--- a/inputs/SphericalCollapse.in
+++ b/inputs/SphericalCollapse.in
@@ -33,4 +33,5 @@ do_tracers = 0          # turn on tracer particles
 
 ascent_interval = 100
 plotfile_interval = 200
+particle_csv_interval = 200
 checkpoint_interval = 1000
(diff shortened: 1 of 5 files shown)


markkrumholz (approved):

chongchonghe:
The amdclang error comes from #1268 and is irrelevant to this PR, so i'm gonna ignore it and merge this PR.

PR #1299, opened 2025-09-03
[WIP] Deterministic particle to mesh deposition

chongchonghe:
### Description

This PR, plus [SumBoundary reimplementation #4658 from AMReX](https://github.com/AMReX-Codes/amrex/pull/4658), will make particles reproducible on a single level. 

### Related issues
Related to #1125, partially resolving it. 

### Checklist


src/particles/PhysicsParticles.hpp
```diff
@@ -215,9 +215,49 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 	void depositMass(const amrex::Vector<amrex::MultiFab *> &rhs, int finest_lev, amrex::Real Gconst) override
 	{
 		if (container_ != nullptr && this->getMassIndex() >= 0) {
+#ifdef QUOKKA_DETERMINISTIC_DEPOSITION
+			// Deterministic version: Sort particles and use local buffers for reproducible ordering
+			container_->SortParticlesByCell();
```
BenWibking: Ok, this won't actually work. Sorting is actually NOT deterministic (you can see my discussion with Andrew Myers on Slack for why).

You do actually need to do Kahan summation cell-by-cell.

chongchonghe: OK. It should be possible and simple to do Kahan summation on particles that interact with the cell for each cell individually.

chongchonghe:
I'm going to implement looping over cells in a tile and doing Kahan summation over all the particles that interact with a cell.

In this first version, I'm trying to reproduce the original ParticleToMesh result with this new approach, without doing Kahan summation. The code works fine: it passes the binary orbit test. However, The results are not identical to that from the dev branch: the Gpot is different by 1e-4. I don't understand how could this happen.  @BenWibking  can you take a look?

BenWibking:
> I'm going to implement looping over cells in a tile and doing Kahan summation over all the particles that interact with a cell.
> 
> In this first version, I'm trying to reproduce the original ParticleToMesh result with this new approach, without doing Kahan summation. The code works fine: it passes the binary orbit test. However, The results are not identical to that from the dev branch: the Gpot is different by 1e-4. I don't understand how could this happen. @BenWibking can you take a look?

I think this is expected. The rhs is not the same to ~1e-12 or so due to different orders of summation. Then the Poisson solve is not necessarily the same up to the relative tolerance of the Poisson solve, which is ~1e-5:
https://github.com/quokka-astro/quokka/blob/b05442caa730e55e3e0a3ddc433f4ceb6d945d60/src/simulation.hpp#L187

The relative tolerance is actually the tolerance on the residual, which is NOT the same as the tolerance on the solution itself. Those can be different by up to the condition number of the linear system being solved.

So I am not surprised by this.

src/particles/PhysicsParticles.hpp
```diff
@@ -215,9 +215,107 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 	void depositMass(const amrex::Vector<amrex::MultiFab *> &rhs, int finest_lev, amrex::Real Gconst) override
 	{
 		if (container_ != nullptr && this->getMassIndex() >= 0) {
+#ifdef QUOKKA_DETERMINISTIC_DEPOSITION
+			// Deterministic cell-centric version with Kahan summation for GPU reproducibility
+			// Algorithm:
+			// 1. Loop over cells in deterministic order (ParallelFor over grid cells)
```
BenWibking: This loop is not deterministic, but it doesn't matter.

chongchonghe:
After implementing Kahan summation, it's still not reproducible. The relative error between multiple runs does not change. 

With Kahan summation:
```
Relative difference in gpot_min for all 8 tiles:
[-0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00]
[-0.0e+00 -2.1e-16 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00]
Relative difference in gpot_max for all 8 tiles:
[-1.6e-16 -0.0e+00 -1.6e-16 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00]
[-1.6e-16 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00]
```

Without Kahan summation:
```
Relative difference in gpot_min for all 8 tiles:
[-0.0e+00 -2.1e-16 -0.0e+00 -0.0e+00 -0.0e+00 -2.1e-16 -0.0e+00 -2.1e-16]
[-0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -2.1e-16]
Relative difference in gpot_max for all 8 tiles:
[-0.0e+00 -0.0e+00 -1.6e-16 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00]
[-0.0e+00 -0.0e+00 -1.6e-16 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00 -0.0e+00]
```

I speculate another cause is `SumBoundary`. I don't think it's shown to be deterministic since it's not used anywhere else. 

line 308 of PhysicsParticles.hpp :
```
// Sum boundary cell values to real cells
buffer_rhs.SumBoundary(container_->Geom(lev).periodicity());
```

BenWibking:
Can you output the buffer multifab itself and run `fcompare` on that? I think that will be more informative that checking gpot.

BenWibking:
One thing to note is that Kahan summation does not actually make it deterministic. It just suppresses the rounding error by many orders of magnitude. So it may be necessary to round the result to lower precision after doing Kahan summation. This reduces accuracy, but will make it reproducible.

BenWibking:
But again, please check the buffer directly.

chongchonghe:
> Can you output the buffer multifab itself and run `fcompare` on that? I think that will be more informative that checking gpot.

The hydro components of the Multifab at snapshot 0 should be exactly the same. Later steps they are different because of the difference in gpot and acceleration. So, I don't think fcompare can give you any more useful information.

BenWibking:
> > Can you output the buffer multifab itself and run `fcompare` on that? I think that will be more informative that checking gpot.
> 
> The hydro components of the Multifab at snapshot 0 should be exactly the same. Later steps they are different because of the difference in gpot and acceleration. So, I don't think fcompare can give you any more useful information.

No, I mean you should add a WriteSingleLevelPlotfile call with the buffer itself at timestep 0 and compare that.

chongchonghe:
> One thing to note is that Kahan summation does not actually make it deterministic. It just suppresses the rounding error by many orders of magnitude. So it may be necessary to round the result to lower precision after doing Kahan summation. This reduces accuracy, but will make it reproducible.

The relative error is 1e-16 without Kahan, so by suppressing it by orders of magnitude the error should go away, or at least only exist at the last digit, but it didn't. 

I will try reducing accuracy to 1e-15 and see if the error goes away.

BenWibking:
> > One thing to note is that Kahan summation does not actually make it deterministic. It just suppresses the rounding error by many orders of magnitude. So it may be necessary to round the result to lower precision after doing Kahan summation. This reduces accuracy, but will make it reproducible.
> 
> The relative error is 1e-16 without Kahan, so by suppressing it by orders of magnitude the error should go away, or at least only exist at the last digit, but it didn't.
> 
> I will try reducing accuracy to 1e-15 and see if the error goes away.

Ok, but you have to check the buffer directly. The Poisson solver can amplify the error potentially by many orders of magnitude.

BenWibking:
This is the correct way to round to lower precision: https://stackoverflow.com/a/19431713
For the proof it works, see: https://people.eecs.berkeley.edu/~wkahan/CS279/Veltkamp.pdf

src/math/float.hpp
```diff
@@ -0,0 +1,44 @@
...
+#define FLOAT_HPP_
+
+#include <cmath>
+#include <exception>
+#include <stdexcept>
+
+#include "AMReX_BLassert.H"
+#include "AMReX_GpuQualifiers.H"
+
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto round_to_sigfigs_stable(amrex::Real x, int N) -> amrex::Real
```
BenWibking: This should not be done with log/pow. You should implement this method: https://stackoverflow.com/questions/19431172/reduce-mantissa-bit-width/19431713#19431713

chongchonghe:
After trying two ways to do rounding to 11 significant figures, the results did not change with respect to previous ones. The error is about 1e-16. 

~I tend to believe the cause of non-determinism is SumBoundary.~
`SumBoundary` is not the cause because i tried removing `buffer_rhs.SumBoundary()` but the error still exists.

src/math/float.hpp
chongchonghe: ok, changed it. the error is still there.

src/particles/PhysicsParticles.hpp
```diff
@@ -215,9 +215,116 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
+							buffer_arr(i, j, k, 0) = c - (c - sum);
+						});
+					}
+				}
+
+				// Sum boundary cell values to real cells
+				buffer_rhs.SumBoundary(container_->Geom(lev).periodicity());
+
+				// Add buffer_rhs to rhs
+				amrex::MultiFab::Add(*rhs[lev], buffer_rhs, 0, 0, 1, nGrow);
```
BenWibking: This is not quite enough to correctly handle AMR.

`amrex::ParticleToMesh` does some special things to ensure mass conservation at coarse-fine boundaries:
https://github.com/AMReX-Codes/amrex/blob/2304c89fefd99efc144585b0fcebccb3f5a62e10/Src/AmrCore/AMReX_AmrParticles.H#L208

These need to be duplicated here.

chongchonghe: OK, I will take a look later. We have to solve the reproducibility issue first, otherwise we'd better just use ParticleToMesh.

chongchonghe:
OK, this partially worked. On a single level (max_level=0), this cell-by-cell + Kahan summation trick made particle deposition 100% reproducible. However, when max_level=1, this trick didn't work. This means the course-fine communication is not deterministic. @BenWibking @markkrumholz

markkrumholz:
OK, well that is progress at least.

chongchonghe:
On a single level, particle deposition is *only* non-deterministic when there are multiple tiles, and Kahan summation made it deterministic. So, the non-determinism is a combination of the race condition on the particle deposition and multifab communication. Kahan summation solved the problem of non-associative float point addition but did not solve the problem of non-deterministic multifab communication.

BenWibking:
Unless there's a bug somehow, multifab communication should be deterministic. Pure hydro sims would not be deterministic if it were not the case that multifab communication is deterministic. I think the cause is probably more subtle.

BenWibking:
Is Redistribute called in the AMR case?

It is the case that *particle* communication is nondeterministic, since the order the particles are copied into MPI buffers depends on atomic writes. That should be fixed by your other PR: https://github.com/quokka-astro/quokka/pull/1135

It just can't be the case the multifab comms are nondeterministic. Every AMReX code's test suite would break on AMR problems if that were the case, including Castro's, which has AMR problems that run on GPU that are 100% reproducible in their nightly test runs.

chongchonghe:
After numerous tests, I come to the following clear conclusions:

There are three sources of non-determinism in Quokka: `ParticleToMesh`, `SumBoundary` and coarse-fine interpolation. The `ParticleToMesh` operation was made deterministic via cell-by-cell Kahan summation in this PR. The other two still exist.

These conclusions are supported by the following test results, all done with the Kahan summation turned on:

- lev_max=0, nGrow=0: YES
- lev_max=0, nGrow=1, with SumBoundary: NO
- lev_max=0, nGrow=1, without SumBoundary: YES
- lev_max=1, nGrow=0: NO
- lev_max=1, nGrow=1, without SumBoundary: NO

nGrow: the number of ghost cells in buffer_rhs. "YES": reproducible. "NO": not reproducible.

I may be missing something, but I believe Quokka's reproducibility with AMR on has never been tested: among all the regression tests, DiskGalaxy is the only one with multiple levels. We should add AMR to at least one of the other tests so that we progressively test the reproducibility of each individual modules: pure hydro, hydro + gravity, hydro + gravity + AMR, hydro + gravity + AMR + particles. I'm going to create a PR to update the regression tests.

I speculate that hydro+gravity+AMR tests will NOT be reproducible. If they turn out to be reproducible, then it means the coarse-fine communication in the Poisson solver is not reproducible.

chongchonghe:
> Is Redistribute called in the AMR case?
> It is the case that particle communication is nondeterministic, since the order the particles are copied into MPI buffers depends on atomic writes. That should be fixed by your other PR: https://github.com/quokka-astro/quokka/pull/1135

Redistribute is called, but it can't be the only cause. I tried removing all the redistribute calls and the same problem exits. Since we don't rely on a determined order to process the particles, non-deterministic redistribution doesn't matter.

BenWibking:
> I may be missing something, but I believe Quokka's reproducibility with AMR on has never been tested: among all the regression tests, DiskGalaxy is the only one with multiple levels. We should add AMR to at least one of the other tests so that we progressively test the reproducibility of each individual modules: pure hydro, hydro + gravity, hydro + gravity + AMR, hydro + gravity + AMR + particles. I'm going to create a PR to update the regression tests.

That's a good idea. I manually tested the reproducibility of the blast_amr.in test when investigating the race condition in the MHD AMR test, and it appeared to be reproducible when I tested it, but we should definitely add a hydro + AMR test and a hydro + gravity + AMR test to the regression tests.

chongchonghe:
This shows how SumBoundary is not reproducible :

> relative difference on the `rhs` after SumBoundary between two runs:

buffer_rhs_differences

chongchonghe:
Now it's very clear what the causes of irreproducibility are and the path way to resolve it. Cell-by-cell deposition with Kahan summation makes the mass deposition reproducible in individual tiles. When MultiFabs is involved, `SumBoundary` makes it not reproducible. However, this is independent of particles and we don't have to solve it for the sink particle paper. I don't know how hard it is to make `SumBoundary` deterministic; I'll wait for Weiqun's reply on Slack. 

I think we can claim victory and finish the paper. We can write that looping over all particles in a tile with Kahan summation for each individual cell is required to make particle deposition deterministic. 

@BenWibking @markkrumholz

chongchonghe:
> > I may be missing something, but I believe Quokka's reproducibility with AMR on has never been tested: among all the regression tests, DiskGalaxy is the only one with multiple levels. We should add AMR to at least one of the other tests so that we progressively test the reproducibility of each individual modules: pure hydro, hydro + gravity, hydro + gravity + AMR, hydro + gravity + AMR + particles. I'm going to create a PR to update the regression tests.
> 
> That's a good idea. I manually tested the reproducibility of the blast_amr.in test when investigating the race condition in the MHD AMR test, and it appeared to be reproducible when I tested it, but we should definitely add a hydro + AMR test and a hydro + gravity + AMR test to the regression tests.

@BenWibking  Can you create a PR to do that?

chongchonghe:
The fact that SumBoundary is not deterministic makes sense and does not conflict with the fact that pure hydro + gravity is always deterministic. Hydro + gravity only uses `FillBoundary` and do not use `SumBoundary`.`FillBoundary` is trivially deterministic because it's a simple copy and paste and does not depend on the order of processing. However, the determinisity of `SumBoundary` replies on a determined order of summation because ghost cells from multiple tiles can add to one cell.

BenWibking:
> The fact that SumBoundary is not deterministic makes sense and does not conflict with the fact that pure hydro + gravity is always deterministic. Hydro + gravity only uses `FillBoundary` and do not use `SumBoundary`.`FillBoundary` is trivially deterministic because it's a simple copy and paste and does not depend on the order of processing. However, the determinisity of `SumBoundary` replies on a determined order of summation because ghost cells from multiple tiles can add to one cell.

Ok, that makes sense. It's not AMR that causes the indeterminism then. It's SumBoundary.

I think we do need to show that it's fully deterministic in a real simulation before submitting the paper. I think Weiqun should be able to make SumBoundary deterministic more or less as soon as he reads the Slack message.

BenWibking:
> > > I may be missing something, but I believe Quokka's reproducibility with AMR on has never been tested: among all the regression tests, DiskGalaxy is the only one with multiple levels. We should add AMR to at least one of the other tests so that we progressively test the reproducibility of each individual modules: pure hydro, hydro + gravity, hydro + gravity + AMR, hydro + gravity + AMR + particles. I'm going to create a PR to update the regression tests.
> > 
> > 
> > That's a good idea. I manually tested the reproducibility of the blast_amr.in test when investigating the race condition in the MHD AMR test, and it appeared to be reproducible when I tested it, but we should definitely add a hydro + AMR test and a hydro + gravity + AMR test to the regression tests.
> 
> @BenWibking Can you create a PR to do that?

No, but you can 😄

BenWibking:
https://github.com/AMReX-Codes/amrex/issues/4644

chongchonghe:
> Ok, that makes sense. It's not AMR that causes the indeterminism then. It's SumBoundary.

No. Both AMR and SumBoundary cause indeterminism. See my list of tests above. Particularly, 'lev_max=1, nGrow=0' is not reproducible.

chongchonghe:
@BenWibking As I predicted, hydro+gravity+AMR is not reproducible: https://quokka-astro.github.io/regression-testing-results/2025-09-08-008/RandomBlastAMR-GPU.html

BenWibking:
> @BenWibking As I predicted, hydro+gravity+AMR is not reproducible: https://quokka-astro.github.io/regression-testing-results/2025-09-08-008/RandomBlastAMR-GPU.html

Hm. I am not sure what could cause this. Maybe there is a race condition.

chongchonghe:
@BenWibking Hydro+AMR is reproducible. Hydro+Gravity+AMR is not. https://quokka-astro.github.io/regression-testing-results/2025-09-09-001/index.html

BenWibking:
Well, this is confusing. I'm not sure what specifically about AMR+gravity would cause this.

BenWibking:
Is the RandomBlast-AMR problem reproducible if you turn off self-gravity?

chongchonghe:
> Is the RandomBlast-AMR problem reproducible if you turn off self-gravity?

It should be, because RandomBlast-AMR without self-gravity would be the same as SedovAMR.

BenWibking:
> > Is the RandomBlast-AMR problem reproducible if you turn off self-gravity?
> 
> It should be, because RandomBlast-AMR without self-gravity would be the same as SedovAMR.

It should be, but is it? I think it's helpful to know.

chongchonghe:
This PR, plus [SumBoundary reimplementation #4658 from AMReX](https://github.com/AMReX-Codes/amrex/pull/4658), will make particles reproducible on a single level. 

@BenWibking I implemented everything in the multilevel `amrex::ParticleToMesh`, including coarse-fine interpolation, fine-to-coarse summation, and double-count correction. Please check if this looks fine.

chongchonghe:
we should use #1323

PR #1310, opened 2025-09-10
Setting custom boundary conditions on faces based on direction

lizmcole:
### Description
In order to set custom boundary conditions to face variables, we need the x, y, or z direction information. This PR adds quokka::direction dir to setCustomBoundaryConditionsFaceVar which is called in problems, and fixes the functions that contain this function so that the direction is passed in. 

New test: Brio Wu Shocktube

### Related issues

### Checklist
- [ x] I have added tests for any new physics that this PR adds to the code.


lizmcole:
https://github.com/user-attachments/assets/745d5584-229d-4ab7-90a4-7938f0a3cba4

BenWibking:
> BrioWuShock.mov

It looks like there is a shock launched from the left and right boundaries, which suggests there may be a subtle inconsistency in the implementation of the boundary condition there.

The post-shock oscillations suggest that the shock flattening needs to be adjusted for MHD. The pressure calculation in the shock flattener may need to include the magnetic pressure.

lizmcole:
Here is where it currently stands (this was run on Setonix). The boundaries seem to be working now, and it still has the oscillations but that is probably a later PR to deal with what sort reconstruction is needed to flatten those out, as this test just demonstrates that custom boundary setting of face-centered values seems to be working correctly.

lizmcole:
https://github.com/user-attachments/assets/6c0f97c4-de6d-4e0b-a9a6-32222c428b20

markkrumholz (approved):

src/problems/BrioWuShockTube/test_brio_wu_shocktube.cpp
```diff
@@ -0,0 +1,307 @@
...
+		consVar(i, j, k, RadSystem<MHDShocktubeProblem>::x2GasMomentum_index) = 0.;
+		consVar(i, j, k, RadSystem<MHDShocktubeProblem>::x3GasMomentum_index) = 0.;
+	}
+}
+
+template <>
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+AMRSimulation<MHDShocktubeProblem>::setCustomBoundaryConditionsFaceVar(const amrex::IntVect &iv, amrex::Array4<amrex::Real> const &consVar_fc, int /*dcomp*/,
+								       int /*numcomp*/, amrex::GeometryData const &geom, const amrex::Real /*time*/,
+								       const amrex::BCRec * /*bcr*/, int /*bcomp*/, int /*orig_comp*/, quokka::direction dir)
```
BenWibking, line 200: @lizmcole @markkrumholz I thought the design was going to use a template parameter here? Is there a reason to use a function argument instead? Using a function argument may reduce performance compared to a template argument, since additional registers have to be used to store the value of the function argument.

src/problems/BrioWuShockTube/test_brio_wu_shocktube.cpp
```diff
@@ -0,0 +1,307 @@
...
+#if (AMREX_SPACEDIM == 3)
+	auto [i, j, k] = iv.toArray();
+#endif
+	amrex::Box const &box = geom.Domain();
+	amrex::GpuArray<int, 3> lo = box.loVect3d();
+	amrex::GpuArray<int, 3> hi = box.hiVect3d();
+
+	// Use direction to determine which boundaries to check
+	switch (dir) {
+		case quokka::direction::x:
```
BenWibking, line 220: @lizmcole @markkrumholz When B_x is transverse to a domain boundary, it looks like it is left unfilled here? I think this is incorrect.

PR #1312, opened 2025-09-10
update Physics_Indices

chongchonghe:
### Description
Clean up calculations in `Physics_Indices`. 

### Related issues
None.

### Checklist


BenWibking (approved):
LGTM, assuming GPU tests pass.

chongchonghe:
Unlikely to fail on GPU since nothing will be different on GPU.

BenWibking:
Ok, except the NVIDIA compiler has some dumb restrictions on the use of constexpr in GPU kernels. Maybe it's not an issue in this case, but it's good to check.

chongchonghe:
> Ok, except the NVIDIA compiler has some dumb restrictions on the use of constexpr in GPU kernels. Maybe it's not an issue in this case, but it's good to check.

Sure. In case it fails, I should be able to fix it in a few minutes.

PR #1313, opened 2025-09-10
quokka::BCType and quokka::BC

chongchonghe:
### Description
One-liner `BCs_cc` calculation:

Same BC for all directions:
```
auto BCs_cc = quokka::BC<ParticleProblem>(amrex::BCType::int_dir);
```

Different BC in different directions:
```
auto BCs_cc = quokka::BC<PulseProblem>(amrex::BCType::foextrap, amrex::BCType::int_dir, amrex::BCType::int_dir);
```

Changes:
- Created `util/BC.hpp` for easier BC generation. 
- Updated all test problems to use this new API. Some test problems use mixed BC for upper and lower bound and this new API does not apply, so I kept them as unchanged. I don't plan to deal with them right now. 

### Related issues
None.

### Checklist

### DIFF
--- a/src/util/BC.hpp
+++ b/src/util/BC.hpp
@@ -1,105 +1,105 @@
 #ifndef BC_HPP_
 #define BC_HPP_
 
-#include "AMReX_BC_TYPES.H"
 #include "AMReX_BCRec.H"
+#include "AMReX_BC_TYPES.H"
 #include "AMReX_Vector.H"
 #include "hydro/hydro_system.hpp"
 #include "physics_numVars.hpp"
 #include "radiation/radiation_system.hpp"
 #include <array>
 
-namespace quokka {
+namespace quokka
+{
 
-namespace BoundaryCondition {
-    // Special boundary condition for reflecting walls
-    // Uses both reflect_odd and reflect_even depending on component
-    constexpr int reflecting = 8881;
+namespace BoundaryCondition
+{
+// Special boundary condition for reflecting walls
+// Uses both reflect_odd and reflect_even depending on component
+constexpr int reflecting = 8881;
 } // namespace BoundaryCondition
 
-namespace detail {
+namespace detail
+{
 
-    // Check if a component is a normal component (momentum or radiation flux) in a given dimension
-    template <typename problem_t>
-    constexpr bool isNormalComponent(int n, int dim) {
-        // Check radiation flux components if radiation is enabled
-        if constexpr (Physics_Traits<problem_t>::is_radiation_enabled) {
-            // Check gas momentum components in RadSystem
-            if ((n == RadSystem<problem_t>::x1GasMomentum_index) && (dim == 0)) {
-                return true;
-            }
-            if ((n == RadSystem<problem_t>::x2GasMomentum_index) && (dim == 1)) {
-                return true;
-            }
-            if ((n == RadSystem<problem_t>::x3GasMomentum_index) && (dim == 2)) {
-                return true;
-            }
-            
-            // Check radiation flux components for all groups
-            for (int g = 0; g < Physics_Traits<problem_t>::nGroups; ++g) {
-                if ((n == RadSystem<problem_t>::x1RadFlux_index + Physics_NumVars::numRadVarsPerGroup * g) && (dim == 0)) {
-                    return true;
-                }
-                if ((n == RadSystem<problem_t>::x2RadFlux_index + Physics_NumVars::numRadVarsPerGroup * g) && (dim == 1)) {
-                    return true;
-                }
-                if ((n == RadSystem<problem_t>::x3RadFlux_index + Physics_NumVars::numRadVarsPerGroup * g) && (dim == 2)) {
-                    return true;
-                }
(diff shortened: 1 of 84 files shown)


chongchonghe:
Took me 1 hour of tedious work plus $5 of Claude Code API calls. I don't know why I have to do this; must be OCD. 

CleanShot 2025-09-10 at 20 18 35@2x

src/util/BC.hpp
```diff
@@ -0,0 +1,106 @@
...
+#include <array>
+
+namespace quokka
+{
+
+namespace BoundaryCondition
+{
+// Special boundary condition for reflecting walls
+// Uses both reflect_odd and reflect_even depending on component
+constexpr int reflecting = 8881;
```
BenWibking: This should be an empty struct, rather than a random number.

BenWibking (changes requested):
This is a nice change. One code implementation detail comment below:

src/util/BC.hpp
chongchonghe: reflecting has to be a new integer, because it chooses refl_odd and refl_even for different components. No?

BenWibking: Strong types are better than magic numbers: https://www.fluentcpp.com/2016/12/08/strong-types-for-strong-interfaces/

BenWibking: You could make new strong types (just empty structs) for each BC type inside the `quokka::BoundaryConditions` namespace. Then no magic numbers are needed.

chongchonghe: OK, AI messed this up and I missed it. I meant to use a enum class for quokka-specific BCs.

chongchonghe: > You could make new strong types (just empty structs) for each BC type inside the quokka::BoundaryConditions namespace. Then no magic numbers are needed.

This won't work, unless i'm missing somoething. later in this file, i have to compare bcs[i] with reflecting, and bcs[i] takes amrex::BCType (an integer enum) most of the time. So, reflecting has to be an integer.

src/util/BC.hpp
```diff
@@ -0,0 +1,106 @@
...
+}
+} // namespace detail
+
+// Three parameter version - sets each dimension separately
+template <typename problem_t> amrex::Vector<amrex::BCRec> BC(int bc_x, int bc_y, int bc_z)
+{
+	const int ncomp_cc = Physics_Indices<problem_t>::nvarTotal_cc;
+	amrex::Vector<amrex::BCRec> BCs_cc(ncomp_cc);
+
+	std::array<int, 3> bcs = {bc_x, bc_y, bc_z};
```
BenWibking, line 95: This doesn't have to be an integer.

src/util/BC.hpp
```diff
@@ -0,0 +1,106 @@
...
+				// and reflect_even for all other components (including tangential momentum)
+				if (detail::isNormalComponent<problem_t>(n, i)) {
+					BCs_cc[n].setLo(i, amrex::BCType::reflect_odd);
+					BCs_cc[n].setHi(i, amrex::BCType::reflect_odd);
+				} else {
+					BCs_cc[n].setLo(i, amrex::BCType::reflect_even);
+					BCs_cc[n].setHi(i, amrex::BCType::reflect_even);
+				}
+			} else {
+				// For all other boundaries, use the AMReX BC type directly
```
BenWibking: You could set each type explicitly here, translating from, e.g. `quokka::BoundaryCondition::Periodic` to `amrex::BCType::int_dir`.

BenWibking (commented):
You could translate each boundary condition type from quokka types to AMReX types explicitly.

src/util/BC.hpp
BenWibking: See comment below.

src/util/BC.hpp
chongchonghe: but it will depend on the runtime parameter amrex.is_periodic, so calling it Periodic can be confusing.

chongchonghe:
> You could translate each boundary condition type from quokka types to AMReX types explicitly.

I am trying to keep ti simple, using integers inherited from amrex::BCType. I could do a direct mapping from amrex::BCType to quokka::BCType, and add new types to quokka::BCType. How does that sound? @BenWibking

src/util/BC.hpp
BenWibking: We could remove that from the input files, and set it based on the compile-time settings.

chongchonghe: This is what I wanted to do for a long time. The current control of boundary condition is very confusing because amrex.is_periodic overrides BCs_cc. However, this is independent of this PR, so I'll do that in a separate PR.

chongchonghe:
I favor simplicity over extra complexity, so I prefer the straightforward integer enum approach, similar to the one used in amrex. 

I created a new PR for the int enum approach #1314

src/util/BC.hpp
chongchonghe: but integer is simple and it works, so I would prefer it over strong types.

PR #1316, opened 2025-09-12
Periodic BC for self-gravity

### LINKED ISSUE #700: support periodic boundary conditions for self-gravity
**Describe the proposal**
We currently only support open boundary conditions for the Poisson equation using the `amrex::OpenBC` Poisson solver. We can also use `amrex::MLMG` to do the Poisson solve for periodic BCs.

N/A

**Additional context**
Needed for cosmological PopIII stars.

chongchonghe:
### Description

Multi-level Poisson solver with periodic boundary condition using `amrex::MLMG`. Tested for single-levle and multi-level full periodic boundary condition. Currently does not support mixed periodic and Dirichlet BCs. 

More changes:
- add two new runtime parameters: poisson_reltol and poisson_abstol 
- updated all tests in src/problems/BinaryOrbitCIC/ to use periodic boundary conditions

#### How does Quokka deal with boundary conditions?

Quokka set the boundary conditions for the hydro variables following this logic:
1. Read `geometry.is_periodic` from the input file or command line arguments. For a dimension with is_periodic = 1, will use periodic boundary condition.
2. For dimensions with is_periodic = 0, Quokka will use whatever is set by `BCs_cc` in the cpp file for the **hydro** variables.
   - If set to `int_dir`, the code will crash with 'maxmum hydro steps reached, aborting' error due to nan values in boundayr cells.
   - If set to `ext_dir`, use either Dirichlet or NSCBC boundary condition, and the user has to specify `setCustomBoundaryConditions`
   - For everything else (reflect_even, reflect_odd, reflecting, foextrap), no need for setCustomBoundaryConditions.

If self-gravity is enabled, Quokka set the boundary condition of the Poisson solver following this logic:
- `geometry.is_periodic = 0 0 0`: open-box BC.
- `geometry.is_periodic = 1 1 1`: fully periodic BC in all dimensions.
- `geometry.is_periodic = 1 1 0 or 1 0 1 or 0 1 1`: periodic BC in two dimensions and Dirichlet BC in the other dimension. The `amrex::MLMG` solver starts from an initial guess `φ = 0` and relax to convergence. The value of this initial const does not matter because the gravitational potential has gauge freedom, i.e. gravitational force F = -∇φ is gauge invariant (independent of φ + constant)
- `geometry.is_periodic = 0 0 1 or 0 1 0 or 1 0 0`. periodic BC in one dimension and Dirichlet BC with phi = const in the other two. This is an infinite tube setup, suitable for simulating star formation in isolated, infinite filaments.

Updated the tests in BinaryOrbitCIC to validate all combinations of periodic and Dirichlet BCs. 

Controlling boundary conditions with both `geometry.is_periodic` and `BCs_cc` with one overriding another is quite confusing, so I'm gonna fix it in a future PR. 

### Related issues
Closes #700

### Checklist


BenWibking:
> periodic BC in two dimensions and Dirichlet BC with phi = const. in the other dimension, and the value of this const does not matter because the gravitational potential has gauge freedom; gravitational force F = -∇φ is gauge invariant (independent of φ + constant)

It might be helpful to document that this is only true when the desired solution has a constant potential at the boundary. For free-space boundary conditions, this approximation will fail to hold when the planar symmetry is broken and there is significant mass near the boundary.

src/simulation.hpp
```diff
@@ -1366,15 +1365,122 @@ template <typename problem_t> void AMRSimulation<problem_t>::calculateGpotAllLev
...
+
 		if (verbose) {
 			amrex::Print() << "\n";
 		}
 
 		// check for NaN
 		for (int lev = 0; lev <= finest_level; ++lev) {
-			AMREX_ALWAYS_ASSERT(!phi[lev].contains_nan()); // NOTE: this fails when multiple levels are fully refined
+			AMREX_ALWAYS_ASSERT_WITH_MESSAGE(!phi[lev].contains_nan(), fmt::format("NaN detected in phi at level {} after Poisson solve",
+											       lev)); // NOTE: this fails when multiple levels are fully refined
```
BenWibking: This comment only applies to the OpenBC solver.

BenWibking (changes requested):
LGTM. Minor comment below.

chongchonghe:
> > periodic BC in two dimensions and Dirichlet BC with phi = const. in the other dimension, and the value of this const does not matter because the gravitational potential has gauge freedom; gravitational force F = -∇φ is gauge invariant (independent of φ + constant)
> 
> It might be helpful to document that this is only true when the desired solution has a constant potential at the boundary. For free-space boundary conditions, this approximation will fail to hold when the planar symmetry is broken and there is significant mass near the boundary.

I believe as long as the Poisson solver comes to convergence, the solution is the correct solution to the periodic disk geometry, whether or not it's uniform. In the solution, `phi` at the upper `z` plane won't be uniform; it's accurate to whatever it should be, up to a constant.

chongchonghe:
Don't need to run GPU tests because there is only commentary change since previous successful test.

BenWibking:
> > > periodic BC in two dimensions and Dirichlet BC with phi = const. in the other dimension, and the value of this const does not matter because the gravitational potential has gauge freedom; gravitational force F = -∇φ is gauge invariant (independent of φ + constant)
> > 
> > 
> > It might be helpful to document that this is only true when the desired solution has a constant potential at the boundary. For free-space boundary conditions, this approximation will fail to hold when the planar symmetry is broken and there is significant mass near the boundary.
> 
> I believe as long as the Poisson solver comes to convergence, the solution is the correct solution to the periodic disk geometry, whether or not it's uniform. In the solution, `phi` at the upper `z` plane won't be uniform; it's accurate to whatever it should be, up to a constant.

I don't think this is the case. The phi at the upper z plane should be very nearly constant (due to the difference between the cell center and the upper boundary face, which is where the Dirichlet BCs are specified).

BenWibking (approved):

PR #1323, opened 2025-09-17
Making ParticleToMesh reproducible by rounding off rhs at 12th digit. 

chongchonghe:
### Description

This completely solves the particle reproducibility issue on a single level (multilevel yet to be tested). 

The solution is simple. Assuming there is no more than an order of a few particles in a cell, we can probably estimate the upper bound on the roundoff error caused by float-point addition. From my tests, it's always at or below 1e-13. We simply roundoff the `rhs_buffer` to the 12th digit before adding it to the real `rhs`. Since this is done after `SumBoundary`, a deterministic `SumBoundary` is not required. 

The loss of accuracy on `rhs_buffer` doesn't matter, because particle-mesh interaction always have huge systematics. Particularly, the CIC algorithm with cubic kernel is very crude and I speculate it's systematics is at least 1% (Ben can confirm this). Other particle-mesh interaction recipes, like sink accretion and SN, has a systematic of 1% to 20%. 

Tested with the spherical_collapse problem the same way I did in #1299. The PR gives reproducible results with 100000 particles in a 32^3 box with 4^3 fabs, i.e. ~3 particles per cell. 

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -1341,13 +1341,13 @@ template <typename problem_t> void AMRSimulation<problem_t>::roundoffMultiFab(am
 	// 1. Multiplying by factor = 2^15 + 1 = 32769 to shift significant bits
 	// 2. The multiplication and subsequent operations naturally truncate lower-order bits
 	// 3. The final subtraction c - (c - sum) recovers the rounded value
-	constexpr unsigned int digit_to_remove = 15;  // Remove 15 bits from mantissa
+	constexpr unsigned int digit_to_remove = 15;					 // Remove 15 bits from mantissa
 	constexpr auto factor = static_cast<amrex::Real>((1ULL << digit_to_remove) + 1); // 2^15 + 1 = 32769
-	
+
 	// Get array accessor for all patches at once
 	auto arr = mf.arrays();
 	const int ncomp = mf.nComp();
-	
+
 	// Apply roundoff algorithm to every grid point and component in parallel
 	amrex::ParallelFor(mf, [=] AMREX_GPU_DEVICE(int bx, int i, int j, int k) noexcept {
 		// Process all components at this grid point
@@ -1402,15 +1402,15 @@ template <typename problem_t> void AMRSimulation<problem_t>::calculateGpotAllLev
 					rhs_buffer[lev].define(grids[lev], dmap[lev], ncomp, nghost_rhs);
 					rhs_buffer[lev].setVal(0);
 				}
-				
+
 				// deposit mass into temporary buffer
 				particleRegister_.depositMass(amrex::GetVecOfPtrs(rhs_buffer), finest_level, Gconst_);
-				
+
 				// apply roundoff to buffer before adding to rhs
 				for (int lev = 0; lev <= finest_level; ++lev) {
 					roundoffMultiFab(rhs_buffer[lev]);
 				}
-				
+
 				// add buffer to rhs
 				for (int lev = 0; lev <= finest_level; ++lev) {
 					amrex::MultiFab::Add(rhs[lev], rhs_buffer[lev], 0, 0, ncomp, 0);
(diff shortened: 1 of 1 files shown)


BenWibking:
This would certainly be simpler.

My concern is that we don't have any rigorous error bounds that would enable us to compute the minimum number of bits to round.

Is it possible to empirically determine a function that returns the number of bits that should be chopped off given the machine epsilon and the number of particles per cell?

markkrumholz:
@BenWibking, @chongchonghe and I just met and we think we have figured out a way to get rigorous bounds that will guarantee reproducibility. The trick is as follows: for each FAB where we're going to deposit particles, we create a temporary array to hold error bounds -- let's call it `e` -- along with the sums `s` that will be the amount we're depositing in each cell.

When we do a deposition from a particle to a cell, we update both `s` and `e` as follows:
`s += ds`
`e += MACHINE_EPS * abs(s)`
where `ds` is the amount we're depositing from that particle and `MACHINE_EPS` is machine precision, corresponding to putting a 1 in the least significant digit. Then once we have processed all particles, we compute an upper bound on the relative error in each cell for which `s` is non-zero as
`r = e / s`.
Finally, we round each `s` by removing a number of bits given by
`nremove = log_2(r) + n_mantissa + 2`
where `n_mantissa` is the number of bits in the mantissa of our floating point representation.

The idea behind this algorithm is that `e` represents an absolute upper bound on the error that could accumulate due to the non-associativity of floating arithmetic -- each addition can only change the least significant bit, so `MACHINE_EPS * abs(s)` is an upper bound on the absolute error accumulated. Then `r` gives the relative error (which is conceivably large is there are lots of cancellations so `s` is much smaller than the sum of the absolute values of the `ds`'s). Then the calculation of `nremove` ensures that we're removing enough bits to hide those errors: if we just add a few numbers and there are no cancellations, then `r` will be comparable to `MACHINE_EPS`, and `log_2(r)` will be equal to the negative of the number of bits in the mantissa, so we will get `nremove` set to 2 (for safety) and we will kill the 2 least significant bits. Larger values of `r` will lead to more bits being removed, but in a way that guarantees that the amount of rounding is always at least 2 digits bigger than our upper bound on the error.

chongchonghe:
@markkrumholz @BenWibking Actually, nothing more has to be done for this PR, and it's ready for review and merge. 

Here is why. For pure-positive feedback, which is the case for mass deposition, the relative error on the sum is

$$
\frac{\sum_i \epsilon m_i}{\sum_i m_i} = \epsilon
$$

So, chopping off the last three digits (in base 10 numbers) is a universally valid solution to make mass deposition deterministic, no matter how many particles are in a cell. 

In the next PR, I'll make all other particle feedback mechanisms deterministic, and that's where tracking the absolute error is required.

markkrumholz:
OK, agreed. I will review this PR as soon as I'm not in a meeting (which may not be until tonight).

chongchonghe:
In fact, it's trivial to make all feedback reproducible by doing the same thing as I've already done with gravity, and no tracking of absolute error is required. This is because, in `particleMeshInteraction`, i do sink accretion first, then star formation, then SN feedback, and all of them are pure positive or negative feedback, so the upper bound on the relative error is always `epsilon`.

markkrumholz (approved):

markkrumholz:
I agree that this is fine for all existing implementations. However, we cannot rule out the possibility that a future particle implementation could have opposite signs, and there are clear examples where that would happen -- the most obvious one I can think of is in a PIC code where you both ions and electrons contributing to the charge field. I would therefore like to leave at least the description of the algorithm in the paper, with perhaps a note that this can be simplified if we know that a particular deposition is always single-signed.

chongchonghe:
let me merge this and update the regression test so that all regression test will pass. 

The only non-trivial thing to do for the rest is that I have to keep track of the mass taking off in stochastic star formation, communicate them, and then deposite the negative mass to the cells, instead of subtracting the mass on the fly.

chongchonghe:
Yes, I agree, it's possible in the future we want to do both positive and negative feedback in one routine, and we would need to do error tracking. I'll implement that in the next couple of days.

chongchonghe:
>  I would therefore like to leave at least the description of the algorithm in the paper, with perhaps a note that this can be simplified if we know that a particular deposition is always single-signed.

Yes. I would simply say that there is a fixed upper bound on the relative error, which is `epsilon`, for pure positive or pure negative feedback. For feedback with mixed signs, we'll have to keep track of the absolute error and compute how many digit we need to chop off.

PR #1328, opened 2025-09-20
Robust reproducible particle operations: gravity, accretion, and SN

chongchonghe:
### Overview

This PR addresses long-standing issues with non-deterministic regression tests in particle deposition routines, particularly under stress tests and on GPUs (e.g., Setonix and avatargpu). Previous fixes (see #1323) were insufficient: the key operation `result = c - (c - sum);` was optimized away by compilers, failing to enforce consistent rounding and reproducibility. The new approach uses a combination of `volatile` and explicit mantissa truncation to robustly zero the last N bits, ensuring deterministic results for sufficiently large number of cell updates.

### Motivation & Problem Statement

Regression tests, such as SphericalCollapse and DiskGalaxy, previously failed to reproduce results reliably when the number of particles per cell increased or when stress-tested over many steps. The root cause was floating-point non-determinism: roundoffMultiFab did not behave consistently on some GPUs, and compiler optimizations bypassed intended rounding logic. Initial tests were flawed due to insufficiently strict initial conditions, which masked the issue until more particles were introduced.

### Approach & Algorithm

- **Volatile workaround**: The PR adopts the portable volatile pattern:
  ```cpp
  volatile double t = c - sum;
  double result = c - t;
  ```
  This prevents the compiler from optimizing away the rounding, forcing two FP operations with a memory round-trip. See discussion for caveats and portability notes.

- **Mantissa truncation**: The code now explicitly truncates the mantissa bits to ensure reproducibility for a sufficient number of cell updates. The truncation bit is determined by the formula:
  ```
  N_chopoff = log2(N_par) + log2(N_margin) + 2
  ```
  where:
    - N_par: number of particles interacting with the cell
    - N_margin: a redundancy number set via runtime parameter `particles.reproducibility_roundoff_redundancy` (default 20). This number can be estimated as the number of cells that interact with multiple particles multiplied by the number of time steps you want to evolve while ensuing determinism, plus a small extra margin. 

  For SphericalCollapse, this simplifies to `N_margin = 16 + log2(N_step)`.

- **Probability of non-reproducibility**: The probability that a cell deposition is non-reproducible is bounded by `2^-M`, where M is the safety margin in the truncation. To guarantee reproducibility, set M such that `nstep * ncell * 2^-M << 1`.

- **Generalization**: Mark suggests expressing the truncation rule in terms of an error bound r (see paper appendix):
  ```
  Truncate at bit: log2(r) - N_mantissa + M
  ```
  This formulation applies regardless of whether summed numbers are the same sign.

- **Statistical argument**: Once a problem passes reproducibility for several runs, it is extremely unlikely for a particle to ever overlap with a float-point boundary, as the chance of n consecutive reproducible runs with a hidden boundary overlap is `1/2^n`.

- **Random walk refinement**: For random sums, the upper bound on error scales as `sqrt(N_par) * epsilon`, so N_par may be replaced by N_par/2 in the formula for a more accurate estimate.

### Implementation Notes

- The new logic avoids rounding for cells interacting with 0 or 1 particles, which are the majority in most runs.
- TODO comments have been added to clarify that the scheme only works for single-sign depositions and would need revision for PIC simulations.
- The regression tests have been updated: SphericalCollapse is now reproducible up to 1000 steps with 1M stars in 128³ cells (N=28).

Hopeful the DiskGalaxy test will be deterministic as well with proper choice of `particles.reproducibility_roundoff_redundancy`.

### Related issues
Addresses #1125 . We will know this PR resolves that issue once we run the regression tests. 

### Checklist


BenWibking:
This is not a good solution. Your new rounding implementation can affect both the exponent and mantissa, which will cause some weird issues as you mention.

Fortunately, there is a better solution: you can use `volatile`:
```
volatile double t = c - sum;
double result = c - t;
```

BenWibking:
Explanation:
```
▌ Is voltaile guaranteed to prevent optimization portably?

> Short answer: Yes for this pattern, but it’s a blunt tool.

  - What volatile guarantees
      - C/C++ require that accesses to a volatile object are performed as written (a real store and a real load of the declared type).
      - Because a volatile object can change “outside” the compiler’s view, the compiler may not assume t == c - sum after volatile double
  t = c - sum;, so it cannot fold c - t into sum.
      - Practically, this forces two FP operations with a store/load in between, yielding two roundings at the object’s precision.
  - Example that fences reassociation
      - volatile double t = c - sum;
      - double result = c - t;
  - Caveats
      - Volatile is not a general optimization barrier: compilers can still reorder non‑volatile work around it, as long as they preserve
  the order of volatile accesses.
      - It is not a concurrency primitive (no synchronization/atomicity).
      - With “fast-math” enabled, other algebraic transforms elsewhere may still occur; volatile only fences at the volatile access.
      - It inhibits vectorization and other optimizations and introduces a forced memory round-trip.
  - Prefer when possible
      - Avoid -ffast-math; or use strict FP modes/pragmas (e.g., #pragma STDC FENV_ACCESS ON, Clang #pragma clang fp reassociate(off),
  MSVC /fp:precise or /fp:strict) if available, to scope the constraint more precisely.

  Bottom line: Using a volatile temporary is a portable, standard‑conforming way to prevent the specific optimization of c - (c - sum) into
  sum, but it’s heavy-handed. If your toolchain supports FP pragmas/attributes, they’re a cleaner, narrower fix.
```

I don't think it's a good idea to do what Codex suggests here -- that would require adding different pragmas for every compiler we use.

BenWibking:
If we avoid volatile, we would have to implement three separate versions of this line of code:

CUDA (using math intrinsics that map directly to GPU assembly instructions)
```
double t = __dsub_rn(c, sum);
double result = __dsub_rn(c, t);
```

AMD GPU and Clang CPU:
```
#pragma clang fp reassociate(off)
double result = c - (c - sum);
#pragma clang fp reassociate(on)
```

GCC CPU:
```
#pragma GCC push_options
#pragma GCC optimize("no-fast-math","no-associative-math","no-tree-reassoc","rounding-math","trapping-math")
double result = c - (c - sum);
#pragma GCC pop_options
```

markkrumholz:
Here is another potential solution to do bit-rounding: 

1. Use reinterpret_cast to cast the float to an unsigned int.
2. Bit shift by however many binary decimal places we want to remove, then bit shift back -- this will have the effect of zeroing out some number of binary digits.
3. Use reinterpret_cast to convert the resulting int back to float.

This is guaranteed not to be optimized out, is compiler-independent, avoids issues with volatile, and will let us precisely control the number of bits that get rounded. The only subtlety involves endianness: whether we bit shift left and then right, or right and then left, depends on the endianness of the system, and we therefore need a little macro to determine the floating endianness. With C++20 this can be done in a standard way at compile time using std::endian. With earlier C++ it requires writing a little macro function, but should still be straightforward.

BenWibking:
reinterpret_cast from/to anything except char* is undefined behavior. This only works with gcc, which allows it because a lot of old code did this.

BenWibking:
The only legal way to do this in C++ is to use `std::bit_cast`, but that requires C++20: https://en.cppreference.com/w/cpp/numeric/bit_cast.html

markkrumholz:
We can reinterpret_cast to char* and then to int. Basically, I just want to convert to a type that allows use of the bit bit shift operator.

BenWibking:
> We can reinterpret_cast to char* and then to int. Basically, I just want to convert to a type that allows use of the bit bit shift operator.

This is still illegal. You *cannot* reinterpret_cast from one semantically-meaningful type to a different one, even transitively.

markkrumholz:
Actually, as an alternative, we don't even need to go through int. Just reinterpret cast to char*, then set the last character to \0 (where last depends on endianness), then reinterpret_cast back. That's the equivalent of zeroing out the last 8 bits.

BenWibking:
```
Many compilers issue "strict aliasing" warnings in such cases, even though technically such constructs run afoul of something other than the paragraph commonly known as the "strict aliasing rule".

The purpose of strict aliasing and related rules is to enable type-based alias analysis, which would be decimated if a program can validly create a situation where two pointers to unrelated types (e.g., an int* and a float*) could simultaneously exist and both can be used to load or store the same memory (see [this email on SG12 reflector](https://web.archive.org/web/20201128194944/http://www.open-std.org/pipermail/ub/2016-February/000565.html)). Thus, any technique that is seemingly capable of creating such a situation necessarily invokes undefined behavior.

When it is needed to interpret the bytes of an object as a value of a different type, [std::memcpy](https://en.cppreference.com/w/cpp/string/byte/memcpy.html) or [std::bit_cast](https://en.cppreference.com/w/cpp/numeric/bit_cast.html)(since C++20) can be used:

double d = 0.1;
std::int64_t n;
static_assert(sizeof n == sizeof d);
// n = *reinterpret_cast<std::int64_t*>(&d); // Undefined behavior
std::memcpy(&n, &d, sizeof d);               // OK
std::bit_cast<std::int64_t>(d);          // also OK
```
This is really, truly the only legal way to do it in C++. I tried all of the hacks, and they break with compilers we are using in production.

Reference: https://en.cppreference.com/w/cpp/language/reinterpret_cast.html

markkrumholz:
If you want to do things in fewer than 1-byte increments, you just need to do the arithmetic to figure out what character is produced by zeroing out only some of the bits in the final byte.

markkrumholz:
Alternately, the example you put in @BenWibking shows another valid way to do it: create an int64_t, use memcpy to copy the bits from the float to it, bit shift it, then use memcpy to copy the bits back to the float.

chongchonghe:
I tried a few tricks to avoid compiler optimization. Once example is shown below. I also tried other methods without volatile.  However, it does not produce determinitic results. I think there is a fundamental issue with this simple algorithm. 
```
      volatile amrex::Real vc = c;
      volatile amrex::Real vsum = sum;
      volatile amrex::Real tmp = vc - vsum;
      arr[bx](i, j, k, n) = vc - tmp;
```

The algorithm currently implemented in this PR is more stable, but it also failed after hundreds of steps in a test with 1M stars. I have to set the accuracy to 1e-6 to make it stably run for 1000 steps. 

I think precise bit-wise manipulation is required to make it reliably deterministic.

BenWibking:
This is a classic computer science algorithm invented in the 1960s. I think it can't be wrong. If this doesn't work, there has to be a compiler bug.

Can you list exactly which compilers and compiler versions don't work with the code above? Then we can try to construct minimal failing/working examples using https://godbolt.org/ that will show the generated assembly.

BenWibking:
Otherwise, I agree with Mark - I think using `std::memcpy`, bit shifting and `std::memcpy` back is the next best option.

BenWibking:
This is a complete minimal example that shows rounding with this algorithm using `volatile` using Clang: https://godbolt.org/z/csE8ozqbj

You can immediately verify that it produces two separate subtraction instructions, and rounds the output correctly.

Can you send me a link to the exact compiler you are using where this doens't work?

chongchonghe:
> This is a complete minimal example that shows rounding with this algorithm using `volatile` using Clang: https://godbolt.org/z/csE8ozqbj
> 
> You can immediately verify that it produces two separate subtraction instructions, and rounds the output correctly.
> 
> Can you send me a link to the exact compiler you are using where this doens't work?

This algorithm successfully modifies the MultiFab (the first volatile is necessary to make a difference on GPU):
```cpp
			volatile amrex::Real c = factor * val;
			volatile amrex::Real a = c - val;
			arr[bx](i, j, k, n) = c - a;
```

However, even with `digit_to_remove = 25`, it occasionally produces indeterministic results, with the modified value varies by 3e-10 between two runs, which is an order of magnitude lower than `2^-(52-25+1)=3.7e-9`, so this is unlikely due to a difference in the last reserved bit.

BenWibking:
> > This is a complete minimal example that shows rounding with this algorithm using `volatile` using Clang: https://godbolt.org/z/csE8ozqbj
> 
> > 
> 
> > You can immediately verify that it produces two separate subtraction instructions, and rounds the output correctly.
> 
> > 
> 
> > Can you send me a link to the exact compiler you are using where this doens't work?
> 
> 
> 
> This algorithm successfully modifies the MultiFab (the first volatile is necessary to make a difference on GPU):
> 
> ```cpp
> 
> 			volatile amrex::Real c = factor * val;
> 
> 			volatile amrex::Real a = c - val;
> 
> 			arr[bx](i, j, k, n) = c - a;
> 
> ```
> 
> 
> 
> However, even with `digit_to_remove = 25`, it occasionally produces indeterministic results, with the modified value varies by 3e-10 between two runs, which is an order of magnitude lower than `2^-(52-25+1)=3.7e-9`, so this is unlikely due to a difference in the last reserved bit. 

I see. Ok, I'm guessing that a small rounding error is being amplified by subsequent evolution. It might help to try to reproduce this with a small number of steps and then see what happens at the first step where things diverge fr a reference solution.

Does using volatile with Kahan summation reduce the rounding error?

chongchonghe:
Actually, i made a mistake in the calculation (or, it's because fcompare report is confusing). the relative difference after roundoff is indeed about 4e-9. So, the inconsistency exists at the last bit we preserve, which is unavoidable in this algorithm.

chongchonghe:
in binary notation, the special case I mentioned above is unavoidable as well. e.g. two mantissas, 01111111 and 1000000 is only different by 1/128, but by rounding off 5 digits, we get 011 and 100, which is different by the precision we deman, i.e. 2^(-2). 

@BenWibking Do you know the theory behind this? Is this roundoff algorithm guaranteed to give you deterministic result when two numbers are different by less than the precision? Does it handle the special case i mentioned?

chongchonghe:
> Does using volatile with Kahan summation reduce the rounding error?

I am not trying Kahan at all, since we can't do kahan with this particle-by-particle approach.

chongchonghe:
BTW, I was wrong when I said the relative error of naive summation is $\epsilon$. It can be shown that the upper bound of the relative error in the sum of all positive numbers is actually $(N-1) \epsilon$. My original statement on Slack on Thursday was correct. Therefore, a track of the number of additions is necessary. We would count the number of values added, `N`, not the sum of $\epsilon x_i$ (which is trivially `epsilon sum`).

markkrumholz:
> BTW, I was wrong when I said the relative error of naive summation is ϵ . It can be shown that the upper bound of the relative error in the sum of all positive numbers is actually ( N − 1 ) ϵ . My original statement on Slack on Thursday was correct. Therefore, a track of the number of additions is necessary. We would count the number of values added, `N`, not the sum of ϵ x i (which is trivially N ϵ s u m ).

OK, but if you're going to do that, you might as well just keep track of the sum of the epsilon's, because it's the same amount of work and then it will work even when the numbers are not all the same sign.

But I don't have a solution to your first issue, because I realize now that this is a fundamental issue for any rounding approach. Rounding always involves breaking at a boundary, where you map things on one side of the boundary to one value and things on the other side to another, and there is no way to guarantee that floating point associativity issues won't every. land right on the boundary. The best you can do is make the bins wider to reduce the probability of this happening, but you can never guarantee it won't happen.

markkrumholz:
I should add, though, that the ability to reduce the probability of a non-repeatable round to arbitrarily small values is probably all we need for the purposes of code testing and reproducibility. Here's how I'm thinking about this: if do a non-associative addition such that different runs of the code may produce results that differ at the Nth binary decimal place, then if we round to the Mth binary decimal place, for some M > N, then the probability that a given sum lands right at the edge of a rounding interval is 1 in 2^(M-N). So we cannot guarantee perfect reproducibility for arbitrary numbers of particles or time steps, but if we encounter a situation where we want to enforce reproducibility for a finite number of steps, we can make M sufficiently large compared to N that the probability that two different runs will give different results at the bit level is as small as we want.

chongchonghe:
Exactly. I agree. We can do a quick estimation of the probably of this happening. In 128^3 cells, across 1000 steps, this is 1e9 occurrence of this roundoff. That means we need to round off extra 9 digits to make it very safe that this edge case won't happen. Since in practice, only a small fraction of the cells will interact with more than 1 particles, this could drop by several orders of magnitude. However, I think this is still totally fine as long as we do not allow the relative error to go above, say, 1e-4.

BenWibking:
Right, I agree. I think the best we can do with a post-summation rounding algorithm is to make it very improbable that it is non-reproducible. However, it appears to be not quite as improbable as I had hoped.

I will ask Weiqun if it's possible to implement the fixed-point superaccumulation algorithm in AMReX.

chongchonghe:
> OK, but if you're going to do that, you might as well just keep track of the sum of the epsilon's, because it's the same amount of work and then it will work even when the numbers are not all the same sign.

Yes, tracking `epsilon | x |` is helpful when there are cancelations.

chongchonghe:
By the way, regarding Kahan summation. The only advantage Kahan summation has over naive summation is that it reduces the relative error from `N epsilon` to `2 epsilon`. Note that this `N epsilon` in the naive addition is the **upper bound** and only happens when all the errors are at the same direction. In most cases, the actually relative error is close to `epsilon` because of cancellation (of positive and negative roundoff errors). If we can put an upper bound to the maximum number of particles in a cell at, say, 1000, then is reduces the relative error by at most 1000 in extremely rare cases, which still doesn't matter much in our cases. Additionally, Kahan summation does not solve the roundoff issue we are facing. Therefore, I think we can put Kahan summation off the table.

BenWibking:
> By the way, regarding Kahan summation. The only advantage Kahan summation has over naive summation is that it reduces the relative error from `N epsilon` to `2 epsilon`. Note that this `N epsilon` in the naive addition is the **upper bound** and only happens when all the errors are at the same direction. In most cases, the actually relative error is close to `epsilon` because of cancellation (of positive and negative roundoff errors). If we can put an upper bound to the maximum number of particles in a cell at, say, 1000, then is reduces the relative error by at most 1000 in extremely rare cases, which still doesn't matter much in our cases. Additionally, Kahan summation does not solve the roundoff issue we are facing. Therefore, I think we can put Kahan summation off the table.

I think that is probably the case, but I think it would be useful to compute the number of rounding bits needed to make our planned MW simulation reproducible with probability > 0.99 for both summation methods. If neither method can make the whole simulation reproducible for the entire planned runtime, then I agree Kahan is definitely not worth it.

chongchonghe:
@markkrumholz @BenWibking This is ready for your review.

markkrumholz:
It looks like the method you have implemented works by counting, so it will only work if the quantities being deposited are strictly same-sign. Is that correct?

chongchonghe:
> It looks like the method you have implemented works by counting, so it will only work if the quantities being deposited are strictly same-sign. Is that correct?

Yes, it works for all currently existing feedbacks, and it has the advantage that we can avoid rounding the cells that interact with 0 or 1 particles, which are the vast majority of the cells in most runs. I tested it on setonix and the spherical collapse test is deterministic.

src/particles/particle_utils.hpp
```diff
@@ -86,32 +88,88 @@ inline void roundoffMultiFab(amrex::MultiFab &mf)
...
-			// This is mathematically equivalent to sum, but with reduced floating-point precision
-			arr[bx](i, j, k, n) = c - (c - sum);
+			const auto val = arr[bx](i, j, k, n);
+
+			if (std::abs(val) < tiny) {
+				arr[bx](i, j, k, n) = 0.0;
+				continue;
+			}
+
+			volatile amrex::Real const c = factor * val;
```
markkrumholz: I would just add a comment here explaining the purpose of the volatile declaration, since otherwise someone in the future is going to look at it and be confused.

markkrumholz:
OK, that's fine. I would suggest just adding a TODO comment in the source code as a reminder to ourselves that the scheme you have implemented only works with single-sign depositions, so for example it would not work with PIC simulations.

I also left one other minor inline comment.

Once you have dealt with those two, let me know and I will approve.

markkrumholz:
I guess it needs to pass clang-tidy too, so let me know one you have done that as well, so that I don't have to re-review.

src/particles/particle_utils.hpp
chongchonghe: done. I added comments at line 138.

chongchonghe:
> OK, that's fine. I would suggest just adding a TODO comment in the source code as a reminder to ourselves that the scheme you have implemented only works with single-sign depositions, so for example it would not work with PIC simulations.
> 
> I also left one other minor inline comment.
> 
> Once you have dealt with those two, let me know and I will approve.

OK. I'll also do cell count for sink accretion and SN feedback. Working on it right now.

src/particles/particle_utils.hpp
chongchonghe: done.

markkrumholz (approved):

PR #1332, opened 2025-09-23
Rotated Linear Alfven Wave Test Problem

AstroKriel:
### Description
This PR extends the linear Alfven-wave test, allowing the wavefront to be rotated to essentially arbitrary orientations in the domain. The wavefront direction is set by integer modes (nx, ny, nz); the background-field alignment can also be adjusted. Face-centered B is initialised from a magnetic vector potential so the discrete divergence constraint (div B = 0) holds at t=0. An orthonormal basis maps between the math reference frame and the problem (grid) frame.

Details about the setup: 
[alfven-wave-setup.pdf](https://github.com/user-attachments/files/22631263/alfven-wave-setup.pdf)

As confirmation that the problem evolves correctly, here is a video of a non–grid-aligned scenario with (nx, ny, nz) = (1, 2, 3) and theta = 45 deg (k and b0 misaligned). This problem evolves as expected and matches the analytic linear solution within the existing tolerance (error norm < 0.005).

alfven-wave-nx=1-ny=2-nz=3-angle=45

### Checklist


AstroKriel:
I'll provide some diagnostic plots (confirming we see the right behaviour) and math to support the new ICs, soon

AstroKriel:
@BenWibking it looks like the linear Alfven wave test problem is too big to run with the address sanitiser turned on. Do we need to run every MHD test problem with it enabled, or would it be sufficient to run ASAN on for only a single MHD test problem?

BenWibking:
> @BenWibking it looks like the linear Alfven wave test problem is too big to run with the address sanitiser turned on. Do we need to run every MHD test problem with it enabled, or would it be sufficient to run ASAN on for only a single MHD test problem?

You could change it so it only runs for a single MHD wave test.

BenWibking:
@AstroKriel Is this ready to be reviewed, or are you still working on it?

AstroKriel:
@BenWibking This is done and ready for review. See the description for relevant details about how I changed the ICs.

markkrumholz:
@BenWibking, did you want to be the one to review this? I can if you don't want to or have the time.

BenWibking:
> @BenWibking, did you want to be the one to review this? I can if you don't want to or have the time.

No, feel free to review this. I don't have time at the moment.

src/problems/AlfvenWaveLinear/test_alfven_wave_linear.cpp
```diff
@@ -35,134 +38,207 @@ template <> struct quokka::EOS_Traits<AlfvenWaveLinear> {
...
+{
+	return {vfield1[1] * vfield2[2] - vfield1[2] * vfield2[1], vfield1[2] * vfield2[0] - vfield1[0] * vfield2[2],
+		vfield1[0] * vfield2[1] - vfield1[1] * vfield2[0]};
+}
 
-// background states
-constexpr double bg_density = 1.0;
-constexpr double bg_pressure = sound_speed * sound_speed * bg_density / gamma_gas;
-constexpr double bg_mag_amplitude = 1.0;
+AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE void normaliseVector(std::array<amrex::Real, 3> &vfield)
```
markkrumholz: Minor thing, but in keeping with the convention in the rest of the code I suggest using US rather than Australian English spellings.

src/problems/AlfvenWaveLinear/test_alfven_wave_linear.cpp
```diff
@@ -35,134 +38,207 @@ template <> struct quokka::EOS_Traits<AlfvenWaveLinear> {
...
+AMREX_GPU_MANAGED std::array<amrex::Real, 3> k_dir_prf{1.0, 0.0, 0.0};		// NOLINT
+AMREX_GPU_MANAGED std::array<amrex::Real, 3> inplane_dir_prf{0.0, 1.0, 0.0};	// NOLINT
+AMREX_GPU_MANAGED std::array<amrex::Real, 3> outofplane_dir_prf{0.0, 0.0, 1.0}; // NOLINT
 
-AMREX_GPU_DEVICE auto computeMagneticVectorPotential_x(double /*x1*/, double /*x2*/, double /*x3*/, double /*time*/) -> double { return 0.0; }
+// wavefront
+AMREX_GPU_MANAGED double k_magn = 2.0 * M_PI; // NOLINT
 
-AMREX_GPU_DEVICE auto computeMagneticVectorPotential_y(double x1, double /*x2*/, double /*x3*/, double time) -> double
+AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE auto rotatePRF2MRF(const std::array<amrex::Real, 3> &vec_prf) -> std::array<amrex::Real, 3>
```
markkrumholz, line 150: I think you should add some comments for this function and its companion inverse function explaining what this function is actually doing, since it is far from obvious in just looking at the expressions that the underlying operation here is multiplying a vector by a rotation matrix.

markkrumholz (changes requested):
This looks basically good, just left some minor comments.

Also, I'd like to suggest that we have two versions of this test: one grid-aligned and that can therefore be run pretty fast as essentially a 1d test, and one rotated and thus needing to use a larger 3d grid. I think there is no need to include the non-grid aligned case in the automated test suite that is part of the automatic pipeline; it's getting to the point where that pipeline takes enough time to run that we shouldn't just automatically put everything in it.

So my proposal is that we fork this test in two. Use the same exact setup, but have two tests, one rotated and one non-rotated, and only put the latter into the automated test suite.

src/problems/AlfvenWaveLinear/test_alfven_wave_linear.cpp
AstroKriel: Quokkas are an AUSTRALIAN animal!! But OK. ;-)

AstroKriel:
@markkrumholz Thanks, this makes sense. I’ve split the test into two versions: (1) a simpler nearly grid-aligned case (nx,ny,nz)=(1,1,0), theta=0, run at 16^3 as part of the automated test suite, and (2) a more complex rotated case (nx,ny,nz)=(1,2,3), theta=45, run at 64^3 as part of the regression suite.

src/problems/AlfvenWaveLinear/test_alfven_wave_linear.cpp
markkrumholz: I just want there to be one language in the codebase -- I do not want us to go the way of Ramses, where the core hydro routines are all commented in French because Romain Teyssier wrote them. Instead of sL and sR, Ramses has sD and sG -- droit and gauche -- and that way lies madness. ;-)

src/problems/AlfvenWaveLinear/test_alfven_wave_linear.cpp
AstroKriel: Done.

markkrumholz:
For the automated test suite I was suggesting something like 64 x 8 x 8 -- that way we have enough resolution to tell if the accuracy is getting screwed up, but since the domain is thin it's still cheap.

markkrumholz (approved):

PR #1340, opened 2025-09-26
[WIP] Emit warning when bitwise reproduciblity is not possible

BenWibking:
### Description
This warns the code user when any settings have been enabled that use atomic operations, which are not bitwise reproducible. The user should not expect identical runs to produce bitwise identical results.

### Related issues
N/A

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -560,6 +560,22 @@ template <typename problem_t> void AMRSimulation<problem_t>::initialize()
 
 	readParameters();
 
+	bool wroteWarning = false;
+	if (max_level > 0) {
+		amrex::Print() << "\n[Warning] [Reproducibility] Adaptive mesh refinement is enabled; reflux corrections use atomic updates, so results are not bitwise reproducible.\n";
+		wroteWarning = true;
+	}
+
+	if constexpr (Particle_Traits<problem_t>::particle_switch != ParticleSwitch::None) {
+		amrex::Print() << (wroteWarning ? "" : "\n")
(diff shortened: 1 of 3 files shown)


chongchonghe:
All particle operations implemented to date are reproducible as long as you set `reproducibility_roundoff_redundancy` to a proper value (#1328). This is not different from mass deposition for self gravity.

src/simulation.hpp
```diff
@@ -560,6 +560,24 @@ template <typename problem_t> void AMRSimulation<problem_t>::initialize()
...
+	if (max_level > 0) {
+		amrex::Print() << "\n[Warning] [Reproducibility] Adaptive mesh refinement is enabled; reflux corrections use atomic updates, so results are "
+				  "not bitwise reproducible.\n";
+		wroteWarning = true;
+	}
+
+	if constexpr (Particle_Traits<problem_t>::particle_switch != ParticleSwitch::None) {
+		amrex::Print()
+		    << (wroteWarning ? "" : "\n")
+		    << "[Warning] [Reproducibility] Particle deposition uses atomic reductions; runs with physical particles are not bitwise reproducible.\n";
```
chongchonghe, line 573: We could explain more clearly. 
All particle operations implemented to date are reproducible as long as you set reproducibility_roundoff_redundancy to a proper value (https://github.com/quokka-astro/quokka/pull/1328). This is not different from mass deposition for self gravity.

chongchonghe (changes requested):

BenWibking:
We can make AMR reproducible by switching to `amrex::FluxRegister` (instead of `amrex::YAFluxRegister`), so I'm putting this PR on hold until that's done.

PR #1343, opened 2025-09-26
Rewrite hydro retry algorithm

BenWibking:
### Description
This rewrites the hydro retry algorithm so that it never rolls back the state after an accepted substep. Instead, if a non-initial substep fails during a retry, it will reduce the timestep only for the remaining interval between the current accepted time and the final level time `time + dt_lev`.

Also, we now only update the flux register and tracer particles after a substep has been accepted. This avoids the need to backup and restore the flux register and the tracer particles, which reduces the overall memory usage of the hydro update.

Pseudocode implementation:
```
time = currentLevelTime
max_total_substeps = 1 << max_retries
accepted_state_cc = snapshotCellCenteredState()
accepted_state_fc = snapshotFaceCenteredState()

completed_units = 0
cur_retry_level = 0
while completed_units < max_total_substeps and cur_retry_level <= max_retries:
    total_substeps = 1 << cur_retry_level
    units_per_substep = max_total_substeps // total_substeps
    start_substep = completed_units // units_per_substep
    dt_substep = dt_lev / total_substeps

    for substep_index in range(start_substep, total_substeps):
        current_substep_time = time + substep_index * dt_substep
        substep_success = advanceHydroAtLevel(current_substep_time, dt_substep)
        if not substep_success:
            restoreHydroState()
            cur_retry_level += 1
            break
        completed_units += units_per_substep
        updateAcceptedHydroState()

if completed_units < max_total_substeps:
    writeFatalDiagnostics()
    abortHydroStep()
```

### Related issues
N/A

### Checklist

### DIFF
--- a/docs/markdown/hydro_integrator.md
+++ b/docs/markdown/hydro_integrator.md
@@ -6,8 +6,8 @@ The hydro integrator advances the conservative gas (and optionally MHD) variable
 - **Source splitting:** Each accepted substep begins and ends with `addStrangSplitSourcesWithBuiltin`, applying the problem-specific source terms for half a timestep.
 - **Runge–Kutta stages:** `advanceHydroAtLevel` implements an SSP-RK2 integrator (with an optional first-order Euler fallback) over two stages. Stage 1 advances from the old state to an intermediate state; Stage 2 starts from the intermediate state and computes the second stage fluxes. The final update is done from the initial state using the average from the Stage 1 and Stage 2 fluxes.
 - **Flux register coupling:** When refluxing is enabled the combined RK fluxes are accumulated via `incrementFluxRegisters`, guaranteeing flux conservation across AMR coarse/fine interfaces.
-- **Tracer and dual-energy support:** Face-averaged velocities produced during the update are reused to advance tracer particles at the end of the step. At each stage, face-centered velocities are used to update the auxiliary internal energy and perform a dual energy sync.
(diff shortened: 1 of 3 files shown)


BenWibking:
Need to check:
```
  - src/QuokkaSimulation.hpp:1477 – High: we accumulate the stage-one flux into flux_rk2 (and avgFaceVel) before running FOFC, but never update those
  aggregates after the limiter rewrites fluxArrays/faceVel. Later we reflux with incrementFluxRegisters(..., flux_rk2, dt_lev) and advect tracers
  off avgFaceVel, so any cell that tripped the first-order fallback still sends the high-order (unstable) flux/velocity downstream. The previous
  implementation incremented the registers directly from the already-corrected fluxArrays, so this regression lets bad fluxes leak into the conserved
  state, reflux registers, and tracer motion. Please either defer the Saxpy until after FOFC or reapply it with the corrected fluxes before we call
  incrementFluxRegisters / use avgFaceVel.
```

```
  - Bug (high): The new refluxing path now feeds flux_rk2 directly into incrementFluxRegisters, but the SSP-RK1/RK2 first-order fallback never updates
  flux_rk2 after it swaps the stage-1 fluxes to their FO versions (replaceFluxes(fluxArrays, FOfluxArrays, redoFlag);). As a result, whenever FOFC
  fires in stage 1 the cell update uses the diffusive flux, while the flux register still accumulates the pre-fallback high-order flux (and in SSP-RK1
  there is no stage-2 pass to clean this up). That breaks coarse/fine conservation and tracer advection in exactly the regimes where the fallback is
  protecting us. (src/QuokkaSimulation.hpp:1471, src/QuokkaSimulation.hpp:1528, src/QuokkaSimulation.hpp:1695)
```

markkrumholz (approved):

PR #1346, opened 2025-09-28
Pass time to reference soln

AstroKriel:
### Description
This PR modifies computeReferenceSolution and computeAfterEvolve to accept the current simulation time, which allows the reference solution to be evaluated at the actual end time, making the final comparison correct for time-dependent problems.

### Checklist

### DIFF
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -823,7 +823,8 @@ void QuokkaSimulation<problem_t>::computeReferenceSolution(amrex::MultiFab &ref,
 
 template <typename problem_t>
 void QuokkaSimulation<problem_t>::computeReferenceSolution_fc(amrex::MultiFab &ref, amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &dx,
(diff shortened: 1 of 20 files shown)


chongchonghe (commented):
`computeReferenceSolution` is a member function of QuokkaSimulation, so it has access to all member functions and variables of QuokkaSimulation, including the current time as `tNew_[0]`. See examples in test_hydro_sms.cpp . For the same reason, `dx` and `prob_lo` are not necessary, but I guess Ben included them for convenience (so that you use `dx` instead of `geom[0].CellSizeArray()`).

BenWibking:
> `computeReferenceSolution` is a member function of QuokkaSimulation, so it has access to all member functions and variables of QuokkaSimulation, including the current time as `tNew_[0]`. See examples in test_hydro_sms.cpp . For the same reason, `dx` and `prob_lo` are not necessary, but I guess Ben included them for convenience (so that you use `dx` instead of `geom[0].CellSizeArray()`).

Right. Modifying the function signature is not necessary to solve this problem.

Issue #1062, opened 2025-05-11
set output cadence separately for fields and particles


chongchonghe:
**Describe the proposal**

Add a new runtime parameter `particle_file_interval` to control the cadence of particle data outputs. In some cases, like GMC sims, we want to dump particle data more frequently than plotfiles. In some cases, we want the opposite. 

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions you've considered.

**Additional context**
Add any other context about the feature proposal here.

BenWibking:
I think what we want to do is have a DiagPlotfile class that inherits from DiagBase (https://github.com/quokka-astro/quokka/blob/development/src/io/DiagBase.H).

Then for each instance of DiagPlotfile in the inputs file, you can turn on/off multifab outputs and particle outputs individually. We can also add the ability to set the fields we output for each instance separately, and which species to output for each instance. That means we can output different particle species at difference intervals.

BenWibking:
The reason I think this is preferable is because in cases where the particles and multifabs are output at the same coarse steps, then we just want to keep them inside the same pltXXXXX directory. If they are not output at the same cadence, then we put them in distinct folders for each output.

chongchonghe:
Sounds good. I'll do that in the next couple of days. It will be useful for some of the tests I'm running now.

chongchonghe:
@BenWibking Here is my design. Please let me know what you think.

Example 1: 
```
plotfile_interval = 2
```
This will write plotfiles with all fields and particle data every 2 coarse steps. This is our current design. 

Example 2:
```
plotfile.type = field particle
plotfile.interval = 2 # Output cadence for all types (in number of coarse steps)
```
This is the same as Example 1: output fileds and particles every 2 steps.

Example 3:
```
plotfile.type = field particle
plotfile.field.interval    = 1 # Output cadence for field
plotfile.particle.interval    = 2 # Output cadence for field
```
This will write plotfiles every 1 coarse steps, but will only include particles every 2 steps. 

Example 4: 
```
plotfile.type = field particle
plotfile.field.interval    = 1 # Output cadence for fields
plotfile.particle.interval    = 2 # Output cadence for particles
plotfile.field.names = gasDensity gasInternalEnergy # List of variables to include in output (default to all)
plotfile.particle.names = CIC_particles # list of particle types to include in output (default to all types)
```

We can also add `plotfile.time_interval`, `plotfile.particle.time_interval` to output plotfiles by time interval.

[...]

Issue #1356, opened 2025-09-29
Automate ProjectionPlot 


chongchonghe:
**Describe the proposal**

We can automate projection plot calculation and ask the user to specify the state variables to store at runtime.

**Additional context**
#1339

BenWibking:
@chongchonghe Did you implement `ComputeProjections`: https://quokka-astro.github.io/quokka/insitu_analysis/#2d-projections ?

chongchonghe:
Oh, I missed that. I see, i have to write computePlaneProjection myself.

chongchonghe:
but this shouldn't be necessary. It's straight forward to do projections of existing components. We can either specify which components to compute projections for by giving a list of integers or strings representing the default variable names.

BenWibking:
Yes, setting things at runtime would be better.

For the shock-cloud problem, I wanted to compute projections of derived variables like this:
https://github.com/quokka-astro/quokka/blob/b642c024d70c60eba2f4d5e2ac5c4238f79015a1/src/problems/ShockCloud/cloud.cpp#L628

chongchonghe:
I'll create a PR for that soon. Here is my idea. Please let me know what you think. 

I'll define a runtime parameter projection.vars as a list of strings. For variable that belong to an existing components, say, one of the six hydro components listed below, it will simply do a line integral. Since all of them are volume density of some quantity, their line integral is meaningful. 

`std::vector<std::string> hydroNames = {"gasDensity", "x-GasMomentum", "y-GasMomentum", "z-GasMomentum", "gasEnergy", "gasInternalEnergy"};`

For undefined components, the user has to define how to compute it in the problem generator, using this format:

```
if (var == "nH") {
  const Real rho = state(i, j, k, HydroSystem<ShockCloud>::density_index);
  return  rho / m_H;
} else if (var == xxx) {
  ....
}
```

chongchonghe:
> `return (H_mass_fraction * rho) / m_H;`

If `H_mass_fraction` is a constant, and it seems like it is, this is simply the density surface density times `H_mass_fraction/m_H`.

BenWibking:
I think it should probably always do a line integral, with an option to be either mass-weighted or volume-weighted (like the histograms).

I think allowing runtime selection of either the conserved quantities or the derived vars would be ideal. Then I can convert all of the existing projections in ShockCloud to be of derived vars.

For now, the derived vars probably still have to be defined with the `ComputeDerivedVars` function in the problem file.

[...]

PR #1366, opened 2025-10-01
DiagPlotfile for Plotfile control with selected particle types

chongchonghe:
### Description
Define `DiagPlotfile` that derives from `DiagBase` to control pltfile like this:

```
quokka.diagnostics = plt slice_z 

quokka.plt.type = PlotFile         # pltfile for the whole box
quokka.plt.file = plt              # Output file prefix (default "plt", must end in "plt")
quokka.plt.int  = 32                # Output cadence (in number of coarse steps)
quokka.plt.particles = CICRad_particles CIC_particles  # select particle types to output; default is all particles
```

When `PlotFile` is set, `plotfile_interval` will be ignored (set to -1). 

Define `DiagProjectionPlot` to control projection plots like this:
```
quokka.proj.type = ProjectionPlot   # Projection plot diagnostic
quokka.proj.file = proj_plt         # Output file prefix (must end in "plt")
quokka.proj.int  = 2                # Output cadence (in number of coarse steps)
quokka.proj.dirs = x                # Projection directions (x, y, z)
quokka.proj.particles = CIC_particles # List of particle types to include in output (default to none)
```

This PR also include include particles in projection plots. To prevent it from saving a particle data that is too large, this is skipped when a particle type has more than 10,000 stars. 

Also added metadata to projection plot so that YT can interpret it as Quokka data. 

More features to add in the future:
- Select fields to output via `quokka.plt.field_names`
- Make projection plot work the same way, and also include particles

### Related issues
Partially addresses #1062 

### Checklist


BenWibking:
This is a nice feature.

Is there any way to get rid of the old plotfile output code path and instead have `plotfile_interval` create a DiagPlotfile without the user having to modify their input files? Then we could get rid of the old plotfile code path and avoid having two independent plotfile code paths that both need to be tested.

chongchonghe:
> This is a nice feature.
> 
> Is there any way to get rid of the old plotfile output code path and instead have `plotfile_interval` create a DiagPlotfile without the user having to modify their input files? Then we could get rid of the old plotfile code path and avoid having two independent plotfile code paths that both need to be tested.

Yes, I think it is possible. That would be equivalent to the following default values:
```
quokka.diagnostics = plt
quokka.plt.type = PlotFile
quokka.plt.file = plt
```
and `quokka.slice_z.int = plotfile_interval`. 

Then, by setting `quokka.diagnostics = slicex`, we add `slicex` to `quokka.diagnostics`. `quokka.plt.int` will override `plotfile_interval`.

chongchonghe:
I'm also trying to make projection plot work like this so that we can have particles in projection plots. Does it sound good to you? @BenWibking

BenWibking:
> I'm also trying to make projection plot work like this so that we can have particles in projection plots. Does it sound good to you? @BenWibking

I guess that's fine, but I don't immediately see how it's useful. These will be very large if particles are included.

chongchonghe:
> > I'm also trying to make projection plot work like this so that we can have particles in projection plots. Does it sound good to you? @BenWibking
> 
> I guess that's fine, but I don't immediately see how it's useful. These will be very large if particles are included.

This is useful because it makes sense to display particles on top of the projection plots, like the video i showed in slack. Additional, for resolved star formation simulations I'm going to run in the next couple of years, there won't be more than 1 million stars, like not more than 100,000 stars, so the file size will be limited. 

To resolve your concern, I can skip saving a particle type if it has more than, say 10000, particles.

chongchonghe:
@BenWibking  This is ready for your review. There are duplicated code since DiagPlotfile reimplements WritePlotfile. I'll keep the original WritePlotfile for now for backward compatibility. I'll remove it later when this is thoroughly tested and there is no issue with this new API.

src/simulation.hpp
```diff
@@ -2950,7 +2948,46 @@ template <typename problem_t> void AMRSimulation<problem_t>::doDiagnostics()
...
+				// Write cell-centered data and particles
+				plotfileDiag->writePlotfile(istep[0], tNew_[0], finestLevel(), mf_cc_ptr, varnames, Geom(0, finestLevel()), istep, refRatio(),
+							    particleRegister_, do_tracers, TracerPC.get(), simulationMetadata_);
+
+				// Write face-centered data if present
+				plotfileDiag->writePlotfileFC<problem_t>(amrex::Concatenate(plotfileDiag->getDiagFileName(), istep[0], 5), finestLevel(),
+									 mf_fc_ptr, varnames_fc, Geom(0, finestLevel()), tNew_[0], istep, refRatio(),
+									 simulationMetadata_);
+			} else {
+				// Check if this is a DiagProjectionPlot - if so, call its special writeProjection method
```
BenWibking: Is there any way we can avoid this for the DiagProjectionPlot? It would be better to do this in the processDiag() function.

src/io/projection.hpp
```diff
@@ -106,7 +116,34 @@ auto ComputePlaneProjection(amrex::Vector<amrex::MultiFab> const &state_new, con
...
+	// If no particle types specified, skip particle output
+	if (particleTypes.empty()) {
+		return;
+	}
+
+	// Construct the plotfile name (same directory as field data)
+	const std::string filename = amrex::Concatenate(basename, istep, 5);
+
+	// Write particles using the filtered method
+	constexpr int max_particles_to_write = 10000;
```
BenWibking: If users have to turn on particle output for the projection file explicitly, then maybe limiting to the first 1e4 particles isn't necessary.

src/simulation.hpp
```diff
@@ -2950,7 +2948,46 @@ template <typename problem_t> void AMRSimulation<problem_t>::doDiagnostics()
 
 	for (const auto &diag : m_diagnostics) {
 		if (diag->doDiag(tNew_[0], istep[0])) {
-			diag->processDiag(istep[0], tNew_[0], GetVecOfConstPtrs(diagMFVec), m_diagVars, simulationMetadata_);
+			// Check if this is a DiagPlotfile - if so, call its special writePlotfile method
```
BenWibking: I assume it's not currently possible to do this inside processDiag(). Is that because of the face-centered vars, the particles, or both?

BenWibking (changes requested):
My only major concern is that this creates special ways to do things outside of processDiag. See below.

src/simulation.hpp
chongchonghe: it's possible if we update the base class to take more parameters, which will not be used by DiagFramePlot.

BenWibking: I think that would be a better long-term solution. What additional parameters are needed?

src/io/projection.hpp
chongchonghe: I agree. Removed.

src/io/DiagPlotfile.cpp
```diff
@@ -0,0 +1,123 @@
...
+	quokka::ScopedVisMFNOutFiles scoped_nfiles(m_nfiles);
+
+	// Write the main plotfile data using standard AMReX format
+	amrex::WriteMultiLevelPlotfile(plotfilename, finest_level + 1, a_state, a_varNames, a_geoms, a_time, a_istep, a_refRatio);
+
+	// Write metadata file (inside the plotfile directory)
+	WriteMetadataFile(plotfilename + "/metadata.yaml", simulationMetadata);
+
+	// Write tracer particles if enabled
+	if (do_tracers != 0) {
```
BenWibking: You can remove `do_tracers` and just rely on the nullptr check to see if there are tracers.

DIAGNOSTIC_API_CHANGES.md
BenWibking, line 1: Did you accidentally commit this?

chongchonghe: yes

chongchonghe:
@BenWibking I've unified the API for plotfile, projection, and regular diagnostics. However, a minor issue is there are too many arguments. We unavoidably have to pass everything to this single function, including particles, tracer particles, ProjectionData, and MultiFabs. It works fine, though. 

My idea for a future PR is to get rid of processDiag from DiagBase, and define it in DiagFramePlane. Then, we define processProj in DiagProjectionPlot.cpp, define processPlot in DiagPlotfile.cpp. These three functions don't share much, so it makes sense to define them separately. I'll also implement the automatic projection machine we discussed in #1356 .

BenWibking:
> @BenWibking I've unified the API for plotfile, projection, and regular diagnostics. However, a minor issue is there are too many arguments. We unavoidably have to pass everything to this single function, including particles, tracer particles, ProjectionData, and MultiFabs. It works fine, though.
> 
> My idea for a future PR is to get rid of processDiag from DiagBase, and define it in DiagFramePlane. Then, we define processProj in DiagProjectionPlot.cpp, define processPlot in DiagPlotfile.cpp. These three functions don't share much, so it makes sense to define them separately. I'll also implement the automatic projection machine we discussed in #1356 .

Yeah, it's not a great interface. I'm not sure what the best solution is.

That could make sense. What about DiagPDF?

BenWibking (dismissed):

src/io/DiagPlotfile.cpp
chongchonghe: Done.

chongchonghe:
> > @BenWibking I've unified the API for plotfile, projection, and regular diagnostics. However, a minor issue is there are too many arguments. We unavoidably have to pass everything to this single function, including particles, tracer particles, ProjectionData, and MultiFabs. It works fine, though.
> > My idea for a future PR is to get rid of processDiag from DiagBase, and define it in DiagFramePlane. Then, we define processProj in DiagProjectionPlot.cpp, define processPlot in DiagPlotfile.cpp. These three functions don't share much, so it makes sense to define them separately. I'll also implement the automatic projection machine we discussed in #1356 .
> 
> Yeah, it's not a great interface. I'm not sure what the best solution is.
> 
> That could make sense. What about DiagPDF?

DiagPDF takes `a_state` (vector of MultiFab), so it's similar to DiagFramePlane. 

I think trying to fix those four classes of diagnostics into one framework is just not a good idea. I propose a new design. Please let me know what you think.

What DiagBase can really help is to provide some help tools to parse runtime parameters and check for storing data at a snapshot. It also allows `DiagBase::create` to create concrete subclass. I would rewrite it to do exact that, and write the actually `processDiag` as a member function of QuokkaSimulation. Then, we don't need to pass anything from QuokkaSimulation to processDiag.

BenWibking:
> > > @BenWibking I've unified the API for plotfile, projection, and regular diagnostics. However, a minor issue is there are too many arguments. We unavoidably have to pass everything to this single function, including particles, tracer particles, ProjectionData, and MultiFabs. It works fine, though.
> > > My idea for a future PR is to get rid of processDiag from DiagBase, and define it in DiagFramePlane. Then, we define processProj in DiagProjectionPlot.cpp, define processPlot in DiagPlotfile.cpp. These three functions don't share much, so it makes sense to define them separately. I'll also implement the automatic projection machine we discussed in #1356 .
> > 
> > 
> > Yeah, it's not a great interface. I'm not sure what the best solution is.
> > That could make sense. What about DiagPDF?
> 
> DiagPDF takes `a_state` (vector of MultiFab), so it's similar to DiagFramePlane.
> 
> I think trying to fix those four classes of diagnostics into one framework is just not a good idea. I propose a new design. Please let me know what you think.
> 
> What DiagBase can really help is to provide some help tools to parse runtime parameters and check for storing data at a snapshot. It allows `DiagBase::create` to create concrete subclass. I would rewrite it to do exact that, and write the actually `processDiag` as a member function of QuokkaSimulation. Then, we don't need to pass anything from QuokkaSimulation to processDiag.

What if we have QuokkaSimulation pass itself to DiagBase and make it a friend class. Then DiagBase can just access all the member variables of QuokkaSimulation/AMRSimulation directly.

chongchonghe:
> What if we have QuokkaSimulation pass itself to DiagBase and make it a friend class. Then DiagBase can just access all the member variables of QuokkaSimulation/AMRSimulation directly.

That's also a good solution. Is there performance issue of this approach?

BenWibking:
> > What if we have QuokkaSimulation pass itself to DiagBase and make it a friend class. Then DiagBase can just access all the member variables of QuokkaSimulation/AMRSimulation directly.
> 
> That's also a good solution. Is there performance issue of this approach?

No.

Here's a complete example showing this pattern: https://chatgpt.com/s/t_68dea8ee40a88191b91adf3123be0cec

chongchonghe:
"[clang-tidy-review / build (pull_request)](https://github.com/quokka-astro/quokka/actions/runs/18215030905/job/51862443204?pr=1366)Failing after 207m" 

207m is absurd. Can we disable clang-tidy and rely on SonarCloud? @BenWibking 
From what i understand, SonarCloud covers most of the static-analysis and style check. The advantage of clang-tidy is that it includes many deep checks and auto fix for compiler based issues, but this is useful as a VSCode plugin and not very useful as a mandatory github workflow.

chongchonghe:
@BenWibking I've implemented everything we discussed. It's ready for another review and merge.

BenWibking:
> "[clang-tidy-review / build (pull_request)](https://github.com/quokka-astro/quokka/actions/runs/18215030905/job/51862443204?pr=1366)Failing after 207m"
> 
> 207m is absurd. Can we disable clang-tidy and rely on SonarCloud? @BenWibking From what i understand, SonarCloud covers most of the static-analysis and style check. The advantage of clang-tidy is that it includes many deep checks and auto fix for compiler based issues, but this is useful as a VSCode plugin and not very useful as a mandatory github workflow.

I really don't want to do that. clang-tidy has found issues in both Quokka and in AMReX itself. The AMReX clang-tidy workflow is much faster than ours, for some reason. We could replace our clang-tidy.yml file with theirs. Feel free to open a PR to do that.

src/simulation.hpp
```diff
@@ -2946,11 +2952,48 @@ template <typename problem_t> void AMRSimulation<problem_t>::doDiagnostics()
...
-			diag->processDiag(istep[0], tNew_[0], GetVecOfConstPtrs(diagMFVec), m_diagVars, simulationMetadata_);
+			// Set common diagnostic data (including simulation pointer)
+			diag->setDiagData(this, &diagMFVec_ptr, &m_diagVars, &geoms, &ref_ratio, &simulationMetadata_);
+
+			// Check if this is a DiagPlotfile - call template version
+			auto *plotfileDiag = dynamic_cast<DiagPlotfile *>(diag.get());
+			if (plotfileDiag != nullptr) {
+				plotfileDiag->processDiag<problem_t>(istep[0], tNew_[0]);
+			} else {
+				// Check if this is a DiagProjectionPlot - call template version for each direction
```
BenWibking: Can DiagProjectionPlot just call `m_sim->ComputeProjections(dir)` itself? I think that would eliminate all of the diagnostic-specific code here, and let you just call `diag->processDiag()` for all of them.

BenWibking (changes requested):
Nice! Looks a lot cleaner than the previous version. I think it can be simplified further, so that `->processDiag()` can be called the same way for all of the diag classes. See comment below.

src/simulation.hpp
chongchonghe: OK, this is a good idea. However, it won't get rid of the `dynamic_cast` and if-else clauses, because we have to cast it to specific class, and we can't do something like we did in PhysicsParticles because Diagnostics are based on Factory<DiagBase>::Create() before problem_t is known. I think this is fine since the amount of diagnostic types are not going to grow.

chongchonghe: Implemented and it worked fine.

src/simulation.hpp
```diff
@@ -2946,11 +2952,46 @@ template <typename problem_t> void AMRSimulation<problem_t>::doDiagnostics()
...
+
 	for (const auto &diag : m_diagnostics) {
 		if (diag->doDiag(tNew_[0], istep[0])) {
-			diag->processDiag(istep[0], tNew_[0], GetVecOfConstPtrs(diagMFVec), m_diagVars, simulationMetadata_);
+			// Set common diagnostic data (including simulation pointer)
+			diag->setDiagData(this, &diagMFVec_ptr, &m_diagVars, &geoms, &ref_ratio, &simulationMetadata_);
+
+			// Call the appropriate template processDiag for each diagnostic type
+			// All diagnostics now have a unified API: processDiag<problem_t>(nstep, time)
+			auto *plotfileDiag = dynamic_cast<DiagPlotfile *>(diag.get());
```
BenWibking, line 2969: This looks reasonable. However, I think this dynamic dispatch workaround is not actually necessary for DiagBase and its child classes, since (unlike the AMReX ParticleContainers) we can just mark `processDiag` as virtual in `DiagBase`.

Calling a virtual function once for the whole simulation has a _completely_ negligible performance cost. It's only an issue if you call virtual functions for every cell (i.e., inside a ParallelFor), or maybe in some very extreme cases like an $N^2$ loop between all pairs of FABs, per FAB.

chongchonghe: The problem is `processDiag` cannot see `typename problem_t`, so a template implementation  "processDiagImpl" is unavoidable. 

I asked Claude to make `processDiag` pure virtual, and the result is the code becomes more complicated. I pushed it to a different branch, take a look here: https://github.com/quokka-astro/quokka/commit/1f78fbafaa90da7d038533e5d1a7d6bb71ca7c3b 
I think the current implementation is cleaner.

chongchonghe:
@claude review this PR.

BenWibking:
@codex Review this PR.

src/io/DiagPlotfile.H
```diff
@@ -0,0 +1,151 @@
...
+	std::array<amrex::Vector<amrex::MultiFab>, AMREX_SPACEDIM> mf_fc = sim->PlotFileMF_fc(sim->nghost_fc_);
+	auto const varnames_fc = sim->GetPlotfileVarNames_fc();
+
+	auto const finest_level = sim->finestLevel();
+	auto const geoms = sim->Geom(0, finest_level);
+	auto const refRatio = sim->refRatio();
+
+#ifdef QUOKKA_USE_OPENPMD
+	// Write using OpenPMD format
+	quokka::OpenPMDOutput::WriteFile(varnames, finest_level + 1, mf_cc_ptr, geoms, m_diagfile, a_time, sim->istep[0]);
```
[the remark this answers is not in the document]
chongchonghe: Addressed.

BenWibking (approved):

PR #1368, opened 2025-10-02
Add a bash script to do pre-commit reformatting

chongchonghe:
### Description

Run `./scripts/bash/format.sh` to reformat the entire database according to pre-commit standards. Developers are advised to do this regularly after commits to prevent GitHub from automatically reformatting the code you pushed to the cloud, which can lead to conflicts.

### Related issues
None.

### Checklist


scripts/bash/format.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Check if pre-commit is installed, offer to install if missing
+if ! command -v pre-commit &> /dev/null; then
+    echo "pre-commit is not installed."
+    # Install in virtual environment or with --user flag
+    if [[ -z "$VIRTUAL_ENV" ]]; then
+        read -p "Install via 'pip install --user pre-commit'? (y/n) " -n 1 -r
+        echo
+        if [[ $REPLY =~ ^[Yy]$ ]]; then
+            pip install --user pre-commit
```
BenWibking: Can this use `uv pip install` if `uv` exists on `PATH`?

scripts/bash/format.sh
```diff
@@ -0,0 +1,50 @@
...
+        fi
+    fi
+fi
+
+# Run pre-commit
+pre-commit install
+pre-commit run --all-files
+
+# Always uninstall hooks when done
+pre-commit uninstall
```
BenWibking: What if the user had the hooks installed before running this script? I think you don't want to uninstall them in that case.

scripts/bash/format.sh
chongchonghe: OK. I can add a check for uv environment first.

scripts/bash/format.sh
chongchonghe: Well, you should not use this script if you have pre-commit previously installed. I'll add a message to tell the user 'pre-commit' is uninstalled. 

BTW, having pre-commit hooks enabled by default never worked well for me. It hangs 'git commit' in VScode.

BenWibking: Ok, that's a good point.

No, I haven't been able to get it working properly. I was hoping one day I would re-enable it and it would just work.

BenWibking (approved):

PR #1373, opened 2025-10-05
Radiation feedback from stars via fast_log table interpolation

chongchonghe:
### Description

This PR implements a comprehensive 2D table interpolation infrastructure for stellar radiation feedback, enabling stochastic stellar population particles to interpolate `(age, mass)` to multi-group stellar luminosity. The implementation provides GPU-accelerated n-dimensional (1D-4D) table interpolation with support for multiple coordinate spacing types and automatic logarithmic transformations. The data format follows the discussion in #1243 and currently supports CSV input, with future plans to extend `H5Reader` and migrate the cooling module to this infrastructure.

### Key Features

**DataTable Infrastructure (`src/util/DataTable.hpp`)**:
- Extended `DataTable` to support multiple output dimensions (template parameter `Nout`)
- Implemented three coordinate spacing types: `linear`, `log`, and `fast_log` for both input coordinates and output values
- Automatic coordinate transformation: input values are transformed to log10 space for interpolation when `log` or `fast_log` spacing is specified
- Automatic output transformation: interpolated values are converted back from log10 space when output spacing is `log` or `fast_log`

**Particle Radiation System (`src/particles/`)**:
- New `particle_radiation.hpp`: defines `LuminosityTables` class and `LuminosityUpdate` class for stellar luminosity evolution
- New `particle_update.hpp`: traits-based system for particle property updates via `ParticlePropertyUpdateTraits`
- Specialization for `StochasticStellarPop` particles that updates luminosity from tables using `(age, mass)` interpolation
- Support for multi-group radiation with automatic conversion between physical units (years, solar masses, erg/s) and code units

**Simulation Integration (`src/simulation.hpp`, `src/QuokkaSimulation.hpp`)**:
- Runtime parameters `particles.use_luminosity_table` and `particles.rad_table` for enabling and specifying luminosity table files
- Runtime parameter `particles.rad_table_output_spacing` for controlling output value storage (`linear`, `log`, or `fast_log`)
- Automatic table loading and validation during initialization
- Metadata validation: enforces input names ("age", "mass"), units ("year", "Msun"), and output units ("erg/s")

**Test Problem (`src/problems/ParticleRadiation/`)**:
- New `particle_radiation.cpp`: test problem for radiation from stellar particles with multi-group radiation transport
- Tests energy conservation by tracking total gas + radiation energy over multiple timesteps
- Validates luminosity interpolation against expected values for different spacing types
- Three test configurations:
  - `ParticleRadiation`: baseline test with linear spacing
  - `ParticleRadiationLog`: test with logarithmic coordinate spacing
  - `ParticleRadiationFastlog`: test with fast_log spacing (using FastMath approximations)

**Example Data Files** (`inputs/`):
- `lum_demo_2groups.csv`: Linear spacing example
- `lum_demo_2groups_log.csv`: Logarithmic spacing example  
- `lum_demo_2groups_fastlog.csv`: Fast logarithmic spacing example

### CSV Data Format

CSV data files for table interpolation follow this structure (see `inputs/lum_demo_2groups_fastlog.csv` for an example):

- `Ndim` (integer) — number of input dimensions
- `Nx` (comma-separated integers) — number of grid points for each input dimension  
- `Nout` (integer) — number of output dimensions
- `input_names` (comma-separated strings, Ndim entries) — names of input quantities (e.g., "age", "mass")
- `output_names` (comma-separated strings, Nout entries) — names of output quantities (e.g., "luminosity_group_0", "luminosity_group_1")
- `input_units` (comma-separated strings, Ndim entries) — units of input quantities (empty strings for dimensionless)
- `output_units` (comma-separated strings, Nout entries) — units of output quantities (empty strings for dimensionless)
- `xlo` (comma-separated reals, Ndim entries) — lower bounds of input ranges (values in physical units, not logarithms)
- `xhi` (comma-separated reals, Ndim entries) — upper bounds of input ranges (values in physical units, not logarithms)
- `spacing` (comma-separated strings, Ndim entries) — spacing type for each dimension: `linear`, `log`, or `fast_log`
- **Remaining lines**: `ydata` — output data in row-major order
  - For 2D tables: `Nx[1]` rows × `Nx[0]` columns (last dimension varies fastest)
  - For 3D tables: `(Nx[2] × Nx[1])` rows × `Nx[0]` columns
  - For 4D tables: `(Nx[3] × Nx[2] × Nx[1])` rows × `Nx[0]` columns

**Note**: When `spacing[dim]` is `log` or `fast_log`, the values in `xlo` and `xhi` should still be in physical units (not logarithms). The table infrastructure handles the log transformation internally.

### Cleanup

- Removed obsolete `RadParticle` and `RadParticle2D` test problems (superseded by unified `ParticleRadiation` test)
- Updated `Advection`, `Advection2D`, and `AdvectionSemiellipse` tests to include `nGroups` constant for consistency

### Related issues

### Checklist


chongchonghe:
@claude Write a documentation for the new features added in this PR. Use my PR description as a reference.

chongchonghe:
@codex Write a PR description for the changes to the quokka particle module and DataTable class.

chongchonghe:
@markkrumholz @BenWibking This PR is ready for your review.

markkrumholz:
Looks like there is a sonar cloud code quality fail. Can you address that first just so that I don’t have to re-review?

chongchonghe:
> Looks like there is a sonar cloud code quality fail. Can you address that first just so that I don’t have to re-review?

I resolved some of them, and most of the rest comments are irrelevant/wrong, like misreading comments as 'commented out code' and asking me to remove it. I think we should disable checks like this. 
In any case, it's ready for review. I may continue to make minor changes to let it pass style check.

inputs/lum_demo_2groups.csv
```diff
@@ -0,0 +1,18 @@
+2
```
markkrumholz, line 1: I recommend adding comments explaining what each quantity is, so for example the first few lines would read

```
2     # Number of input dimensions
3,4  # Number of elements in each dimension
2     # Number of output quantities
```

and so forth. This will make the demo easier to follow.

inputs/lum_demo_2groups_fastlog.csv
```diff
@@ -0,0 +1,18 @@
+2
```
markkrumholz, line 1: Same comment as on the previous csv demo file.

src/util/DataTable.hpp
```diff
@@ -81,25 +92,27 @@ template <int Ndim, int Nout = 1> struct DataTableGpuConst {
...
 		for (int dim = 0; dim < Ndim; ++dim) {
-			// Get table bounds for this dimension - assumes uniform grid spacing
-			amrex::Real const coord_start = coords[dim](coords[dim].begin); // First coordinate, begin = 0
-			amrex::Real const coord_end = coords[dim](coords[dim].end - 1); // Last coordinate, end = size
+			// Get table bounds - use precomputed coord_min/coord_max instead of accessing coords table
+			amrex::Real const coord_start = coord_min[dim];
+			amrex::Real const coord_end = coord_max[dim];
 
 			// Clamp coordinates to valid table bounds (extrapolation not supported)
 			amrex::Real clamped_coord = amrex::max(coord_start, amrex::min(point[dim], coord_end));
```
markkrumholz: I think there should be a choice about what happens on out-of-bounds. Clamping is one option, but I can see this causing hard-to-diagnose problems. I think there should be a template parameter that describes the out-of-bounds policy. Right now it can just have two allowed values -- clamp and fail -- with the former corresponding to this implementation, and the latter to throwing a runtime error if an out of bounds value is passed. In the future if we want we can extend this to support extrapolation.

inputs/lum_demo_2groups.csv
chongchonghe: We can rely on documentations. I'll add to the official documentation. 
Adding inline comments to CSV files is not a common feature. It will make our data files not parsable with other tools.

src/util/DataTable.hpp
```diff
@@ -116,11 +129,36 @@ template <int Ndim, int Nout = 1> struct DataTableGpuConst {
 	[[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto interpolate(const std::array<amrex::Real, Ndim> &point) const
 	    -> std::array<amrex::Real, Nout>
 	{
+		// Take log or fast_log if the spacing types are log or fast_log
+		std::array<amrex::Real, Ndim> point_{};
+		for (int dim = 0; dim < Ndim; ++dim) {
+			if (spacing_types[dim] == SpacingType::linear) {
+				point_[dim] = point[dim];
+			} else if (spacing_types[dim] == SpacingType::log) {
```
markkrumholz, line 152: Is there a reason to use log10 here, instead of just log? Similarly for fastlog, I think you should use the fastlg function, which is a log base 2. Since you'll be inverting these functions anyway, might as well save the extra floating operation required to go from a log base e or base 2 to base 10.

markkrumholz (changes requested):
There is a conceptual issue here with how fastlog is implemented. The problem is that fast log10 and fast pow10 are not exactly inverses, and your code treats them as if they are. The place this comes up is that in your code, if the user request fastlog output spacing, your code applies a fast log10 to the table data, and then inverts that by doing a fast pow10 on the interpolated output. That won't work, in the sense that it will not have the property that if someone specifies exactly one of the input points, the output will be exactly the table value for that point, because fastpow(fastlog(x)) != x.

The way to fix this is that, rather than just taking fastlog of the data when you store it, you need to define a function that numerically finds the solution y to fastpow(y) = x. That way you can store y, and when you apply fastpow to it, you will get back exactly x to machine precision. This is probably easiest to do with a simple Newton iteration, using fastlog(x) as your initial guess for y.

src/util/DataTable.hpp
chongchonghe: OK, done.

src/util/DataTable.hpp
chongchonghe: That's a good point. I've implemented it.

src/util/DataTable.hpp
chongchonghe: OK, done. I've added a oob_policy typename to DataTableGpuConst, so the user can specify the policy at declaration. However, both cooling and luminosity tables still use the default clamp.

inputs/lum_demo_2groups_fastlog.csv
chongchonghe: Also see my reply to your previous comment.

chongchonghe:
> There is a conceptual issue here with how fastlog is implemented. The problem is that fast log10 and fast pow10 are not exactly inverses, and your code treats them as if they are. The place this comes up is that in your code, if the user request fastlog output spacing, your code applies a fast log10 to the table data, and then inverts that by doing a fast pow10 on the interpolated output. That won't work, in the sense that it will not have the property that if someone specifies exactly one of the input points, the output will be exactly the table value for that point, because fastpow(fastlog(x)) != x.
> 
> The way to fix this is that, rather than just taking fastlog of the data when you store it, you need to define a function that numerically finds the solution y to fastpow(y) = x. That way you can store y, and when you apply fastpow to it, you will get back exactly x to machine precision. This is probably easiest to do with a simple Newton iteration, using fastlog(x) as your initial guess for y.

We've talked about this in Leiden but I didn't think that through carefully, so I was not thinking about implementing it. Now I think about it more, I realized this might not work as you expected. What you described will be true if we do interpolation based on the fastlog of the base coordinates, which is irregular. However, since we reply on uniform grids in the log-log space, the base grids are not the same. Linear spacing in fastlog space does not match fastlog of logrithmically-spaced points.

An obvious solution to the problem is to use irregular grids in fastlog space so that the grids are inverse_fast_pow2 of the base grids. Then, the question is, is the extra cost of binary search of coordinates justified by the extra accuracy or better performance of fastlog.

Another solution is to prepare the data in fastlog grids at the first place. This has the benefits of both performance and accuracy, but it makes data preparation harder.

chongchonghe:
For cooling and luminosity, I don't think this exact reversibility at sampled data points matters. I think it's a fundamental issue of FastLog approach, and even if you get perfect accuracy at grid points, all other points will still be inaccurate anyway.  If we demand high accuracy, perhaps the solution is to get a more accurate FastLog algorithm, or use std::log10.

BenWibking:
> For cooling and luminosity, I don't think this exact reversibility at sampled data points matters. I think it's a fundamental issue of FastLog approach, and even if you get perfect accuracy at grid points, all other points will still be inaccurate anyway. If we demand high accuracy, perhaps the solution is to get a more accurate FastLog algorithm, or use std::log10.

This is not true, and these are VERY important points:
* First: Exact reversibility really does matter, because near the mantissa rounding boundary, the naive round-trip `inverse_fastlog(fastlog(x))` will be off by 25%.
* Second: the accuracy of FastLog as an approximation to true log does not matter at all. That is whole point of FastLog. This is explained in great detail in the paper: https://ui.adsabs.harvard.edu/abs/2025ApJS..277...65H/abstract. If this point is unclear, this is really worth reading.

BenWibking:
@codex Review this PR.

markkrumholz:
> For cooling and luminosity, I don't think this exact reversibility at sampled data points matters. I think it's a fundamental issue of FastLog approach, and even if you get perfect accuracy at grid points, all other points will still be inaccurate anyway. If we demand high accuracy, perhaps the solution is to get a more accurate FastLog algorithm, or use std::log10.

@chongchonghe, you are correct that if one does fastlog interpolation for the inputs, then the input data points need to be uniformly spaced in fastlog and not in log. The code I have generated for JAFF tables in fact does that, and I can do the same for the stellar data from slug. But once you do that I don't think you will get worse accuracy from fastlog interpolation than from log interpolation. That is, there is no reason to believe that interpolating in log of the data is more accurate than interpolating in fastlog of the data, except in the special case where linear interpolation in the logarithm is exact (which happens only if the relationship is an exact powerlaw of constant slope). For JAFF, I have explicitly added error control that adaptively refines the table as needed to reach a target accuracy in the interpolation, and this procedure works equally well for log- or fastlog-spaced data.

chongchonghe:
@markkrumholz OK, I see. That's what you meant when you said "preparing tables in fastlog space" earlier. That makes sense. I'll implement that now.

chongchonghe:
@markkrumholz @BenWibking I've implemented this inverse_fastlog and the test result looks fine.

Here's how i did it: 
Given bounds [x_min, x_max], I compute [fastlog_x_min, fastlog_x_max], x_sample = np.linspace(fastlog_x_min, fastlog_x_max, N). Then, I get the fastlog_inverse of the x_sample, and compute the exact output Y on those points. Lastly, the table is written as:
x_min
x_max
Y

Indeed, I get machine accuracy in the radiation_particle test when interpolating only on the grid boundary (x_max = 100 Msun).

However, the setup for the fastlog test problem is very peculiar; I have to let the data to interpolate from fall exactly on the boundary in order to test machine accuracy interpolation (I also manually confirmed that it's not because of out-of-bound clamping). For example, if the mass grids are [1, 10, 100], we will get exact interpolation at 1 and 100, but not on 10. If I ask for interpolation at mass = 10, the error is still 1.4%, same as what I originally get where I generate data without doing any fastlog or inverse_fastlog at all. Therefore, I'm still not convinced that doing this inverse_fastlog trick when preparing the data is doing much good. It makes interpolation exact on all the sample grids, but the accuracy on other points are not necessarily improved.

I understand this will help a lot when we use non-uniformly sample grids in log space and we sample data on critical points, so I agree it's worth implementing. It doesn't add any extra compute anyway.

Everything is implemented and it's ready for your review again.

BenWibking:
For the standard base-2 fastlog, the inverse issue should arise around powers of two, when the exponent is either rounded back to the exact original value (or not).

If you don't do the Newton-Raphson inverse, you get severe error spikes at powers of 2, as shown in this figure (middle panel):

image

src/simulation.hpp
```diff
@@ -826,6 +834,59 @@ template <typename problem_t> void AMRSimulation<problem_t>::readParameters()
...
+				for (int i = 0; i < nGroups; ++i) {
+					AMREX_ALWAYS_ASSERT_WITH_MESSAGE(luminosityTables_.luminosity.output_unit(i) == "erg/s",
+									 fmt::format("Luminosity table output unit {} must be 'erg/s', got '{}'", i,
+										     luminosityTables_.luminosity.output_unit(i)));
+				}
+
+				// Set global pointer for access from particle update functions
+				quokka::g_luminosity_tables_ptr<nGroups> = &luminosityTables_;
+			}
+		}
```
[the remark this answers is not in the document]
chongchonghe: @BenWibking Do you know if this is true? It looks reasonable, because `readParameters()` is in AMRSimulation<problem_t>::initialize(), which is run in the very beginning, before `setInitialConditions()` where is `InitPhyParticles()` is located in. However, apparently the tests will fail if this is true. What's going on?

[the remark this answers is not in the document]
BenWibking: If the tests pass, then I think it's wrong.

chongchonghe:
@BenWibking So, the last figure is when you do inverse_fastlog, the same way as I described above? OK, I see. 

I just realized that you did `inverse_fast_log2` in your resampled cooling table scripts. Sorry I missed that.

src/simulation.hpp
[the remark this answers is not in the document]
BenWibking: I have seen it get confused about complex or non-obvious control flow in other contexts.

markkrumholz:
Just to close the loop on the question of accuracy: I agree that there is no guarantee that you'll get higher accuracy at arbitrary points for finite grids if your data are properly sampled to be uniform in fastlog. But an important point is that if you properly sample to be uniform in fastlog, the error is reducible by using a finer grid. That is, the interpolation error you're getting now is ~1%, but if I wanted it to be 10^-4, and we are properly sampling in a uniform grid in fastlog, then I can just use more grid points. By contrast, if the grid is uniform in log but being sampled in fastlog, the error is never going to be any better than the ~1% difference between fastlog and true log. So while it won't make a difference at ~1% error, I suspect if you repeat the test with a denser grid that gives you accuracy to 10^-4, you will see a significant difference: the proper solution with uniform spacing in fastlog will improve with finer sampling, while the approximate solution will not.

In any event, I will go re-review the PR now.

inputs/lum_demo_2groups.csv
markkrumholz: OK, fair point. I withdraw my suggestion.

chongchonghe:
> Just to close the loop on the question of accuracy: I agree that there is no guarantee that you'll get higher accuracy at arbitrary points for finite grids if your data are properly sampled to be uniform in fastlog. But an important point is that if you properly sample to be uniform in fastlog, the error is reducible by using a finer grid. That is, the interpolation error you're getting now is ~1%, but if I wanted it to be 10^-4, and we are properly sampling in a uniform grid in fastlog, then I can just use more grid points. By contrast, if the grid is uniform in log but being sampled in fastlog, the error is never going to be any better than the ~1% difference between fastlog and true log. So while it won't make a difference at ~1% error, I suspect if you repeat the test with a denser grid that gives you accuracy to 10^-4, you will see a significant difference: the proper solution with uniform spacing in fastlog will improve with finer sampling, while the approximate solution will not.
> 
> In any event, I will go re-review the PR now.

Oh, that makes sense. I got it. It's like you have a net (the fastlog space) covering a surface (the real log space), and by forcing the fastlog to perfectly match the real log at grid points, you are effectively sticking more and more joints onto the surface, and eventually you get higher and higher accuracy across the whole space. That's a beautiful idea.

markkrumholz (dismissed):
Thanks for addressing the fastlog issue. This implementation looks good.

chongchonghe:
@markkrumholz I fixed two bugs to make the GPU tests pass. Can you approve it again?

markkrumholz (approved):

PR #1383, opened 2025-10-10
Add runtime param tol and relative convergence check

chongchonghe:
### Description
1. Make residual tolenrance for Newton-Raphason iteration a runtime parameter `radiation.iteration_tolerance`
2. Add a parameter `radiation.iteration_tolerance_rel` for relative convergence check. This can be useful in some problems. Default is '-1' (disabled). 

### Related issues

### Checklist

### DIFF
--- a/inputs/RadTube.in
+++ b/inputs/RadTube.in
@@ -22,3 +22,5 @@ do_subcycle = 0
 suppress_output = 1
 
 plotfile_interval = 10000000
+
+radiation.iteration_tolerance = 1.0e-13
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -161,7 +161,7 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 
 	bool use_wavespeed_correction_ = false;
 	bool print_rad_counter_ = false;
-	amrex::Real radiation_iteration_tolerance_ = 1e-11; // tolerance for the Newton-Raphson iteration residuals
+	amrex::Real radiation_iteration_tolerance_ = 1e-11;    // tolerance for the Newton-Raphson iteration residuals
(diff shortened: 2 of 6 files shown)


BenWibking:
It might be useful to make the convergence criteria check into a functor. Then you could choose which one you want to use at compile time. That should reduce the register usage and make it possibly significantly faster. Also, it will make it easy to experiment with other criteria in the future.

chongchonghe:
> It might be useful to make the convergence criteria check into a functor. Then you could choose which one you want to use at compile time. That should reduce the register usage and make it possibly significantly faster. Also, it will make it easy to experiment with other criteria in the future.

Probably not worth doing it right now as the radiation solver is already a mess. I have an idea to rewrite the whole radiation-matter coupling module using a better structure: similar to Microphysics, I'll use a struct to hold the state variables and use a 'burner' function for the iteration solver. It's a few days of work; I'll do it in the beginning of November, when I'm done with my talks and proposals.

This PR is necessary to keep my tall box simulation running, so it's helpful for now.

BenWibking (approved):

BenWibking:
> > It might be useful to make the convergence criteria check into a functor. Then you could choose which one you want to use at compile time. That should reduce the register usage and make it possibly significantly faster. Also, it will make it easy to experiment with other criteria in the future.
> 
> Probably not worth doing it right now as the radiation solver is already a mess. I have an idea to rewrite the whole radiation-matter coupling module using a better structure: similar to Microphysics, I'll use a struct to hold the state variables and use a 'burner' function for the iteration solver. It's a few days of work; I'll do it in the beginning of November, when I'm done with my talks and proposals.
> 
> This PR is necessary to keep my tall box simulation running, so it's helpful for now.

Great, that is exactly what I was thinking as well. You could consider using the interface I wrote here: https://github.com/BenWibking/cppvode-header-only. It includes both VODE and backward Euler integrators.

PR #1190, opened 2025-07-09
[WIP] Add DerivedFieldFactory system for configurable derived fields

BenWibking:
Implements a factory pattern for derived field computation that allows users to configure derived fields through ParmParse input files. This addresses issue #1064 by providing a unified system for derived field computation.

## Key Features
- Factory pattern based on existing DiagBase architecture
- Built-in fields: temperature, vorticity, B-field divergence, sound speed, number density
- AMReX parser integration for user-defined mathematical expressions
- ParmParse configuration with field-specific parameters
- Backward compatibility with existing problem-specific derived fields

## Usage Example
```
derived_fields.fields = "temperature" "vorticity" "custom_field"
temperature.gamma = 1.4
custom_field.expression = "rho * (momx^2 + momy^2) / (2 * rho^2)"
```

## Implementation
- Follows existing factory patterns in the codebase
- Integrates with existing `ComputeDerivedVar` system
- Maintains backward compatibility
- Includes comprehensive example configuration

Closes #1064

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1390, opened 2025-10-12
use the correct temperature derivation for quokka dataset

chongchonghe:
### Description
Update python script `quick_plot` to use a more accurate temperature derivation, using 'total_energy - kinetic_energy' to derive internal energy. 

### Related issues

### Checklist

### DIFF
--- a/scripts/python/quick_plot
+++ b/scripts/python/quick_plot
@@ -115,7 +115,7 @@ assert yt.__version__ >= "4.3.0", "yt version must be >= 4.3.0"
 yt.set_log_level(40)
 
 m_u = 1.660539e-24 * unyt.g
-
+kelvin = unyt.K
 
 def plot_one(pltdir, args, suffix_idx=None):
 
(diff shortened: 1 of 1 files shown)


scripts/python/quick_plot
```diff
@@ -190,8 +190,14 @@ def plot_one(pltdir, args, suffix_idx=None):
...
         gamma = 5.0 / 3.0
-        ds.add_field(field, function=lambda field, data: data[("gas", "internal_energy_density")] * (gamma - 1.0) / (
-            data[('gas', 'density')] / mean_molecular_weight_per_H_atom * k_B), units="K", sampling_type="cell")
+        def temperature_function(field, data):
+            etot = data[("gas", "energy_density")]
+            density = data[("gas", "density")]
+            kinetic_energy = 0.5 * density * (data[("gas", "velocity_x")]**2 + data[("gas", "velocity_y")]**2 + data[("gas", "velocity_z")]**2)
+            eint = etot - kinetic_energy
+            # eint = data[("gas", "internal_energy_density")]
+            return eint * (gamma - 1.0) / (density / mean_molecular_weight_per_H_atom * k_B)
```
BenWibking: This is not correct. The mean molecular weight will depend on the cooling table. This is why I strongly advocate adding a derived 'temperature' field in the plotfile itself and not attempt to recompute this in post-processing.

chongchonghe: I agree. Let's do that.

BenWibking: My idea was to add it as an 'automatic' derived field here: https://github.com/quokka-astro/quokka/pull/1190. I haven't had time to finish this PR, though.

BenWibking:
@chongchonghe Is this PR needed anymore?

chongchonghe:
@BenWibking Yes, this PR is needed. I've updated the script to use ('boxlib', 'temperature') if it exists, and fall back to "total_energy - kinetic_energy" derivation otherwise. Before we add 'temperature' to the automatic derived field list, we need a way to compute the gas temperature, and this new method is an improvement over the previous method using 'gas_internal_energy'.

markkrumholz (approved):

PR #1398, opened 2025-10-14
trigger outputs when sentinel files are created in working dir

BenWibking:
### Description
This adds a check for the file `outputNow` in the simulation working directory after each coarse step. If it exists, a plotfile is written and then `outputNow` is deleted. An analogous feature is added for checkpoints with the `checkpointNow` file.

The output code for plotfiles and checkpoints is modified to ensure that only one checkpoint or plotfile is written per coarse step. Previously, it was possible that multiple outputs of the same type were written if both the step interval and simulation time interval were specified. Now each output trigger checks whether an output of the same type has already been triggered for the current step (for the step interval trigger, the simulation time interval trigger, and the sentinel file trigger).

This is inspired by the analogous Enzo feature.

### Related issues
N/A

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -1322,7 +1322,8 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 			if (amrex::ParallelDescriptor::IOProcessor()) {
 				std::error_code remove_ec;
 				if (!std::filesystem::remove(sentinel_path, remove_ec) && remove_ec) {
-					amrex::Print() << "[WARNING] Failed to remove sentinel '" << sentinel_path.string() << "': " << remove_ec.message() << '\n';
+					amrex::Print()
+					    << "[WARNING] Failed to remove sentinel '" << sentinel_path.string() << "': " << remove_ec.message() << '\n';
 				}
 			}
 		}
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -1294,9 +1294,12 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 		doDiagnostics();
 
 		// Writing Plot files at time intervals
-		if (last_plot_file_step != step + 1 && plotTimeInterval_ > 0 && next_plot_file_time <= cur_time) {
+		if (plotTimeInterval_ > 0 && next_plot_file_time <= cur_time) {
 			next_plot_file_time += plotTimeInterval_;
-			WritePlotFile();
+			if (last_plot_file_step != step + 1) {
+				last_plot_file_step = step + 1;
+				WritePlotFile();
+			}
 		}
 
 		// Forced plotfile output via sentinel file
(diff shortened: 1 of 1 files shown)


markkrumholz (changes requested):
I'd suggest two changes:

1. Handle the special case where someone adds an outputNow file during the same time step when a plot file would have been written anyway based on the output interval. Right now the logic in the code is such that, if you added an outputNow file during the same time step when a plot file would have been written anyway for other reasons, the result will be that two copies of the same plot file are written. This should be avoided.

2. I suggest writing out the particle file too any time you are doing an extra plot file output.

chongchonghe:
> 2. I suggest writing out the particle file too any time you are doing an extra plot file output.

`WritePlotfile()` already includes writing particle files.

BenWibking:
I will fix (1).

As ChongChong says, I think (2) is already done. They are written out in binary form inside plt00000/.

markkrumholz:
OK, agreed.

chongchonghe:
Perhaps add an option to write checkpoint as well? This is helpful when the server will shut down soon, or when it's reaching time limit soon, and we want to save a checkpoint to restart from. 

I also recommend doing this check less often, like once every 10 or 30 steps. The average coarse step update time is ~1 sec, which means the snapshot will start to write immediately after you touch a file, and immediately delete it. We probably want it to wait a few seconds.

BenWibking:
> Perhaps add an option to write checkpoint as well? This is helpful when the server will shut down soon, or when it's reaching time limit soon, and we want to save a checkpoint to restart from.

I can add that.

> 
> I also recommend doing this check less often, like once every 10 or 30 steps. The average coarse step update time is ~1 sec, which means the snapshot will start to write immediately after you touch a file, and immediately delete it. We probably want it to wait a few seconds.

I don't see the disadvantage to checking after each step, unless the filesystem is really slow/broken.

Our current applications take ~1 s per coarse step, but I can imagine that is not necessarily the case once there are a lot of radiation energy groups, dust particles, doing a Poisson solve every step, etc.

BenWibking:
Hmm, Lustre actually is kind of slow with file metadata, so maybe it is a good idea to only check every ~10 steps or so. I'll update it.

BenWibking:
Actually, ChatGPT says it's a few milliseconds per check:
```

› What is the typical overhead to check for the existence of the outputNow file on a Lustre filesystem?

• On the Lustre systems I’ve used, a best-case stat()-style check against a file already in metadata cache is basically free—a small fraction of a
  millisecond. The catch is metadata RPC latency: once Lustre has to hit the MDT, the round trip is on the order of 1–3 ms per lookup under light load,
  and it can stretch to 5 ms+ if the metadata servers are busy. So if outputNow lives on the same MDT as your run directory and the check happens once per
  timestep, you’re typically paying a few milliseconds at worst—usually lost in the noise of an AMR step.
```

I think that's negligible.

BenWibking:
@markkrumholz @chongchonghe I've updated this PR to address both of your comments. I've kept the particle writes and filesystem query cadence the same, as I don't think there are issues there (see above).

markkrumholz (approved):

chongchonghe:
> Actually, ChatGPT says it's a few milliseconds per check:
> 
> ```
> 
> › What is the typical overhead to check for the existence of the outputNow file on a Lustre filesystem?
> 
> • On the Lustre systems I’ve used, a best-case stat()-style check against a file already in metadata cache is basically free—a small fraction of a
>   millisecond. The catch is metadata RPC latency: once Lustre has to hit the MDT, the round trip is on the order of 1–3 ms per lookup under light load,
>   and it can stretch to 5 ms+ if the metadata servers are busy. So if outputNow lives on the same MDT as your run directory and the check happens once per
>   timestep, you’re typically paying a few milliseconds at worst—usually lost in the noise of an AMR step.
> ```
> 
> I think that's negligible.

A few ms would be 1% of the coarse step time for hydro+gravity runs, according to your INCITE benchmark and I don't think that's negligible. There is virtually no benefits of checking it every coarse step. Instead, we can think of when we write a outputNow file, how long do we want to wait for the code to write a snapshot. It's probably of order 10 seconds.

BenWibking:
> A few ms would be 1% of the coarse step time for hydro+gravity runs, according to your INCITE benchmark and I don't think that's negligible. There is virtually no benefits of checking it every coarse step. Instead, we can think of when we write a outputNow file, how long do we want to wait for the code to write a snapshot. It's probably of order 10 seconds.

Why do you want to wait? I want it to write it now.

PR #1399, opened 2025-10-15
add momentum sponge for low densities

BenWibking:
### Description
Implement a density-triggered momentum sponge to prevent velocities in low density regions from unreasonably limiting the timestep.  Note that unlike the old speed limit feature of Quokka (not in the current version), this damps the velocity to zero below a threshold density.

Adapted from the implementation in Castro (https://amrex-astro.github.io/Castro/docs/sponge.html).

**TODO: add test problem.**

### Related issues
N/A

### Checklist


BenWibking:
Ok, I realized that damping to zero velocity is undesired for most of our problems. We should damp only to a "speed limit" velocity, rather than to zero, since we still need to have fast outflows in some low density supernova remnants.

BenWibking:
Closing for now. We can re-open if it turns out to be needed.

PR #1420, opened 2025-10-31
Adding magnetic energy to energy calculations in hydro

lizmcole:
### Description
Total energy is thermal + kinetic + magnetic. Hydro_system.hpp needs to be updated to include the magnetic energy portion. In cases where there is no magnetic field, this will simply be 0, but in MHD cases it is nonzero. 

### Checklist


lizmcole:
To do: I need to update the problem files that use ComputePressure and ComputeSoundSpeed from the hydro module (not the EOS versions). 

The test case for this is the Fast Wave, which should have a lower L-norm error after the fix. An analytic solution to the Brio Wu shock will be added later.

BenWibking:
I've created a separate helper function `HydroSystem::ComputeMagneticEnergy` so the code to compute E_mag isn't duplicated multiple times.

I've also changed the runtime assertion to a `static_assert`.

BenWibking:
Ok, it fails because of the perturbation to the AlfvenWaveCircular test that puts it above the threshold.

It also fails due to first-capture issues with nvcc.

BenWibking:
@lizmcole Is this PR ready to be reviewed?

chongchonghe:
@lizmcole Please run the magic command `azp run` to trigger GPU tests frequently. I usually do it every time I make some changes to the code that may affect the accuracy of the test problems. This is the easier way to find GPU-related errors.

chongchonghe:
My previous PR renamed all the cpp files, causing a conflict in this PR. You need to merge in the development branch.

lizmcole:
Okay, it should be ready for review, passed all tests and the GPU tests.

markkrumholz (approved):

PR #1447, opened 2025-11-22
Add runtime parameter to toggle particle splitting on restart

chongchonghe:
Add a new runtime parameter `particles.split_particles_on_restart_refine` to control whether particles are split when restarting with refinement. It makes sense to split dark matter particles, but it doesn't make sense to split individual stellar particles.

### Changes
- Add boolean member variable `splitParticlesOnRestartRefine_` (defaults to true)
- Add runtime parameter `particles.split_particles_on_restart_refine`
- Modify particle splitting logic to check this parameter
- When false, particles are not split during restart with refinement
- Maintains backward compatibility with default behavior

Related to #1100

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -199,7 +199,7 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 	amrex::Real reltolPoisson_ = 1.0e-5;			     // default
 	amrex::Real abstolPoisson_ = 1.0e-5;			     // default (scaled by minimum RHS value)
 	int poissonSupercycleInterval_ = 1;			     // number of coarse steps between Poisson solves (default: 1)
-	bool splitParticlesOnRestartRefine_ = true; // whether to split particles when restarting with refinement
+	bool splitParticlesOnRestartRefine_ = true;		     // whether to split particles when restarting with refinement
 	amrex::Vector<amrex::MultiFab> phi;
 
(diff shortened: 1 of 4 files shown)


src/simulation.hpp
BenWibking, line 1: Can you add a test case to the CheckpointRestartRefine GitHub action for this?

BenWibking (changes requested):
Only one requested addition:

src/simulation.hpp
chongchonghe: there was a bug in the previous version of the code that I need to fix.

chongchonghe: turns out to be a bug only on macOS: #1449 . I'll ignore it for now and proceed with this PR.

chongchonghe: The current CheckpointRestartRefine script does not test particles at all. Do you mean you want to test particle spliting as well in that script? Either way, I tested this PR on my computer and it worked as expected.  I think the changes in this PR are trivial, so I don't think we need a regression test for that.

chongchonghe: I have added a new test `BinaryOrbitCICRefactor2` inside BinaryOrbitCIC to test this new feature.

chongchonghe:
@BenWibking Can you review this PR?

chongchonghe:
@BenWibking @markkrumholz This is ready for review and merge.

BenWibking (approved):

PR #1459, opened 2025-11-26
Abort simulation when max signal speed becomes unphysically large

chongchonghe:
Simulations occasionally reach unphysically high velocities that stall execution, resulting in waste of computational resources and user time. To address this issue, this PR introduces an optional mechanism that aborts the simulation when the signal speed exceeds a user-specified threshold. 

A new runtime parameter, `signal_speed_abort`, defines the maximum allowable signal speed (in code unit). If the maximum signal speed at any timestep exceeds this threshold, the code will abort with the following message. By default, `signal_speed_abort = -1`, meaning no abort will be triggered regardless of the signal speed.

```
Coarse STEP 4 at t = 104455.656 (1.103333333e-06%) starts ...
...[level 0] estimated hydro timestep: 2.987793e+10
...[level 0]    hydro timestep limited at cell [16, 15, 14] with signal speed = 2.478628e+08
...[level 0]    cell density = 3.161355e-27, |v| = 2.088636e+03, cs = 2.478607e+08
...[level 0] estimated particle timestep: 1.797693e+308
...[level 0]    max particle velocity: 0.000000e+00
...[level 0]    particle timestep limited at position [nan, nan, nan]
...[level 1]    cell density = 6.469539e-32, |v| = 1.292989e+08, cs = 1.889988e+08
[FATAL] Maximum signal speed (3.183e+08 code unit) exceeded abort threshold (2.800e+08 code unit) on level 1 at cell [34, 28, 32]
amrex::Abort::0::[FATAL] Maximum signal speed (3.183e+08 code unit) exceeded abort threshold (2.800e+08 code unit) on level 1 at cell [34, 28, 32] !!!
SIGABRT
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
  Proc: [[9078,0],0]
  Errorcode: 6
```

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -179,8 +179,8 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 	amrex::Real stopTime_ = 1.0;	      // default
 	amrex::Real cflNumber_ = 0.3;	        // default
 	amrex::Real particleCflNumber_ = 0.5;   // default
-	amrex::Real signalSpeedAbortThreshold_ = 5.0e8;	 // default 5000 km/s in cm/s; <=0 disables
-	amrex::Real particleSpeedAbortThreshold_ = 5.0e8; // default 5000 km/s in cm/s; <=0 disables
+	amrex::Real signalSpeedAbortThreshold_ = -1.0;
+	amrex::Real particleSpeedAbortThreshold_ = -1.0;
 	amrex::Real dtToleranceFactor_ = 1.1; // default
 	amrex::Real dtCutoff_ = 0.0;	      // default: no cutoff (disabled when 0)
 	amrex::Long cycleCount_ = 0;
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -177,8 +177,10 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 	amrex::Vector<amrex::Real> tOld_;     // for state_old_cc_
 	amrex::Vector<amrex::Real> dt_;	      // timestep for each level
 	amrex::Real stopTime_ = 1.0;	      // default
-	amrex::Real cflNumber_ = 0.3;	      // default
-	amrex::Real particleCflNumber_ = 0.5; // default
+	amrex::Real cflNumber_ = 0.3;	        // default
+	amrex::Real particleCflNumber_ = 0.5;   // default
+	amrex::Real signalSpeedAbortThreshold_ = 5.0e8;	 // default 5000 km/s in cm/s; <=0 disables
+	amrex::Real particleSpeedAbortThreshold_ = 5.0e8; // default 5000 km/s in cm/s; <=0 disables
 	amrex::Real dtToleranceFactor_ = 1.1; // default
 	amrex::Real dtCutoff_ = 0.0;	      // default: no cutoff (disabled when 0)
 	amrex::Long cycleCount_ = 0;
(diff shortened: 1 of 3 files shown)


markkrumholz (changes requested):
Rather than tying this to a specific unit system and expressing the ceiling in km/s, I suggest leaving the unit system unspecified so that the max speed can be expressed in whatever unit system the simulation is running in. This will be cm/s in CGS, but that way you can also use the same machinery for other unit systems. There are clear use cases where one might want to have a ceiling in, say, a dimensionless turbulence simulation.

chongchonghe:
> Rather than tying this to a specific unit system and expressing the ceiling in km/s, I suggest leaving the unit system unspecified so that the max speed can be expressed in whatever unit system the simulation is running in. This will be cm/s in CGS, but that way you can also use the same machinery for other unit systems. There are clear use cases where one might want to have a ceiling in, say, a dimensionless turbulence simulation.

`vmax = 3e8` is reasonable for cgs units, but it won't necessarily make sense for a dimensionless system. I can't think of a universal threshold for dimensionless systems, unless we use the speed of light `c`, but we don't use `c` at all for pure hydro sims. This is fine for this PR, since we just let the user define it for specific problems. If we want to enforce a threshold (see #1460 ), it only applies to CGS units. 
I'll make this change.

markkrumholz:
> > Rather than tying this to a specific unit system and expressing the ceiling in km/s, I suggest leaving the unit system unspecified so that the max speed can be expressed in whatever unit system the simulation is running in. This will be cm/s in CGS, but that way you can also use the same machinery for other unit systems. There are clear use cases where one might want to have a ceiling in, say, a dimensionless turbulence simulation.
> 
> `vmax = 3e8` is reasonable for cgs units, but it won't necessarily make sense for a dimensionless system. I can't think of a universal threshold for dimensionless systems, unless we use the speed of light `c`, but we don't use `c` at all for pure hydro sims. This is fine for this PR, since we just let the user define it for specific problems. If we want to enforce a threshold (see #1460 ), it only applies to CGS units. I'll make this change.

I don't think it is reasonable to set a single, universal threshold even in a single unit system -- a ridiculous speed in a galaxy simulation may look very different than a ridiculous speed in a protoplanetary disk simulation -- so my preferred option is that this ceiling is simply off by default, and that it be turned on in appropriate units problem-by-problem where needed.

src/simulation.hpp
```diff
@@ -983,6 +989,13 @@ template <typename problem_t> auto AMRSimulation<problem_t>::computeTimestepAtLe
 	computeMaxSignalLocal(lev);
 	const amrex::Real domain_signal_max = max_signal_speed_[lev].norminf();
 	const amrex::IntVect domain_signal_maxloc = max_signal_speed_[lev].maxIndex(0);
+	if (signalSpeedAbort_ > 0.0 && domain_signal_max > signalSpeedAbort_) {
```
markkrumholz, line 992: Looks good except that I would ditch the reference to a specific unit.

src/simulation.hpp
```diff
@@ -1002,17 +1015,23 @@ template <typename problem_t> auto AMRSimulation<problem_t>::computeTimestepAtLe
...
 		AMREX_ALWAYS_ASSERT(!std::isnan(max_particle_speed.value));
 		AMREX_ALWAYS_ASSERT(std::isfinite(max_particle_speed.value));
+		amrex::IntVect particle_cell_idx{AMREX_D_DECL(-1, -1, -1)};
+		amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &dxinv = geom[lev].InvCellSizeArray();
+		amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &prob_lo = geom[lev].ProbLoArray();
+		for (int i = 0; i < AMREX_SPACEDIM; ++i) {
+			particle_cell_idx[i] = static_cast<int>(dxinv[i] * (max_particle_speed.index[i] - prob_lo[i]));
+		}
+		if (particleSpeedAbort_ > 0.0 && max_particle_speed.value > particleSpeedAbort_) {
+			const std::string abort_msg =
```
markkrumholz: Again, I would not refer to specific unit.

src/simulation.hpp
```diff
@@ -738,6 +740,10 @@ template <typename problem_t> void AMRSimulation<problem_t>::readParameters()
 	// Default CFL number for particles == 0.5, set to whatever is in the file
 	pp.query("particle_cfl", particleCflNumber_);
 
+	// Abort when signal speed exceeds this threshold (specified in km/s, CGS only)
```
markkrumholz: I don't think this should be cgs only, and even for the changes you have already made it's not in km/s.

markkrumholz:
Just to summarize: I would prefer an implementation that is unit-agnostic, and where the value specified in the input file is assumed to be in the same unit system used for all other input parameters, whatever that may be, and where this option does not require that the simulation use any particular unit system. By default there is no maximum speed, and the maximum can be turned on in the input file, using whatever unit system is in place for that problem.

chongchonghe:
> Just to summarize: I would prefer an implementation that is unit-agnostic, and where the value specified in the input file is assumed to be in the same unit system used for all other input parameters, whatever that may be, and where this option does not require that the simulation use any particular unit system. By default there is no maximum speed, and the maximum can be turned on in the input file, using whatever unit system is in place for that problem.

That's actually what I didn't in my previous push. I forgot to update the inline comments, though. I've corrected that.

markkrumholz (approved):

chongchonghe:
There is a `Erroneous arithmetic operation` error on GPUs that I didn't catch. I'll fix it ASAP.

chongchonghe:
@claude Identify potential cause of `Erroneous arithmetic operation` introduced in this PR.

PR #1465, opened 2025-11-27
set amr.n_error_buf = 0 for static mesh refinement

chongchonghe:
### Description
I believe the correct value of `amr.n_error_buf` for static mesh refinement is 0. @BenWibking Do you agree?

I my tall-box simulations (1 kpc widt and 4 kpc tall), using `n_error_buf=3` causes the refinement tags to extend three cells beyond the intende region, which in turn creates an extra layer of `64^3` grids for refinement. For ths problem, I want a 1 kpc-tall box at level 2 and 2 kpc-tall box at level 1; however, the buffer expands the max-level box beyond these targets. Setting `n_error_buf=0` gives the correct grid geometry.

*Left: `n_error_buf=3`. Right: `n_error_buf=0`*
plt00008 n_error_buf3 plt00008 n_error_buf0

### Checklist

### DIFF
--- a/inputs/DiskGalaxy.in
+++ b/inputs/DiskGalaxy.in
@@ -17,7 +17,7 @@ amr.v              = 1       # verbosity in Amr
 amr.n_cell          = 128 128 128
 amr.blocking_factor = 16    # grid size must be divisible by this
 amr.max_grid_size   = 64    # at least 128 for GPUs
-amr.n_error_buf     = 3     # minimum 3 cell buffer around tagged cells
+amr.n_error_buf     = 0     # no cell buffer around tagged cells
 amr.grid_eff        = 0.7
 regrid_interval     = -1    # static refinement only (do not regrid)
 


BenWibking:
Yes, I agree this should be changed for the DiskGalaxy problem.

I don't want to change the default because most of the time users will expect a buffer with AMR, but I think it might be good to add a note in the AMR documentation page (https://quokka-astro.github.io/quokka/amr_grid_refinement/) about this.

BenWibking:
The figure you have in this PR would be useful to have in the docs.

BenWibking:
The AMR grids are not correct when setting this to 0 for the DiskGalaxy problem. I don't know why.

BenWibking:
We can merge this once this PR gets merged in AMReX: https://github.com/AMReX-Codes/amrex/pull/4903.

BenWibking:
I am going to close this PR. We need 3 cells to match the grid sizes we ran the scaling tests with, but it's not necessary for other problems.

PR #1469, opened 2025-11-28
Runtime calculator

chongchonghe:
### Description
I wrote a HTML+JS simulation runtime calculator for Quokka. It computes the estimated simulation runtime and resource usage given some simulation parameters. See the screenshot below. Everything should be self-explanatory. You can access it through the Quokka documentation page. 

The easiest way to experiment with it is to download [this zip file](https://github.com/user-attachments/files/23823340/quokka-runtime-calculator.zip), unzip it, and open the index.html file in your browser.

I vibe coded this using Antigravity. Thanks to Google for providing plenty of free-tier tokens. 

CleanShot 2025-11-28 at 22 23 12@2x

### Related issues

### Checklist

### DIFF
--- a/docs/markdown/html/runtime-calculator/index.html
+++ b/docs/markdown/html/runtime-calculator/index.html
@@ -1,0 +1,190 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>QUOKKA Runtime Calculator</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link href="https://fonts.googleapis.com/css2?family=Outfit:wght@300;400;500;600;700&family=JetBrains+Mono:wght@400;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="style.css">
+</head>
+<body>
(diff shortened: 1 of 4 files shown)


BenWibking (changes requested):
I tested this and it works well. Two suggestions:
* Can you make it more explicit to the user that this assumes no timestep subcycling? I know it's in a tooltip description, but that's easy to miss.
* Can you have it recalculate the virtual grid sizes on refined levels if the base level grid is modified? If I change level zero from 128^3 to 256^3, the higher levels should ideally update themselves to keep the same refinement ratios.

chongchonghe:
> I tested this and it works well. Two suggestions:
> 
> * Can you make it more explicit to the user that this assumes no timestep subcycling? I know it's in a tooltip description, but that's easy to miss.

Good point. I've updated it. New screenshot:
CleanShot 2025-11-29 at 12 03 16@2x

> * Can you have it recalculate the virtual grid sizes on refined levels if the base level grid is modified? If I change level zero from 128^3 to 256^3, the higher levels should ideally update themselves to keep the same refinement ratios.

To do that, you'll have to delete all the fine levels and re-add them. It will be recalculated. Currently, a new level added will have dimensions equal to 2 times the previous level. I don't want to force all existing fine levels to update when I change the base level, because there are scenarios where the user wants to add more cells in the base level, say, change from 1:4 to 1:8 in a tall box, and keep the fine levels unchanged.

BenWibking (approved):

PR #1481, opened 2025-12-02
Fix ComputeSoundSpeed for particles with MHD

### LINKED ISSUE #1467: particles+MHD is broken
**Describe the bug**
The particle creation machinery computed the sound speed without subtracting the magnetic energy. Previously, this was getting the wrong answer, but now it causes the simulation to crash (a good thing -- otherwise we might have missed this for a long time!):
```
 1::Assertion `cons_fc != nullptr' failed, file "/Users/benwibking/amrex_codes/quokka/src/hydro/hydro_system.hpp", line 556, Msg:
  ComputeMagneticEnergy called without face-centered fields for MHD problem !!!
  SIGABRT
```

**To Reproduce**
1. Compile the DiskGalaxy problem with MHD enabled
2. Run it
3. See error

**Additional context**
None

BenWibking:
### Description
Modifies ComputeSoundSpeed in the particle creation machinery to correctly subtract magnetic energy.

### Related issues
Fixes https://github.com/quokka-astro/quokka/issues/1467.

### Checklist


BenWibking:
The tests fail due to a ROCm bug: https://github.com/ROCm/rocm-systems/issues/2141.

BenWibking:
On moth:
```
/data/mash/cche/azp-agent-in-docker-moth/azp-agent-moth/_work/1/s/src/particles/PhysicsParticles.hpp:8:10: fatal error: 'ranges' file not found
    8 | #include <ranges>
      |          ^~~~~~~~
1 warning and 1 error generated when compiling for gfx90a.
```

This is because the default toolchain on SUSE 15.6 does not support `<ranges>`...

BenWibking:
Depends on https://github.com/quokka-astro/quokka/pull/1484.

BenWibking:
I'm not sure how to fix the SonarCloud warnings -- they are all reporting as "new code" warnings but as far as I can tell, all of the reported warnings are about code that's not changed in this PR.

BenWibking:
@chongchonghe This changes quite a bit of the particles infrastructure you wrote, so I would appreciate your close review of this.

BenWibking:
@chongchonghe I've fixed the issue found by Codex. It is ready to be reviewed.

BenWibking:
@codex review

BenWibking:
@markkrumholz @chongchonghe Can you review this PR?

src/particles/particle_accretion.hpp
```diff
@@ -381,14 +421,25 @@ void UpdateParticleMassAndMomentumInBox(const typename ContainerType::ParIterTyp
...
 		const auto &local_scale_down = scale_down.array(pti);
+		std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM> local_state_fc{};
+		bool has_face_fields = false;
+		if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
+			AMREX_ALWAYS_ASSERT_WITH_MESSAGE(state_fc != nullptr, "Face-centered state is required for MHD sink accretion");
+			local_state_fc[0] = (*state_fc)[0].array(pti);
+#if AMREX_SPACEDIM >= 2
+			local_state_fc[1] = (*state_fc)[1].array(pti);
+#endif
+			local_state_fc[2] = (*state_fc)[2].array(pti);
```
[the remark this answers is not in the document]
BenWibking: It's accessible.

chongchonghe:
> @markkrumholz @chongchonghe Can you review this PR?

Reviewing it right now

src/simulation.hpp
```diff
@@ -1811,10 +1812,21 @@ template <typename problem_t> void AMRSimulation<problem_t>::particleMeshInterac
 	// Assume all SN progenitors are at the finest level
 	const int lev = finest_level;
 
-	// Fill boundary before computing accretion rate. This is necessary because the computation of accretion rate uses 3 ghost cells, but the
+	// Fill ghost zones before computing accretion rate. This is necessary because the computation of accretion rate uses 3 ghost cells, but the
 	// boundaries are not filled at this point.
 	// TODO(cch): fill the hydro variables but not radiation variables, since radiation variables are used in accretion
-	state_new_cc_[lev].FillBoundary(geom[lev].periodicity());
+	fillBoundaryConditions(state_new_cc_[lev], state_new_cc_[lev], lev, time, quokka::centering::cc, quokka::direction::na, InterpHookNone, InterpHookNone,
```
chongchonghe, line 1818: Question: since this handles cc variables only, isn't is the same as the simpler `state_new_cc_[lev].FillBoundary(geom[lev].periodicity());` ?

src/particles/particle_accretion.hpp
```diff
@@ -253,7 +280,17 @@ void ComputeScaleDown(amrex::MultiFab &state, amrex::MultiFab &accretion_rate, a
 		// Jeans-violating cell that is not in a accretion zone emerging at the beginning of a step.
 		if (accretion_rate_cell > std::numeric_limits<double>::min()) {
 			// Compute Jeans density rho_J = J^2 * pi * cs^2 / (G * dx^2)
-			double cs_cell = HydroSystem<problem_t>::ComputeSoundSpeed(local_state_arr[bx], i, j, k);
+			std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM> cons_fc{};
+			std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM> const *cons_fc_ptr = nullptr;
+			if (has_face_fields) {
+				cons_fc[0] = state_fc_x0[bx];
+#if AMREX_SPACEDIM >= 2
```
chongchonghe: All functions in this file is inside `if AMREX_SPACEDIM == 3`, so you can omit this.

chongchonghe: Same applies to all other `if AMREX_SPACEDIM ==` in this file.

src/particles/particle_accretion.hpp
```diff
@@ -133,7 +133,8 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto compute_accretion_kernel(const dou
 template <typename ContainerType, typename problem_t>
 void ComputeAccretionRateInBox(const typename ContainerType::ParIterType &pti, const amrex::Array4<const amrex::Real> &local_state,
 			       const amrex::Array4<amrex::Real> &local_accretion_rate, const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> &plo,
-			       const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> &dx, amrex::Real /*time*/, amrex::Real dt, int /*mass_index*/)
+			       const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> &dx, std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM> state_fc,
+			       bool has_face_fields, amrex::Real /*time*/, amrex::Real dt, int /*mass_index*/)
```
chongchonghe: This `has_face_fields` parameter seems unnecessary, since it can be derived from `if constexpr (Physics_Traits<problem_t>::is_mhd_enabled)` .

src/simulation.hpp
```diff
@@ -1811,10 +1812,21 @@ template <typename problem_t> void AMRSimulation<problem_t>::particleMeshInterac
...
+			       FillPatchType::fillpatch_function);
+
+	std::array<amrex::MultiFab, AMREX_SPACEDIM> const *state_fc_ptr = nullptr;
+	if constexpr (Physics_Indices<problem_t>::nvarTotal_fc > 0) {
+		for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+			fillBoundaryConditions(state_new_fc_[lev][idim], state_new_fc_[lev][idim], lev, time, quokka::centering::fc,
+					       static_cast<quokka::direction>(idim), InterpHookNone, InterpHookNone, FillPatchType::fillpatch_function);
+		}
+
+		state_fc_ptr = &state_new_fc_[lev];
```
chongchonghe, line 1828: I think the code can be made cleaner by adding a `state_fc_ptr != nullptr` here, and then inside all the functions below, only check for state_fc_ptr != nullptr; there is no need to check if MHD is enabled, and in cases where MHD is disabled, you can pass the nullprt all the way down to ComputeSoundSpeed.

src/particles/particle_accretion.hpp
chongchonghe: Actually, see my other comments in simulation.hpp . `has_face_fields` and `if constexpr (Physics_Traits<problem_t>::is_mhd_enabled)` are unnecessary. It's safe to simply check if state_fc is nullptr. And I think you can pass state_fc as a pointer here.

chongchonghe (changes requested):
I think the code can be simplified by removing all the MHD/`has_face_fields` checks in `src/particles`, because none of the particle operations care about if MHD is enabled or not; you can simply pass a pointer `state_fc`, whether null or not, all the way through to ComputeSoundSpeed. 

Reading the whole logic, it appears to me that if MHD is enabled, `state_new_fc_` won't be empty/null and the face-centered state is always passed to the particle operations. Is this right? Then, if MHD is enabled, you check for nullptr once before all the particle operations, and then safely pass it to functions in src/particles.

chongchonghe:
@BenWibking I previously said I could take this task, but I didn't have time to do it. Sorry about that. 
If you wish, I am happy to help clean it up as I suggested and push to this PR.

src/simulation.hpp
BenWibking: No, that only fills ghost cells that are internal to the Multifab. So any ghost cells at coarse-fine AMR boundaries or at physical boundaries are left unfilled.

BenWibking:
> @BenWibking I previously said I could take this task, but I didn't have time to do it. Sorry about that. If you wish, I am happy to help clean it up as I suggested and push to this PR.

@chongchonghe Yes, that would be great. One thing to note is that the earlier usage of `mf.FillBoundary(geom.periodicity())` was not correct. That left coarse-fine boundaries and physical boundaries unfilled, since multifabs don't know about either of those.

chongchonghe:
@BenWibking  I have made the changes in #1511

src/simulation.hpp
```diff
@@ -1811,10 +1812,23 @@ template <typename problem_t> void AMRSimulation<problem_t>::particleMeshInterac
 	// Assume all SN progenitors are at the finest level
 	const int lev = finest_level;
```
BenWibking, line 1813: @chongchonghe We should not assume this in general. There may be a rare event that happens outside the finest level due to a hypervelocity star.

BenWibking:
@codex review

BenWibking:
@chongchonghe I think something in your changes caused the GPU tests to fail:
```
The following tests FAILED:
	  6 - ParticleSink (Failed)
	 12 - ParticleSinkFormation (Failed)
	 13 - SN (Failed)
	 14 - SN_long_grackle (Failed)
	 15 - SN_long_resampled (Failed)
```
https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=13334&view=logs&j=655dacad-9e02-5f72-e0cf-043ebfb7cecc&t=040ddf27-e411-58f7-ab97-dfe1d2a20f3a&l=314
and
https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=13335&view=logs&j=0410d313-00ef-550d-37c8-facbf7a1cb10&t=62d0f0fc-52cd-5dfc-847e-751a25f0b5e6

It looks like an array-out-of-bounds error.

src/simulation.hpp
BenWibking: Is this code only used for sink particles? As far as I can tell, that is the case, but your comments say 'SN'. Is the comment wrong here?

chongchonghe: It applies to both SN, sink accretion, and star formation; they currently all exist only at the highest level. For star formation, this makes sense because of Jeans refinement criteria and Truelove condition for star formation. For sink accretion, the reason I am doing this is that since the stars form at the highest level, they will accrete most of their mass at the highest level. Additionally, we should enforce stars to live at the finest level in sink simulations. 

For SN feedback, I agree it's not a very good assumption that they only occur at the finest level. However, I believe it's not important for our disk galaxy simulations because we use uniform resolution at the MW disk, and SN explosions off the disk at low-density regions will have nearly no effect at all. One of the SILCC papers demonstrates that if a significant fraction of SN feedback happens at low-density regions, the feedback failed to regulate star formation. That said, I will relax this assumption in future PRs. 

@markkrumholz Do you have any comments on this?

markkrumholz: I agree that for our intended application it is probably not a huge error to omit SNe that occur outside of the max level region, but in general I do not think we want to disable star formation feedback outside the max level region. For accreting sink particles it is a slightly trickier call -- we may want to force refinement to max level around anything that is accreting (which is what ORION does), in which case it is moot. But we definitely don't want to enforce max level around all stars, which means that SNe should not require that we be at max level.

chongchonghe: Agreed. I'll put implementing SN at all levels higher in my TODO list. 

There is another subtlety. In the current implementation, one unexpected phenomenon might happen: when a hypervelocity star gets ejected off the disk and comes back in >3 Myr, it will immediately explode as it enters the finest-level zone. Then, we might see lots of unphysical SNR at the finest-level box boundary.

BenWibking: I think that will be very rare. I actually don't worry about that at all for our upcoming runs.

chongchonghe:
> @chongchonghe I think something in your changes caused the GPU tests to fail:
> 
> ```
> The following tests FAILED:
> 	  6 - ParticleSink (Failed)
> 	 12 - ParticleSinkFormation (Failed)
> 	 13 - SN (Failed)
> 	 14 - SN_long_grackle (Failed)
> 	 15 - SN_long_resampled (Failed)
> ```
> 
> https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=13334&view=logs&j=655dacad-9e02-5f72-e0cf-043ebfb7cecc&t=040ddf27-e411-58f7-ab97-dfe1d2a20f3a&l=314 and https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=13335&view=logs&j=0410d313-00ef-550d-37c8-facbf7a1cb10&t=62d0f0fc-52cd-5dfc-847e-751a25f0b5e6
> 
> It looks like an array-out-of-bounds error.

It turns out many of the intricacy in your original implementation is necessary to get it work on GPUs. Sorry about the back and forth. 

Now I have fixed the GPU errors. I think it's ready to merge once all GPU tests pass.

chongchonghe (approved):

PR #1485, opened 2025-12-03
Fix small b-field degeneracy in HLLD solver

### LINKED ISSUE #746: add small fix to HLLD solver from Athena++
**Describe the bug**
One of the degeneracy checks in the HLLD solver should not be used. See: https://github.com/PrincetonUniversity/athena/pull/620

**Additional context**
From https://github.com/PrincetonUniversity/athena/pull/620:

> It turned out that we should not remove the double-star states even when the normal magnetic field bx is very weak. If this branch is selected when bx is actually non-zero, it may return unphysical flux. In fact, this check is not mentioned in Miyoshi & Kusano 2005 but already existed in the old Athena,
> 
> I do not think this has affected many people, as this is problematic only when the magnetic fields are very weak and the flow is not smooth. The probability that this branch is selected is very low in a realistic flow.

AstroKriel:
### Description
This PR removes the ad-hoc "weak magnetic field" degeneracy branch in the HLLD MHD Riemann solver, where the double-star states were collapsed onto the single-star states when the normal field component was very weak. The double-star states are now always computed using the Miyoshi & Kusano (2005) eqs. (59)–(63), bringing Quokka’s HLLD implementation in line with the Athena++/Parthenon implementation (see [Athena++'s PR discussion](https://github.com/PrincetonUniversity/athena/pull/620)). This should avoid unphysical fluxes in non-smooth regions when the field is weak, while leaving the rest of the solver unchanged. I have not added a dedicated correctness test for this change, but the logic now follows the published HLLD derivation and the upstream Athena++ fix.

### Related issues
Fixes https://github.com/quokka-astro/quokka/issues/746

### Checklist

### DIFF
--- a/src/hydro/HLLD.hpp
+++ b/src/hydro/HLLD.hpp
@@ -209,49 +209,57 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto HLLD(quokka::HydroState<N_scalars, N_ms
 	// MK5: eqn (48)
 	u_star_R.E = (siui_R * u_R.E - ptot_R * sR.u + ptot_star * spds[2] + bx * (sR.u * bx + (sR.v * u_R.by + sR.w * u_R.bz) - vb_star_R)) * sism_inv_R;
 
-	// if Bx is near zero, then u_i^dstar = u_i^star
-	if (0.5 * bx_sq < DELTA * ptot_star) {
-		u_dstar_L = u_star_L;
-		u_dstar_R = u_star_R;
-	} else {
-		double rho_sum_inv = 1.0 / (rho_sqrt_L + rho_sqrt_R);
-		double bx_sign = (bx > 0.0 ? 1.0 : -1.0);
-		u_dstar_L.rho = u_star_L.rho;
-		u_dstar_R.rho = u_star_R.rho;
-		u_dstar_L.mx = u_star_L.mx;
-		u_dstar_R.mx = u_star_R.mx;
-		u_dstar_L.Eint = u_star_L.Eint;
-		u_dstar_R.Eint = u_star_R.Eint;
-		for (int n = 0; n < N_scalars; ++n) {
-			u_dstar_L.scalar[n] = u_star_L.scalar[n];
-			u_dstar_R.scalar[n] = u_star_R.scalar[n];
-		}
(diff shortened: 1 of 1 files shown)


src/hydro/HLLD.hpp
```diff
@@ -209,50 +209,50 @@ AMREX_FORCE_INLINE AMREX_GPU_DEVICE auto HLLD(quokka::HydroState<N_scalars, N_ms
...
-		// MK5: eqn (62)
-		tmp = rho_sum_inv * (rho_sqrt_L * u_star_R.bz + rho_sqrt_R * u_star_L.bz +
-				     bx_sign * rho_sqrt_L * rho_sqrt_R * ((u_star_R.mz * u_star_rho_inv_R) - (u_star_L.mz * u_star_rho_inv_L)));
-		u_dstar_L.bz = tmp;
-		u_dstar_R.bz = tmp;
-		// MK5: eqn (63)
-		tmp = spds[2] * bx + (u_dstar_L.my * u_dstar_L.by + u_dstar_L.mz * u_dstar_L.bz) / u_dstar_L.rho;
-		u_dstar_L.E = u_star_L.E - rho_sqrt_L * bx_sign * (vb_star_L - tmp);
-		u_dstar_R.E = u_star_R.E + rho_sqrt_R * bx_sign * (vb_star_R - tmp);
+	// if Bx is near zero, compute double-star states using MK5 eqns (59)–(63)
```
BenWibking: This comment is not needed anymore, is it?

src/hydro/HLLD.hpp
BenWibking, line 16: `DELTA` isn't needed anymore, is it?

BenWibking (changes requested):

src/hydro/HLLD.hpp
AstroKriel: It's used in a couple more places: lines 166 and 190.

src/hydro/HLLD.hpp
AstroKriel: I've removed the "if Bx is near zero" qualifier from the comment

BenWibking: Can you push this change?

BenWibking (approved):

PR #1486, opened 2025-12-03
Fix fextract with > 1 MPI rank

BenWibking:
### Description
fextract did not work correctly with more than 1 MPI rank. This caused multiple tests to fail when run with more than 1 rank.

In src/util/fextract.cpp (MPI section around the Gatherv calls), each rank previously built its slice with local offsets starting at zero, and the IO rank simply concatenated those chunks in MPI-rank order. That means the gathered pos/data were scrambled relative to physical coordinate order and didn't match a single-rank run; no global offsets or sorting was applied.

### Related issues
N/A

### Checklist


BenWibking:
@codex review

chongchonghe:
I thought this problem was fixed in #1194 , no?

BenWibking:
> I thought this problem was fixed in #1194 , no?

Unfortunately, no. There was still an indexing issue. See PR description.

BenWibking:
> > I thought this problem was fixed in #1194 , no?
> 
> Unfortunately, no. There was still an indexing issue. See PR description.

Actually, let me double check something.

chongchonghe:
I attempted to fix the MPI issue in fextract, but Claude Code figured out that the first commit 530d8160a5976f52e8682df07fc636c1bb83301c in this PR already fixed the issue. I verified it worked just fine. See #1615 . Do you want to continue and complete this PR?

This simple trick would do what you tried to do with the test_fextract_driver.cpp, right?
```
add_test(NAME ${JOB_NAME} COMMAND mpirun -np 2 $<TARGET_FILE:${JOB_NAME}> ../inputs/${JOB_NAME}.in ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
```

I implemented this for the RadDust test in my PR. We can also merge in my PR if you want. It's ready for review.

chongchonghe:
Here is a summary from Claude Code:

# Fix MPI Issue in fextract Function

## Summary

Fixed an MPI bug in the `fextract` function that caused crashes when extracting data from domains decomposed across multiple MPI ranks. This affected tests like DustDamping and RadDust when run with grids large enough to be split across ranks.

## Root Cause

The `fextract` function used **global domain indices** instead of **local box indices** to store extracted data. When the domain was decomposed across MPI ranks, this caused array out-of-bounds errors because:

1. Each MPI rank allocates an array sized for its local data only
2. The indexing calculation used global coordinates (e.g., `i - lo0.x` where `lo0.x = 0` is the global domain lower bound)
3. When rank 1 tried to store data using its global indices, it exceeded the allocated array size

**Example**: With a 128×8×8 grid split across 2 MPI ranks:
- Rank 0 owns cells i=[0, 63], allocates 64-element array
- Rank 1 owns cells i=[64, 127], allocates 64-element array
- But rank 1 computed index as `64 - 0 = 64`, writing beyond its 64-element array (indices 0-63)

## Solution

Applied the fix from commit `530d8160` (from `BenWibking/fix-fextract-mpi` branch):

1. **Two-pass algorithm**: First pass calculates per-box offsets and total point		s
2. **Local coordinate system**: Use `bx.smallEnd(idir)` as the reference point for each box instead of the global domain lower bound
3. **Contiguous indexing**: Store data using `offset + (i - start_dir)` where `start_dir` is the local box start

This ensures each rank uses indices 0 to N-1 for its N local points, regardless of their global coordinates.

## Changes

- Modified `src/util/fextract.cpp` to use local box coordinates for indexing
- Added bounds checking assertions to catch future indexing errors

## Testing

Both tests now pass with MPI:
- ✓ **DustDamping**: Passes with 2 MPI ranks (128×8×8 grid)
- ✓ **RadDust**: Passes with 2 MPI ranks (128×8×8 grid)
- ✓ Both tests continue to work correctly without MPI (single rank)

## Notes

This is the same fix from commit `530d8160` by Ben Wibking that was on the `BenWibking/fix-fextract-mpi` branch but had not been merged into `development`. The fix resolves a critical bug that prevented multi-rank MPI runs of any test using `fextract` on decomposed domains.

chongchonghe:
@BenWibking Can you take a look at this issue? Fixing `fextract` issue is important for doing answer checks on multiple MPI ranks, so it's helpful to address it as soon as possible.

PR #1488, opened 2025-12-04
Optimize clang-tidy workflow using ccache

chongchonghe:
### Description

This PR optimizes the clang-tidy workflow to significantly reduce execution time, adopting the strategy used by WarpX.

Changes:

- Replaced Action: Switched from ZedThree/clang-tidy-review to a manual workflow which integrates ccache to cache compilation results between runs.
- Selective Linting: Added a helper script (.github/workflows/scripts/makeMakefileForClangTidy.py) that parses ccache logs to run clang-tidy only on files that miss the cache (i.e., changed files).

Impact:

First Run: Will be slow as it populates the cache.
Subsequent Runs: Will be drastically faster, checking only modified files.

### Related issues

### Checklist


chongchonghe:
This makes clang-tidy a lot faster, but with a tradeoff, as Gemini 3 Pro explains it:

The key difference is incremental checking between commits within the same Pull Request. Here is the breakdown:

1. The Old Strategy (Cumulative) The previous workflow checked "files changed in the PR relative to the development branch".

- Scenario: You open a PR that modifies 100 files.
- Run 1: The CI checks all 100 files.
- Run 2: You push a fix for a typo in 1 file.
- The Problem: The "diff with development" still contains all 100 files. The workflow sees that 100 files are "changed" in this PR, so it runs clang-tidy on all 100 of them again. This is why it takes hours every time.

2. The New Strategy (Incremental) The new workflow uses ccache to remember the state of the previous CI run for this PR.

- Scenario: You open a PR that modifies 100 files.
- Run 1: ccache is empty. It compiles 100 files (cache misses). The script sees 100 misses and runs clang-tidy on all 100 files. (This run is still slow).
- Run 2: You push a fix for 1 file.
- The Solution: GitHub Actions restores the ccache from Run 1.
  - ccache sees that 99 files haven't changed since the last run, so it retrieves them from the cache (Hits).
  - It only recompiles the 1 file you touched (Miss).
  - Our script detects only 1 cache miss and runs clang-tidy on just that single file.
- Result: The check finishes in minutes instead of hours.

Question: In Run 2, does it print out all the tidy warnings/errors that existed in previous runs?

Answer: No. This is the trade-off for the speed increase.

- The workflow identifies which files to check by parsing the ccache log for "Cache Misses" (files that had to be recompiled).
- If you have a warning in File_A but you only modify File_B in Run 2:
  - File_A is a Cache Hit (skipped by clang-tidy).
  - File_B is a Cache Miss (checked by clang-tidy).
- Consequence: You will only see warnings/errors for the files you are actively working on in that specific push. You won't see a persistent list of "all current warnings" for the whole PR unless you clear the cache.

chongchonghe:
@claude review

BenWibking:
@codex Review this PR

BenWibking:
I don't think it makes sense to only check the last commit for issues. If this were the case, you could just create a new commit to pass the check without changing any files.

chongchonghe:
> I don't think it makes sense to only check the last commit for issues. If this were the case, you could just create a new commit to pass the check without changing any files.

@BenWibking  I agree. What did you suggest last time? You mentioned some code has a much faster tidy workflow, but I don't remember it.

BenWibking:
> > I don't think it makes sense to only check the last commit for issues. If this were the case, you could just create a new commit to pass the check without changing any files.
> 
> @BenWibking I agree. What did you suggest last time? You mentioned some code has a much faster tidy workflow, but I don't remember it.

AMReX runs clang-tidy as part of its other workflows, e.g.: https://github.com/AMReX-Codes/amrex/blob/507c73f5c724394c3d6fd5ef7e88f59d2dc11faf/.github/workflows/gcc.yml#L64

BenWibking:
@chongchonghe I've tried to use both ccache and ensure the warnings are produced for all changed lines in a given PR: https://github.com/quokka-astro/quokka/pull/1494

PR #1491, opened 2025-12-04
add Gadi profile (gpuhopper)

BenWibking:
### Description
Adds an environment profile script for compiling/running on Gadi (Hopper) nodes. Tested on Gadi.

Performance measurement for `HydroBlast3D` on 1 gpuhopper node:
```
Performance figure-of-merit: 0.001111289931 μs/zone-update [899.855179 Mupdates/s]
```
(With tuning, it should be possible to get ~2x this performance.)

### Related issues
N/A

### Checklist


chongchonghe:
What's the reason to delete so many pbs scripts? @BenWibking

BenWibking:
> What's the reason to delete so many pbs scripts? @BenWibking

They were all old. I didn't think anyone was using them. They all only work for gpuvolta. Since gpuhopper is so much faster, I didn't think they were useful anymore.

chongchonghe:
OK, no one that I know is using Gadi V100, so it's fine to remove them.

markkrumholz:
I would actually prefer to leave the scripts for gpuvolta in place. We aren't currently using them much, but there is no guarantee that we might not want to run some stuff on Gadi next year, and the only GPU queue with any significant number of cards is gpuvolta.

BenWibking:
> I would actually prefer to leave the scripts for gpuvolta in place. We aren't currently using them much, but there is no guarantee that we might not want to run some stuff on Gadi next year, and the only GPU queue with any significant number of cards is gpuvolta.

@markkrumholz I've moved the old PBS scripts to `scripts/pbs/gpuvolta`. I deleted a few scripts that I think were only ever used for debugging.

markkrumholz (approved):

PR #1494, opened 2025-12-05
Rewrite clang-tidy workflow

BenWibking:
### Description
This rewrites the clang-tidy workflow so we control all the steps, and use ccache so it doesn't take as long.

### Related issues
N/A

### Checklist


chongchonghe:
@BenWibking Seems like it's not really using cache. I tested it in #1495 . The compilation with clang-tidy is as slow the second time as it was the first time.

BenWibking:
> @BenWibking Seems like it's not really using cache. I tested it in #1495 . The compilation with clang-tidy is as slow the second time as it was the first time.

Huh, weird. I will debug.

PR #1501, opened 2025-12-08
Handle negative values in fastlog table interpolation

chongchonghe:
### Description
Fixed an insecure logarithm calculation in table interpolation that triggers an error only on ROCm 7.1

### Related issues
Will fix CI failure on moth: [https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=13269&view=logs&[…]-550d-37c8-facbf7a1cb10&t=62d0f0fc-52cd-5dfc-847e-751a25f0b5e6](https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=13269&view=logs&j=0410d313-00ef-550d-37c8-facbf7a1cb10&t=62d0f0fc-52cd-5dfc-847e-751a25f0b5e6) 

### Checklist


src/util/DataTable.hpp
```diff
@@ -998,11 +1000,17 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> c
...
+				if (output_spacing == SpacingType::fast_log) {
+					return FastMath::inverse_pow2(x);
+				}
+				return std::log(x);
+			}
+			if (x < -1.0e-30 * max_val) {
+				amrex::Print() << "[Warning] DataTable: Negative value " << x << " found in log output (max value: " << max_val << ")\n";
 			}
-			return std::log(x);
+			return -10000.0;
```
BenWibking, line 1010: I'm not sure if this is a good idea. If this happens, it should really be NaN, or the code should abort.

chongchonghe: My idea is that, since the interpolation x bound is certainly above 1e-200, or log value above -460, so -10000 will be below the lower bound in log space. Then, it's up to the OOB_policy to decide.

markkrumholz: Just trying to understand the issue here -- under what circumstances is a negative value being passed to this lambda function? I can see it being necessary to floor near-zero values that are introduced due to finite numerical precision, but if that is what is going on I would prefer that we apply an explicit it floor function with the floor level specified, rather than hiding it in the definition of the log function.

BenWibking: Ok, that is true, but this situation really should not happen unless there is a bug elsewhere in the code, which this will hide.

chongchonghe: In some situations, like the SLUG table @markkrumholz generated, negative numbers are unavoidable due to interpolation or numerical integration. In those situations, it's usually fine to interpolate it to the lower bound.

markkrumholz: Whether it is a bug or not depends on the circumstances. I have certainly see situations where the "right answer" to a numerical computation might be zero to below the target absolute precision, in which case it is possible that a negative could be produced. But if that is the case, then the correct way to handle this is to explicitly sanitize to enforce positivity.

BenWibking: I would prefer to apply a floor when reading the table or somewhere else before passing it to this function.

markkrumholz: I see -- so are the negative numbers in the table data that you're loading, or positions to which you're being asked to interpolate?

BenWibking: Ok, I just saw Mark's comment above. I think we're making the same point.

markkrumholz: OK, so attempting to converge on this: if the issue is that the data table being loaded has numerical zeros that are negative and thus blow up the log function, I think the right way to handle that is to introduce an explicit clipping function that can be applied to the data at load time. This clipping function can be used to do more or less what this function is intending to do: if values in the data table are less than 1e-(something big) * the max value, floor them. The clipping value can be provided as an input when the table is loaded, with a default value of zero set meaning that no clipping is performed. This way we make the clipping explicit.

chongchonghe: yes, that is what I was trying to do

chongchonghe: I set 'something big' to 30, but I'm open to any suggestions.

markkrumholz: OK, that's fine. I think @BenWibking would just prefer that the clipping be made explicit and placed in a function for that purpose, with explicit user control over the clipping level.

markkrumholz: I'd specifically suggest that the clipping level be an optional argument to the constructor, with a default of zero which means that no clipping is performed.

chongchonghe: > I'd specifically suggest that the clipping level be an optional argument to the constructor, with a default of zero which means that no clipping is performed.

Sure, I will do that.

chongchonghe:
@BenWibking Do you think it's possible this code can produce nan values on GPUs when x is very small? I suspect that's the real cause of the error, considering that it never triggers any assertion error. 
```c++
			if (x > 0.0) {
				if (output_spacing == SpacingType::fast_log) {
					return FastMath::inverse_pow2(x);
				}
				return std::log(x);
			}
```

BenWibking:
> @BenWibking Do you think it's possible this code can produce nan values on GPUs when x is very small? I suspect that's the real cause of the error, considering that it never triggers any assertion error.
> 
> ```c++
> 			if (x > 0.0) {
> 				if (output_spacing == SpacingType::fast_log) {
> 					return FastMath::inverse_pow2(x);
> 				}
> 				return std::log(x);
> 			}
> ```

Unfortunately, "weird thing happening on GPU" is always possible. You just have to check if it *is* happening.

markkrumholz:
Is it the FastLog case or the std::log case that is producing NaN's?

chongchonghe:
> Is it the FastLog case or the std::log case that is producing NaN's?

Linear, Log, FastLog all failed with the same error.

markkrumholz:
> > Is it the FastLog case or the std::log case that is producing NaN's?
> 
> Linear, Log, FastLog all failed with the same error.

Wait, Linear is producing NaN's?! If that's the case then I don't see how it has anything to do with the log function.

chongchonghe:
> > > Is it the FastLog case or the std::log case that is producing NaN's?
> > 
> > 
> > Linear, Log, FastLog all failed with the same error.
> 
> Wait, Linear is producing NaN's?! If that's the case then I don't see how it has anything to do with the log function.

True. I just realized both FastLog and Linear failed after you asked that question. I'll have to debug different part of the code.

chongchonghe:
The issue is on particle initialization and is fixed in #1502 .

chongchonghe:
This kind of issue happened multiple times. The problem is, AMReX's `InitFromAsciiFile` allows the number of real components read from file to be smaller than the number of real components in the particle type. When this happens, Quokka will fail with FPE error or iteration failure. 

To avoid errors like this, I plan to write a thin wrapper around `InitFromAsciiFile` and throw an error if the number of real components in the file does not match the particle's trait. @BenWibking  Do you have better suggestion?

BenWibking:
> This kind of issue happened multiple times. The problem is, AMReX's `InitFromAsciiFile` allows the number of real components read from file to be smaller than the number of real components in the particle type. When this happens, Quokka will fail with FPE error or iteration failure.
> 
> To avoid errors like this, I plan to write a thin wrapper around `InitFromAsciiFile` and throw an error if the number of real components in the file does not match the particle's trait. @BenWibking Do you have better suggestion?

I think the best solution is to submit a PR that adds a check for this in AMReX.

BenWibking:
@markkrumholz @chongchonghe With this experience in mind, I have changed my mind and I think we should *not* floor the tables at runtime, since this in fact did cover up a bug that was happening elsewhere.

However, since we require the tables to have a specific format, we can provide a helper script in `scripts/python/` that reads in the table and outputs a floored table to a new HDF5 file. This makes it almost as easy to use, while also allowing it to fail if something really has gone awry elsewhere.

PR #1521, opened 2025-12-17
Revert "Bump actions/cache from 4.3.0 to 5.0.1"

BenWibking:
Reverts quokka-astro/quokka#1512.


BenWibking:
Let's see if this fixes the CI failures.

markkrumholz (approved):

markkrumholz:
It's not immediately error-ing out like it was before, so that is encouraging. I've approved, so if it passes then it should get merged automatically.

markkrumholz:
Nope, sadly still failing with the same error.

PR #1525, opened 2025-12-18
add user-defined dust stopping time

mxchen39:
### Description

This PR adds an optional Picard iteration for the implicit dust–gas drag update and refactors the stopping time interface to use a user-defined implementation.

#### 1. Picard iteration for dust–gas drag

- adds Picard iteration to ensure the dust–gas drag update remains implicit when the stopping time depends on relative speed between dust and gas.
- new runtime parameters:
  - `dust.enable_iter_stoptime`
  - `dust.print_iteration_counts`

#### 2. User-defined dust stopping time

- removes the runtime parameter for the reciprocal stopping time.
- users must define a problem-specific stopping time via `DustDrag::ComputeReciprocalStoppingTime`
  (this function should return the reciprocal of the stopping time).
- adds a helper function `DustDrag::ComputeReciprocalStoppingTimeKwok`
  implementing the physical dust stopping time from Kwok (1975), with an optional supersonic correction.

### Related issues

### Checklist


chongchonghe:
@claude review

src/problems/DustyShock/testDustyShock.cpp
```diff
@@ -0,0 +1,400 @@
...
+} // namespace
+
+auto problem_main() -> int
+{
+	const double Lx = 40.0;
+	const double CFL_number = 0.4;
+
+	// Boundary conditions: transmissive (first-order extrapolation)
+	constexpr int nvars = HydroSystem<DustyShock>::nvar_;
+	amrex::Vector<amrex::BCRec> BCs_cc(nvars);
```
chongchonghe: We favour runtime control of Boundary Conditions now. Please see how this is done in other test problems.

src/problems/DustyShock/testDustyShock.cpp
```diff
@@ -0,0 +1,400 @@
...
+	// locate shock position from numerical density gradient
+	std::vector<double> drho(nx > 1 ? nx - 1 : 0);
+	for (int i = 0; i < nx - 1; ++i) {
+		drho[i] = std::abs(rho_g_num[i + 1] - rho_g_num[i]);
+	}
+	int const shock_idx = (drho.empty()) ? 0 : static_cast<int>(std::distance(drho.begin(), std::max_element(drho.begin(), drho.end())));
+	double const shock_pos_numeric = position[shock_idx];
+
+	// Analytic solution parameters (matching previous Python logic)
+	const double v_s = 2.0;
```
chongchonghe: A style recommendation: define v_s, c_s, rho_d_left, v_d_left as global variables (constexpr double) in this test problem and use it in both the initial condition and here to ensure consistancy.

src/problems/DustDamping/testDustDamping.cpp
```diff
@@ -109,6 +109,18 @@ template <> struct Physics_Traits<DustDamping> {
 	static constexpr double radiation_constant = 1.0;
 };
 
+template <>
+AMREX_GPU_HOST_DEVICE auto DustDrag<DustDamping>::ComputeReciprocalStoppingTime(
```
chongchonghe: 1. It seems unncessary to include all three spatial components of velocity since the stopping time is probably direction-independent. Furthermore, if the returned type is one scalar per dust group, the stopping time should depend on the *magnitude* of velocity only. 
2. The second dimension of vel should be `nDustGroups` instead of `nDustGroups + 1`. Note that this is different from the radiation groups because, unlike velocities, radiation groups are defined by bin boundaries, and there is one more boundary than number of bins. 
3. You can simpliy the code here by using `nDustGroups` instead of `Physics_Traits<DustDamping>::nDustGroups`. See how `nGroups` is used in radiation tests. `nGroups` is declared as a global member of the `RadSystem` and is therefore accessible directly in its member functions.

src/dust/DustDrag.hpp
```diff
@@ -0,0 +1,392 @@
...
+					    bool enable_supersonic_correction) -> amrex::GpuArray<amrex::Real, Physics_Traits<problem_t>::nDustGroups>;
+
+	// compute dust-gas drag source terms and update conserved variables
+	static void computeDustDrag(amrex::MultiFab &consVar_cc_mf, std::array<amrex::MultiFab, AMREX_SPACEDIM> const &consVar_fc_mf, amrex::Real dt,
+				    amrex::Real dust_omega_, int enableIterDustStoptime_);
+};
+
+template <typename problem_t>
+AMREX_GPU_HOST_DEVICE auto DustDrag<problem_t>::ComputeReciprocalStoppingTime(
+    amrex::Real /*rho_g*/, amrex::GpuArray<amrex::Real, Physics_Traits<problem_t>::nDustGroups> /*rho_d*/,
```
chongchonghe: Similarly, you can make a copy of nDustGroups as a member variable of DustDrag, call it `nDustGroups_`, and use `nDustGroups_` instead of the long `Physics_Traits<problem_t>::nDustGroups` here. See the usage of `nGroups_` in radiation_system.hpp

Also see my comments in another file about optimizing the arguments of this function.

src/dust/DustDrag.hpp
```diff
@@ -0,0 +1,392 @@
...
+    -> amrex::GpuArray<amrex::Real, Physics_Traits<problem_t>::nDustGroups>
+{
+	constexpr int N = Physics_Traits<problem_t>::nDustGroups;
+	amrex::GpuArray<amrex::Real, N> alpha;
+	alpha.fill(0.0);
+	return alpha;
+}
+
+template <typename problem_t>
+AMREX_GPU_HOST_DEVICE auto DustDrag<problem_t>::ComputeReciprocalStoppingTimeHelper(
```
chongchonghe: my previous comments about vel also applies to here.

src/problems/DustDampingIter/testDustDampingIter.cpp
```diff
@@ -0,0 +1,426 @@
...
+	for (size_t i = 0; i < t.size(); ++i) {
+		outfile << t[i] << " " << v_gas[i] << " " << v_dust1[i] << " " << v_dust2[i] << " " << E_gas[i] << "\n";
+	}
+
+	outfile.close();
+	amrex::Print() << "Reference solution saved to " << filename << "\n";
+}
+
+// read reference solution from file
+auto load_reference_solution(const std::string &filename, std::vector<double> &t_ref, std::vector<double> &v_gas_ref, std::vector<double> &v_dust1_ref,
```
chongchonghe: what's the reason to write the reference solution to the disk and load it? Why not just computing it and compare it with numerical solutions?

src/problems/DustDampingIterNoCorrection/testDustDampingIterNoCorrection.cpp
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	// check if the time points match (the test time points should be a subset of the reference solution time points)
+	if (t_ref.size() < t.size()) {
+		amrex::Print() << "Error: Reference solution has fewer points (" << t_ref.size() << ") than test solution (" << t.size() << ")\n";
+		return 1;
+	}
+
+	// calculate the sampling step size
+	auto step = static_cast<size_t>(t_ref.size() / t.size());
```
chongchonghe: The use of `size_t` type here is strange. Should use `int`.

src/problems/DustDampingIter/testDustDampingIter.cpp
mxchen39: because the input files may be different between reference solution and numerical solution, I don't know how to handle this situation, so I write the reference solution to the disk.

src/problems/DustDamping/testDustDamping.cpp
mxchen39: to 1., do you mean that I need to create a new variable to store magnitude of velocity in `DustDrag::computeDustDrag` and pass the new created variable to `DustDrag::ComputeReciprocalStoppingTime`?
to 2., `nDustGroups` is for dust and `1` is for gas, so the sum is `nDustGroups + 1`.

chongchonghe (changes requested):
Mostly looks good. A major request for change is about the arguments of `ComputeReciprocalStoppingTime`. Please see my comments.

src/problems/DustDamping/testDustDamping.cpp
chongchonghe: > to 1., do you mean that I need to create a new variable to store magnitude of velocity in `DustDrag::computeDustDrag` and pass the new created variable to `DustDrag::ComputeReciprocalStoppingTime`? 

Just compute a velocity magnitude as `GpuArray<Real, nDustGroups_>` before calling `ComputeReciprocalStoppingTime`, and pass it to `ComputeReciprocalStoppingTime`.

> to 2., `nDustGroups` is for dust and `1` is for gas, so the sum is `nDustGroups + 1`.

OK, I see.

src/dust/DustDrag.hpp
chongchonghe: Can you add a documentation to this function explaining what model you are using? I would also rename it to a more descriptive name.

src/problems/DustDamping/testDustDamping.cpp
chongchonghe: I recommend either setting both rho and vel_mag to size `nDustGroups_+1` array, or both separately as `rho_g, rho_d` and `vel_mag_g, vel_mag_d`. This is cleaner than the current case mixing sizes `nDustGroups_` and `nDustGroups_+1`.

src/problems/DustDampingIteration/testDustDampingIteration.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+	// step 1: run reference solution (dt = 0.00001)
+	auto ref_data = run_simulation(dt_ref, enableIterDustStoptime_ref);
+
+	// step 2: run numerical solution (dt = 0.005)
+	auto test_data = run_simulation(dt_test, enableIterDustStoptime_test);
+
+	// step 3: calculate errors
+	// calculate sampling step size
+	const auto step = static_cast<size_t>(std::round(dt_test / dt_ref));
```
chongchonghe: This really should be `int` instead of `size_t`.

chongchonghe:
@mxchen39 There are two remaining things:
1. Add inline comments to `ComputeReciprocalStoppingTimeHelper` to explain what it does, and perhaps rename it to `ComputereciprocalStoppingTimeXXX`, where `XXX` is the author/model name
2. I recommend changing the arguments of `ComputeReciprocalStoppingTime` to either `(rho_g, rho_d, vel_mag_g, vel_mag_d, cs)` or `(rho, vel_mag, cs)` . 

There are more unresolved conversations. Can you go through all of them and mark them as resolved once addressed?

chongchonghe:
@mxchen39 There is also a clang-tidy warning that needs to be fixed. The easiest way to find out the exact warning is to click on "clang-tidy-review / build (pull_request)", "Summary", "Artifacts", and click on the download button to download "clang-tidy-review-output.json"

chongchonghe:
@mxchen39 A few more suggestions:
1. You can change the grids structure of the DustDampling tests to `8 8 8`, then you can use `CFL=0.3` and the time step will be about the size of your fixed time step (0.005). Then, you can compare it to small-step no-iteration benchmark results, and this is a more meaning comparison.

chongchonghe:
@mxchen39 You can probably simplify the tests further by combining the Correction tests and the Iterations tests, i.e. running three tests in a single problem file. Can you try that?

chongchonghe:
@claude review

src/dust/DustDrag.hpp
```diff
@@ -0,0 +1,431 @@
...
+	amrex::GpuArray<amrex::Real, nDustGroups_> alpha;
+
+	for (int g = 0; g < nDustGroups_; ++g) {
+		if (rho_g <= 0.0 || rho_d[g] <= 0.0 || cs <= 0.0) {
+			alpha[g] = 0.0;
+			continue;
+		}
+
+		// compute relative velocity squared between dust group g and gas
+		amrex::Real v_rel_sq = (vel_mag_d[g] - vel_mag_g) * (vel_mag_d[g] - vel_mag_g);
```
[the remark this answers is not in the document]
chongchonghe: @mxchen39 gemini is right. Your original implementation where you pass v_g and v_d as vectors should be correct, but after accepting my suggestion of passing v_mag, this becomes wrong. Perhaps you can pass |v_d - v_g| to the function, assuming a dust drag model depends on the magnitude of the relative velocity.

chongchonghe (approved):

PR #1528, opened 2025-12-18
Add gas halo to DiskGalaxy problem

aditivijayan:
### Description
The edit reads in initial rho, momr, eint, etot from tables. These tables were generated by the cooling flow code of Stern+19 and modified to include Agora vc and Grackle cooling. The modifications to the original code can be accessed here - https://bitbucket.org/aditivijayan/cooling_flow_stern-19/src/master/
The code generates rho, v, T profiles for Mdot=0.99 Msun/yr and rsonic of 0.38 kpc.

### Related issues
Depends on:
* https://github.com/quokka-astro/quokka/pull/1595
* https://github.com/quokka-astro/quokka/pull/1596
* https://github.com/quokka-astro/quokka/pull/1607

### Checklist


BenWibking:
The cooling rates in the Python code are not quite right. It's necessary to add the heating rates to the cooling rates to get a net cooling rate, and also add PE heating and Compton cooling by hand.

You can copy the code from here to do that: https://github.com/quokka-astro/quokka/blob/5fbfea866c972eb7ddf1018ad35e6f98ae6af3b4/extern/cooling/grackle_tables.py#L88

aditivijayan:
> The cooling rates in the Python code are not quite right. It's necessary to add the heating rates to the cooling rates to get a net cooling rate, and also add PE heating and Compton cooling by hand.
> 
> You can copy the code from here to do that:
> 
> https://github.com/quokka-astro/quokka/blob/5fbfea866c972eb7ddf1018ad35e6f98ae6af3b4/extern/cooling/grackle_tables.py#L88

The net cooling rate is negative for some combination of T and nH. How do you propose I estimate dlnLambda_dlnrho and dlnLambda_dlnT? Having negative values of lambda just throws up nan when estimating the gradients.

markkrumholz:
Physically the net cooling rate can indeed be negative for some combinations of n and T, because those correspond to places where the gas is colder than the equilibrium temperature given the radiation field to which it is being subjected. But does the solution actually pass through any of those points? I would guess not. If you need to supply a numerical value there anyway just to avoid numerical or computer issues, I would suggest that you instead just supply d ln(|Lambda|) / d ln T instead of d ln Lambda / d ln T, i.e., just take the absolute value of the cooling rate.

BenWibking:
You can't take the logarithm when it's negative, but you should be able to rewrite it with the chain rule (by physicist math):
$\frac{d ln \Lambda}{d ln T} =\frac{T}{\Lambda} \frac{d \Lambda}{d T}$

Mathematically, this should be the analytic continuation of the function when $\Lambda < 0$.

aditivijayan:
I've pushed the changes to bitbucket. Now profiles include PE and Compton. This is what they look like. 

Screenshot 2025-12-19 at 1 04 14 pm

aditivijayan:
> I've pushed the changes to bitbucket. Now profiles include PE and Compton. This is what they look like.
> 
> Screenshot 2025-12-19 at 1 04 14 pm

Screenshot 2025-12-19 at 2 13 22 pm

Profile regenerated for Z=Zsun to match with the metallicity of the cooling table.

BenWibking:
I don't think it makes sense to continue the solution inward of the sonic point. Can you remove that part of the solution?

BenWibking:
It looks like it's going to hit another sonic point just outside of R200. If you want me to try to debug this, I have some time tomorrow.

markkrumholz:
Stern et al. do in fact continue their solution inward of the sonic point, so I don't think it is necessarily a problem to do so -- and I think we do want to do so, because otherwise we are going to have an artificial density jump in our initial condition. There's no mass in this gas in any event, so I see no harm in continuing the solution inward. I think the second sonic point is just an artifact of the cooling function we're using, which has a temperature that starts to drop after some density rather than rising continuously as Stern et al. assume. Again, I don't think this is necessarily a problem or an error.

aditivijayan:
> I don't think it makes sense to continue the solution inward of the sonic point. Can you remove that part of the solution?

And just copy the solution from r=rsonic? Sure.

aditivijayan:
> Stern et al. do in fact continue their solution inward of the sonic point, so I don't think it is necessarily a problem to do so -- and I think we do want to do so, because otherwise we are going to have an artificial density jump in our initial condition. There's no mass in this gas in any event, so I see no harm in continuing the solution inward. I think the second sonic point is just an artifact of the cooling function we're using, which has a temperature that starts to drop after some density rather than rising continuously as Stern et al. assume. Again, I don't think this is necessarily a problem or an error.

This is correct.

BenWibking:
> > I don't think it makes sense to continue the solution inward of the sonic point. Can you remove that part of the solution?
> 
> And just copy the solution from r=rsonic? Sure.

This would be my preferred solution. It remains continuous in this case (although not differentiable).

I think it's not any less meaningful than the continued solution, since the smoothly-continued solution won't be realized in practice (there will be a shock, it will become time-dependent, etc.).

BenWibking:
> Stern et al. do in fact continue their solution inward of the sonic point, so I don't think it is necessarily a problem to do so -- and I think we do want to do so, because otherwise we are going to have an artificial density jump in our initial condition. There's no mass in this gas in any event, so I see no harm in continuing the solution inward. I think the second sonic point is just an artifact of the cooling function we're using, which has a temperature that starts to drop after some density rather than rising continuously as Stern et al. assume. Again, I don't think this is necessarily a problem or an error.

My mistake. That makes sense.

In terms of what physical solution to choose, I think it might be preferable to choose the sonic point location (or mass accretion rate) to maximize observational agreement. Essentially all the cooling flow solutions give temperatures ~1e6 K, so there's nothing to calibrate there. For density, the observational constraints are weak, but I think one that is as good as any other is the Miller & Bregman (https://ui.adsabs.harvard.edu/abs/2013ApJ...770..118M/abstract) electron density profile, which goes at $n_e^{-1.7}$ (see their paper for the normalization). This is consistent with all the more recent measurements I'm aware of within the (very large) observational uncertainties.

aditivijayan:
> > Stern et al. do in fact continue their solution inward of the sonic point, so I don't think it is necessarily a problem to do so -- and I think we do want to do so, because otherwise we are going to have an artificial density jump in our initial condition. There's no mass in this gas in any event, so I see no harm in continuing the solution inward. I think the second sonic point is just an artifact of the cooling function we're using, which has a temperature that starts to drop after some density rather than rising continuously as Stern et al. assume. Again, I don't think this is necessarily a problem or an error.
> 
> My mistake. That makes sense.
> 
> In terms of what physical solution to choose, I think it might be preferable to choose the sonic point location (or mass accretion rate) to maximize observational agreement. Essentially all the cooling flow solutions give temperatures ~1e6 K, so there's nothing to calibrate there. For density, the observational constraints are weak, but I think one that is as good as any other is the Miller & Bregman (https://ui.adsabs.harvard.edu/abs/2013ApJ...770..118M/abstract) electron density profile, which goes at n e − 1.7 (see their paper for the normalization). This is consistent with all the more recent measurements I'm aware of within the (very large) observational uncertainties.

The normalisation I am getting for n_e (0.024) is within the error range (0.048(+0.085/-0.037). Their accretion rate is Msun/yr for large r.

BenWibking:
I think this looks fine, but I want to run a test on Frontier just to make sure there aren't any hidden GPU surprises.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -218,25 +273,120 @@ template <> void QuokkaSimulation<AgoraGalaxy>::setInitialConditionsOnGrid(quokk
...
 		// integrate density profile over cell volume
 		// TODO(bwibking): use adaptive quadrature with relative tolerance
 		const double cell_vol = dx[0] * dx[1] * dx[2];
 		const double rho_disk = quad_3d(rho_exact, x0, x1, y0, y1, z0, z1) / cell_vol;
+		const double rho_halo = quad_3d(rhoHalo_exact, x0, x1, y0, y1, z0, z1) / cell_vol;
+		const double momx_halo = quad_3d(momx_exact, x0, x1, y0, y1, z0, z1) / cell_vol;
+		const double momy_halo = quad_3d(momy_exact, x0, x1, y0, y1, z0, z1) / cell_vol;
+		const double momz_halo = quad_3d(momz_exact, x0, x1, y0, y1, z0, z1) / cell_vol;
+		const double eint_halo = quad_3d(eintHalo_exact, x0, x1, y0, y1, z0, z1) / cell_vol;
+		const double etot_halo = quad_3d(etotHalo_exact, x0, x1, y0, y1, z0, z1) / cell_vol;
```
BenWibking: I think it would be preferable to interpolate the primitive variables (rho, v_x, v_y, v_z, e_int) and then compute the conserved variables afterward. Otherwise, there is an inconsistency of order $(\Delta x)^2$ in the definition of the primitive variables after the code converts them. This means that, e.g., the temperature won't be exactly what we would have interpolated for the average temperature in a given cell.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -218,25 +273,120 @@ template <> void QuokkaSimulation<AgoraGalaxy>::setInitialConditionsOnGrid(quokk
...
 		double T = NAN;
 
 		// IMPORTANT: transition between disk and halo at the P_halo == P_disk surface
-		if (rho_halo * T_halo > rho_disk * T_disk) {
-			rho = rho_halo;
-			T = T_halo;
-			vx = 0; // velocity is zero in the halo
-			vy = 0;
-		} else { // we are in the disk
+		if (rho_ambient * T_halo < rho_disk * T_disk) { // we are in the disk
```
BenWibking: I think we shouldn't use the `rho_ambient` or `T_halo` that were read in from the input file. That isn't meaningful anymore. We should use the actual interpolated density and temperature from the profile instead. This means that the disk-halo boundary will change compared to before, but I think that's what we want -- that way the pressure is continuous at the boundary between the two.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -263,12 +413,12 @@ template <> void QuokkaSimulation<AgoraGalaxy>::setInitialConditionsOnGrid(quokk
...
-		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::x1Momentum_index) = rho * vx;
-		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::x2Momentum_index) = rho * vy;
-		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::x3Momentum_index) = rho * vz;
-		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::energy_index) = Eint + 0.5 * rho * vsq;
-		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::internalEnergy_index) = Eint;
+		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::density_index) = rho + rho_halo;
+		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::x1Momentum_index) = rho * vx + momx_halo;
+		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::x2Momentum_index) = rho * vy + momy_halo;
+		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::x3Momentum_index) = rho * vz + momz_halo;
+		state_cc(i, j, k, HydroSystem<AgoraGalaxy>::energy_index) = Eint + 0.5 * rho * vsq + etot_halo;
```
BenWibking: The kinetic energies from the two components cannot be summed like this. You have to recompute the total kinetic energy.

BenWibking (changes requested):
A few minor comments below:

src/problems/DiskGalaxy/testDiskGalaxy.cpp
aditivijayan: @markkrumholz had recommended adding up conservative variables during our chat. Happy to make changes if required.

BenWibking: I agree that adding the conserved variables make sense. What I propose is instead to do the interpolation in the primitive variables. Then you can convert to the conserved variables, and then add the disk + halo in the conserved variables.

aditivijayan: A possible issue could be mu. It is taken to be 0.61 in the Stern code.

BenWibking: The motivation for doing this is because interpolating in the primitive vars gives a significantly better answer when a 1D model is close to hydrostatic equilibrium. This is what Castro does.

BenWibking: That could make a small difference. However, the inconsistency in $\mu$ is there either way -- it will just manifest slightly differently in how it will compensate after initialization.

If you want to eliminate it, then you could modify the Stern code to compute the mean molecular weight as a function of density and temperature. But I think that's going to be so small of a correction that it won't really matter.

aditivijayan:
I see that the magnetic field has been initialised. I've added emag to etot.

BenWibking:
> I see that the magnetic field has been initialised. I've added emag to etot.

Great, thanks! I am still debugging some GPU issues with my variable density floor PR. Once I get that fixed, I will test this PR and approve.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -236,25 +269,110 @@ template <> void QuokkaSimulation<AgoraGalaxy>::setInitialConditionsOnGrid(quokk
...
+			double temp_H = NAN;
+			if (R > R_table_min && R < R_table_max) {
+				temp_H = interpolate_value(R, R_table, temp_table, len_table);
+			} else if (R <= R_table_min) {
+				temp_H = temp_inner;
+			} else {
+				temp_H = temp_outer;
+			}
+			return temp_H;
+		};
```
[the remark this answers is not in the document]
BenWibking: This is fine as-is.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -93,31 +107,49 @@ template <> void QuokkaSimulation<AgoraGalaxy>::preCalculateInitialConditions()
...
+	userData_.rho_inner = rho_h[std::distance(radius_h.begin(), min_result)];
+	userData_.velr_inner = velr_h[std::distance(radius_h.begin(), min_result)];
+	userData_.temp_inner = temp_h[std::distance(radius_h.begin(), min_result)];
 
 	auto max_result = std::max_element(radius_h.begin(), radius_h.end());
 	userData_.r_outer = (*max_result) * length_unit;
 	userData_.vcirc_outer = vcirc_h[std::distance(radius_h.begin(), max_result)] * vel_unit;
+	userData_.rho_outer = rho_h[std::distance(radius_h.begin(), max_result)];
+	userData_.velr_outer = velr_h[std::distance(radius_h.begin(), max_result)];
+	userData_.temp_outer = temp_h[std::distance(radius_h.begin(), max_result)];
```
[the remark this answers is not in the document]
BenWibking: This is fine as-is.

BenWibking:
I think there's an issue with the pressure criterion I proposed for switching between disk and halo. When the halo profile is continued to small radii, the pressure switch is not monotonic, so it switches back to the halo profile below about 2 kpc. I will think about a better switch to fix this.

radial_profiles_4panel

BenWibking:
Ok, I think it works now:
radial_profiles_4panel

Working on cleaning up the code now.

BenWibking:
I think there's still an issue with some of the derived vars. The values are not physical in the slices.

Update: The derived vars are fine in the plotfiles, but not in the slices. So the is in the slicing machinery, not in this PR.

chongchonghe:
>  I think there's still an issue with some of the derived vars. The values are not physical in the slices.
> 
> Update: The derived vars are fine in the plotfiles, but not in the slices. So the is in the slicing machinery, not in this PR.

I've seen this. The slice uses linear(?) interpolation that sometimes produces unphysical values. I would recommend changing the slice machinery to piecewise constant.

BenWibking:
> > I think there's still an issue with some of the derived vars. The values are not physical in the slices.
> > Update: The derived vars are fine in the plotfiles, but not in the slices. So the is in the slicing machinery, not in this PR.
> 
> I've seen this. The slice uses linear(?) interpolation that sometimes produces unphysical values. I would recommend changing the slice machinery to piecewise constant.

@chongchonghe I've fixed it. The problem is that the interpolation assumed that the ghost cells were filled, which was not true for the derived vars. I've filled the derived var ghosts for internal boundaries, which is sufficient to fix the slicing interpolation (it always uses internal ghosts).

aditivijayan:
This PR contains a lot of changes I haven't made. Please do let me know if there are modifications to be made to the code changes I made.

BenWibking:
> This PR contains a lot of changes I haven't made. Please do let me know if there are modifications to be made to the code changes I made.

I've been trying to separate out the changes I made into separate PRs:
* https://github.com/quokka-astro/quokka/pull/1597
* https://github.com/quokka-astro/quokka/pull/1595

I don't think anything needs to be changed in the parts of the code you added.

BenWibking:
Once those are merged, I think this PR is ready to merge.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -62,62 +67,107 @@ template <> struct Particle_Traits<AgoraGalaxy> {
...
+		userData_.temp_halo[i] = temp_h[i];
 	}
 
 	// save min/max radii
 	auto min_result = std::min_element(radius_h.begin(), radius_h.end());
 	userData_.r_inner = (*min_result) * length_unit;
 	userData_.vcirc_inner = vcirc_h[std::distance(radius_h.begin(), min_result)] * vel_unit;
+	userData_.rho_inner = rho_h[std::distance(radius_h.begin(), min_result)];
+	userData_.velr_inner = velr_h[std::distance(radius_h.begin(), min_result)];
+	userData_.temp_inner = temp_h[std::distance(radius_h.begin(), min_result)];
```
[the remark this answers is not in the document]
BenWibking: This is not important.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -62,62 +67,107 @@ template <> struct Particle_Traits<AgoraGalaxy> {
...
 	for (std::string line; std::getline(fstream, line);) {
 		std::istringstream iss(line);
 		std::vector<double> values;
 
 		for (double value = NAN; iss >> value;) {
 			values.push_back(value);
 		}
+		if (values.empty()) {
+			continue;
+		}
```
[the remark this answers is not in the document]
BenWibking: I will replace this with `CSVReader` and `quokka::DataTable`.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -432,6 +600,185 @@ template <> void QuokkaSimulation<AgoraGalaxy>::ComputeDerivedVar(int lev, std::
...
+template <> auto QuokkaSimulation<AgoraGalaxy>::ComputeStatistics() -> std::map<std::string, amrex::Real>
+{
+	std::map<std::string, amrex::Real> stats;
+
+	const amrex::Real time_now = tNew_[0];
+	const amrex::Real time_window = 1.0e6 * seconds_per_year;
+	const amrex::Real sfr_g_per_s = particleRegister_.computeSfrAveragedOverTime(time_now, time_window);
+	stats["sfr_1myr"] = (sfr_g_per_s / C::M_solar) * seconds_per_year;
+
+	amrex::ParmParse const pp("agora_galaxy");
```
aditivijayan: @BenWibking did you want to change this to "galaxy" for parsing?

BenWibking: Yes, thanks for the reminder. Will do that :)

BenWibking: Done.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -55,76 +66,106 @@ template <> struct Physics_Traits<AgoraGalaxy> {
...
-		for (double value = NAN; iss >> value;) {
-			values.push_back(value);
-		}
-		Real const R_val = values.at(0);
-		Real const vcirc_val = values.at(1);
-
-		radius_h.push_back(R_val);
-		vcirc_h.push_back(vcirc_val);
-	}
+	auto halo_table = quokka::DataTable<1, 4, quokka::OutOfBounds::clamp>::CSVReader(filename, quokka::SpacingType::linear);
```
BenWibking, line 106: For consistency throughout the code, I changed this to use quokka::DataTable. I need to test this to make sure it works as expected.

BenWibking: Ok, it seems to be a lot slower for some reason. Need to debug.

BenWibking: This is not a show-stopper, since it's only used in the ICs, but it's weird.

BenWibking:
@markkrumholz Can you review this? I've made some changes, so it would be good to have a third person look at it all.

markkrumholz (changes requested):
This generally looks fine except that that changes in the infrastructure to compute the SFR in PhysicsParticles seem to have been duplicated in this PR. Once they are removed, I'm happy to approve.

BenWibking:
> This generally looks fine except that that changes in the infrastructure to compute the SFR in PhysicsParticles seem to have been duplicated in this PR. Once they are removed, I'm happy to approve.

@markkrumholz I've removed those changes.

BenWibking (approved):

BenWibking:
@markkrumholz This PR is ready to be approved.

markkrumholz (approved):

PR #1529, opened 2025-12-18
Let SonarCloud use c++20 standard

chongchonghe:
### Description
Use `gcc-11` and `sonar.cfamily.reportingCppStandardOverride=c++20` in sonarcloud.yml

### Checklist

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -22,7 +22,7 @@ jobs:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of sonarcloud analysis
           
       - name: Install dependencies
-        run: sudo apt-get update && sudo apt-get install python3-dev python3-numpy python3-matplotlib libopenmpi-dev libhdf5-mpi-dev
+        run: sudo apt-get update && sudo apt-get install gcc-11 g++-11 python3-dev python3-numpy python3-matplotlib libopenmpi-dev libhdf5-mpi-dev
         
       - name: Install sonar-scanner and build-wrapper
         uses: SonarSource/sonarcloud-github-c-cpp@44cc4d3d487fbc35e5c29b0a9d717be218d3a0e8 # v3.2.0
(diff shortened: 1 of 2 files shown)


chongchonghe:
Not sure how to test this.

BenWibking:
I think this action is actually not used, since it doesn't work on PRs from forks. Sonarcloud has its own "automatic analysis" that runs instead via a GitHub app.

BenWibking:
It's disabled here: https://github.com/quokka-astro/quokka/actions/workflows/sonarcloud.yml

chongchonghe:
> I think this action is actually not used, since it doesn't work on PRs from forks. Sonarcloud has its own "automatic analysis" that runs instead via a GitHub app.

I think we need control of the SonarCloud if we want to use it. Otherwise, it will throw misleading warnings that are supported by different standards from ours, like C++23. 

Is there a way to configure the 'automatic analysis' to use C++20? If not, I would suggest using sonarcloud.yml . Fork support is not extremely important anyway since sonarcloud is an optional feature.

BenWibking:
I think it will be very confusing if PRs from external developers aren't analyzed by Sonarcloud, but PRs from branches in-repo are.

I will look into how to change the settings for the automatic analysis. Do you have an example of a spurious error from Sonarcloud due to the wrong C++ standard being used?

chongchonghe:
> I will look into how to change the settings for the automatic analysis. Do you have an example of a spurious error from Sonarcloud due to the wrong C++ standard being used?

https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=1516&id=BenWibking_TwoMomentRad&open=AZsiURLk7w-kv06zsJeF

This is a [C++23 feature](https://en.cppreference.com/w/cpp/utility/to_underlying.html):
- Use "std::to_underlying" to cast enums to their underlying type.

This is ignored by 'NOLINT' by still show up in SonarCloud:
- Replace this "enum" with "enum class".

This is annoying since the whole point of `TODO(name)` is to leave it to do later
- Complete the task associated to this "TODO" comment.

If we can control the rule, I would suppress those warning.

chongchonghe:
testing 'check_changes' in CI tests

/azp run

PR #1530, opened 2025-12-19
Add installation guide for macOS

chongchonghe:
### Description
Include a detailed instruction for macOS that helps with debugging MPI issues. 

I've also heard people complaining about `Could NOT find MPI` issues on macOS, and I also see it on my Macbook Air. The new guide will help resolving these kind of issues. 

```
-- Could NOT find MPI_CXX (missing: MPI_CXX_WORKS) 
CMake Error at /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:227 (message):
  Could NOT find MPI (missing: MPI_CXX_FOUND CXX) (found version "3.1")
Call Stack (most recent call first):
  /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:591 (_FPHSA_FAILURE_MESSAGE)
  /opt/homebrew/share/cmake/Modules/FindMPI.cmake:1845 (find_package_handle_standard_args)
  extern/amrex/Tools/CMake/AMReXParallelBackends.cmake:23 (find_package)
  extern/amrex/Src/CMakeLists.txt:40 (include)
```

### Related issues

### Checklist


chongchonghe:
This is fine because I don't want the 'installation_macos.md' page show in the navigation bar. 
```
Markdown files are not reachable via the MkDocs navigation:
  - installation_macos.md
```

@BenWibking @markkrumholz  Ready for review.

markkrumholz (dismissed):

markkrumholz:
> This is fine because I don't want the 'installation_macos.md' page show in the navigation bar.
> 
> ```
> Markdown files are not reachable via the MkDocs navigation:
>   - installation_macos.md
> ```
> 
> @BenWibking @markkrumholz Ready for review.

Looks like this issue is causing the automated tests to fail. So we will either have to force-bypass the rule, remove the rule, or fix the issue raised.

BenWibking:
I put the rule in place because I think it's better to make sure all of the documentation pages are accessible from the menu bar. This might have to be reconsidered if we end up with a large number of pages, say, several dozen.

Why do you want to hide it from the menu?

docs/markdown/installation_macos.md
```diff
@@ -0,0 +1,129 @@
...
+```bash
+python3 -m pip install cmake ninja --user
+```
+
+Using uv (if you have [uv](https://docs.astral.sh/uv/) installed):
+```bash
+uv tool install cmake ninja
+```
+
+Note: For uv, you may also use `uv pip install cmake ninja` if you're working within a virtual environment.
```
BenWibking: If we are providing this level of detail, I think we should recommend setting up a virtual environment inside the Quokka repo as `.venv`. Then `uv` should pick it up automatically when running any Python script from within the repo.

chongchonghe: That's up to the user to decide. I personally prefer to use pure python in uv's .venv path.

chongchonghe:
> I put the rule in place because I think it's better to make sure all of the documentation pages are accessible from the menu bar. This might have to be reconsidered if we end up with a large number of pages, say, several dozen.
> 
> Why do you want to hide it from the menu?

Becuase having a 'Installation' menu and a 'Installation for macOS' menu is confusing. Most of the time, the original instruction should just work for both linux and macOS.

chongchonghe:
> > This is fine because I don't want the 'installation_macos.md' page show in the navigation bar.
> > ```
> > Markdown files are not reachable via the MkDocs navigation:
> >   - installation_macos.md
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @BenWibking @markkrumholz Ready for review.
> 
> Looks like this issue is causing the automated tests to fail. So we will either have to force-bypass the rule, remove the rule, or fix the issue raised.

'Docs toctree coverage' is not a required workflow, so it does not affect merging queue.

BenWibking:
> > > This is fine because I don't want the 'installation_macos.md' page show in the navigation bar.
> > > ```
> > > Markdown files are not reachable via the MkDocs navigation:
> > >   - installation_macos.md
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > @BenWibking @markkrumholz Ready for review.
> > 
> > 
> > Looks like this issue is causing the automated tests to fail. So we will either have to force-bypass the rule, remove the rule, or fix the issue raised.
> 
> 'Docs toctree coverage' is not a required workflow, so it does not affect merging queue.

That's true, but it will then fail on all future PRs.

docs/markdown/installation_macos.md
BenWibking: What do you mean by "use pure python in uv's .venv path"?

BenWibking:
> > I put the rule in place because I think it's better to make sure all of the documentation pages are accessible from the menu bar. This might have to be reconsidered if we end up with a large number of pages, say, several dozen.
> > Why do you want to hide it from the menu?
> 
> Becuase having a 'Installation' menu and a 'Installation for macOS' menu is confusing. Most of the time, the original instruction should just work for both linux and macOS.

It should be possible to make it a sub-menu item.

chongchonghe:
> > > I put the rule in place because I think it's better to make sure all of the documentation pages are accessible from the menu bar. This might have to be reconsidered if we end up with a large number of pages, say, several dozen.
> > > Why do you want to hide it from the menu?
> > 
> > 
> > Becuase having a 'Installation' menu and a 'Installation for macOS' menu is confusing. Most of the time, the original instruction should just work for both linux and macOS.
> 
> It should be possible to make it a sub-menu item.

OK, I've done that.

BenWibking (approved):

PR #1128, opened 2025-06-05
Data table class for convenient interpolation

chongchonghe:
### Description

Implementation of the DataTable class for convenient bilinear interpolation and partial derivatives from a 2D data table. Bilinear interpolation is used in the DataTable cooling module. Partial derivatives are not used or tested.

`DataTableCooling.hpp` is tested and produces identical results as `ResampledCooling.hpp`. This header uses the `interpolate` member function of `DataTableGpuConst` for interpolation. 

I also wrote `numeric_derivative`. It compiles and runs, but I haven't tested its correctness. 

The next steps (in future PRs) are:
1. Extend the DataTable to support arbitrary dimensions via class templates. 
2. Implement a backward Euler solver in `computeCooling`.

#### The algorithm for bilinear interpolation and derivative

With the following coordinate mapping and normalization, 

```
y2  z3 -------- z4     (x1,y2) -------- (x2,y2)
    |     *     |         |     *     |
    | (h,v)     |         | (h,v)     |
    |           |         |           |
y1  z1 -------- z2     (x1,y1) -------- (x2,y1)
   x1          x2

where: h = (x - x1)/(x2 - x1), v = (y - y1)/(y2 - y1)
```

we have the following formula for `f(x, y), df_dx, df_dy` ([this paper](https://www.researchgate.net/publication/222529415_Laplace_spectra_as_fingerprints_for_image_recognition), equations before Fig. 4)

```python
h = (x - x1) / (x2 - x1)
v = (y - y1) / (y2 - y1)
f(x, y) = f(h, v) = (1 - v)((1 - h) z1 + h z2) + v((1 - h) z3 + h z4)

df_dh = v (z4 - z3) + (1 - v) (z2 - z1)
df_dv = h (z4 - z2) + (1 - h) (z3 - z1)

df_dx = f_h / (x2 - x1)
df_dy = f_v / (y2 - y1)
```

The following values are stored as an intermediate result and shared in computing `f(x,y)` and `df_dx, df_dy`: `ix, it, iix, iiy, x1, x2, y1, y2, h, v`. 

### Related issues
Closes #969 

### Checklist
_Before this pull request can be reviewed, all of these tasks should be completed. Denote completed tasks with an `x` inside the square brackets `[ ]` in the Markdown source below:_
- [ ] I have added a description (see above).
- [ ] I have added a link to any related issues (if applicable; see above).
- [ ] I have read the [Contributing Guide](https://github.com/quokka-astro/quokka/blob/development/CONTRIBUTING.md).
[...]

PR #1533, opened 2025-12-20
Migrate ResampledCooling to quokka::DataTable

chongchonghe:
### Description
This finishes what's planned in #1128, #1243 , and #1373 , using the new DataTable class for the resampled Grackle cooling. 

### Checklist


BenWibking:
Is this PR still needed?

chongchonghe:
> Is this PR still needed?

Yes, still needed. Didn't have time to finish it yet. This PR updates your resampled cooling module to use my new DataTable class. This involves updating both the python script and the ResampledCooling.hpp/cpp file to use `quokka::DataTableGpuConst` .

BenWibking:
Ok, that makes sense. I would like to change DataTable to use device memory before merging this. Otherwise, there will be a performance regression.

chongchonghe:
> Ok, that makes sense. I would like to change DataTable to use device memory before merging this. Otherwise, there will be a performance regression.

I'll take a closer look in a couple of days, but my impression is it's already using device memory. The data is copied from CPU to GPU at the beginning of a simulation and stays on GPU.

BenWibking:
> > Ok, that makes sense. I would like to change DataTable to use device memory before merging this. Otherwise, there will be a performance regression.
> 
> I'll take a closer look in a couple of days, but my impression is it's already using device memory. The data is copied from CPU to GPU at the beginning of a simulation and stays on GPU.

No, this statement is not correct. The current implementation does not copy to GPU memory.

See this PR for details: https://github.com/quokka-astro/quokka/pull/1648/changes

PR #1552, opened 2025-12-29
Fix PLM convergence

BenWibking:
### Description
There was a factor-of-two bug in the PLM slopes that preventing PLM convergence at order 2.

I have also switched to use the Sweby $\beta=1.5$ limiter, which is sharper than minmod. (Minmod only converges at order ~1.5 due to severe extrema clipping. MC converges at almost exactly order 2.)

It now converges at nearly order 2, as it should:
```
Running Richardson convergence test for HydroWave:
Resolution      Error Norm
----------      ----------

       128      5.680154e-09
       256      1.544953e-09
       512      4.469956e-10
      1024      1.255872e-10
      2048      3.500900e-11

Reached maximum resolution (nx = 2048) without achieving the target error 2.000e-11

Convergence Rate Analysis:
Resolution Pair Observed Rate   Expected Rate
--------------- -------------   -------------
 128 ->  256             1.88             2.0
 256 ->  512             1.79             2.0
 512 -> 1024             1.83             2.0
1024 -> 2048             1.84             2.0

Overall convergence rate: 1.84
Expected rate: 2.0
```

hydro_wave_convergence_limiters

### Related issues
N/A
### Checklist

### DIFF
--- a/scripts/python/hydro_wave_convergence_analysis.py
+++ b/scripts/python/hydro_wave_convergence_analysis.py
@@ -32,6 +32,13 @@ class FitResult:
     sse: float
 
 
+@dataclass
+class ConvergenceRun:
+    label: str
+    data: Sequence[ConvergenceDatum]
+    fit: FitResult
+
+
 def read_convergence_csv(path: Path) -> List[ConvergenceDatum]:
     if not path.exists():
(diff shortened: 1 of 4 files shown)


chongchonghe (approved):
LGTM.

PR #1567, opened 2026-01-12
Use 7-digit plotfile and checkpoint numbering

BenWibking:
### Description
Switch to using 7 digits for numbering checkpoint and plotfile outputs by default.

Since we regularly run simulations with millions of timesteps, this will avoid changes in the number of digits in output filenames in most cases.

### Related issues
N/A

### Checklist


chongchonghe (approved):
Thanks for doing this. I always wanted to do this. The current 5-digit plotfile name makes the output from `ls` command not sorted correctly. 

Assuming all tests in this PR will pass, there might be more gotchas caused by this change, e.g. I have to rename the regression test benchmark files.

PR #1576, opened 2026-01-16
`setDiodeBC` for MHD and Radiation

chongchonghe:
### Description
Similar to `setConstantDirichletBCLo` and `setConstantDirichletBCHi` introduced in #1574 , define helper functions `setDiodeBCLo` and `setDiodeBCLo` to set diode boundary condition for hydro (MHD and RHD support in progress)

### Related issues

### Checklist


chongchonghe:
@markkrumholz When we talked about first-order extrapolation for outflow,  are you thinking of 'real' first order (linear) extrapolation? However, AMReX's foextrap refers to constant (zero-order) extrapolation : https://erf.readthedocs.io/en/latest/BoundaryConditions.html#ideal-domain-bcs . This is what has been used everywhere in Quokka for extrapolation BC. Therefore, I implemented this in this PR.  I actually think this makes more sense than a linear extrapolation; the latter can introduce non-physical quantities like negative density. 

@BenWibking

BenWibking:
Yes, I think zero-th order extrapolation is better for this reason. It is robust and I don't think the accuracy difference is important for our problems.

chongchonghe:
Just a sanity check. The diode BC for radiation is the same as hydro, right? i.e. foextrap when flux is outward, and reflecting when flux is inward. Does this make sense? @BenWibking @markkrumholz

BenWibking:
> Just a sanity check. The diode BC for radiation is the same as hydro, right? i.e. foextrap when flux is outward, and reflecting when flux is inward. Does this make sense? @BenWibking @markkrumholz

If you really want a diode boundary for radiation, formally yes.

However, think the boundary should basically always be Marshak in practice.

markkrumholz:
I agree with @BenWibking that the correct implementation of diode for radiation is indeed the same as for gas, but that in practice one will almost always want to use Marshak rather than diode for radiation.

PR #1577, opened 2026-01-16
add a SN feedback test in a boosted frame

BenWibking:
### Description
This adds a test for the SN kinetic feedback algorithm in a boosted frame. Note that the gas and the particle velocity must be shifted by a constant $v_0$ for this test.

The result should be identical to the non-boosted frame, but it is not currently.

### Related issues
N/A

### Checklist


BenWibking:
@chongchonghe @markkrumholz This line does not agree with Equation 16 in the paper:
https://github.com/quokka-astro/quokka/blob/e8193fba819e7dbc36527bc26b2b1118be1600ae/src/particles/particle_deposition.hpp#L255

It should be the star particle velocity in the first term, NOT the gas velocity.

BenWibking:
@chongchonghe @markkrumholz I think the limiter (Equation 20) is not Galilean invariant:

https://github.com/quokka-astro/quokka/blob/e8193fba819e7dbc36527bc26b2b1118be1600ae/src/particles/particle_deposition.hpp#L435

https://github.com/quokka-astro/quokka/blob/e8193fba819e7dbc36527bc26b2b1118be1600ae/src/particles/particle_deposition.hpp#L448

BenWibking:
@chongchonghe @markkrumholz I think one way to fix this is to change to a comoving frame in the limiter. One possible algorithm I could imagine is this:
* Pick a reference velocity per cell that shifts under boosts, e.g. v_ref = p/ρ (local gas frame) or a kernel‑averaged gas velocity.
* Transform to that frame:
      * p_rel = p - ρ v_ref
      * Δp_rel = Δp - Δρ v_ref
      * e_rel = e - v_ref·p + 0.5 ρ v_ref^2
      * Δe_rel = Δe - v_ref·Δp + 0.5 Δρ v_ref^2
* Solve the limiter in that frame:
      * e_int + Δe_int + |p_rel + λ Δp_rel|^2 / (2(ρ+Δρ)) = e_rel + Δe_rel
* Apply back in lab frame:
      * Δp_lim = Δρ v_ref + λ Δp_rel (total Δe unchanged, same fallback if no root).

BenWibking:
I actually think ensuring total energy conservation is not important, and can lead to some pathological effects. Hopkins pointed this out in a 2025 paper (with conclusions in opposition to the Hopkins 2018 SN feedback paper): https://scixplorer.org/abs/2025OJAp....8E..44H/abstract

In the Hopkins approach (both 2018 and 2025), the coupling only depends on the gas-star relative velocity, so it is always Galilean invariant.

chongchonghe:
I confirm the SN feedback recipe is not Galilean invariant. Even worse, the gas temperature in a boosted frame is a lot higher. This is a bug. I'll try to fix it.

CleanShot-2026-01-17T12 23 19

BenWibking:
@chongchonghe Ok great.

I should note that the Hopkins method also solves an energy conservation equation, but it includes a term that depends on the gas-star velocity difference. This makes it Galilean-invariant:
```
 E_injection = E_ej + (1/2) m_ej Σ_b w'_b |v_b − v_a|^2
```
Then Hopkins solves an energy conservation equation to determine the momentum vs thermal energy balance.

I've attached my notes on the Hopkins method in case they are useful: [hopkins_cartesian_sn_algorithm.pdf](https://github.com/user-attachments/files/24684298/hopkins_cartesian_sn_algorithm.pdf)

chongchonghe:
A simple trick made it Galilean invariant: #1582 

CleanShot-2026-01-18T20 08 38

PR #1579, opened 2026-01-17
add meaningful names for particle components to plotfile headers

BenWibking:
### Description
Adds meaningful names for particle properties to the plotfiles.

In order for compile-time asserts of particle names to work, the particle creation system is adjusted slightly to be done at compile time.

### Related issues
N/A

### Checklist


chongchonghe (changes requested):
This design is bad because it uses lots of manual checking that is hard to maintain. For example, this snippet
```
	static constexpr bool has_mass_velocity = P == ParticleType::CIC || P == ParticleType::CICRad || P == ParticleType::StochasticStellarPop ||
						  P == ParticleType::Sink || P == ParticleType::Test;
```
is not generic, and requires update when we add a new particle type in the future. In fact, it can be replaced with `if (descriptor->getMassIndex() >= 0)`. And there are more examples like that.

chongchonghe:
@BenWibking I have a very simple solution to this feature. If you need this for the INCITE run, I can create a PR very quick.

BenWibking:
> @BenWibking I have a very simple solution to this feature. If you need this for the INCITE run, I can create a PR very quick.

@chongchonghe Yes, I think we need this. If you have a better solution, please do so.

chongchonghe:
Reimplemented in #1594

PR #1580, opened 2026-01-17
Record birth/death location and ambient density for star particles

BenWibking:
### Description
Save the birth and death coordinates (x,y,z) as particle real properties that are included in plotfiles. Also records the ambient gas density at the death location.

This is needed to check the Kennicutt-Schmidt relation.

### Related issues
Depends on https://github.com/quokka-astro/quokka/pull/1579.

### Checklist


chongchonghe (changes requested):
There are changes from your previous PR about 'meaningful particle component names' that was surpassed by my PR. Can you remove them? Other than that, everything else looks fine to me.

src/particles/PhysicsParticles.hpp
```diff
@@ -36,6 +38,234 @@
 namespace quokka
 {
 
+namespace particle_name_detail
```
chongchonghe: All of this chunk of code (up to line 256) should be removed.

src/particles/PhysicsParticles.hpp
```diff
@@ -705,6 +935,79 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 		    this->getBirthTimeIndex(), this->getMassAtBirthIndex(), state_fc, verbose);
 	}
 #endif // AMREX_SPACEDIM == 3
+
+      private:
+	[[nodiscard]] auto buildBaseRealCompNames() const -> amrex::Vector<std::string>
```
chongchonghe: and these code should be removed as well.

chongchonghe:
In some test problems like `RandomBlast` and `TallBoxSf`, the `createInitialStochasticStellarPopParticles` function is not updated and is still using 7 real components, and they will fail on Setonix due to FPE. However, since all tests passed on moth and avatargpu, I don't recommend fixing it in this PR. I propose we write a wrapper of `InitFromAsciiFile` to fill in extra components with zero. I can do it in another PR.

BenWibking:
> In some test problems like `RandomBlast` and `TallBoxSf`, the `createInitialStochasticStellarPopParticles` function is not updated and is still using 7 real components, and they will fail on Setonix due to FPE. However, since all tests passed on moth and avatargpu, I don't recommend fixing it in this PR. I propose we write a wrapper of `InitFromAsciiFile` to fill in extra components with zero. I can do it in another PR.

@chongchonghe Is there anything else you think is needed for this PR? If not, can you approve?

chongchonghe (dismissed):

BenWibking:
Had to resolve a merge conflict. @chongchonghe can you re-approve?

chongchonghe (approved):

PR #1582, opened 2026-01-18
Make SN feedback Galilean invariant

chongchonghe:
### Description

Progress:
1. fixed a bug on momentum update that Ben pointed out.
2. Identified that the current thermal-only scheme is Galilean invariant (apparently), but the other schemes that involves kinetic/momentum deposition are not, even for a single SN. Need to fix that. 
3. Updated the `depositThermalKineticMomentumSNR` function to do the following. This made the SN feedback Galilean invariant. 
    1. Find the center-of-mass velocity `v_frame` of the supernova remnant (SNR). We define the SNR as the sum of the gas enclosed in the SN feedback region and the ejecta with mass `m_ej`. This defines a COM frame. 
    2. Deposit the SN feedback into the COM frame isotropically.
    3. Transform SNR back into the lab frame.

The `SN` test shows the new scheme is Galilean invariant to machine accuracy:

SN test with two SNs overlapping at the first frame:
CleanShot-2026-01-18T19 05 40

### Related issues

### Checklist


chongchonghe:
CleanShot-2026-01-18T20 08 38

chongchonghe:
RandomBlast test: left: zero initial velocity; right: both gas and particles are boosted to `vx=1e8 cm/s`
CleanShot-2026-01-19T01 17 31

There two run should give exactly the same results up to displacement symmetry. However, there is a time delay between these two runs. I believe this can be explained by the different time steps: the boosted simulation has 3x shorter time step, and the SNR evolution could be sensitive to time step.

BenWibking:
Ok, I agree this basically works.

Can you run the RandomBlast problem (un-boosted) with a different CFL number, say, 0.1 and 0.3, and see if it gives a different SNR evolution (and also SFR over ~Myr timescales)? That might explain what's going on.

chongchonghe:
I think I reproduced the behavior of feedback strength depending on timestep in this simple test. I'll try to fix the bug tomorrow. The simulation with shorter time step undergoes SN earlier. This is wired.

BenWibking:
@chongchonghe I've tried to clean up this PR a bit so it's ready to merge. Let me know if you want to make any other changes or do more testing.

chongchonghe:
For reference, this is the star formation history from the old and new scheme, in a boosted (200 km/s) Tall Box sim:
CleanShot 2026-01-27 at 14 23 05

BenWibking:
The average timestep is a lot larger in my test runs with this PR. This may be due to the fixed artificial viscosity (which would be great), or it might be due to isotropizing the momentum components within the stencil.

BenWibking:
I will look at this in detail tomorrow.

Would it be possible to add a runtime option to choose either i) Galilean-invariant multi-SN feedback, or ii) energy-conserving multi-SN feedback? I would like to rule this out as a cause of the remining small discrepancy that I showed on the telecon.

chongchonghe:
I just pushed a commit where I added a new parameter `SN_use_galilean_invariant`, default to 1. When set to 0, it reduces to the original scheme where SN is not Galilean invariant (for a single SN) but the energy is strictly conserved. When set to 1, it uses the new Galilean-invariant feedback, but the total energy is not conserved due to velocity flattening.

BenWibking (commented):
Based on the tall box tests, I think this PR basically works fine.

I don't think we need to keep the old non-Galilean invariant behavior at all at this point. Do you want to remove that?

chongchonghe:
> Based on the tall box tests, I think this PR basically works fine.
> 
> I don't think we need to keep the old non-Galilean invariant behavior at all at this point. Do you want to remove that?

Yes, i think it makes sense to keep it default. I have removed the extra parameter and now it's ready for review. @BenWibking

BenWibking:
@chongchonghe Can you fix the two failing tests on moth:
```
The following tests FAILED:
	 93 - SN_long_grackle (Failed)
	 94 - SN_long_resampled (Failed)
```

chongchonghe:
@BenWibking Ready

docs/markdown/parameters.md
```diff
@@ -133,6 +133,7 @@ These parameters are read in the `particleParmParse()` function in `src/particle
...
 | particles.sink_particle_use_uniform_kernel    | Boolean (0/1) | `0` (Disabled)                                                                                                | If set to 1, uses uniform accretion kernel in a (7 dx)^3 box for sink particles.                                                                               |
 | particles.SN_scheme                           | String        | `SN_thermal_or_thermal_momentum`                                                                              | Scheme for SN feedback. Options: SN_thermal_only, SN_thermal_or_thermal_momentum, SN_thermal_kinetic_or_thermal_momentum, SN_pure_kinetic_or_thermal_momentum. |
+| particles.SN_use_galilean_invariant           | Boolean (0/1) | `1` (Enabled)                                                                                                 | Use Galilean-invariant formulation for momentum-based SN schemes. If 0, uses energy-conserving (lab-frame) formulation. See [Particles](particles.md) for details. |
```
BenWibking: Can you remove this? I thought this was removed from the code, since it's always Galilean invariant now.

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+- **Birth time**: Simulation time when the particle formed
+- **Death time**: When the particle explodes as a supernova
+- **Mass at birth**: Initial mass before mass loss
+- **Luminosity**: Radiation luminosity (if radiation is enabled)
+
+Each particle also stores an integer **evolution stage** that tracks its lifecycle:
+
+- `SNProgenitor`: High-mass star ($M > 8 M_{\odot}$) that will explode as a supernova
+- `SNRemnant`: Compact remnant left after supernova explosion
+- `LowMassStar`: Low-mass star that will not explode
```
BenWibking: When is a particle put into this stage?

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+- `LowMassComposite`: Composite particle representing a population of low-mass stars
+
+## Star Formation
+
+### Overview
+
+The star formation module adds star particles through a lightweight stochastic prescription that plugs into the Stochastic Stellar Population specialisation.
+
+- Runs once per hydro timestep and evaluates each cell independently.
+- Always spawns a low-mass star particle when a trial succeeds.
```
BenWibking: I thought it formed a LowMassComposite star?

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+   - Deposits $E_{\text{blast}} + E_{\text{kin}}$ as thermal energy
+   - Deposits ejecta mass and momentum $m_{\text{ej}} \vec{v}_{\text{ej}}$
+   - No resolution-dependent momentum injection
+   - Simplest scheme, appropriate when Sedov-Taylor phase is well-resolved. Should only be used for testing. 
+
+2. **`SN_thermal_or_thermal_momentum`** (default):
+   - When $R_M < 0.027$: Pure thermal (like `SN_thermal_only`)
+   - When $0.027 \leq R_M < 1$: Partial momentum injection with $f = 0.529 \sqrt{R_M}$
+   - When $R_M \geq 1$: Full momentum injection with $f = 1$
+   - Based on Kim & Ostriker (2017) calibration: $f^2 = 0.28$ captures 28% kinetic energy fraction in the well-resolved limit
```
BenWibking: This statement is incorrect. `f` is 0.28, *not* $f^2$.

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+1. **`SN_thermal_only`**: Pure thermal energy injection
+   - Deposits $E_{\text{blast}} + E_{\text{kin}}$ as thermal energy
+   - Deposits ejecta mass and momentum $m_{\text{ej}} \vec{v}_{\text{ej}}$
+   - No resolution-dependent momentum injection
+   - Simplest scheme, appropriate when Sedov-Taylor phase is well-resolved. Should only be used for testing. 
+
+2. **`SN_thermal_or_thermal_momentum`** (default):
+   - When $R_M < 0.027$: Pure thermal (like `SN_thermal_only`)
+   - When $0.027 \leq R_M < 1$: Partial momentum injection with $f = 0.529 \sqrt{R_M}$
+   - When $R_M \geq 1$: Full momentum injection with $f = 1$
```
BenWibking: Can you clarify that thermal energy may still be injected in this regime? This is what you said on Slack:
> Actually, let me clarify. We do have a small amount of thermal energy ejected in the unresolved regime. This is because I always assign a fraction (0%, 28%, and 100% in the three regimes) of the asymptotic momentum, p_snr , into a cell, and the rest (1e51 erg - assigned_KE) goes to thermal energy. This is convenient for the limiting step of multiple SN. The original idea was that the code should regulate the hot phase gas by efficient cooling in the unresolved phase.
> and note that p_snr^2 / 2 m_snr = ~70% * 1e51 erg, so there is about 30% thermal in the resolved regime.

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+
+$$
+\Delta \mathbf{p}_{ijk} = f \, p_{\text{snr}} \, W_{ijk} \, \hat{\mathbf{r}}_{ijk}
+$$
+
+This creates an isotropic outward momentum distribution centered on the SN position.
+
+#### Galilean Invariance Option
+
+The parameter `particles.SN_use_galilean_invariant` (default: `1`) controls whether the momentum deposition is Galilean invariant:
```
BenWibking: I thought this option was removed. Can you remove this section if so?

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+\Delta \vec{p}_{ijk} = \left[\rho_{\text{new}} \vec{v}_{\text{COM}} - \vec{p}_{\text{old}}\right] + \vec{p}_{\text{radial}}
+$$
+
+where $\vec{p}_{\text{radial}} = f \, p_{\text{snr}} \, W_{ijk} \, \hat{\mathbf{r}}_{ijk}$ is the radial momentum component.
+
+3. **Energy deposition** includes a cross term:
+
+$$
+\Delta E_{ijk} = \left(E_{\text{blast}} + E_{\text{kin}}\right) W_{ijk} + \vec{v}_{\text{COM}} \cdot \vec{p}_{\text{radial}}
+$$
```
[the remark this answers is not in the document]
BenWibking: Can you address this comment?

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+
+The cross term $\vec{v}_{\text{COM}} \cdot \vec{p}_{\text{radial}}$ accounts for the kinetic energy change from "resetting" the cell velocity to the COM frame before adding radial expansion. This term sums to zero over all cells (momentum is conserved), ensuring Galilean invariance.
+
+**Physical interpretation**: The gas velocity is first transformed to the COM frame, then isotropic expansion is added. This ensures that a boosted observer sees the same SN physics.
+
+##### Lab-frame Formulation (`SN_use_galilean_invariant = 0`)
```
BenWibking: Can you remove this section?

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+- **Mass**: $\Delta \rho_{ijk} = m_{\text{ej}} W_{ijk}$
+- **Momentum**: $\Delta \vec{p}_{ijk} = m_{\text{ej}} \vec{v}_{\text{ej}} W_{ijk}$ (ejecta momentum only)
+- **Energy**: $\Delta E_{ijk} = \left(E_{\text{blast}} + \frac{1}{2} m_{\text{ej}} |\vec{v}_{\text{ej}}|^2\right) W_{ijk}$
+
+### Runtime Parameters
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `particles.SN_scheme` | String | `SN_thermal_or_thermal_momentum` | Feedback scheme (see above) |
+| `particles.SN_use_galilean_invariant` | Boolean | `1` | Use Galilean-invariant formulation for momentum schemes |
```
BenWibking: Remove this parameter.

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Examples
+
+#### Basic SN Test
+
+The `SN` problem provides a canonical test with one or more SN progenitors in a uniform medium:
+
+```
+particles.SN_scheme = SN_thermal_or_thermal_momentum   # adaptive scheme
+particles.SN_use_galilean_invariant = 1                # use Galilean-invariant formulation
```
BenWibking: Remove this parameter

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+particles.verbose = 1                                  # print SN diagnostics
+```
+
+#### Random Blast Test
+
+The `RandomBlast` problem tests multiple SN explosions with various ambient conditions and bulk flows:
+
+```
+particles.SN_scheme = SN_thermal_or_thermal_momentum
+particles.SN_use_galilean_invariant = 1                # critical for moving medium
```
BenWibking: Remove this parameter.

docs/markdown/particles.md
```diff
@@ -0,0 +1,309 @@
...
+```
+
+The boost velocity tests demonstrate the importance of Galilean invariance. With `SN_use_galilean_invariant = 1`, the temperature and velocity profiles are identical in the boosted and rest frames (to numerical precision). With `SN_use_galilean_invariant = 0`, the profiles differ significantly.
```
BenWibking: Remove this sentence.

src/particles/particle_deposition.hpp
```diff
@@ -246,15 +275,31 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE void depositThermalKineticMomentumSNR(
...
-				const double dpy = (delta_rho_i * py / rho) + (momentum_per_cell * r_hat_y);
-				const double dpz = (delta_rho_i * pz / rho) + (momentum_per_cell * r_hat_z);
+				// Compute momentum change for Galilean invariance:
+				// After SN, cell velocity = v_COM + v_radial
+				// p_new = rho_new * v_COM + p_radial
+				// delta_p = p_new - p_old = (rho + delta_rho) * v_COM + p_radial - p_old
+				const double rho_new = rho + delta_rho_i;
+				const double dpx = (rho_new * v_COM_x - px) + p_radial_x;
+				const double dpy = (rho_new * v_COM_y - py) + p_radial_y;
+				const double dpz = (rho_new * v_COM_z - pz) + p_radial_z;
```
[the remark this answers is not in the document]
BenWibking: Is this comment accurate?

ChongChong: are you referring to a Gemini comment? I don't see it.

src/problems/SN/testSN.cpp
```diff
@@ -37,12 +43,14 @@ const double CV = 1. / (gamma_ - 1.) / mu * C::k_B;
...
 const double mass_SNR = 10.0 * C::M_solar;
-const int n_SNR = 2;
+const int n_SNR = 1;
 constexpr double B0 = 1.0e-7; // uniform background field for MHD variant
 
 static double n_amb = 1.0;    // ambient density (g cm^-3) // NOLINT
 static double T_amb = 100.0;  // ambient temperature (K) // NOLINT
 static double t_stop = 3.0e5; // stop time (yr) // NOLINT
+// static amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> boost_velocity{0.0, 0.0, 0.0}; // NOLINT
+static bool skip_checks = false; // NOLINT
```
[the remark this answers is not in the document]
BenWibking: Is this accurate? If so, can you fix?

BenWibking (changes requested):
Comments below

docs/markdown/parameters.md
chongchonghe: yes.

docs/markdown/particles.md
chongchonghe: We are not using `LowMassStar` anywhere in the code, because in the Stochastmic star formation recipe, all massive stars are candidate SN progenitors. It's there for future use, where we want to sample individual stars less massive than 8 solar mass.

chongchonghe: I have updated the text to explain it's not used in the star formation

docs/markdown/particles.md
chongchonghe: You are right. I've corrected it.

docs/markdown/particles.md
chongchonghe: Done.

docs/markdown/particles.md
chongchonghe: $f = 0.529 \sqrt{R_M}$ and $f^2 = 0.28 R_M$

chongchonghe: note that kinetic energy is proportional to f^2

docs/markdown/particles.md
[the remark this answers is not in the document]
chongchonghe: Done.

docs/markdown/particles.md
```diff
@@ -0,0 +1,246 @@
...
+4. Feedback is deposited into a temporary buffer
+5. Buffer is added to the state using a reproducibility-aware algorithm
+6. Particle evolution stage is updated to `SNRemnant`
+
+#### Numerical Reproducibility
+
+The deposition uses a roundoff-resistant algorithm to ensure bit-for-bit reproducibility across different processor counts. The algorithm:
+
+1. Deposits feedback into a temporary buffer with a count field
+2. Applies the Kahan compensated summation algorithm
```
BenWibking: Does the current code actually do Kahan summation? I thought it just rounded.

BenWibking (changes requested):

BenWibking:
@chongchonghe This is the Gemini comment:

Screenshot 2026-02-01 221929

docs/markdown/particles.md
chongchonghe: Oh, go catch. AI wrote this part and I failed to catch it.

chongchonghe:
@BenWibking @markkrumholz 
Here is a discussion about the Galilean invariance, smoothing, and energy conservation. There is some update compared to our previous discussion: Galilean invariance is guaranteed by including the cross term in the energy equation alone. For the momentum update, we can choose to either smooth gas velocity (default) to ensure energy conservation and reduce peak signal speed, or not smooth gas velocity but introduce an extra energy term that is the vector produce of the gas velocity in the COM frame and the radial momentum. In either scheme, the algorithm is reproducible and independent of the order at which the particles are processed.

https://github.com/quokka-astro/quokka/blob/chong/particles/fix-SN-v2/docs/markdown/particles.md#momentum-deposition

I attach a screenshot below because github has issues rendering Markdown equations:

CleanShot-2026-02-02T19 49 00

chongchonghe:
**Extra comments:**
This smoothing could make a significant influence on the SNR evolution in a scenario where two SN event occurs in a single cell at two consecutive steps, both injecting pure momentum. Suppose after the first step, the gas momentum is isotropic; in the second step, due to velocity smoothing, the radial momentum is removed from all cells for the second SN ejection, and they are converted to thermal energy (total energy is conserved). This is the regime where overcooling is an issue if thermal energy is ejected instead of terminal momentum. However, with the non-smoothing scheme, the extra inhomogeneity term can be extraordinarily large because $\delta v_i$ and $p_{\text{radial},i}$ are parallel. 

Although I don't know a solution to both issues, I believe a better option is to do smoothing, because over cooling is a lesser issue than introducing an extra heating term that is twice the kinetic energy: $v p_{\rm snr} = 2 (p_{\rm snr}^2 / 2 \rho)$.

BenWibking:
I think the smoothing is fine. But the paper made a point that it was not smoothed, so I wanted to make sure this was a deliberate choice and documented.

BenWibking (approved):

PR #1589, opened 2026-01-19
Fix bug in SN count

chongchonghe:
### Description
There was a bug in my previous implementation of SN count: the global variable `sn_count_` was never updated and therefore always 0. This PR fixed this bug and also added a global parameter `sn_count_cumulative_` to record the cumulative number of SN events. 

### Checklist

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -212,7 +212,8 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 	int sfh_interval_ = -1;		       // interval for the star formation history
 	amrex::Real sfh_time_interval_ = -1.0; // time interval for the star formation history
 	amrex::Real last_sfh_time_ = 0.0;
-	int sn_count_ = 0; // number of SN explosions in a step (used for diagnostics)
+	int sn_count_ = 0;	      // number of SN explosions in a step (used for diagnostics)
+	int sn_count_cumulative_ = 0; // cumulative number of SN explosions (used for diagnostics)
 
 	amrex::Real densityFloor_ = 0.0; // default
 	amrex::Real tempFloor_ = 0.0;	 // default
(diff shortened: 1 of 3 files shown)


src/simulation.hpp
```diff
@@ -1968,6 +1969,8 @@ template <typename problem_t> void AMRSimulation<problem_t>::particleMeshInterac
 
 	// Deposit the SN particles into the MultiFab
 	const auto [num_sn_explosions, max_velocity] = particleRegister_.depositSN(state_new_cc_[lev], state_fc_ptr, lev, time, dt);
+	sn_count_ = num_sn_explosions;
```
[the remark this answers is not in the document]
chongchonghe: `particleMeshInteraction` runs on all levels at once, so there is no such concern.

src/simulation.hpp
```diff
@@ -212,7 +212,8 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 	int sfh_interval_ = -1;		       // interval for the star formation history
 	amrex::Real sfh_time_interval_ = -1.0; // time interval for the star formation history
 	amrex::Real last_sfh_time_ = 0.0;
-	int sn_count_ = 0; // number of SN explosions in a step (used for diagnostics)
+	int sn_count_ = 0;	      // number of SN explosions in a step (used for diagnostics)
+	int sn_count_cumulative_ = 0; // cumulative number of SN explosions (used for diagnostics)
```
BenWibking, line 216: Do these get saved to `metadata.yaml`?

Ideally they should be, so they get restored for restarted sims.

src/simulation.hpp
BenWibking, line 1972: Is `num_sn_explosions` computed over all MPI ranks, or is this just for this MPI rank?
What about `max_velocity`?

BenWibking: Ok, I see it does this inside depositSN.

src/simulation.hpp
BenWibking: `sn_count_cumulative_` should be saved to metadata.yaml. I will update this PR to do that.

BenWibking: done.

chongchonghe:
@BenWibking Thanks for doing the edit. I'll take a look and merge the PR.

src/simulation.hpp
chongchonghe: Looks good.

src/simulation.hpp
chongchonghe: Not actually. In depositSN I only did the summation over all particles types, not over MPI ranks. I may have missed the MPI communication part. I'll take a look.

chongchonghe: Oh, never mind, I did. Forgot about this.

chongchonghe:
@BenWibking Everything looks fine. Can you approve this PR?

BenWibking (approved):

PR #1591, opened 2026-01-19
update amrex submodule (bugfix for AMR grids)

BenWibking:
### Description
Update AMReX. Includes an important grid generation bugfix needed for INCITE runs.

### Related issues
N/A

### Checklist


BenWibking:
Ok, it looks like changes to `amrex::Array4` broke our code.

BenWibking:
@chongchonghe The GPU tests shouldn't be skipped if the AMReX submodule is updated. We should re-run the GPU tests if AMReX is updated.

chongchonghe:
> @chongchonghe The GPU tests shouldn't be skipped if the AMReX submodule is updated. We should re-run the GPU tests if AMReX is updated.

OK, problem fixed. Depending on #1593 . 
This was the intended behavior, actually. However, the change of submodule amrex is just a pointer change (the commit SHA reference) at extern/amrex itself - not changes to files inside extern/amrex/**. So, add `extern/amrex` to the include list solved the problem.

BenWibking:
The fixed TurbGen works. Now we just need Christoph to merge https://github.com/chfeder/turbulence_generator/pull/4.

chongchonghe:
why do those MHD tests fail?

BenWibking:
It looks like the MHD tests are just timing out. AMReX's new Array4 implementation is slower in debug mode (but faster in release mode).

BenWibking:
Since this is a feature of AMReX, I think we will have to just adjust our tests so they don't take as many steps.

BenWibking:
Ah, wait. This AMReX PR should fix the timeout issue: https://github.com/AMReX-Codes/amrex/pull/4900

BenWibking:
Ok, I've updated amrex so it incorporates the fix for Debug mode slowness. Let's see if everything passes now.

chongchonghe (approved):

PR #1595, opened 2026-01-20
Fix 2D slices of derived vars

BenWibking:
### Description
After computing derived vars, they are averaged down to coarser AMR levels, and then the ghost cells are filled.

This is required for 2D slices of derived vars to work correctly, since the 2D slice machinery does linear interpolation that depends on ghost cell values when the slice plane lies on an AMR box boundary.

Also removes the quadratic interpolation option for 2D slices. This is not useful (and also untested).

### Related issues
N/A

### Checklist


chongchonghe (changes requested):
I think what we really should do is to always do piecewise constant interpolation for slices, because that's what comes out of the simulation. The user could do any kind of the interpolation they want in post process.

docs/markdown/insitu_analysis.md
```diff
@@ -60,7 +60,7 @@ quokka.slice_z.file = slicez_plt       # Output file prefix (should end in "plt"
 quokka.slice_z.normal = 2              # Plane normal (0 == x, 1 == y, 2 == z)
 quokka.slice_z.center = 2.4688e20      # Coordinate in the normal direction
 quokka.slice_z.int    = 10             # Output interval (in number of coarse steps)
```
markkrumholz, line 62: If linear is now the only option, should this parameter be removed entirely? It's confusing to have a parameter for which there is only a single option.

markkrumholz:
> I think what we really should do is to always do piecewise constant interpolation for slices, because that's what comes out of the simulation. The user could do any kind of the interpolation they want in post process.

I disagree and think doing linear is fine, but I think that if there is one option, then it doesn't make sense to have a variable to select that one option. I therefore think we should get rid of the input file variable.

chongchonghe (commented):
OK, I misunderstood what the interpolation does. Since its along the normal direction only, the user can't do interpolation on their own since they don't have the information. I think it's fine, although I would suggest adding an option to do piecewise constant because that's the most stable option, and it would also perfectly align with what `yt.sliceplot` does

chongchonghe:
> > I think what we really should do is to always do piecewise constant interpolation for slices, because that's what comes out of the simulation. The user could do any kind of the interpolation they want in post process.
> 
> I disagree and think doing linear is fine, but I think that if there is one option, then it doesn't make sense to have a variable to select that one option. I therefore think we should get rid of the input file variable.

I misunderstood what the interpolation does. I thought it's doing interpolation in the plane as well for coarse-fine boundaries.   I agree linear interpolation is fine, but I think it's good to have a second option to do piecewise constant.

docs/markdown/insitu_analysis.md
BenWibking: That makes sense. I'll get rid of the parameter.

BenWibking:
I don't think piecewise constant really makes sense for AMR data. You can never have data that is exactly consistent across resolution levels because the cell centers of one level don't line up with any of the cell centers of the next-coarser level.

Piecewise linear is extrema-preserving, and it is only carried out only in the plane-normal direction, so it basically can't go wrong.

chongchonghe (dismissed):
OK, looks fine.

BenWibking:
@chongchonghe @markkrumholz I've removed the `interpolation` parameter, since we now have only one interpolation method. It should be ready for re-review.

chongchonghe (approved):

BenWibking:
It seemed like some of the actions got stuck and failed to run for some reason, so I merged dev again to re-trigger all the actions. Hopefully they will all run now.

chongchonghe:
if it is moth, don't worry about it. moth rebooted just a couple of hours ago and that's why it failed. I will always monitor that.

chongchonghe (approved):

PR #1600, opened 2026-01-21
Refactor script for uploading regression test results

BenWibking:
### Description
This force pushes and overwrites the previous results in the regression testing results website. This should be faster and save disk space. It should also show if there are network or filesystem issues that make this upload slow.

### Related issues
N/A

### Checklist

### DIFF
--- a/.ci/azure-pipelines-regression.yml
+++ b/.ci/azure-pipelines-regression.yml
@@ -38,7 +38,7 @@ jobs:
 
     - template: templates/ccache-prepare.yml
 
-    - script: ./extern/regression_testing/regtest.py --clean_testdir regression/quokka-tests.ini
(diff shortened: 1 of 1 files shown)


chongchonghe (commented):
Testing it right now.Will approve once regression test is successful.

chongchonghe:
@BenWibking The `git reset --mixed` command seems dangerous and left the whole directory untracked (probably due to me interrupting it while the agent is running). 

Anyway, the origin of the 'publish to github' hanging issue is because, whenever 'git' tries to ask the user for input, like to set `--user` and `--email`, or adding an address to known_host, it will wait for the user to confirm, and this causes the workflow to hang indefinitely. I created a PR to solve that issue: #1604

PR #1605, opened 2026-01-21
fix a bug in artificial viscosity

mxchen39:
### Description
In the current implementation of artificial viscosity, the viscosity term is added in the wrong direction. This PR fixes the bug by using the variable `F_canonical` to store intermediate fluxes and then applying the correct permutation.

### Related issues

### Checklist


chongchonghe (commented):
This looks correct to me, but I'll wait @BenWibking to double check.

chongchonghe:
This PR will correct a mistake in #1599

BenWibking (approved):
Thanks for catching this.

PR #1606, opened 2026-01-22
Fix Ghost Cell Handling for Gravitational Potential After Poisson Solve

chongchonghe:
### The bug
The TallBoxSf test on Setonix failed with
```
		AMREX_ALWAYS_ASSERT(!phi_extended.contains_nan());
```
inside `kickParticlesAllLevels`.

### Plausible reason

The MLMG Poisson solver (`mlmg.solve()`) only fills valid (interior) cells of the solution MultiFab `phi`, leaving ghost cells uninitialized. This caused incorrect gravitational accelerations at box boundaries when computing gradients in:
- `applyPoissonGravityAtLevel()` - applies gravity to hydrodynamic state
- `kickParticlesAllLevels()` - computes particle acceleration from potential gradients

Both functions use centered finite differences that require properly filled ghost cells:
```cpp
gx = -0.5 * (phi[i+1] - phi[i-1]) / dx
```

### `calculateGpotAllLevels()` - Ghost Cell Filling After Poisson Solve

**For MLMG solver** (any dimension is periodic):
- Fill periodic boundaries with `FillBoundary()`
- Apply Dirichlet BC (φ = 0) at non-periodic physical boundaries using `setFunctorPhiZero`
- Coarse-fine boundaries are handled by the MLMG solver for valid cells

**For OpenBC solver** (no periodic dimensions):
- No ghost cell filling needed - the solver computes physically-consistent boundary values (φ → 0 at infinity, NOT at domain boundaries)

### `kickParticlesAllLevels()` - Extended Ghost Cells for Particle Acceleration

Particles require `phi_extended` with 3 ghost cells (vs 1 in `phi`) for gradient computation with CIC interpolation.

**Level 0:**
- Copy from `phi`, fill periodic boundaries with `FillBoundary()`
- For MLMG with Dirichlet BC: apply `setFunctorPhiZero` at non-periodic boundaries

**Fine levels:**
- Use `FillPatchTwoLevels` to properly handle coarse-fine interpolation
- For MLMG with Dirichlet BC: use `setFunctorPhiZero` boundary functor
- For fully periodic or OpenBC: use no-op functor `setFunctorParticleAccel`

## Code Changes

- **New functor `setFunctorPhiZero`**: Enforces homogeneous Dirichlet BC (φ = 0) at physical boundaries
- **New functor `setFunctorParticleAccel`**: No-op functor for OpenBC/periodic cases in particle acceleration
- **`calculateGpotAllLevels()`**: Ghost cell filling moved inside `if (use_mlmg_solver)` block; OpenBC needs no filling
- **`kickParticlesAllLevels()`**: Uses `FillPatchTwoLevels` for fine levels with proper coarse-fine interpolation

### Checklist


BenWibking:
This is not correct. `phi` is not zero at physical boundaries when amrex::OpenBC is used. The whole point of OpenBC is to compute the physically-consistent values of `phi` at the boundary that give a value of zero at infinity (NOT at the box boundaries).

This is also not correct at coarse-fine AMR boundaries. The ghost cells must also be consistently filled there using amrex::FillPatchTwoLevels in AMReX_MultiFabUtil.H.

BenWibking:
Actually, MLMG should fill 1 ghost cell layer with the correct solution. If it's not doing this, then I think there's a bug in how we're using it or in AMReX itself.

chongchonghe:
> Actually, MLMG should fill 1 ghost cell layer with the correct solution. If it's not doing this, then I think there's a bug in how we're using it or in AMReX itself.

No, it's not. MLMG does not fill in physical boundaries.

BenWibking:
Ok, it should definitely fill in *internal* boundaries.

Can you modify this PR so it just fills the physical boundaries? OpenBC requires special treatment, and you cannot set `phi=0` for the physical boundaries when using it.

chongchonghe:
> This is not correct. `phi` is not zero at physical boundaries when amrex::OpenBC is used. The whole point of OpenBC is to compute the physically-consistent values of `phi` at the boundary that give a value of zero at infinity (NOT at the box boundaries).

OK, that makes sense. Thanks for pointing out. Corrected.

> This is also not correct at coarse-fine AMR boundaries. The ghost cells must also be consistently filled there using amrex::FillPatchTwoLevels in AMReX_MultiFabUtil.H.

Yes, I did that. One thing to note is that for fine levels, we have to use `setFunctorPhiZero` for fabs that touches the physical boundary.

chongchonghe:
@BenWibking Ready for another review.

BenWibking:
If I remember correctly, we have tried to fix this several times, but some edge case was not handled correctly. Can you create a Discussion post on the AMReX GitHub and explain the issues with ghost cells for MLMG and OpenBC + particles and ask for a clarification? Weiqun should be able to clarify this, and the community will have a permanent answer for these questions going forward. I would like to try to make sure this PR gets it right and we don't need any future fixes for this.

src/simulation.hpp
```diff
@@ -1879,37 +1939,44 @@ template <typename problem_t> void AMRSimulation<problem_t>::kickParticlesAllLev
...
+			if (needs_dirichlet_bc) {
+				// Apply Dirichlet BC (phi=0) at non-periodic boundaries for MLMG solver
+				amrex::GpuBndryFuncFab<setFunctorPhiZero> boundaryFunctor(setFunctorPhiZero{});
+				amrex::PhysBCFunct<amrex::GpuBndryFuncFab<setFunctorPhiZero>> phiBdryFunct(geom[lev], phiBC, boundaryFunctor);
+				phiBdryFunct(phi_extended, 0, 1, phi_extended.nGrowVect(), 0., 0);
+			} else {
+				// Apply physical boundary conditions to phi at open boundaries
+				amrex::GpuBndryFuncFab<setFunctorPhiNoOp> boundaryFunctor(setFunctorPhiNoOp{});
+				amrex::PhysBCFunct<amrex::GpuBndryFuncFab<setFunctorPhiNoOp>> phiBdryFunct(geom[lev], phiBC, boundaryFunctor);
+				phiBdryFunct(phi_extended, 0, 1, phi_extended.nGrowVect(), 0., 0);
```
BenWibking, line 1971: Is this `else` branch needed at all? Can we just get rid of it?

src/simulation.hpp
```diff
@@ -1786,22 +1809,45 @@ template <typename problem_t> void AMRSimulation<problem_t>::calculateGpotAllLev
 			if (verbose) {
 				amrex::Print() << "MLMG converged with final residual norm: " << final_resnorm << "\n";
 			}
+
+			// Fill ghost cells of phi after MLMG solve
+			// For MLMG: fill periodic boundaries and apply Dirichlet BC (phi=0) at non-periodic physical boundaries
+			// Note: coarse-fine boundaries are handled by the MLMG solver for valid cells.
+			for (int lev = 0; lev <= finest_level; ++lev) {
+				// Fill periodic boundaries
+				phi[lev].FillBoundary(geom[lev].periodicity());
```
BenWibking, line 1831: I think periodic boundaries should be filled by MLMG? Did you test this?

BenWibking (changes requested):

chongchonghe:
> If I remember correctly, we have tried to fix this several times, but some edge case was not handled correctly. Can you create a Discussion post on the AMReX GitHub and explain the issues with ghost cells for MLMG and OpenBC + particles and ask for a clarification? Weiqun should be able to clarify this, and the community will have a permanent answer for these questions going forward. I would like to try to make sure this PR gets it right and we don't need any future fixes for this.

@BenWibking 
I've mentioned this issue several times in my previous PRs, but I have never actually addressed it. This issue is very unusual; I was never able to consistently reproduce it. 

I just realized that this error only occurs when there are more than one level fully refines (or touches the domain boundary). This is noted in your comment at simulation.hpp:1826
```
			// NOTE: this fails when multiple levels are fully refined when open boundary condition is used.
			AMREX_ALWAYS_ASSERT_WITH_MESSAGE(!phi[lev].contains_nan(), fmt::format("NaN detected in phi at level {} after Poisson solve", lev));
		}
```

chongchonghe:
So, that is the problem to fix, and I have to fix it for my tall box simulations because I plan to use a AMR structure where level 2 (or up to level 10) will touch the domain boundary at the disk plane. 

This PR (unintentionally) fixed this issue. I'll try to understand exactly what's going on what does this fix. I'll come back to this later this weekend.

chongchonghe:
@BenWibking Replaced by a much simpler solution in #1660 . Can you take a look?

PR #1607, opened 2026-01-22
add helper functions for SFR diagnostics

BenWibking:
### Description
This adds helper functions to compute SFR history diagnostics.

This is useful to write SFR history with multiple sliding windows to the statistics output file `history.txt`.

### Related issues
N/A

### Checklist

### DIFF
--- a/src/particles/PhysicsParticles.hpp
+++ b/src/particles/PhysicsParticles.hpp
@@ -1,6 +1,7 @@
 #ifndef PHYSICS_PARTICLES_HPP_
 #define PHYSICS_PARTICLES_HPP_
 
+#include <algorithm>
 #include <array>
 #include <cstdint>
 #include <map>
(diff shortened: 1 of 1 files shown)


markkrumholz (changes requested):
I'm a bit confused by the logic of this code. It seems like you're averaging over entries in the SF history -- but why? If your goal is to get the mean SFR over some time window, isn't the logical way to do it just to compute the mass now, mass at the start of the time window, and then divide by the time window? That is, why isn't the correct implementation here just

`const amrex::Real mass_now = compute_mass_at_time(history, time_now);`
`const amrex::Real mass_start = compute_mass_at_time(history, time_now - time_window);`
`return (mass_now - mass_start) / (time_now - time_window);`

plus maybe some checks to handle the special case when `time_now - time_window` is smaller than the first entry in the SF history. I don't understand the purpose of the loop starting on line 1113.

BenWibking:
> I'm a bit confused by the logic of this code. It seems like you're averaging over entries in the SF history -- but why? If your goal is to get the mean SFR over some time window, isn't the logical way to do it just to compute the mass now, mass at the start of the time window, and then divide by the time window? That is, why isn't the correct implementation here just
> 
> `const amrex::Real mass_now = compute_mass_at_time(history, time_now);` `const amrex::Real mass_start = compute_mass_at_time(history, time_now - time_window);` `return (mass_now - mass_start) / (time_now - time_window);`
> 
> plus maybe some checks to handle the special case when `time_now - time_window` is smaller than the first entry in the SF history. I don't understand the purpose of the loop starting on line 1113.

Yes, that's a good point. We need to compute the sum of the birth_mass of the stars, but that should work fine.

BenWibking:
@markkrumholz I've simplified the implementation as you've suggested, and summing the particles by their birth mass while filtering out particles that were born after the requested time $t$. It's ready to be re-reviewed.

markkrumholz (approved):

PR #1633, opened 2026-02-04
add optional cap on composite star particle masses

BenWibking:
### Description
This adds an optional cap on the composite star particle masses. If the composite star particle would have a mass exceeding the cap, then it splits the composite star particle into multiple particles of equal mass with randomized velocities (dispersion set to 0.5*cs). In all other respects, the star formation algorithm is unchanged.

The cap is set in the ParmParse input with
```
particles.low_mass_composite_max_mass = 1.99e33 * 1e3   # mass in code units
```

This is needed to avoid dynamical friction effects due to large point masses. 

### Related issues
N/A

### Checklist

### DIFF
--- a/src/particles/particle_creation.hpp
+++ b/src/particles/particle_creation.hpp
@@ -508,9 +508,6 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 						++num_low;
 					}
 				}
-				if (num_low > num_particles) {
-					num_low = num_particles;
-				}
 				const int num_high = num_particles - num_low;
 				const amrex::Real mass_low_each = mass_low_mass_star / static_cast<amrex::Real>(num_low);
 				const bool split_low_mass_composite = (num_low > 1);
--- a/src/particles/particle_creation.hpp
+++ b/src/particles/particle_creation.hpp
@@ -551,7 +551,7 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 						p.idata(evolution_stage_index) = static_cast<int>(StellarEvolutionStage::LowMassComposite);
 						// Set particle mass
 						p.rdata(mass_idx) = mass_low_each;
-						// Set particle postion and velocity
+						// Set particle position and velocity
 						if (split_low_mass_composite) {
 							// Randomize velocity within the cell only when split
 							const amrex::Real rx = amrex::Random(engine);
@@ -696,7 +696,7 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 					state_arr(i, j, k, HydroSystem<problem_t>::x2Momentum_index) *= factor;
 					state_arr(i, j, k, HydroSystem<problem_t>::x3Momentum_index) *= factor;
 
-					// Update internal energy to relect mass change
+					// Update internal energy to reflect mass change
 					state_arr(i, j, k, HydroSystem<problem_t>::internalEnergy_index) *= factor;
 
 					// Update total energy
--- a/src/particles/particle_creation.hpp
+++ b/src/particles/particle_creation.hpp
@@ -585,7 +585,7 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 							double const cos_theta_random = 2. * amrex::Random(engine) - 1.;
 							double const sin_theta_random = std::sqrt(1. - cos_theta_random * cos_theta_random);
 							// Sample phi from a uniform distribution between 0 and 2*pi
-							double const phi_random = (1. - amrex::Random(engine)) * 2. * std::numbers::pi;
+							double const phi_random = amrex::Random(engine) * 2. * std::numbers::pi;
 
 							double const vx_random = v_mag * sin_theta_random * std::cos(phi_random);
 							double const vy_random = v_mag * sin_theta_random * std::sin(phi_random);
--- a/src/particles/particle_creation.hpp
+++ b/src/particles/particle_creation.hpp
@@ -399,6 +399,7 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 		amrex::Real param1 = particle_param1;
 		amrex::Real param2 = particle_param2;
 		amrex::Real eps_ff_ = eps_ff;
+		amrex::Real low_mass_composite_max_mass_ = low_mass_composite_max_mass;
 
 		AMREX_GPU_HOST_DEVICE ParticleChecker(amrex::Real current_time, amrex::Real dt) : current_time(current_time), dt(dt) {}
 
@@ -426,8 +427,17 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 			    random_draw < prob_star_formation) { // Create a particle only if LambdaJ < J*dx and prob_star_formation> random draw
 				const amrex::Real particle_mass = cell_density * cell_volume * eps_star;
 				const amrex::Real m_high_tot = particle_mass * fstar_high;
+				const amrex::Real mass_low_mass_star = particle_mass * (1.0 - fstar_high);
 				amrex::Real const num_high_mass_stars_exp = m_high_tot / m_star_high_avg;
-				num_star = static_cast<int>(1 + (amrex::RandomPoisson(num_high_mass_stars_exp, engine)));
+				const int num_high = static_cast<int>(amrex::RandomPoisson(num_high_mass_stars_exp, engine));
+				int num_low = 1;
+				if ((low_mass_composite_max_mass_ > 0.0) && (mass_low_mass_star > low_mass_composite_max_mass_)) {
+					num_low = static_cast<int>(mass_low_mass_star / low_mass_composite_max_mass_);
+					if (mass_low_mass_star > static_cast<amrex::Real>(num_low) * low_mass_composite_max_mass_) {
+						++num_low;
+					}
+				}
+				num_star = num_low + num_high;
 			}
 			return num_star;
 		}
@@ -448,6 +458,7 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 		amrex::Real param2 = particle_param2;
 		amrex::Real eps_ff_ = eps_ff;
 		amrex::Real stellar_velocity_limit_ = stellar_velocity_limit;
+		amrex::Real low_mass_composite_max_mass_ = low_mass_composite_max_mass;
 
 		AMREX_GPU_HOST_DEVICE
 		ParticleCreator(int mass_index, int birth_time_index, int death_time_index, int processor_id, amrex::Long particle_id_start,
@@ -483,7 +494,20 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 				double total_momy = 0.0;
 				double total_momz = 0.0;
 
-				// p_idx = 0 represents the low mass star and p_idx = 1, 2..  represent the high mass stars
+				int num_low = 1;
+				if ((low_mass_composite_max_mass_ > 0.0) && (mass_low_mass_star > low_mass_composite_max_mass_)) {
+					num_low = static_cast<int>(mass_low_mass_star / low_mass_composite_max_mass_);
+					if (mass_low_mass_star > static_cast<amrex::Real>(num_low) * low_mass_composite_max_mass_) {
+						++num_low;
+					}
+				}
+				if (num_low > num_particles) {
+					num_low = num_particles;
+				}
+				const int num_high = num_particles - num_low;
+				const amrex::Real mass_low_each = mass_low_mass_star / static_cast<amrex::Real>(num_low);
+
+				// p_idx = 0..(num_low-1) represent the low mass composites, p_idx = num_low.. represent the high mass stars
 
 				for (int p_idx = 0; p_idx < num_particles; ++p_idx) {
 					auto &p = particles[p_idx]; // NOLINT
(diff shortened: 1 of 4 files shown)


src/particles/particle_creation.hpp
```diff
@@ -426,8 +427,17 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
 				amrex::Real const num_high_mass_stars_exp = m_high_tot / m_star_high_avg;
-				num_star = static_cast<int>(1 + (amrex::RandomPoisson(num_high_mass_stars_exp, engine)));
+				const int num_high = static_cast<int>(amrex::RandomPoisson(num_high_mass_stars_exp, engine));
+				int num_low = 1;
+				if ((low_mass_composite_max_mass_ > 0.0) && (mass_low_mass_star > low_mass_composite_max_mass_)) {
+					num_low = static_cast<int>(mass_low_mass_star / low_mass_composite_max_mass_);
+					if (mass_low_mass_star > static_cast<amrex::Real>(num_low) * low_mass_composite_max_mass_) {
+						++num_low;
+					}
+				}
```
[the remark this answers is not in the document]
markkrumholz: Given that this is a basically a one-line bit of logic being duplicated, I don't see this as essential.

src/particles/particle_creation.hpp
```diff
@@ -499,10 +523,20 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
 
-					// Low Mass particle position at cell center
-					p.pos(0) = plo[0] + (i + 0.5) * dx[0];
-					p.pos(1) = plo[1] + (j + 0.5) * dx[1];
-					p.pos(2) = plo[2] + (k + 0.5) * dx[2];
+					if (p_idx < num_low) {
+						// Randomize LowMassComposite position within the cell
+						const amrex::Real rx = amrex::Random(engine);
+						const amrex::Real ry = amrex::Random(engine);
+						const amrex::Real rz = amrex::Random(engine);
```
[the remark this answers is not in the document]
markkrumholz: I think the bot is correct on this one.

src/particles/particle_creation.hpp
```diff
@@ -483,7 +494,20 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
+				int num_low = 1;
+				if ((low_mass_composite_max_mass_ > 0.0) && (mass_low_mass_star > low_mass_composite_max_mass_)) {
+					num_low = static_cast<int>(mass_low_mass_star / low_mass_composite_max_mass_);
+					if (mass_low_mass_star > static_cast<amrex::Real>(num_low) * low_mass_composite_max_mass_) {
+						++num_low;
+					}
+				}
+				if (num_low > num_particles) {
+					num_low = num_particles;
+				}
```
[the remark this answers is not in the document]
markkrumholz: Agree

src/particles/particle_creation.hpp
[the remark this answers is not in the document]
BenWibking: That's a good point. I'll revert to the old behavior when it doesn't split the composite particle.

markkrumholz (commented):
Other than the bot comments, my only other comment is that I am not confident that random placement within the cell will be sufficient to prevent the particles from moving together for a long time, given that they all have the same velocity. The only thing that will cause them to separate is tidal forces associated with their different sub-cell positions, which are likely to be weak. In a galaxy simulation, I could easily see the time taken for them to separate being comparable to the galactic orbital period. We could just try it and see, but we might also consider adding a small random velocity to encourage separation.

BenWibking:
> In a galaxy simulation, I could easily see the time taken for them to separate being comparable to the galactic orbital period. We could just try it and see, but we might also consider adding a small random velocity to encourage separation.

@markkrumholz Ok, we could. How would you choose this velocity? If based on the local velocity dispersion, I think we would have to add a cap to prevent something wild in shocks and I wanted to avoid adding another parameter. What are your thoughts on this?

markkrumholz:
> > In a galaxy simulation, I could easily see the time taken for them to separate being comparable to the galactic orbital period. We could just try it and see, but we might also consider adding a small random velocity to encourage separation.
> 
> @markkrumholz Ok, we could. How would you choose this velocity? If based on the local velocity dispersion, I think we would have to add a cap to prevent something wild in shocks and I wanted to avoid adding another parameter. What are your thoughts on this?

To be safe, how about making it 50% of the local sound speed, which we know will be relatively small because this is a Jeans-violating cell? We don't need it to be a lot, just enough that the particles will drift apart by more than a cell in a reasonable amount of time.

BenWibking:
> > > In a galaxy simulation, I could easily see the time taken for them to separate being comparable to the galactic orbital period. We could just try it and see, but we might also consider adding a small random velocity to encourage separation.
> > 
> > 
> > @markkrumholz Ok, we could. How would you choose this velocity? If based on the local velocity dispersion, I think we would have to add a cap to prevent something wild in shocks and I wanted to avoid adding another parameter. What are your thoughts on this?
> 
> To be safe, how about making it 50% of the local sound speed, which we know will be relatively small because this is a Jeans-violating cell? We don't need it to be a lot, just enough that the particles will drift apart by more than a cell in a reasonable amount of time.

Ok, that seems like a reasonable choice.

src/particles/particle_creation.hpp
```diff
@@ -520,13 +554,13 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 					p.rdata(StochasticStellarPopParticleDeathDensityIdx) = unset_position;
 
 					// Low Mass particle mass and velocity
-					p.rdata(mass_idx) = mass_low_mass_star;
+					p.rdata(mass_idx) = (p_idx < num_low) ? mass_low_each : 0.0;
```
chongchonghe: We can put lines 557 to 562 into if (p_idx < num_low) instead of repeating assignment.

src/particles/particle_creation.hpp
```diff
@@ -578,12 +612,16 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
 				}
 
-				if (num_particles > 1) { // Update momentum of the low mass star if there is(are) high mass star(s) in the cell
-					const int p_idx = 0;
-					auto &plow = particles[p_idx]; // NOLINT
-					plow.rdata(mass_idx + 1) = -total_momx / plow.rdata(mass_idx);
-					plow.rdata(mass_idx + 2) = -total_momy / plow.rdata(mass_idx);
-					plow.rdata(mass_idx + 3) = -total_momz / plow.rdata(mass_idx);
+				if (num_high > 0) { // Update momentum of the low mass stars if there is(are) high mass star(s) in the cell
+					const amrex::Real vx_low = -total_momx / mass_low_mass_star;
```
chongchonghe: There is a serious bug here in the original code (and this PR carries it as well). The total momentum of the particles in a cell add up to 0, as long as there are high-mass stars. But they should add to the total mass times the cell velocity, i.e. they should carry the cell momentum. 
To correct this, `plow.rdata(mass_idx + 1)` should be set to either
(1) `(particle_mass * vx - total_momx) / mass_low_mass_star`, or
(2) `(real_particle_total_mass * vx - totla_momx) / mass_low_mass_star`
The diffrence between the two is: (1) would garantees momentum convervation, because `particle_mass * vx` is the momentum taken away from the cell. (2) would violate momenum conservation because `real_particle_total_mass != particle_mass` due to stachastoty, but it will garantee the COM velocity of all the stars is `vx`. I would advocate option (2) because if we violate mass conservatio in a single cell, it's fine to violate momentum conservation in the same way, but it's more important to keep the velocity correct in a rotating disk.

chongchonghe (changes requested):
see my comments in the code. A critical comment is that the particle momentum was set wrong.

BenWibking:
Will merge in https://github.com/quokka-astro/quokka/pull/1634 after it goes though the merge queue, and then try to address the remaining comments tomorrow.

src/particles/particle_creation.hpp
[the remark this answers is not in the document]
BenWibking: done.

src/particles/particle_creation.hpp
```diff
@@ -502,89 +524,119 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
-					// Set particle evolution stage to LowMassComposite if it is a low-mass stellar composite
-					// This gets changed in the for loop below if this is a high mass star
-					p.idata(evolution_stage_index) = static_cast<int>(StellarEvolutionStage::LowMassComposite);
-
-					// Low Mass particle position at cell center
+					// Set particle position to cell center
+					// (Note: if multiple particles are formed, their velocities will be randomized.)
 					p.pos(0) = plo[0] + (i + 0.5) * dx[0];
 					p.pos(1) = plo[1] + (j + 0.5) * dx[1];
 					p.pos(2) = plo[2] + (k + 0.5) * dx[2];
```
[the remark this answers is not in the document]
BenWibking: This is handled with velocity dispersion instead. I will update the PR description.

src/particles/particle_creation.hpp
```diff
@@ -433,8 +435,17 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
+				const amrex::Real mass_low_mass_star = particle_mass * (1.0 - fstar_high);
 				amrex::Real const num_high_mass_stars_exp = m_high_tot / m_star_high_avg;
-				num_star = static_cast<int>(1 + (amrex::RandomPoisson(num_high_mass_stars_exp, engine)));
+				const int num_high = static_cast<int>(amrex::RandomPoisson(num_high_mass_stars_exp, engine));
+				int num_low = 1;
+				if ((low_mass_composite_max_mass_ > 0.0) && (mass_low_mass_star > low_mass_composite_max_mass_)) {
+					num_low = static_cast<int>(mass_low_mass_star / low_mass_composite_max_mass_);
+					if (mass_low_mass_star > static_cast<amrex::Real>(num_low) * low_mass_composite_max_mass_) {
+						++num_low;
+					}
```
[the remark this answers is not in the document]
BenWibking: Fixed.

BenWibking:
Since I moved some code blocks around and added braces, this PR may be significantly easier to read in split mode:
https://github.com/quokka-astro/quokka/pull/1633/changes?diff=split

src/particles/particle_creation.hpp
```diff
@@ -502,89 +524,119 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
+						p.idata(evolution_stage_index) = static_cast<int>(StellarEvolutionStage::LowMassComposite);
+						// Set particle mass
+						p.rdata(mass_idx) = mass_low_each;
+						// Set particle position and velocity
+						if (split_low_mass_composite) {
+							// Randomize velocity within the cell only when split
+							const amrex::Real rx = amrex::Random(engine);
+							const amrex::Real ry = amrex::Random(engine);
+							const amrex::Real rz = amrex::Random(engine);
+							// randomize velocity with velocity dispersion ~0.5*cs
```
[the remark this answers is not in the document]
markkrumholz: I think this is just a hallucination. The code looks correct as written to me.

markkrumholz (dismissed):

chongchonghe:
I am reviewing it now

src/particles/particle_creation.hpp
```diff
@@ -502,89 +515,119 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
+							const amrex::Real cs = HydroSystem<problem_t>::ComputeSoundSpeed(state_arr, i, j, k, fab_fc);
+							constexpr amrex::Real vdisp_norm = 0.5 / std::numbers::sqrt3;
+							p.rdata(StochasticStellarPopParticleVxIdx) = vx + vdisp_norm * (rx - 0.5) * cs;
```
chongchonghe: should this be `(2 * rx - 1)` ?

src/particles/particle_creation.hpp
```diff
@@ -502,89 +515,119 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
 						p.rdata(mass_at_birth_idx) = p.rdata(mass_idx);
 					}
 				}
 
-				if (num_particles > 1) { // Update momentum of the low mass star if there is(are) high mass star(s) in the cell
-					const int p_idx = 0;
-					auto &plow = particles[p_idx]; // NOLINT
-
+				// Update momentum of the low mass composite star particles if there is(are) high mass star(s)
+				if (num_high >= 1) {
```
chongchonghe: Recommend change:

```c++
if (num_high >= 1 || split_low_mass_composite)
```

this is because, when there are low-mass particle splitting, the random component of the velocity does not exactly cancel (with a typical residual of ~0.5 cs). This makes the COM velocity deviate from the target velocity (vx vy  vx). Therefore, we want to do this correct in this situation.

chongchonghe: when num_high = 0, `total_high_momx` will be 0, but this is fine.

chongchonghe (changes requested):

chongchonghe:
ParticleSF test failed, probably due to the validation is not valid anymore. I'll try to fix it.

chongchonghe:
> ParticleSF test failed, probably due to the validation is not valid anymore. I'll try to fix it.

Turned out to be a minor bug in your code. I fixed it and pushed. `low_mass_composite_max_mass` should default to max() of Real so that `std::ceil(mass_low_mass_star / low_mass_composite_max_mass_)` is 1 by default.

chongchonghe:
Both GPU tests passed.

src/particles/particle_creation.hpp
BenWibking: I don't think so. Why?

src/particles/particle_creation.hpp
BenWibking: That makes sense, I'll change it.

src/particles/particle_creation.hpp
BenWibking: I want the 3D velocity dispersion to be `0.5*cs`, so each component should span a range of `0.5*cs/sqrt(3)`.

src/particles/particle_creation.hpp
BenWibking: Fixed.

BenWibking:
Hmm, GitHub appears to be having problems today.

src/particles/particle_creation.hpp
BenWibking: Ok, not a Gaussian. Will fix.

The standard deviation of a uniform [0, 1] distribution is sqrt(1/12) ~ 0.29.

BenWibking: Use $v_{esc} = \sqrt{2 G M_{sum} / \Delta x}$. For 10^4 Msun and 5 pc, this is about 4 km/s. This will ensure the cluster of particles is unbound.

Note: we also need to do this when splitting particles on restart.

BenWibking: Fixed.

src/particles/particle_creation.hpp
```diff
@@ -502,89 +515,124 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
...
+
+							// Draw velocity from the power-law distribution with inverse transform sampling
+							double v_mag = amrex::Random(engine) * (vmax_pow - vmin_pow) + vmin_pow;
+							v_mag = std::pow(v_mag, 1. / (1. - beta)) * km_per_s; // Convert to km/s
+
+							// Sample cos theta from a uniform distribution between -1 to 1
+							double const cos_theta_random = 2. * amrex::Random(engine) - 1.;
+							double const sin_theta_random = std::sqrt(1. - cos_theta_random * cos_theta_random);
+							// Sample phi from a uniform distribution between 0 and 2*pi
+							double const phi_random = (1. - amrex::Random(engine)) * 2. * std::numbers::pi;
```
markkrumholz: Not a big deal, but why `(1. - amrex::Random(engine))` and not just `amrex::Random(engine)`?

BenWibking: I don't know. This was @chongchonghe 's original expression.

BenWibking: I can simplify it.

BenWibking: Fixed.

markkrumholz (approved):

BenWibking:
@chongchonghe Since you requested changes in an earlier version, it requires your re-review as well.

chongchonghe (approved):
Looks good. The new velocity dispersion looks correct.

src/particles/particle_creation.hpp
chongchonghe: hah, i didn't write the stochastic star formation part.

PR #1639, opened 2026-02-07
Set CFL to 0.3

chongchonghe:
### Description
Somehow, after the recent update of the SN module, RandomBlast regression test failed with 'Hydro update exceeded max_retries on level 1'. Changing `CFL` from 0.2 to 0.3 somehow fixed the problem. 

### Related issues

### Checklist

### DIFF
--- a/inputs/RandomBlastAMR_regression.in
+++ b/inputs/RandomBlastAMR_regression.in
@@ -26,7 +26,7 @@ problem.part_fn = "./particles_stochastic_n100.txt"
 do_reflux = 1
 do_subcycle = 0
 
-cfl = 0.2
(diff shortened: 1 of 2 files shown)


BenWibking:
Hmm, this is very odd, especially since the full disk galaxy simulations with feedback work fine. At what step does it crash? It may be that it just needs artificial viscosity turned on, or the dt_shrink parameter set.

BenWibking:
I'll approve this since it gets the testing back online.

Can you upload the logs from your run, so if we see this again, there is some context for debugging it?

BenWibking (approved):

PR #1643, opened 2026-02-09
Add `ignore_return` Parameter

chongchonghe:
## Summary

Added a new runtime parameter `ignore_return` that forces `main()` to return 0 regardless of the return code from `problem_main()`. This allows tests that are expected to fail validation checks to still pass in CI/CD pipelines.

## Changes

- **src/main.cpp**: Added logic to read `ignore_return` parameter and return 0 when enabled

## Usage

Add to any input file:
```
ignore_return = true
```

Or pass as command-line argument:
```
./TestName inputs/test.in ignore_return=true
```

### Checklist


src/main.cpp
```diff
@@ -75,6 +75,11 @@ auto main(int argc, char **argv) -> int
 
 	amrex::Real const start_time = amrex::ParallelDescriptor::second();
 
+	// Check if we should ignore the return code from problem_main
+	int ignore_return = 0;
```
BenWibking: Can you make this `bool` instead? ParmParse supports this now, so there's no good reason to use `int` instead of `bool` if we actually mean `bool`.

BenWibking (changes requested):

src/main.cpp
chongchonghe: done

BenWibking (approved):

PR #1652, opened 2026-02-13
Add velocity dispersion to split particles on restart

BenWibking:
### Description
Perturb the velocities of particles that are split on restart so that they have a velocity dispersion that makes them gravitationally unbound from each other.

### Related issues
N/A

### Checklist


BenWibking:
`BinaryOrbitCICSplit` failed. I think it's because this change does not conserve momentum.

BenWibking:
Ok, so the star particles are split on restart, which is not what we want:
```
Splitting CIC_particles using split_factor = 8
Splitting StochasticStellarPop_particles using split_factor = 8
```

BenWibking:
Still fails:
```
Number of particles: 16
max deviation from initial particle separation = 2.562781038 cell widths.
Test failed (split particles)
```
This works on my Mac laptop, so I'm not sure what's going on.

chongchonghe:
@BenWibking The same test failed on my Mac, and I'm pretty sure it will fail on any platform. The velocity dispersion is huge (I estimate of order 1000 km/s), so the splitted cluster of stars immediately evaporate. The test expects the splitted particles remain bound for the simulation period and check for largest separation in the end. that's why it failed.

src/particles/PhysicsParticles.hpp
```diff
@@ -421,46 +422,71 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
 						p_new.cpu() = cpu_id;
 						p_new.id() = pid + idx * splitFactor + idx_new;
-						p_new.pos(0) = x0 + dx[0] * amrex::Random(rng);
-						p_new.pos(1) = y0 + dx[1] * amrex::Random(rng);
-						p_new.pos(2) = z0 + dx[2] * amrex::Random(rng);
 						p_new.rdata(mass_idx) = old_mass / static_cast<amrex::Real>(splitFactor);
+
+						if (has_velocity_components) {
+							// Apply a component-wise uniform velocity kick with RMS set by cell-scale escape speed.
+							const amrex::Real v_esc = std::sqrt(2.0 * C::Gconst * old_mass / dx_cell);
```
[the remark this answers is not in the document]
chongchonghe: Right. We should add a safeguard to use this only in CGS units.

chongchonghe:
@BenWibking The BinaryOrbitCIC test problem is not ideal for testing your new particle splitting algorithm. The box size is very small, dx = 0.05 AU, so the velocity dispersion sqrt(2 G M / dx) is huge. We should either disable the answer check for the `BinaryOrbitCICSplit` test, or set the 'max_expected_distance' to the actually velocity dispersion (about 582 km/s) times simulation time, plus the initial separation.

chongchonghe:
> Still fails:
> 
> ```
> Number of particles: 16
> max deviation from initial particle separation = 2.562781038 cell widths.
> Test failed (split particles)
> ```
> 
> This works on my Mac laptop, so I'm not sure what's going on.

You are probably testing BinaryOrbitCIC.in, but this failed test is using BinaryOrbit_split.in

BenWibking:
> > Still fails:
> > ```
> > Number of particles: 16
> > max deviation from initial particle separation = 2.562781038 cell widths.
> > Test failed (split particles)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This works on my Mac laptop, so I'm not sure what's going on.
> 
> You are probably testing BinaryOrbitCIC.in, but this failed test is using BinaryOrbit_split.in



> > Still fails:
> > ```
> > Number of particles: 16
> > max deviation from initial particle separation = 2.562781038 cell widths.
> > Test failed (split particles)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This works on my Mac laptop, so I'm not sure what's going on.
> 
> You are probably testing BinaryOrbitCIC.in, but this failed test is using BinaryOrbit_split.in

~~No, I am testing with BinaryOrbit_split.in.~~

Ah, you are right.

BenWibking:
I will get to this on Monday my time.

chongchonghe:
This should fix the problem

src/problems/BinaryOrbitCIC/testBinaryOrbitCIC.cpp
```diff
@@ -126,16 +128,85 @@ template <> void QuokkaSimulation<BinaryOrbit>::computeAfterTimestep()
...
-				for (size_t i = 0; i < real_data.size(); ++i) {
-					for (size_t j = i + 1; j < real_data.size(); ++j) {
-						const auto &p1 = real_data[i];
-						const auto &p2 = real_data[j];
-						const double dx = p1[0] - p2[0]; // position x
-						const double dy = p1[1] - p2[1]; // position y
-						const double dz = p1[2] - p2[2]; // position z
-						const double pair_dist = std::sqrt((dx * dx) + (dy * dy) + (dz * dz));
-						dist = std::max(dist, pair_dist);
+				if (do_split_particles) {
```
chongchonghe: I recommend we remove this if clause and use my original 'max separation among all pairs' approach since the star cluster explode immediately and expands at close to the velocity dispersion we put in. Here is the test log after remove this if clause:

```
Number of particles: 16
Maximum particle mass: 2.5e+33
max deviation from initial particle separation = 13.19828324 cell widths.
    Expected maximum deviation = 15.66096231 cell widths.
Test passed
elapsed time: 1.20137 seconds.
AMReX (26.02-21-gbe8dfebe4bea) finalized
```

chongchonghe:
@BenWibking @markkrumholz I think we should use virial velocity instead of escape velocity. Due to the inaccuracy of the gravity integrator, the splitted particles immediately expands quickly at the first step after splitting, and this process is unresolved by the time integrator of our explicit scheme. That's why we are seeing the splitted particle cluster immediately exploded.

markkrumholz:
Is this just a factor of sqrt(2) reduction in the velocity dispersion? If so, I have no objections.

BenWibking:
> @BenWibking @markkrumholz I think we should use virial velocity instead of escape velocity. Due to the inaccuracy of the gravity integrator, the splitted particles immediately expands quickly at the first step after splitting, and this process is unresolved by the time integrator of our explicit scheme. That's why we are seeing the splitted particle cluster immediately exploded.

That makes sense to me.

src/particles/PhysicsParticles.hpp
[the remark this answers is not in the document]
BenWibking: Fixed.

src/particles/PhysicsParticles.hpp
```diff
@@ -421,46 +426,71 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
 						p_new.cpu() = cpu_id;
 						p_new.id() = pid + idx * splitFactor + idx_new;
-						p_new.pos(0) = x0 + dx[0] * amrex::Random(rng);
-						p_new.pos(1) = y0 + dx[1] * amrex::Random(rng);
-						p_new.pos(2) = z0 + dx[2] * amrex::Random(rng);
 						p_new.rdata(mass_idx) = old_mass / static_cast<amrex::Real>(splitFactor);
+
+						if (has_velocity_components) {
+							// Apply a component-wise uniform velocity kick with RMS set by cell-scale virial velocity.
+							const amrex::Real v_virial = std::sqrt(C::Gconst * old_mass / dx_cell);
```
BenWibking, line 471: Updated to use virial velocity.

src/particles/PhysicsParticles.hpp
```diff
@@ -407,6 +408,10 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
 
 	void splitParticles(int const lev, int const splitFactor) override
 	{
+		static_assert(Physics_Traits<problem_t>::UnitSystem == UnitSystem::CGS,
+			      "The current implementation of velocity kick in particle splitting assumes cgs units."
+			      "Please implement the appropriate scaling for other unit systems.");
+
```
BenWibking, line 414: Added static_assert to use CGS units.

BenWibking:
@chongchonghe @markkrumholz This is ready to be re-reviewed.

markkrumholz (changes requested):
I think there is a subtle issue with how the kick velocity is being selected: it induces a grid asymmetry. The current scheme is that it is drawn from a uniform square distribution, so the maximum possible velocity in grid-aligned directions is smaller than in directions at an angle to the grid. That is, for kicks that wind up in the x direction, the maximum possible kick is `vdisp_norm/2`, while for kicks that are at a 45 degree angle in x, y, and z, the maximum is larger by a factor of sqrt(3). This could cause subtle grid-imprinting effects. If we wanted to avoid this asymmetry but still draw from a bounded tophat distribution, an alternative implementation would be to draw the magnitude from a tophat distribution from -0.5 to 0.5 as is being done here, then choose the cosine of the polar angle mu from a uniform distribution in -1 to 1, and the azimuthal angle phi from a uniform distribution in [0, 2pi]. This would produce an isotropic kick distribution direction.

BenWibking:
Ok, that's a good point. ~~I can sample in $|v|$ with a top-hat and then sample on the sphere uniformly. I'll get to this tomorrow.~~

BenWibking:
It turns out it's more efficient to do rejection sampling, since it avoids transcendental functions. So we sample uniformly from the volume [-1,1]^3 and then reject samples outside the sphere of unit radius. Then we scale by the virial velocity.

BenWibking:
@markkrumholz This PR can be re-reviewed. I've implemented rejection sampling to get uniform sampling within the unit sphere, and then scaled by the virial velocity.

markkrumholz (dismissed):

BenWibking:
Do-while loops apparently offend clang-tidy:
```
/github/workspace/src/particles/PhysicsParticles.hpp:476:8: warning: avoid do-while loops [cppcoreguidelines-avoid-do-while]
  476 |                                                         do {
      |                                                         ^
```

If there are no objections, I am going to just suppress the warning with `// NOLINT`.

markkrumholz:
> Do-while loops apparently offend clang-tidy:
> 
> ```
> /github/workspace/src/particles/PhysicsParticles.hpp:476:8: warning: avoid do-while loops [cppcoreguidelines-avoid-do-while]
>   476 |                                                         do {
>       |                                                         ^
> ```
> 
> If there are no objections, I am going to just suppress the warning with `// NOLINT`.

Agreed, this is a silly objection.

markkrumholz (approved):

PR #1664, opened 2026-02-22
Add field_names and fc_dirs variable selection to DiagPlotfile

chongchonghe:
# Add Variable Selection to DiagPlotfile and Fix Initial Snapshot

Two improvements to the diagnostics system (`quokka.diagnostics`):

### Variable selection (`DiagPlotfile`)

- **`field_names`**: Selects which cell-centered variables (including cell-averaged face-centered quantities like `x-BField`) appear in the main plotfile. Empty (default) = all variables.
- **`include_fc_fields`**: Boolean (`true`/`false`) controlling whether `fc_vars/` is written. When MHD is enabled and `include_fc_fields = true` (default), all face-centered directions are written.

### Initial snapshot fix (`DiagBase`)

All diagnostics now write an initial snapshot at step 0 whenever `int` or `time_int` is positive. Previously, time-based diagnostics (`time_int`) skipped the initial state.

### Face-centered variables in slice plots (`DiagFramePlane`) — not implemented

Adding raw fc variable support to `DiagFramePlane` is non-trivial and left for future work. Three issues make it hard:

1. **Separate data path**: `DiagFramePlane` accesses data exclusively through `m_diagMF`, a cc MultiFab built by the `addVars`/`prepare` pipeline. FC MultiFabs have a different `ixType()` and are invisible to this pipeline; they would need to be fetched directly from the simulation via `PlotFileMF_fc()`.
2. **Interpolation stencil**: The current 3-point stencil assumes cell-centered indexing. Face-centered data shifts the index convention, and when the cut-plane normal matches the fc direction the plane falls exactly on a face, requiring different logic.
3. **2D writer assumes cc boxes**: `Write2DMultiLevelPlotfile` / `VisMF2D` / `write_2D_header` all assume cell-centered box types throughout.

**Workaround**: the cell-averaged BField values (`x-BField`, `y-BField`, `z-BField`) are already available as cc variables and can be listed in `field_names` today.

## Changes

- `src/io/DiagPlotfile.H`: Added `m_varNames` and `m_includeFcFields` member variables; updated `processDiag` template to filter cc output and toggle fc output
- `src/io/DiagPlotfile.cpp`: Added reading of `field_names` and `include_fc_fields` from input file in `init()`
- `src/io/DiagBase.cpp`: Fixed initial snapshot — all diagnostics now output at step 0 when any interval is active, matching the behaviour of `plotfile_interval`/`plottime_interval`
- `inputs/ParticleSinkFormation.in`: Added `field_names` and `include_fc_fields` to demonstrate the feature

## Example

```ini
quokka.plt.field_names = gasDensity x-GasMomentum y-GasMomentum z-GasMomentum temperature
quokka.plt.include_fc_fields = false   # suppress fc_vars output
```


BenWibking:
Specifying `field_names` seems like a very useful feature.

What is the use case for specifying `fc_dirs`? I don't immediately see when it would be useful to only output a subset of the face-centered directions.

chongchonghe:
> Specifying `field_names` seems like a very useful feature.
> 
> What is the use case for specifying `fc_dirs`? I don't immediately see when it would be useful to only output a subset of the face-centered directions.

This is equivalent to selecting variables for fc quantities. I agree it's not immediately useful to select a subset of directions, but the real application is to either leave it unspecified (default to include all), or set it to an empty string to turn it off. I think we should provide an option to turn off original fc variables and use only the cc MHD variables to save disk space, because I can think of scenarios where I don't need MHD variables at faces.

chongchonghe:
So I will change the design so that empty means turning off

BenWibking:
> So I will change the design so that empty means turning off

How about just a boolean variable then? If true, it includes the face-centered vars, if false, it doesn't.

I think it's not obvious that the user is allowed to set it to empty. So I think this is clearer and handles the use case you have in mind.

chongchonghe:
@BenWibking Ready

chongchonghe:
Both GPU tests passed. The failure is irrelevant.

BenWibking (approved):

PR #1715, opened 2026-03-03
add separate dust density floor and dust density check

mxchen39:
### Description
Currently, dust in QUOKKA uses the same density floor as the gas. This PR introduces a separate density floor for dust, controlled by the runtime parameter `dust.density_floor`, with a default value of 0.0.

This PR also adds dust density checks to `HydroSystem<problem_t>::isStateValid`. If a negative dust density is detected, it will trigger FOFC.

### Related issues

### Checklist

### DIFF
--- a/src/hydro/hydro_system.hpp
+++ b/src/hydro/hydro_system.hpp
@@ -1070,14 +1070,12 @@ void HydroSystem<problem_t>::EnforceLimits(amrex::Real const densityFloor, amrex
 		}
 
 		// Enforce dust density floor
+		amrex::Real const dust_floor = dustDensityFloor;
 		if constexpr (Physics_Traits<problem_t>::is_dust_enabled) {
 			for (int g = 0; g < Physics_Traits<problem_t>::nDustGroups; ++g) {
-				amrex::Real const dust_floor = dustDensityFloor;
-				for (int g = 0; g < Physics_Traits<problem_t>::nDustGroups; ++g) {
-					amrex::Real dust_rho = state[bx](i, j, k, dustDensity_index + g * numDustVars_);
-					if (dust_rho < dust_floor) {
-						state[bx](i, j, k, dustDensity_index + g * numDustVars_) = dust_floor;
-					}
+				amrex::Real dust_rho = state[bx](i, j, k, dustDensity_index + g * numDustVars_);
+				if (dust_rho < dust_floor) {
+					state[bx](i, j, k, dustDensity_index + g * numDustVars_) = dust_floor;
 				}
 			}
 		}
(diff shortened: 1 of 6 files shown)


chongchonghe:
I think we do need a dust density floor, because we use dust density to compute velocity, v=p_d/rho_d. If rho_d is zero, it will cause division-by-zero error. Close to zero values will cause stability issue.

FOFC also handles dust transport, right? So, you can do what gemini suggested: report invalid state when dust density goes below zero, and do FOFC correction. Although, it's worth discussing whether or not to do FOFC correction for dust independently of gas.

mxchen39:
I see. or we could keep a dust density floor, but define a separate parameter `dust.density_floor` that could be usually much smaller than gas density floor.

mxchen39:
which of the two options do you prefer? @chongchonghe

chongchonghe:
Since dust density does not change in the dust-gas coupling step and only changes in the transport step, it's probably fine and optimal to do flux correction for dust separately. This would require a separate `redoFlag` for dust update. However, since the correction step is cheap as it only applies to bad cells, it's probably not worthing doing it separately. So, the only change we need is to add check for dust density in `isStateValid`, like what gemini suggested.

chongchonghe:
I recommend doing both: adding a `dust.density_floor` that defaults to 0 and add dust density check to `isStateValid`

src/hydro/hydro_system.hpp
```diff
@@ -726,7 +726,14 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto HydroSystem<problem_t>::isStateValid(am
 {
 	// check if cons(i, j, k) is a valid state
 	const amrex::Real rho = cons(i, j, k, density_index);
-	bool const isDensityPositive = (rho > 0.);
+	bool isDensityPositive = (rho > 0.);
+
+	if constexpr (Physics_Traits<problem_t>::is_dust_enabled) {
+		for (int g = 0; g < Physics_Traits<problem_t>::nDustGroups; ++g) {
+			const amrex::Real dust_rho = cons(i, j, k, dustDensity_index + g * numDustVars_);
+			isDensityPositive = isDensityPositive && (dust_rho >= 0.);
```
chongchonghe: should be "dust_rho > 0." . rho=0 is not allowed because you need to derive velocity from momentum, v=p/m

chongchonghe (changes requested):
I left one comment about "dust_rho > 0."
Also, it's usually a good idea to update a test problem to test new features. You can modify one of the dust problems to use a positive value of `dust.density_floor`

chongchonghe (approved):

PR #1719, opened 2026-03-06
Implement photoionization and add tests

James471:
### Description
Implements photochemistry in quokka and changes the Microphysics hash to a compatible version. Also adds the one zone test and the stromgren sphere test problem

### Related issues
Depends on https://github.com/AMReX-Astro/Microphysics/pull/1975 

### Checklist


BenWibking:
Ok, something went wrong with a merge into this branch. Essentially the last several weeks of PRs have somehow been included here as new commits.

James471:
Is it fine now? I had performed a merge from dev to my branch for some tests to pass, since I had branched from a very old commit of dev, and after that format.h had been removed causing tests to fail. I have now rebased, so the commit history should look normal.

BenWibking:
> Is it fine now? I had performed a merge from dev to my branch for some tests to pass, since I had branched from a very old commit of dev, and after that format.h had been removed causing tests to fail. I have now rebased, so the commit history should look normal.

Yeah, it looks fine now.

extern/photoionization-julia/fraction.jl
BenWibking, line 1: I think it might make sense to move `photoionization-julia` to `extern/photoionization-julia`. Otherwise, as you've seen, the linters will try to run on it, and those are not configured for Julia code.

src/problems/OneZonePhotoionization/testOneZonePhotoionization.cpp
```diff
@@ -0,0 +1,378 @@
+//==============================================================================
```
chongchonghe, line 1: I would recommend renaming this folder to something like "PhotoionizationOneZone". RadStreaming looks like a stale name from the RadSteaming test you inherited from.

src/problems/RadStreamingPhotoionization/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+              ../../radiation/photochemistry.cpp ${photoionization_sources})
+
+# this will add #define PHOTOCHEMISTRY
+target_compile_definitions(${JOB_NAME} PUBLIC PHOTOCHEMISTRY)
+
+if(AMReX_GPU_BACKEND MATCHES "CUDA")
+  setup_target_for_cuda_compilation(${JOB_NAME})
+endif(AMReX_GPU_BACKEND MATCHES "CUDA")
+
+add_test(NAME RadStreamingPhotoionizationNoEnergy COMMAND ${JOB_NAME} ../inputs/PhotoionizationTests/StreamingNoEnergyChange.toml ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
```
chongchonghe: To follow the pattern of other test problems, can you make these changes?
1. renamed `RadStreamingPhotoionizationNoEnergy` to `PhotoionizationOneZone`. See my other comments about name change. We want at one test with the same name as the folder name. This makes the `quokka` CLI works smoothly. 
2. move `../inputs/PhotoionizationTests/StreamingNoEnergyChange.toml` to `../inputs/PhotoionizationOneZone.toml` 
3. update the name and toml file name of the second test as well.

src/problems/RadStreamingPhotoionization/testRadStreamingPhotoionization.cpp
```diff
@@ -0,0 +1,378 @@
...
+	static constexpr int numMassScalars = NumSpec;		     // number of mass scalars
+	static constexpr int numPassiveScalars = numMassScalars + 0; // number of passive scalars
+	static constexpr bool is_radiation_enabled = true;
+	static constexpr bool is_dust_enabled = false;
+	static constexpr int nDustGroups = 1; // number of dust groups
+	// face-centred
+	static constexpr bool is_mhd_enabled = false;
+	static constexpr int nGroups = 1; // number of radiation groups
+	static constexpr UnitSystem unit_system = UnitSystem::CGS;
+	static constexpr double boltzmann_constant = C::k_B;
```
chongchonghe: this four lines can be removed. with `unit_system = UnitSystem::CGS`, all constants are automatically set.

src/problems/StromgrenSphere/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+                ../../radiation/photochemistry.cpp ${photoionization_sources})
+
+  # this will add #define CHEMISTRY
+  target_compile_definitions(${JOB_NAME} PUBLIC PHOTOCHEMISTRY)
+
+  if(AMReX_GPU_BACKEND MATCHES "CUDA")
+    setup_target_for_cuda_compilation(${JOB_NAME})
+  endif(AMReX_GPU_BACKEND MATCHES "CUDA")
+
+  add_test(NAME StromgrenSphereNoRecombination COMMAND ${JOB_NAME} ../inputs/PhotoionizationTests/StromgrenNoRecombination.toml ${QuokkaTestParams} WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests)
```
chongchonghe: Same here.

chongchonghe (changes requested):

chongchonghe:
This looks really good in general! I made some quick comments and will continue to review tomorrow.

James471:
Thanks ChongChong! I'll make the changes suggested by you and Ben towards the end collectively once you have finished reviewing.

inputs/PhotoionizationTests/StromgrenNoRecombination.toml
```diff
@@ -0,0 +1,52 @@
+# *****************************************************************
+# Problem size and geometry
+# *****************************************************************
+geometry.prob_lo     =  0.0  0.0  0.0 
+geometry.prob_hi     =  5e13 5e13 5e13
+quokka.bc = reflecting reflecting reflecting
```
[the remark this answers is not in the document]
chongchonghe: Although this will pass the best because Quokka is taking this an old .in file, since we are starting to use the new input format toml, can you reformat this to toml syntax, e.g. `geometry.prob_lo = [0.0, 0.0, 0.0]` ? Just check other examples.

src/problems/RadStreamingPhotoionization/testRadStreamingPhotoionization.cpp
```diff
@@ -0,0 +1,378 @@
...
+	static constexpr double gravitational_constant = C::Gconst;
+	static constexpr double c_light = c;
+	static constexpr double radiation_constant = C::a_rad;
+};
+
+template <> struct RadSystem_Traits<PhotoionizationStreamingProblem> {
+	static constexpr double c_hat_over_c = chat / c;
+	static constexpr double Erad_floor = 0.0;
+	static constexpr int beta_order = 0;
+	AMREX_GPU_HOST_DEVICE static constexpr amrex::GpuArray<double, NumChemActiveRadGroups + 1> ChemActiveRadFreqBounds()
```
chongchonghe: I recommend a simpler name "chemical bands" and "thermal bands", so this could be something like `chemRadBoundaries`

chongchonghe:
@claude review

src/problems/RadStreamingPhotoionization/testRadStreamingPhotoionization.cpp
chongchonghe: Also, I don't think this `AMREX_GPU_HOST_DEVICE` macro is useful here. This macro is for functions, not arrays. The name `amrex::GpuArray` already suggests it's GPU-aware

src/problems/OneZonePhotoionization/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+# Species enums
+set(SPECIES_ENUM
+"e = 0,
+H,
+Hp"
+)
+
+# Atomic masses (in atomic mass units)
+set(AION
+"1.0,
```
chongchonghe, line 31: Why is the atomic weight of electron 1? It's plausible that this is not used in the network, but it's worth explaining why all atomic masses are 1 here.

src/problems/OneZonePhotoionization/testOneZonePhotoionization.cpp
```diff
@@ -0,0 +1,378 @@
...
+	std::string filename = "../src/problems/RadStreamingPhotoionization/photoionization-julia/";
+	if (energy_switch == 0) {
+		filename += "no_energy.csv";
+	} else if (energy_switch == 1) {
+		filename += "with_energy.csv";
+	} else {
+		amrex::Abort("Invalid energy_switch parameter");
+	}
+
+	std::ifstream file(filename);
```
chongchonghe, line 298: Here, for safety and easy of debugging, we can check for file existance, and if failed, warn the user that you should run this test in the `<quokka-root>tests/` folder.

src/problems/OneZonePhotoionization/testOneZonePhotoionization.cpp
```diff
@@ -0,0 +1,378 @@
...
+		amrex::Print() << "Energy error norm exceeds tolerance!" << '\n';
+		errors += 1;
+	}
+	if (errors > 0) {
+		status = 1;
+	}
+
+	// Cleanup and exit
+	amrex::Print() << "Finished." << '\n';
+	return status;
```
chongchonghe, line 435: Can you use matplotlib to make a figure of time evolution of some quantity, e.g. xHII v.s. time? See examples in testRadDust.cpp

src/radiation/photochemistry.hpp
```diff
@@ -0,0 +1,156 @@
...
+				quanta_energy = RadSystem<problem_t>::GetChemActiveRadiationGroupQuanta(nn);
+				photochemstate.rn[0 + MicrophysicsNumRadVarsPerGroup * nn] =
+				    state(i, j, k, RadSystem<problem_t>::radEnergy_index + Physics_NumVars::numRadVarsPerGroup * nn) / quanta_energy;
+				// // TODO(james471): Add check for isotropy
+				photochemstate.rn[1 + MicrophysicsNumRadVarsPerGroup * nn] = 1.0_rt;
+			}
+			photochemstate.rho = rho;
+			photochemstate.e = Eint / rho;
+
+			// dont do photochemistry in cells with densities below the minimum density specified
```
chongchonghe, line 72: do you intend to leave these block of code commented out? "dont do photochemistry in cells with densities below the minimum density" seems a useful choice. If you need to turn this on and off during test, it might be worth defining a static constexpr `skip_chemistry_below_rho_mean = False` on the top of the file.

src/radiation/photochemistry.hpp
```diff
@@ -0,0 +1,156 @@
...
+	// Start off by assuming a successful burn.
+	int photochem_burn_success = 1;
+
+	amrex::Gpu::Buffer<int> d_num_failed({0});
+	auto *p_num_failed = d_num_failed.data();
+
+	int num_failed = 0;
+
+	auto dt_stage = dt / stage;
+	auto energy_update_factor = stage / 1.0_rt;
```
chongchonghe: use `static_cast<Real>`

src/problems/StromgrenSphere/testStromgrenSphere.cpp
```diff
@@ -0,0 +1,555 @@
...
+		const amrex::Real dt = dt_target / static_cast<amrex::Real>(N);
+		amrex::Real R = R0;
+
+		for (int step = 0; step < N; ++step) {
+			const amrex::Real k1 = rhs(R);
+			const amrex::Real k2 = rhs(R + 0.5_rt * dt * k1);
+			const amrex::Real k3 = rhs(R + 0.5_rt * dt * k2);
+			const amrex::Real k4 = rhs(R + dt * k3);
+			R += (dt / 6.0_rt) * (k1 + 2.0_rt * k2 + 2.0_rt * k3 + k4);
+			if (R < 0.0_rt)
```
chongchonghe: Use `std::max` instead of `<` (fix available)clang-tidy

src/radiation/radiation_system.hpp
```diff
@@ -647,6 +652,22 @@ void RadSystem<problem_t>::ConservedToPrimitive(amrex::Array4<const amrex::Real>
...
+{
+	auto const freq_bounds = RadSystem_Traits<problem_t>::ChemActiveRadFreqBounds();
+	amrex::Real freq_low = freq_bounds[group_index];
+	amrex::Real freq_high = freq_bounds[group_index + 1];
+	return GetRadiationGroupQuanta(freq_low, freq_high);
+}
+#endif
+
+template <typename problem_t>
+AMREX_GPU_HOST_DEVICE auto RadSystem<problem_t>::GetRadiationGroupQuanta(amrex::Real freq_low, amrex::Real freq_high) -> amrex::Real
```
chongchonghe: This is a little over-abstraction. We don't need to abstract it into a function if it's used only once.

src/radiation/photochemistry.hpp
```diff
@@ -0,0 +1,156 @@
...
+		});
+
+#if defined(AMREX_USE_HIP)
+		amrex::Gpu::streamSynchronize(); // otherwise HIP may fail to allocate the necessary resources.
+#endif
+	}
+
+	num_failed = *(d_num_failed.copyToHost());
+
+	photochem_burn_success = !num_failed;
```
chongchonghe:
```suggestion
	photochem_burn_success = num_failed == 0;
```

src/problems/StromgrenSphere/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+      ${CMAKE_CURRENT_BINARY_DIR}/network_properties.H
+      @ONLY
+  )
+
+
+  add_executable(
+    ${JOB_NAME} testStromgrenSphere.cpp "${QuokkaSourcesNoEOS}"
+                ../../radiation/photochemistry.cpp ${photoionization_sources})
+
+  # this will add #define CHEMISTRY
```
chongchonghe: type? CHEMISTRY -> PHOTOCHEMISTRY

inputs/PhotoionizationTests/StreamingNoEnergyChange.toml
```diff
@@ -0,0 +1,49 @@
+# *****************************************************************
+# Problem size and geometry
+# *****************************************************************
+geometry.prob_lo     =  0.0  0.0  0.0 
+geometry.prob_hi     =  100000000000.0  100000000000.0  100000000000.0
+geometry.is_periodic =  1    1    1
```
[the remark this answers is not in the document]
chongchonghe: This and all other comments from gemini about TOML makes sense. Just need to following the example in any other .toml file to convert this .in format to .toml format.

chongchonghe (changes requested):
I have finished a review of this PR. It generally looks really good, and I think it will be ready to merge once you addressed all the comments and after Mark and/or Ben review. 

I left many comments, and all are all easy and quick fixes. One format change is required: since we don't do subfolders in the inputs/ folder, to make it consistent with the rest of the code, can you move the toml files in inputs/PhotoionizationTests to inputs/ ? Then, you should be able to use the scripts/bash/quokka script to do quick build and run, because the quokka script assumes a unified file structure. Also need to rewrite the .toml file into TOML format, as gemini pointed out.

BenWibking:
auggie review

BenWibking:
augment review

src/radiation/photochemistry.hpp
James471: Following what Piyush did with chemistry, I am just uncommenting the code out. It is a piece from his chemistry code and is suppose to never be activated unless the density limits are explicitly specified. Otherwise the limits are just the maximum and minimum representable real values.

src/networks/photoionization/actual_rhs.H
```diff
@@ -0,0 +1,187 @@
...
+{
+	return C::k_B * T * get_recombination_coefficient(T) * (0.684 - 0.0416 * std::log(T / 1.0e4));
+}
+
+AMREX_GPU_HOST_DEVICE AMREX_INLINE auto get_ion_ff_cooling_coefficient(Real T) -> Real { return 1.4e-27 * std::sqrt(T) + 1.0e-19 * std::exp(-118348.0 / T); }
+
+AMREX_GPU_HOST_DEVICE AMREX_INLINE auto get_ionization_fraction_scaling(Real n_HI, Real n_HII) -> Real
+{
+	// Calculate ionization fraction
+	Real x_ionization = n_HII / (n_HI + n_HII);
```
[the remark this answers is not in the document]
chongchonghe: is this comment resolved?

src/radiation/photochemistry.hpp
```diff
@@ -0,0 +1,156 @@
...
+	// Start off by assuming a successful burn.
+	int photochem_burn_success = 1;
+
+	amrex::Gpu::Buffer<int> d_num_failed({0});
+	auto *p_num_failed = d_num_failed.data();
+
+	int num_failed = 0;
+
+	auto dt_stage = dt / static_cast<Real>(stage);
+	auto energy_update_factor = static_cast<Real>(stage) / 1.0_rt;
```
chongchonghe: minor comment: you don't need `/ 1.0_rt` once you use static_cast. `Real` will be the float type you compile the code with.

src/networks/photoionization/actual_rhs.H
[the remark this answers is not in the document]
chongchonghe:
```suggestion
	if (assume_sharp_ionization_front == 1) {
		if (n_HII <= 0.0_rt) {
			return 1.0_rt;
		}
		return (n_HI + n_HII) / n_HII;
	}
	return 1.0_rt;
```

chongchonghe:
# Review: Quokka PR #1719 — Implement photoionization and add tests
# (including Microphysics PR #1975 — Feature/radiation-VODE)

Reviewer: Claude Sonnet 4.6 (automated)
Review type: Feature — SPEC → VERIFY → EDGE

---

## SPEC

**What the PR adds:**

- A new `photoionization` chemistry network in Microphysics (PR #1975) that extends the VODE integrator to evolve radiation number densities and fluxes alongside species abundances.
- A new `photochemistry` module in Quokka (`src/radiation/photochemistry.hpp`, `.cpp`) that calls the VODE burner inside the radiation subcycle loop, converting radiation energy ↔ photon number density, running the burn, then converting back.
- Two test problems: **OneZonePhotoionization** (ODE-only, no transport) and **StromgrenSphere** (3D radiation + photochemistry).
- A multi-gamma EOS (`EOS/photoionization/`) that handles electron, neutral hydrogen, and ionized hydrogen with their physical masses.
- A bug fix in `write_probin.py`: the `get_next_line` function was silently stripping the last character of lines that have no `#` comment (returned `line[:pos]` where `pos == -1`).

---

## VERIFY / EDGE

| Phase | What ran | Result |
|---|---|---|
| VERIFY – OneZonePhotoionizationNoEnergy | `OneZonePhotoionization inputs/OneZonePhotoionization.toml` | **PASS** — L1 error norms: species ~1e-12, photons ~1e-14, energy = 0 |
| VERIFY – OneZonePhotoionizationWithEnergy | same binary, `OneZonePhotoionizationWithEnergy.toml` | **PASS** — L1 error norms: species ~5e-9, photons ~3e-11, energy ~1e-8 |
| VERIFY – StromgrenSphere (no recombination) | `StromgrenSphere inputs/StromgrenNoRecombination.toml` | **PASS** — photon conservation error 3.5e-14 (tolerance 1e-6) |
| VERIFY – StromgrenSphere (T-independent recombination) | `StromgrenSphere inputs/StromgrenTempIndependentRecombination.toml` | **PASS** — ionization front radii within tolerance of analytical RK4 solution at all output times |
| EDGE – PrimordialChem regression | `PrimordialChem inputs/PrimordialChem.toml` | **PASS** — no regression from VODE tolerance changes |

---

## Feature Assessment

The design is sound. Radiation number densities and fluxes are threaded through `burn_t` behind `#ifdef PHOTOCHEMISTRY`, keeping the footprint zero for non-photochemistry networks. The photon flux variable is stored as a normalized quantity (initial value 1.0) and the post-burn fluxes are scaled component-wise; this is an unusual but internally consistent convention documented by a comment in `actual_rhs.H`. The StromgrenSphere test validates the full photon-transport + ionization-front expansion against an analytical RK4 reference solution, giving high confidence in the end-to-end implementation.

---

## Bugs Found and Fixed

### 1. `numerical_jacobian.H` (Microphysics) — three-way loop defect
**File:** `integration/utils/numerical_jacobian.H:82-88`

The radiation contribution to the Jacobian normalization factor has three bugs that together make the loop a dead no-op (and would cause out-of-bounds access if it ran):

```cpp
// Bug: upper bound NumRadEqs (=2) is less than loop start net_ienuc+1 (=5); loop never runs
for (int i = net_ienuc + 1; i <= NumRadEqs; i += MicrophysicsNumRadVarsPerGroup) {
    // Bug: state.rn[i-1] → rn[4] when i=5, but rn[] has only NumRadEqs=2 elements
    w = 1.0_rt / (rtol_rad_num * std::abs(state.rn[i-1]) + atol_rad_num);
    for (int j = i + 1; j <= i + MicrophysicsNumRadVarsPerGroup; j++) {
        // Bug: j reaches i+2=7 (beyond int_neqs=6) and rn[j-1] is again out of bounds
        w = 1.0_rt / (rtol_rad_flux * std::abs(state.rn[j-1]) + atol_rad_flux);
    }
}
```

Correct version:
```cpp
for (int i = net_ienuc + 1; i <= net_ienuc + NumRadEqs; i += MicrophysicsNumRadVarsPerGroup) {
    w = 1.0_rt / (rtol_rad_num * std::abs(state.rn[i - net_ienuc - 1]) + atol_rad_num);
    for (int j = i + 1; j <= i + MicrophysicsNumRadVarsPerGroup - 1; j++) {
        w = 1.0_rt / (rtol_rad_flux * std::abs(state.rn[j - net_ienuc - 1]) + atol_rad_flux);
    }
}
```

**Impact:** This function is only called when `integrator.jacobian = 1` (finite-difference Jacobian). All current photoionization test inputs use `integrator.jacobian = 0` (analytical Jacobian), so no test surfaces the bug. A TODO comment already flagged this code as needing work. A fix has been applied in branch `chong/micro-photo-fix` in `extern/Microphysics`.

---

## PR-Introduced Concerns

### 2. Global VODE default tolerance changes (Microphysics `integration/_parameters`)

The PR changes the default VODE tolerances for **all** Microphysics networks:

| Parameter | Old default | New default | Direction |
|---|---|---|---|
| `rtol_spec` | 1e-12 | 1e-10 | relaxed |
| `rtol_enuc` | 1e-6 | 1e-10 | tightened |
| `atol_spec` | 1e-8 | 1e-10 | tightened |
| `atol_enuc` | 1e-6 | **1e-25** | tightened 10¹⁹× |

The `atol_enuc = 1e-25` is extremely restrictive. The change was motivated by photoionization's need for tight absolute tolerances (energy densities are tiny in photochemistry units), but these are *global* defaults. PrimordialChem still passes with the new defaults, but the change could silently degrade performance or change convergence behavior for any code that relies on the defaults.

**Recommendation:** Move the tight photoionization tolerances to the photoionization-specific input files (they already set `integrator.atol_enuc = 1e-14` in the TOML inputs, overriding the default). Revert the global defaults to something closer to the pre-existing values, or add a comment explaining why the global change is safe for all networks.

### 3. `actual_rhs.H` — hardcoded VODE indices 5 and 6

```cpp
ydot(5) = -photoionization_term;         // photon number density
ydot(6) = -n_HI * c * sigma_photo * n_gamma_flux;  // photon flux
```

The indices 5 and 6 are correct only because `net_ienuc = NumSpec + 1 = 4` for this specific 3-species network. Using `ydot(net_ienuc + 1)` and `ydot(net_ienuc + 2)` (and the corresponding fix in `actual_jac`) would be more robust. The same hardcoded indices appear throughout the Jacobian entries. The code is correct today but will silently break if the network is ever extended with additional species.

### 4. Incomplete analytical Jacobian for energy equation

`jac(net_ienuc, net_ienuc) = 0.0_rt` — the Jacobian does not capture the feedback of internal energy on temperature, and therefore on all temperature-dependent rate coefficients (recombination, collisional ionization, heating, cooling). VODE will still converge using the approximate Jacobian but may require more function evaluations when energy changes drive significant temperature variation.

### 5. CMake configuration produces a harmless but misleading error

During `cmake`, the photoionization branch calls `write_network.py -s ""` because `networkfile` is commented out. This outputs:
```
network_param_file.py: ERROR: file .../src/networks/photoionization/.net does not exist
```
The warning is harmless because `configure_file` in the problem CMakeLists immediately overwrites `network_properties.H` with the correct quokka template. The warning is confusing and could mask real failures. Consider either setting `networkfile` to a valid dummy path or skipping the `write_network.py` call for the photoionization case.

---

None found.

---

## Minor Items

- `photochemistry.hpp` was missing a trailing newline (fixed in branch `chong/photo-fix` in quokka).
- `inputs/OneZonePhotoionization.toml` and `inputs/OneZonePhotoionizationWithEnergy.toml` were missing trailing newlines (fixed in `chong/photo-fix`).
- The `stage` parameter in `computePhotoChemistry` exists for future Strang-splitting support but `QuokkaSimulation.hpp` always passes `stage=1`, making the `energy_update_factor` / `dt_stage` scaling for `stage=2` dead code at present.
- `GetChemBandQuanta` uses the arithmetic mean of the band's frequency bounds, which is an approximation for wide ionization bands.

---

## Summary

All four test cases pass: two OneZonePhotoionization variants and two StromgrenSphere variants (no-recombination and T-independent recombination). PrimordialChem shows no regression. The implementation is well-structured. The most important item to address before merging is the global VODE tolerance change (concern #2) — in particular `atol_enuc = 1e-25` affects all Microphysics networks. The `numerical_jacobian.H` bug (Bug #1) has been fixed in `chong/micro-photo-fix`.

BenWibking:
Ok, I think this explains why the stochiometric invariants weren't preserved to machine precision. The Jacobian must be exact for that to happen -- you cannot use an approximate Jacobian if you want to preserve linear invariants. It must include a full chain-rule propagation of everything.

BenWibking:
@chongchonghe Can you review with Opus instead of Sonnet?

BenWibking:
Given the multiple Jacobian bugs, I think it would be better to generate the Jacobian using SymPy or with forward-mode autodiff using the `autodiff` library: https://autodiff.github.io/#forward-mode

James471:
> Given the multiple Jacobian bugs, I think it would be better to generate the Jacobian using SymPy or with forward-mode autodiff using the `autodiff` library: https://autodiff.github.io/#forward-mode

Thanks Ben! I am happy to do it, however, the Jacobain has already been fixed in the draft RSLA PR https://github.com/quokka-astro/quokka/pull/1892. The other bug is in the numerical jacobian calculation which I never finished implementing since we were not using it. ChongChong, Mark and I have discussed this and I will be fixing it.

chongchonghe (dismissed):

BenWibking:
This is what Codex says:
```
• The patch has runtime correctness issues for default photochemistry energy handling and reduced-speed radiation, and it registers a
  3D-only problem in 2D builds. These should be addressed before considering the change correct.

  Full review comments:

  - [P2] Avoid treating generated energy as final energy — /Users/benwibking/amrex_codes/quokka/src/radiation/photochemistry.hpp:112-
    112
    When a photochemistry input omits integrator.subtract_internal_energy = 0, Microphysics uses its default of returning only the
    generated energy after burner(). This eos_input_re call interprets that delta as the final specific energy, so the later dEint =
    photochemstate.e * rho - Eint subtracts the initial internal energy again (or zeroes it for energy_switch=0); force that parameter
    for photochemistry or reconstruct from the final temperature as the chemistry path does.

  - [P2] Use the reduced light speed in the chemistry RHS — /Users/benwibking/amrex_codes/quokka/src/networks/photoionization/
    actual_rhs.H:82-82
    For PHOTOCHEMISTRY problems that set RadSystem_Traits::c_hat_over_c < 1, computePhotoChemistry stores the reduced speed in
    state.c_hat, but the RHS and Jacobian hard-code C::c_light. Those runs will absorb photons and heat gas on a full-c timescale
    while radiation transport/source terms use the reduced speed, making ionization fronts too fast; use state.c_hat consistently here
    and in the Jacobian.

  - [P2] Restrict StromgrenSphere target to 3D — /Users/benwibking/amrex_codes/quokka/src/problems/StromgrenSphere/CMakeLists.txt:1-1
    With the documented 2d preset this block still adds and registers StromgrenSphere, but the problem implementation unconditionally
    uses z coordinates (prob_lo[2], dx[2]) and 3D source normalization/inputs. The 2D target is therefore invalid; guard it with
    AMReX_SPACEDIM EQUAL 3 unless the problem is made dimension-aware.
```

BenWibking (approved):

PR #1430, opened 2025-11-05
Rename all problem cpp files

chongchonghe:
### Description
Renamed problem cpp files to test${JOB_NAME}.cpp , where JOB_NAME is a unique identifier of each test problem and is the same as the directory name. Also defined a `quokka_add_problem` function in ProblemHelpers.cmake to unify most CMakeLists.txt files. 

### Related issues
Related to #1006 

### Checklist
_Before this pull request can be reviewed, all of these tasks should be completed. Denote completed tasks with an `x` inside the square brackets `[ ]` in the Markdown source below:_
- [x] I have added a description (see above).
- [x] I have added a link to any related issues (if applicable; see above).
- [x] I have read the [Contributing Guide](https://github.com/quokka-astro/quokka/blob/development/CONTRIBUTING.md).
- [x] I have added tests for any new physics that this PR adds to the code.
- [ ] *(For quokka-astro org members)* I have manually triggered the GPU tests with the magic comment `/azp run`.

PR #1721, opened 2026-03-07
[WIP] Standardise MHD test problems

AstroKriel:
### Description

WIP: this will be broken up into smaller PRs.

In this PR I standardise MHD tests and workflows:

- Renamed MHD problem directories, setup scripts, and input files to consistently use the `MHD` prefix (e.g., `AlfvenWaveLinear` to `MHDAlfvenWaveLinear`), matching the convention already used by `MHDBalsaraVortex`, `MHDBlast`, etc.
- Updated `CMakeLists` files to use the `quokka_add_problem` helper macro, and introduced a curated `MHD` test label set and a smaller `MHD-ASAN` subset for the address sanitiser workflow
- Fixed the MHD CI workflows to use `ctest -L` label filtering, replacing broken `canary -k` keyword filtering which was silently selecting the wrong tests
- Removed explicit boundary condition and periodicity setup from MHD problem files, as `QuokkaSimulation` now handles this internally
- Cleaned up MHD input files: removed redundant reconstruction order parameters, standardised convergence test interpolation orders, and disabled unnecessary test output
- Standardised file header comments across all MHD test scripts with correct `\file` names and `\brief` descriptions
- Added `MHDBlast` regression input and updated regression config

### Related issues

None.

### Checklist


AstroKriel:
@BenWibking and @chongchonghe, this PR is ready for review. Please refer to the PR description for changes made and why some tests fail.

BenWibking:
The tests run with the configuration you've specified in this PR. There is something that's not working right. It needs to be fixed before it can be reviewed.

AstroKriel:
> The tests run with the configuration you've specified in this PR. There is something that's not working right. It needs to be fixed before it can be reviewed.

The MHD tests are currently failing across all PRs (including #1717), not because of a bug in those PRs, but because the test discovery in the workflow is broken, and the fix cannot be applied without integrating this PR.

From what I have been able to gather, the MHD workflow currently discovers which executables to build by grepping CMakeLists.txt files for `set(JOB_NAME ...)`. This pattern no longer exists in any MHD CMakeLists.txt file, so none of the executables can be found, and so no tests run. This broke when PR #1430 introduced the `quokka_add_problem` macro (which doesn't use `set(JOB_NAME ...)`), but went unnoticed because MHD files weren't migrated to the new macro. This PR, amongst other things, migrates the MHD CMakeLists.txt files to remove the old pattern and use `quokka_add_problem`.

AstroKriel:
@BenWibking (and @chongchonghe), after digging into this some more, I think the MHD failures may be caused by a `canary-wm` version change rather than anything in these PRs. Version 26.3.5 was released on March 6 and the workflow installs it unpinned (`pip install canary-wm`). In the CI logs for this PR, canary reports "Generated 0 test specs from 1 generators" immediately after startup, so it finds the cmake build directory but can't read any tests from it. The build itself succeeds fine.

If this is right, then it would affect all canary-based workflows, not just the MHD ones, and it would explain why the problem appears across all recent PRs. My earlier comment about the MHD discovery grep was wrong... the old grep would have found the executables fine on development-based branches.

This workflow is a little foreign to me, though, so does this seem consistent with what you guys are seeing? Would pinning `canary-wm` to 25.10.7 be the right fix/test here?

BenWibking:
> @BenWibking (and @chongchonghe), after digging into this some more, I think the MHD failures may be caused by a `canary-wm` version change rather than anything in these PRs. Version 26.3.5 was released on March 6 and the workflow installs it unpinned (`pip install canary-wm`). In the CI logs for this PR, canary reports "Generated 0 test specs from 1 generators" immediately after startup, so it finds the cmake build directory but can't read any tests from it. The build itself succeeds fine.
> 
> If this is right, then it would affect all canary-based workflows, not just the MHD ones, and it would explain why the problem appears across all recent PRs. My earlier comment about the MHD discovery grep was wrong... the old grep would have found the executables fine on development-based branches.
> 
> This workflow is a little foreign to me, though, so does this seem consistent with what you guys are seeing? Would pinning `canary-wm` to 25.10.7 be the right fix/test here?

Yes, I think that's the right fix. A Canary update broke the CI before, and I should have pinned it then.

BenWibking:
CI issues fixed in https://github.com/quokka-astro/quokka/pull/1729.

BenWibking (changes requested):
This changes a large amount of code. Can I suggest that you *not* change the *.cpp files themselves? Whether or not `MHD` is included in the problem struct name does not matter for users in any way.

AstroKriel:
> This changes a large amount of code. Can I suggest that you _not_ change the *.cpp files themselves? Whether or not `MHD` is included in the problem struct name does not matter for users in any way.

The goal of this PR is to for full consistency across folder names, filenames, struct names, and input files. If someone sees `MHDCurrentSheet` in a test output or CMake file, they should be able to find `MHDCurrentSheet/testMHDCurrentSheet.cpp` and `MHDCurrentSheet.in` without guessing. Partial renaming would leave things in an inconsistent state that's harder to fix later on. This sets a clearer standard (already partially implemented) for how new tests should be structured going forwards. I understand its a lot of changes, but its a one-time cost in the name of clarity.

BenWibking:
> > This changes a large amount of code. Can I suggest that you _not_ change the *.cpp files themselves? Whether or not `MHD` is included in the problem struct name does not matter for users in any way.
> 
> The goal of this PR is to for full consistency across folder names, filenames, struct names, and input files. If someone sees `MHDCurrentSheet` in a test output or CMake file, they should be able to find `MHDCurrentSheet/testMHDCurrentSheet.cpp` and `MHDCurrentSheet.in` without guessing. Partial renaming would leave things in an inconsistent state that's harder to fix later on. This sets a clearer standard (already partially implemented) for how new tests should be structured going forwards. I understand its a lot of changes, but its a one-time cost in the name of clarity.

Changing the filenames is fine. What I meant to say was: can you avoid changing the *contents* of the *.cpp files?

AstroKriel:
> > > This changes a large amount of code. Can I suggest that you _not_ change the *.cpp files themselves? Whether or not `MHD` is included in the problem struct name does not matter for users in any way.
> > 
> > 
> > The goal of this PR is to for full consistency across folder names, filenames, struct names, and input files. If someone sees `MHDCurrentSheet` in a test output or CMake file, they should be able to find `MHDCurrentSheet/testMHDCurrentSheet.cpp` and `MHDCurrentSheet.in` without guessing. Partial renaming would leave things in an inconsistent state that's harder to fix later on. This sets a clearer standard (already partially implemented) for how new tests should be structured going forwards. I understand its a lot of changes, but its a one-time cost in the name of clarity.
> 
> Changing the filenames is fine. What I meant to say was: can you avoid changing the _contents_ of the *.cpp files?

Ah, I see -- sure. I'll revert the struct names. The dates and author attributions were wrong is some files. Are you happy for those changes to stay?

BenWibking:
Yeah, that's fine.

AstroKriel:
@chongchonghe and @BenWibking, the error message I mentioned in the telecom did appear; see line 217 of MHD-WaveTests: `Erroneous arithmetic operation`. Is it obvious to either of you why this might happen?

BenWibking:
> @chongchonghe and @BenWibking, the error message I mentioned in the telecom did appear; see line 217 of MHD-WaveTests: `Erroneous arithmetic operation`. Is it obvious to either of you why this might happen?

The most common reason to see this is if the program attempts to divide by zero. If a NAN is produced for some other reason (like sqrt of a negative number), this will also happen.

AstroKriel:
I have changed the default scheme to what we recommend for MHD: EMF reconstruction via `Quokka2026`, and EMF averaging via `Balsara2025`. This is a change that is needed anyway, but also avoids the issue fixed in #1748.

BenWibking:
Some of the test problems should be set to use the other schemes. Otherwise we can't verify that they work.

chongchonghe:
@AstroKriel Did #1748 fix the problem we identified earlier today? If it fixed the FPE error for all MHD tests, then you can use `FelkerStone2017` scheme in some of the tests (in this PR), and they will not fail. You'll have to merge that branch to this PR first.

AstroKriel:
> Some of the test problems should be set to use the other schemes. Otherwise we can't verify that they work.

I agree we should test different scheme combinations. I'm, however, not convinced that randomly assigning schemes across unrelated test problems verifies correctness. Dedicated tests on a problem with a known solution seems like a more reliable approach. That way, if a test fails, you can verify if the test is flawed, or one of the schemes degraded.

I've now added three variants of the linear Alfvén wave test, each explicitly specifying a different EMF scheme combination, so all three schemes are covered in the MHD wave test workflows on every PR.

AstroKriel:
IMO, if someone modifies a particular EMF scheme, then they are responsible for running all tests using that scheme, and making sure it passes before merging.

BenWibking:
> IMO, if someone modifies a particular EMF scheme, then they are responsible for running all tests using that scheme, and making sure it passes before merging.

I agree that is the ideal, but the point of the automated tests is to prevent accidents from happening. Therefore, we need at least one test for each scheme.

I also agree in principle that the scheme variations should be done for the same test problem. The problem in practice is the (already very long) test suite may take even longer to run. If the tests can still run exceeding the time limit, I have no objection to doing this.

AstroKriel:
FYI, I plan on breaking this PR up into smaller, bitesize changes, that I will push through before this PR goes through.

BenWibking:
Ok, this PR has sort of exploded into an enormously large set of changes.

I want to say that as a general rule, pure stylistic changes are not and cannot be a priority for Quokka. They do not affect the functionality, performance, correctness, or maintainability of the code, and they can absorb a disproprotionately large amount of maintainers' (often extremely limited!) review time because they often change a large number of lines of code in a way that tends to introduce new bugs.

Could you isolate any changes that affect performance, functionality, maintainability, or correctness in separate PRs? I will be happy to review those with high priority.

chongchonghe:
@AstroKriel To make your life easier, here is my honest suggestion. You could break this PR into multiple (2 or 3) smaller PRs. The first one should be immediately mergable. 

- PR1: The first PR should be pure stylish change -- no physics/numerics touched: renaming the problem names (prefix problem names with MHD, etc), updating regression tests to use the right name, adding _regression to regression toml files, etc. You can also update the CMakeLists.txt files to use the `quokka_add_problem` helper, but this may break some tests because it adds the 'FPE' trap that was previously missing. So, if your other PR fixes the FPE error in MHD convergence tests, then you can keep the `quokka_add_problem` change in this PR. Otherwise, remove those changes from this PR so that this PR has absolutely zero non-stylish change. You should also remove the changes related to `mhd.emf_compute_scheme` and `mhd.emf_averaging_scheme` . They should be in PR2. 

- PR2: change the default `emfComputingScheme_` and `emfAveragingScheme_` if that is what you want, and update the toml files to use the right schemes for each MHD tests. 

- PR3: update CMakeLists to use `quokka_add_problem` (which turns on FPE) if you didn't do that in PR1, and make sure no tests fail FPE. 

If you think I can do any help, just come to my office on Monday afternoon. I can help you finish PR1 and PR2 very quickly.

BenWibking:
I agree with ChongChong's suggestion.

In fact, I think PR3 is an absolutely critical priority. Bugs in the testing infrastructure should be really fixed ASAP, because if they aren't, it leads to an avalanche of bugs that have to be fixed later.

PR2 will require some discussion. It is a normal priority PR.

PR1 just renames things and is therefore a low priority.

PR #1732, opened 2026-03-11
Guard against FPE in convergence rate estimate

AstroKriel:
### Description
This PR fixes a bug that leads to the "Erroneous arithmetic operation" (SIGFPE) crash in the MHD convergence tests when run with canary. Two root causes are fixed here: (1) `src/util/richardson.hpp` computed `std::log(errors[i-1] / errors[i])` without guarding against zero, negative, or non-finite error norms -- a zero error norm causes an FE_DIVBYZERO trap; guards using `std::isfinite` and positivity checks are now applied before all log ratio computations, with "N/A" printed and the test marked failed in degenerate cases. (2) `src/problems/FastWaveConvergence/testFastWaveConvergence.cpp` unconditionally computed `epsilon` via division by `sinθ`, which is undefined when sinθ = 0; `epsilon` is now reassigned to a safe value inside the existing `sinθ < tiny` branch.

### Related issues
#1721

### Checklist

### DIFF
--- a/src/util/richardson.hpp
+++ b/src/util/richardson.hpp
@@ -109,6 +109,10 @@ template <typename Callable> auto run(const Parameters &params, Callable &&runTe
 	if (resolutions.size() >= 2) {
 		if (!std::isfinite(errors[0]) || !std::isfinite(errors.back()) || errors.back() <= 0.0 || errors[0] <= 0.0) {
 			amrex::Print() << "\nOverall convergence rate: N/A (non-positive or non-finite errors)\n";
+			amrex::Print() << "Problematic values:\n";
+			for (int i = 0; i < resolutions.size(); ++i) {
+				amrex::Print() << std::format("  nx={:4d}  error={:.6e}\n", resolutions[i], errors[i]);
+			}
 		} else {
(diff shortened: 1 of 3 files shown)


src/util/richardson.hpp
```diff
@@ -102,15 +107,19 @@ template <typename Callable> auto run(const Parameters &params, Callable &&runTe
 	}
 
 	if (resolutions.size() >= 2) {
-		double const overall_log_error_ratio = std::log(errors[0] / errors.back());
-		double const overall_log_dx_ratio = std::log(dx_values[0] / dx_values.back());
-		double const overall_rate = overall_log_error_ratio / overall_log_dx_ratio;
+		if (!std::isfinite(errors[0]) || !std::isfinite(errors.back()) || errors.back() <= 0.0 || errors[0] <= 0.0) {
+			amrex::Print() << "\nOverall convergence rate: N/A (non-positive or non-finite errors)\n";
```
BenWibking, line 111: It would be useful to print the actual values of all `errors[]` here for debugging.

BenWibking (approved):

PR #1733, opened 2026-03-11
Add git worktree support to `run-cuda-container.sh`

chongchonghe:
### Description

`scripts/bash/run-cuda-container.sh` now works correctly when run from git worktrees with different basenames (e.g. `quokka`, `quokka2`).

Two changes were made:

- **Worktree-aware root detection**: Replace the manual `.git` directory walk with `git rev-parse --show-toplevel`, which correctly identifies the root of the current worktree.
- **Per-worktree containers**: Container names now include `basename "$REPO_ROOT"` (e.g. `quokka-amd64-cuda-quokka2`), so each worktree gets its own container with the correct volume mount. Multiple worktrees can build independently without interfering with each other.

### Checklist

### DIFF
--- a/scripts/bash/run-cuda-container.sh
+++ b/scripts/bash/run-cuda-container.sh
@@ -55,7 +55,7 @@ TARGETS=""
 
 # Detect repository root
 # Use git to find the worktree root (works for both regular repos and git worktrees)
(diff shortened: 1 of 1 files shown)


BenWibking (changes requested):
I agree with the AI comments.

BenWibking (approved):

PR #1734, opened 2026-03-11
Add documentation for sink particles

chongchonghe:
### Description
Sink formation and accretion is missing from the documentation. This PR adds it. 

### Checklist

### DIFF
--- a/docs/markdown/particles.md
+++ b/docs/markdown/particles.md
@@ -88,7 +88,7 @@ r_{\text{acc}} / 2 & r_{\text{BH}} > r_{\text{acc}} / 2.
 \end{cases}
 $$
 
-The accretion rate deposited in each cell is $\dot{M} \, w_i / \sum_i w_i$. Optionally, a uniform kernel ($w = 1$ for all cells in the $(7 \Delta x)^3$ stencil) can be used for testing by setting `particles.sink_particle_use_uniform_kernel = 1`.
+The accretion rate deposited in each cell is $-\dot{M} \, w_i / \sum_i w_i$ (negative means accreting). Optionally, a uniform kernel ($w = 1$ for all cells in the $(7 \Delta x)^3$ stencil) can be used for testing by setting `particles.sink_particle_use_uniform_kernel = 1`.
 
 ### Accretion limiters
 
--- a/docs/markdown/particles.md
+++ b/docs/markdown/particles.md
@@ -43,7 +43,7 @@ The particle inherits the gas velocity of the parent cell. The cell density is t
 
 ### Practical considerations
 
-- The local maximum check uses a strict inequality ($\rho_{\text{cell}} > \rho_{\text{neighbor}}$), so ties do not prevent formation. In practice, exact density ties only occur in artificial initial conditions.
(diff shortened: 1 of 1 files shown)


docs/markdown/particles.md
```diff
@@ -1,5 +1,141 @@
...
+
+$$
+m_{\text{particle}} = (\rho - \rho_J) \, \Delta x^3
+$$
+
+The particle inherits the gas velocity of the parent cell. The cell density is then set to $\rho_J$, and all conserved quantities (momentum, energy, internal energy) are scaled by the factor $\rho_J / \rho$.
+
+### Practical considerations
+
+- The local maximum check uses a strict inequality ($\rho_{\text{neighbor}} > \rho_{\text{cell}}$), so ties do not prevent formation. In practice, exact density ties only occur in artificial initial conditions.
```
BenWibking: This wording seems to contradict that above, where it says "the cell has the highest density among all cells within a sphere".

docs/markdown/particles.md
```diff
@@ -1,5 +1,141 @@
...
+
+### Accretion rate
+
+The Bondi-Hoyle radius is
+
+$$
+r_{\text{BH}} = \frac{G M}{v_\infty^2 + c_{f,\infty}^2},
+$$
+
+where $M$ is the particle mass, $v_\infty$ is the gas velocity relative to the particle (mass-weighted mean over the accretion zone), and $c_{f,\infty}$ is the fast magnetosonic speed (mass-weighted mean over the accretion zone). For MHD, $c_f^2 = c_s^2 (1 + 2/\beta)$; for non-MHD, $c_f = c_s$.
```
BenWibking: The fast magnetosonic speed depends on the orientation of the magnetic field with respect to the velocity. This approximation does not, so I think it should not be called the fast magnetosonic speed.

BenWibking (changes requested):

docs/markdown/particles.md
chongchonghe: typo, should be ($\rho_{\text{cell}} > \rho_{\text{neighbor}}$)

docs/markdown/particles.md
chongchonghe: It's the upper bound (when the wave is perpendicular to the magnetic field), so it actually accounds for directions. I've updated the texts.

docs/markdown/particles.md
```diff
@@ -1,5 +1,141 @@
...
+
+$$
+m_{\text{particle}} = (\rho - \rho_J) \, \Delta x^3
+$$
+
+The particle inherits the gas velocity of the parent cell. The cell density is then set to $\rho_J$, and all conserved quantities (momentum, energy, internal energy) are scaled by the factor $\rho_J / \rho$.
+
+### Practical considerations
+
+- The local maximum check uses a strict inequality ($\rho_{\text{cell}} > \rho_{\text{neighbor}}$), so ties do not prevent formation. In practice, exact density ties only occur in artificial initial conditions.
```
BenWibking: Ok, now this seems contradictory. If a strict inequality is used, it would seem that ties would prevent sink formation. Is there something missing here?

chongchonghe: OK, it's actually `>=` instead of `>`. So, it will create multiple stars in neighbor cells when there is a tie. 
BTW, no local method can prevent this. Since this only occurs in artificial initial condition, we can ignore special cases.

BenWibking (approved):

PR #1746, opened 2026-03-18
Add angular momentum tracking to sink particles

chongchonghe:
### Description

- Add angular momentum components (`Lx`, `Ly`, `Lz`) to sink particle real data, bringing the total from 5 to 8 real components (`mass`, `vx`, `vy`, `vz`, `mdot`, `Lx`, `Ly`, `Lz`).
- During accretion, accumulate angular momentum from each gas cell: `ΔL = Δm × (r_cell × v_rel)`, where `r_cell` is the cell center position relative to the sink particle and `v_rel = v_gas - v_particle` (relative velocity for Galilean invariance).
- Wire `angMomIndex_` through `PhysicsParticleDescriptorBase` and `PhysicsParticleDescriptor` so any future particle type can opt in to angular momentum tracking.
- Validated with the `ParticleSink` test problem, which passes all 3 phases (density accuracy, Galilean invariance, and mass conservation), and prints per-particle angular momentum after evolution.

### Files Changed

- `src/particles/particle_types.hpp`: Added `Lx`, `Ly`, `Lz` to `SinkParticleRealIdx`; updated `SinkParticleRealComps` and units data.
- `src/particles/particle_accretion.hpp`: Added angular momentum accumulation in `UpdateParticleMassAndMomentumInBox`; threaded `ang_mom_index` through `UpdateParticleMassAndMomentum` and `applyAccretion`.
- `src/particles/PhysicsParticles.hpp`: Added `angMomIndex_` field and getter to the descriptor base class; updated constructors; passed `SinkParticleLxIdx` for sink particles.
- `src/problems/ParticleSink/testParticleSink.cpp`: Added Galilean invariance check for angular momentum (Phase 1 vs Phase 2, 3% tolerance, sorted by particle position); prints per-particle angular momentum after evolution.

### Related issues

None.

### Checklist

### DIFF
--- a/src/particles/particle_accretion.hpp
+++ b/src/particles/particle_accretion.hpp
@@ -575,8 +575,7 @@ void applyAccretion(ContainerType *container, amrex::MultiFab &state, amrex::Mul
 	ComputeScaleDown<problem_t>(state, state_accretion_rate, scale_down, geom, state_fc);
 
 	// Step 3: Update particle mass, momentum, accretion rate, and angular momentum
-	UpdateParticleMassAndMomentum<ContainerType, problem_t>(container, state, scale_down, state_fc, lev, mass_index, time, dt, mdot_index,
-								ang_mom_index);
+	UpdateParticleMassAndMomentum<ContainerType, problem_t>(container, state, scale_down, state_fc, lev, mass_index, time, dt, mdot_index, ang_mom_index);
 
 	// Step 4: Update the hydro state. We do this at last because the original state is needed for updating particles in step 3.
 	UpdateHydroState<problem_t>(state, state_accretion_rate);
--- a/src/particles/particle_accretion.hpp
+++ b/src/particles/particle_accretion.hpp
@@ -183,9 +183,9 @@ void ComputeAccretionRateInBox(const typename ContainerType::ParIterType &pti, c
 		for (int ii = ix - stencil_size; ii <= ix + stencil_size; ++ii) {
 			for (int jj = iy - stencil_size; jj <= iy + stencil_size; ++jj) {
 				for (int kk = iz - stencil_size; kk <= iz + stencil_size; ++kk) {
-					const double x = p.pos(0) - plo[0] - (ii + static_cast<amrex::Real>(0.5)) * dx[0];
-					const double y = p.pos(1) - plo[1] - (jj + static_cast<amrex::Real>(0.5)) * dx[1];
-					const double z = p.pos(2) - plo[2] - (kk + static_cast<amrex::Real>(0.5)) * dx[2];
+					const double x = plo[0] + (ii + static_cast<amrex::Real>(0.5)) * dx[0] - p.pos(0);
+					const double y = plo[1] + (jj + static_cast<amrex::Real>(0.5)) * dx[1] - p.pos(1);
+					const double z = plo[2] + (kk + static_cast<amrex::Real>(0.5)) * dx[2] - p.pos(2);
 					const double r_sqr = x * x + y * y + z * z;
 					double r_acc_sqr = stencil_size * stencil_size * dx_max * dx_max;
 					if (use_uniform_kernel) {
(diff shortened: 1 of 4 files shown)


src/particles/particle_accretion.hpp
```diff
@@ -412,13 +421,33 @@ void UpdateParticleMassAndMomentumInBox(const typename ContainerType::ParIterTyp
...
+					// relative to the unboosted frame, breaking exact Galilean invariance of L.
+					// A potential solution is to use the pre-drift (rewound) particle position when
+					// computing r_cell for angular momentum, e.g.:
+					//   const double x_rewind = (p.pos(0) - pvx * dt) - plo[0] - (ii + 0.5) * dx[0];
+					// This might restore Galilean invariance for all particles. However, this breaks
+					// the order of operations of the current scheme, so it is not necessarily the
+					// right thing to do.
+					accreted_ang_mom_x += accreted_mass_cell * (z * vy - y * vz);
+					accreted_ang_mom_y += accreted_mass_cell * (x * vz - z * vx);
+					accreted_ang_mom_z += accreted_mass_cell * (y * vx - x * vy);
```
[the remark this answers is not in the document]
chongchonghe: You are right.

src/particles/particle_accretion.hpp
```diff
@@ -412,13 +421,33 @@ void UpdateParticleMassAndMomentumInBox(const typename ContainerType::ParIterTyp
 					// M_dot_cell is negative, so we multiply it by -1 to get the accreted mass
 					const double accreted_mass_cell = -M_dot_cell * dt * scale_down_factor;
 					const double rho = local_state(ii, jj, kk, HydroSystem<problem_t>::density_index);
-					const double vx = local_state(ii, jj, kk, HydroSystem<problem_t>::x1Momentum_index) / rho;
-					const double vy = local_state(ii, jj, kk, HydroSystem<problem_t>::x2Momentum_index) / rho;
-					const double vz = local_state(ii, jj, kk, HydroSystem<problem_t>::x3Momentum_index) / rho;
+					const double vx = local_state(ii, jj, kk, HydroSystem<problem_t>::x1Momentum_index) / rho - pvx;
+					const double vy = local_state(ii, jj, kk, HydroSystem<problem_t>::x2Momentum_index) / rho - pvy;
+					const double vz = local_state(ii, jj, kk, HydroSystem<problem_t>::x3Momentum_index) / rho - pvz;
```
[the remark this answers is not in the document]
chongchonghe: fixed

src/particles/particle_accretion.hpp
```diff
@@ -412,13 +421,36 @@ void UpdateParticleMassAndMomentumInBox(const typename ContainerType::ParIterTyp
...
+					// the particle mover has already advanced the particle by v_par*dt before this
+					// accretion step is called. In a boosted frame, this shifts r_cell by -v_par*dt
+					// relative to the unboosted frame, breaking exact Galilean invariance of L.
+					// A potential solution is to use the pre-drift (rewound) particle position when
+					// computing r_cell for angular momentum, e.g.:
+					//   const double x_rewind = (p.pos(0) - pvx * dt) - plo[0] - (ii + 0.5) * dx[0];
+					// This might restore Galilean invariance for all particles. However, this breaks
+					// the order of operations of the current scheme, so it is not necessarily the
+					// right thing to do.
+					accreted_ang_mom_x += accreted_mass_cell * (z * vy_rel - y * vz_rel);
```
markkrumholz: At first I thought there was a sign error here, but now I think it's correct but that you've used a confusing convention, and it would be better to change your convention. The source of my confusion was that I expected you to write the specific angular momentum of the accreting material as r x v, the usual way of writing it. You write it as -r x v, because you defined r as the vector from the cell center position to the particle position, reversing the usual convention that r is the vector from the location of the object whose angular momentum you're computing. So I don't think this is wrong, but I do think that the double-inversion of the sign convention is likely to cause future readers as much confusion as it just caused me. I therefore think it would be better to rewrite this code by flipping the sign convention on the definitions of x, y, and z and then flipping the sign convention here, so that r has its usual meaning -- the vector pointing from the particle to the position of the material whose angular momentum is being computed. Note that this will not require changes elsewhere, because in the existing codebase you only ever use the magnitude of r, which does not depend on the sign convention.

src/particles/particle_accretion.hpp
```diff
@@ -433,13 +465,18 @@ void UpdateParticleMassAndMomentumInBox(const typename ContainerType::ParIterTyp
 		if (mdot_index >= 0) {
 			p.rdata(mdot_index) = accreted_mass / dt;
 		}
+		if (ang_mom_index >= 0) {
```
markkrumholz, line 470: It's a minor optimization, but you could move all of the code calculating the accreted angular momentum inside this if block, so that it doesn't get executed if angular momentum isn't being used.

markkrumholz (changes requested):
Mostly looks fine. I just left two minor comments inline, of which I regard the first -- about the sign convention you're using -- as the more important. The other is a suggestion but not critical. Please let me know once you've looked at these and I can re-review.

src/particles/particle_accretion.hpp
chongchonghe: Makes sense. Changed.

src/particles/particle_accretion.hpp
chongchonghe: OK, done.

src/particles/particle_accretion.hpp
chongchonghe: Done. I changed the definition of r everywhere anyway, just to make it consistent.

chongchonghe:
@markkrumholz I've made the changes you suggested. Ready for review.

markkrumholz (approved):

PR #1751, opened 2026-03-21
[WIP] Add Quokka CLI tools

BenWibking:
### Description

**Note: This is a proof-of-concept create for conceptual and user experience feedback. The implementation will be rewritten.**

This PR adds an initial `quokka` CLI for common local workflows. It introduces a Python entry point plus shell bootstrap/activation scripts, profile configuration via `quokka.toml`, and profile-aware commands for `build`/`run`/`test`/`regression`/`tidy`/`format`/`status`.

**Example usage:**
```bash
./scripts/bash/install-quokka-bootstrap.sh
source scripts/bash/quokka-activate.sh host-default
quokka list profiles
quokka list problems
quokka build <problem>
quokka run <problem> --input <input-file>
```

### Related issues
N/A

### Checklist


docs/markdown/quokka_cli_v1.md
```diff
@@ -0,0 +1,1072 @@
+# Quokka CLI v1 Design
+
+Date: 2026-03-21
```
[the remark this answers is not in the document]
BenWibking: This is wrong.

scripts/python/quokka_cli.py
chongchonghe, line 1: what is this for? is this python script necessary?

BenWibking: This script provides the implementation for all of the `quokka <cmd>` commands.

chongchonghe: I think this kind of verbose scripts is what we should try to avoid when we do agentic coding, which I assume is what is happening here. This 4000+ lines of Python code is impossible for a human to read, and by human, I mean, me. It seems to me that 80+% of the functionality of this script can be done with <200 lines of bash scripts, and it will be much more readable and easier to maintain.

BenWibking: I agree. It's just a proof of concept at this point. At this point, your feedback on the user experience would be useful. It's currently marked as a draft because I will rewrite everything before requesting a review.

BenWibking: The real implementation will be a stack of 5-10 PRs with functionality broken out into separate scripts.

chongchonghe:
I like your idea of adding a quokka CLI to PATH so that you can call it everywhere, but I would prefer a lightweight bash script that handles basic operations: config, build, run. It can be done with ~100 lines of bash scripts. You don't need 'listing targets' because you can do that with `ls -d src/problems/*`. `ctest targets` list is also super long and confusing for LLM. Also, LLM would prefer running the commands directly with `/path/to/exe init.toml` and pipe it to head or tail, instead of running with `ninja test` and reading from a log file that is deeply hidden. 

So, here is my design:

```
quokka config <preset> [--root <path>]
quokka build <preset> <problem> [--root <path>]
quokka run <preset> <problem> [--input <file>] [--fpe] [--root <path>] # default input file is <problem>.toml
quokka list <preset> [--root <path>] # run `ls -d /path/to/quokka/src/problems/*` or `cmake --build . --target help`
quokka target <preset> [--root <path>]
```

and 

```
Presets:
1d
3d
1d-debug
3d-debug
```

Implemented in #1671 with 210 lines of bash script.

BenWibking:
> I like your idea of adding a quokka CLI to PATH so that you can call it everywhere, but I would prefer a lightweight bash script that handles basic operations: config, build, run. It can be done with ~100 lines of bash scripts. You don't need 'listing targets' because you can do that with `ls -d src/problems/*`. `ctest targets` list is also super long and confusing for LLM. Also, LLM would prefer running the commands directly with `/path/to/exe init.toml` and pipe it to head or tail, instead of running with `ninja test` and reading from a log file that is deeply hidden.
> 
> So, here is my design:
> 
> ```
> quokka config <preset> [--root <path>]
> quokka build <preset> <problem> [--root <path>]
> quokka run <preset> <problem> [--input <file>] [--fpe] [--root <path>] # default input file is <problem>.toml
> quokka list <preset> [--root <path>] # run `ls -d /path/to/quokka/src/problems/*` or `cmake --build . --target help`
> quokka target <preset> [--root <path>]
> ```
> 
> and
> 
> ```
> Presets:
> 1d
> 3d
> 1d-debug
> 3d-debug
> ```
> 
> Implemented in #1671 with 210 lines of bash script.

Ok, I agree your implementation is better :)

There are a few features I still want to add to your set of Bash functions. Some of these could instead be independent Python scripts:
* Provide a `quokka list` function that filters out un-buildable problems. I think this is not important for LLMs but I think it is important for new users. There have been some complaints about this in the past, and I think making this as easy as possible in the user CLI is important. This can be done with `ls src/problems`, but we also need a way to easily filter out problems that can't be built in the current profile/preset without knowing CMake magic incantations. For instance, it should be possible to list only 1D problems if the current build is configured with `AMReX_SPACEDIM=1`.
* I think an alias for running the tests (i.e., `ctest`) is useful, just to provide an easy on-ramp for users who don't know anything about CMake or CTest. It is also potentially useful to provide a version of this that runs only problems that match a given regex (i.e., `ctest -R`). A good use case for this is to re-run all the MHD tests locally after modifying the MHD code.
* `quokka doctor` should diagnose problems with missing dependencies or Python packages (e.g., matplotlib). But that could be its own script.

chongchonghe:
@BenWibking Replied to your comments in my PR #1671

BenWibking:
Superceded by https://github.com/quokka-astro/quokka/pull/1671.

PR #1756, opened 2026-03-25
Add documentation for the indices of state components

chongchonghe:
### Description
Provide a comprehensive documentation of the state components. Useful for developers. 

### Related issues

### Checklist

### DIFF
--- a/docs/markdown/components.md
+++ b/docs/markdown/components.md
@@ -1,0 +1,188 @@
+# State Variable Component Indices
+
+Quokka stores all cell-centred (cc) conserved quantities in a single `MultiFab` and all face-centred (fc) quantities in a separate per-dimension `MultiFab` array. The component layout within these arrays is controlled by `Physics_Traits`, `Physics_NumVars`, and `Physics_Indices`.
+
+## Configuration structs
+
+### `Physics_NumVars`
+
+Defined in `src/physics_numVars.hpp`. Holds hard-coded variable counts per module.
+
+
+| Constant              | Value                                 | Meaning                                                                                             |
+| --------------------- | ------------------------------------- | --------------------------------------------------------------------------------------------------- |
+| `numHydroVars`        | 6                                     | Conserved hydro variables: density, 3 momenta, total energy, internal energy                        |
+| `numRadVarsPerGroup`  | 4                                     | Radiation variables per photon group: energy density, 3 flux components                             |
+| `numDustVarsPerGroup` | 4                                     | Dust variables per dust group: density, 3 momenta                                                   |
+| `numMHDVars_per_dim`  | 1                                     | Face-centred MHD variables per spatial dimension (the magnetic field component normal to that face) |
+| `numMHDVars_tot`      | `AMREX_SPACEDIM * numMHDVars_per_dim` | Total face-centred MHD variables across all dimensions                                              |
+
+
+### `Physics_Traits<problem_t>`
+
+Defined in `src/physics_info.hpp`. User-specialized per problem to enable/disable physics modules and set group counts.
+
+
+| Field                                                                           | Meaning                                                                             |
+| ------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------- |
+| `is_hydro_enabled`                                                              | Enable hydrodynamics                                                                |
+| `is_radiation_enabled`                                                          | Enable radiation transport                                                          |
+| `is_dust_enabled`                                                               | Enable dust dynamics                                                                |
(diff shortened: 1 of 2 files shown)


chongchonghe:
@BenWibking Most of the added documentation is pretty straightforward. Perhaps the only part that requires your check is on the passive scalar part. Can you confirm my understanding is right? `passive scalar = mass scalar + other passive (non-mass) scalar`, and the first `numMassScalars` components are mass scalar and the other  `numPassiveScalars - numMassScalars` components are non-mass scalars.

BenWibking:
> @BenWibking Most of the added documentation is pretty straightforward. Perhaps the only part that requires your check is on the passive scalar part. Can you confirm my understanding is right? `passive scalar = mass scalar + other passive (non-mass) scalar`, and the first `numMassScalars` components are mass scalar and the other `numPassiveScalars - numMassScalars` components are non-mass scalars.

I also need to double check...
I think the mass scalars were added for @psharda 's PopIII problem, and haven't yet been used for anything else.

markkrumholz:
> > @BenWibking Most of the added documentation is pretty straightforward. Perhaps the only part that requires your check is on the passive scalar part. Can you confirm my understanding is right? `passive scalar = mass scalar + other passive (non-mass) scalar`, and the first `numMassScalars` components are mass scalar and the other `numPassiveScalars - numMassScalars` components are non-mass scalars.
> 
> I also need to double check... I think the mass scalars were added for @psharda 's PopIII problem, and haven't yet been used for anything else.

I agree that mass scalars have thus far only been used by @psharda, but I think we'll use them any time we have chemistry. The conceptual difference between mass scalars and ordinary passive scalars is that ordinary passive scalars can have arbitrary values and need not obey any particular constraint equations, while mass scalars represent partial densities and thus the sum of the mass scalar must add up to the total density. They may also obey other constraints such as conservation of atomic number, which will be enforced in the fluxes between cells.

BenWibking:
> > > @BenWibking Most of the added documentation is pretty straightforward. Perhaps the only part that requires your check is on the passive scalar part. Can you confirm my understanding is right? `passive scalar = mass scalar + other passive (non-mass) scalar`, and the first `numMassScalars` components are mass scalar and the other `numPassiveScalars - numMassScalars` components are non-mass scalars.
> > 
> > 
> > I also need to double check... I think the mass scalars were added for @psharda 's PopIII problem, and haven't yet been used for anything else.
> 
> I agree that mass scalars have thus far only been used by @psharda, but I think we'll use them any time we have chemistry. The conceptual difference between mass scalars and ordinary passive scalars is that ordinary passive scalars can have arbitrary values and need not obey any particular constraint equations, while mass scalars represent partial densities and thus the sum of the mass scalar must add up to the total density.

This is my understanding of how they currently work, and I definitely think this is useful to have. A more important point I wanted to make is that we could change how the indexing works for them without breaking too many things. When chemistry is on, I think we currently assume that all the mass scalars represent all of the chemical species in the network and vice versa.

docs/markdown/components.md
```diff
@@ -0,0 +1,182 @@
...
+| 3     | `x3Momentum_index`     | z-momentum                                        |
+| 4     | `energy_index`         | Total energy density                              |
+| 5     | `internalEnergy_index` | Auxiliary internal energy (dual energy formalism) |
+
+
+### Passive scalar block (`numPassiveScalars` variables, starting at `pscalarFirstIndex = 6`)
+
+| Index            | Name            | Quantity                                       |
+|------------------|-----------------|------------------------------------------------|
+| 6 .. 6+Nms-1     | `scalar0_index` | Mass scalars (advected with density)           |
```
BenWibking: I think "advected with density" is somewhat ambiguous. I think it would be clearer to say that the mass scalar fluxes are renormalized so that their sum is equal to the gas density flux.

docs/markdown/components.md
```diff
@@ -0,0 +1,182 @@
...
+| 6+Nms .. 6+Nps-1 |                 | Other passive scalars (advected independently) |
+
+where `Nms = numMassScalars` and `Nps = numPassiveScalars`. Mass scalars are the first `numMassScalars` entries of the passive scalar block; they are advected proportional to the density field. When `numMassScalars = 0`, the first component of the passive scalar block is 'other passive scalars'.
```
BenWibking: I would also adjust the wording here.

BenWibking (commented):
I think the description of the indexing is correct, but I would adjust the wording to describe how the mass scalars are advected.

chongchonghe:
@BenWibking @markkrumholz OK, done. Ready for review again.

markkrumholz (approved):

PR #1759, opened 2026-03-27
Make nghost counts compile-time constants in AMRSimulation

chongchonghe:
### Motivation
- Make the number of ghost cells a compile-time constant so it can be used in constant expressions and to avoid runtime initialization issues for template-based physics traits.

### Description
- Change `nghost_cc_` and `nghost_fc_` from non-const `int` members to `static constexpr int` members with `nghost_cc_ = Physics_Traits<problem_t>::is_mhd_enabled ? 7 : 6` and `nghost_fc_ = nghost_cc_`.

### Testing
- Project build (CI) and the existing unit/integration test suite were run and succeeded without failures.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c11859b850832796bb4dbce6e62572)

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -549,8 +549,8 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 
 	// Nghost = number of ghost cells for each array
 	// For our new scheme MHD-scheme, we need 7 ghosts for MHD (4 base + 3 for EMF) or 6 otherwise
-	static constexpr int nghost_cc_ = Physics_Traits<problem_t>::is_mhd_enabled ? 7 : 6;
-	static constexpr int nghost_fc_ = nghost_cc_;
+	const int nghost_cc_ = Physics_Traits<problem_t>::is_mhd_enabled ? 7 : 6;
+	const int nghost_fc_ = nghost_cc_;
 
 	amrex::Vector<std::string> componentNames_cc_;
 	amrex::Vector<std::string> componentNames_fc_flat_;
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -549,8 +549,8 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 
 	// Nghost = number of ghost cells for each array
 	// For our new scheme MHD-scheme, we need 7 ghosts for MHD (4 base + 3 for EMF) or 6 otherwise
-	int nghost_cc_ = Physics_Traits<problem_t>::is_mhd_enabled ? 7 : 6;
-	int nghost_fc_ = nghost_cc_;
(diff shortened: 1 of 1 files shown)


chongchonghe:
@claude review

chongchonghe:
@claude review

BenWibking:
The one downside of making it a compile-time constant is that we can no longer add the option to change the number of ghosts depending on the EMF reconstruction method.

chongchonghe:
> The one downside of making it a compile-time constant is that we can no longer add the option to change the number of ghosts depending on the EMF reconstruction method.

OK, but currently these variables are not modified anywhere. It really has to be `const` to be consistent with the rest of the code. Making them 'constexpr' doesn't have any real benefits anyway (except that VSCode would show its value when you hover over it).

BenWibking:
> > The one downside of making it a compile-time constant is that we can no longer add the option to change the number of ghosts depending on the EMF reconstruction method.
> 
> OK, but currently these variables are not modified anywhere. It really has to be `const` to be consistent with the rest of the code. Making them 'constexpr' doesn't have any real benefits anyway (except that VSCode would show its value when you hover over it).

Ok, but that would prevent making this a runtime parameter. `const` seems to be the worst option. You get neither the compile-time optimizations of `constexpr` nor the runtime flexibility of non-const.

chongchonghe:
> > > The one downside of making it a compile-time constant is that we can no longer add the option to change the number of ghosts depending on the EMF reconstruction method.
> > 
> > 
> > OK, but currently these variables are not modified anywhere. It really has to be `const` to be consistent with the rest of the code. Making them 'constexpr' doesn't have any real benefits anyway (except that VSCode would show its value when you hover over it).
> 
> Ok, but that would prevent making this a runtime parameter. `const` seems to be the worst option. You get neither the compile-time optimizations of `constexpr` nor the runtime flexibility of non-const.

I think you are overthinking here. The reason I am suggesting this change is simple: `nghost_cc_` and `nghost_fc_` are hard-coded constants in the current development branch, so they should be 'const' or 'constexpr'. If you run a quality check in the code, you'll see lines 552-553 will fail. If in the future, you want to add the ability to set `nghost_cc_` at runtime, you should declare it as a public member of `AMRSimulation` and call ParmParse on it. But before that, they should be const/constexpr.

chongchonghe:
Sorry, never mind. I ignored the fact that `nghost_cc_` is const-qualified. The value is actually effectively constant because it's initialized from a compile-time trait and never modified. 

I'll close this PR. However, my other comments still make sense. We can make it constexpr if we won't ever change it, or make it change with EMF reconstruction. In the latter case, you have to move it to a public member.

PR #1761, opened 2026-03-28
[ADR] Centralized conserved-to-primitive conversions

BenWibking:
### Description
This ADR describes a proposal for centralizing conserved-to-primitive conversions (and vice versa). This will be needed for both cosmology and for the Boris correction in MHD.

### Related issues
N/A

### Checklist

### DIFF
--- a/docs/markdown/adrs/ADR-0001-centralize-state-conversions.md
+++ b/docs/markdown/adrs/ADR-0001-centralize-state-conversions.md
@@ -1,0 +1,181 @@
+# ADR-0001: Centralize Conserved-To-Primitive And Primitive-To-Conserved State Conversions
+Date: 2026-03-28 • Status: Proposed
+
+## Context
+
+Quokka currently assumes a mostly uniform interpretation of the stored hydro state:
+
+- `rho`
+- `rho v`
+- `rho E`
+- `rho e`
+
+That assumption is embedded in multiple places across the codebase. Some modules already use shared helpers such as `ConservedToPrimitive`, but other paths directly interpret conserved components by
+
+- dividing momentum by density to recover velocity,
+- subtracting kinetic energy from total energy to recover internal energy,
+- computing temperature or pressure from locally reconstructed quantities,
+- applying source terms directly to conserved components.
+
+This is acceptable while the stored state is always the physical conserved state, but it becomes unsafe if Quokka adopts a different storage convention such as cosmological comoving variables:
+
+- `rho`
+- `a rho v`
+- `a^2 rho E`
+- `a^2 rho e`
+
+In that case, any module that directly interprets conserved components as physical variables will silently become inconsistent.
+
+This issue affects multiple subsystems:
+
+- hydro reconstruction and flux evaluation,
+- dual-energy bookkeeping,
+- floors and validity checks,
+- MHD state interpretation and any Boris-corrected momentum-to-velocity transforms,
+- source modules such as cooling and chemistry,
+- gravity coupling,
+- diagnostics and derived variables,
+- future cosmology support.
+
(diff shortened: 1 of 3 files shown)


markkrumholz:
This seems like a useful proposal, and I agree that centralizing the conversions would be useful. The only minor issue I see is that the function naming conventions should be symmetric -- either `ConservedToPrimitive` and `PrimitiveToConserved` or `cons2prim` and `prim2cons` -- rather than mixed as proposed in this document. We could even unify the two into a single templated function using something like

`enum ConversionDirection { cons2prim, prim2cons };`
`template <> ConvertState<cons2prim>(`
`     amrex::MultiFab const &input_cc_mf,`
`     std::array<amrex::MultiFab, AMREX_SPACEDIM> const &input_fc_mf,`
`     amrex::MultiFab &outVar_cc_mf, `
`     std::array<amrex::MultiFab, AMREX_SPACEDIM> &outVar_fc_mf,`
`     int nghost)`
`{ ... }`
`template <> ConvertState<prim2cons><prim2cons>(`
`     amrex::MultiFab const &input_cc_mf,`
`     std::array<amrex::MultiFab, AMREX_SPACEDIM> const &input_fc_mf,`
`     amrex::MultiFab &outVar_cc_mf, `
`     std::array<amrex::MultiFab, AMREX_SPACEDIM> &outVar_fc_mf,`
`     int nghost)`
`{ ... }`

docs/markdown/adrs/ADR-0001-centralize-state-conversions.md
```diff
@@ -0,0 +1,181 @@
...
+
+- No restart-format or plotfile-format changes are required by this ADR alone.
+- Problem modules and source modules may need updates if they currently infer physical state directly from conserved components.
+
+## Rollout & Testing
+
+### Rollout plan
+
+1. Identify the canonical conversion API surface.
+2. Implement or formalize the inverse `prim2cons` mapping alongside `ConservedToPrimitive`.
```
BenWibking, line 157: The naming should be symmetric. I will update this.

BenWibking:
> This seems like a useful proposal, and I agree that centralizing the conversions would be useful. The only minor issue I see is that the function naming conventions should be symmetric -- either `ConservedToPrimitive` and `PrimitiveToConserved` or `cons2prim` and `prim2cons` -- rather than mixed as proposed in this document. We could even unify the two into a single templated function using something like
> 
> `enum ConversionDirection { cons2prim, prim2cons };` `template <> ConvertState<cons2prim>(` ` amrex::MultiFab const &input_cc_mf,` ` std::array<amrex::MultiFab, AMREX_SPACEDIM> const &input_fc_mf,` ` amrex::MultiFab &outVar_cc_mf,` ` std::array<amrex::MultiFab, AMREX_SPACEDIM> &outVar_fc_mf,` ` int nghost)` `{ ... }` `template <> ConvertState<prim2cons><prim2cons>(` ` amrex::MultiFab const &input_cc_mf,` ` std::array<amrex::MultiFab, AMREX_SPACEDIM> const &input_fc_mf,` ` amrex::MultiFab &outVar_cc_mf,` ` std::array<amrex::MultiFab, AMREX_SPACEDIM> &outVar_fc_mf,` ` int nghost)` `{ ... }`

I agree the naming should be symmetric. I've added a note and will update that.

However, I'm not totally sure what we gain from templating this. Is the idea to guarantee that the function arguments are identical between the two functions at compile-time?

I think we want to have the core version of these two functions operator on individual cells, since I think we don't want to convert a whole multifab or fab for memory usage reasons in many cases. We could require converting a whole fab at a time, but I think that would not be worth the memory cost.

PR #1767, opened 2026-03-30
Add SPH-like kernel function for radiation deposition

chongchonghe:
Replaces the linear (CIC) particle-to-mesh interpolator for radiation deposition with an SPH-like Wendland C2 kernel, using a support radius of N=2 cells.

The key advantages over CIC are: (1) the spherical kernel avoids the quadrilateral artifacts produced by CIC's axis-aligned stencil, and (2) the wider support smooths deposition over more cells. The key advantage over a Gaussian kernel is **GPU performance**: the Wendland C2 kernel is a polynomial and requires no `std::exp` call, which is expensive on GPUs. The small stencil (N=2, 5³ bounding box) limits the per-deposition cost.

### New: `ParticleInterpolator::WendlandC2<N>` (`src/particles/particle_deposition.hpp`)

A standalone SPH-like interpolator (does not inherit from AMReX `Base` — spherical cutoff breaks separability):

- Template parameter `N` (default 2) controls the support radius: kernel weight is non-zero only within a sphere of radius `N·dx` centered on the particle
- Uses the Wendland C2 kernel from `src/math/quadrature.hpp`: `W(r) = (21/2π)(1-r)⁴(4r+1)` for `r ∈ [0,1]` — a pure polynomial, no transcendental functions
- Weights are re-normalised over the discrete stencil (sum = 1), ensuring strict energy conservation
- `static_assert` upper bound raised from N≤6 to N≤7 (hard limit of 8 ghost cells, independent of MHD method)

### Updated: `RadDeposition` (`src/particles/particle_deposition.hpp`)

Switched from `ParticleInterpolator::Linear` to `ParticleInterpolator::WendlandC2<>` (N=2).

### Updated: `QuokkaSimulation` (`src/QuokkaSimulation.hpp`)

Reduced `radEnergySource` ghost cell count from 4 to 3 (matching N=2: stencil radius + 1 for possible particle drift).

## Validation

All three `ParticleRadiation` test variants pass with relative energy conservation errors of ~10⁻¹⁶, well within the 10⁻¹³ tolerance.

## Checklist


chongchonghe:
@claude review

src/particles/particle_deposition.hpp
```diff
@@ -47,6 +47,102 @@ struct NearestEight : public Base<NearestEight, amrex::Real> {
...
+ *  The kernel weight at distance r (in units of dx) is
+ *      W(r) = exp(-r^2 / (2 sigma^2))   for r <= N
+ *           = 0                           for r >  N
+ *  and the weights are normalised so that their sum over the sphere equals 1.
+ *
+ *  This struct does NOT inherit from amrex::ParticleInterpolator::Base because
+ *  the spherical cutoff breaks the separability assumed by Base::ParticleToMesh.
+ */
+template <int N = 3> struct Gaussian {
+	static_assert(N >= 1 && N <= 6, "N must be between 1 and 6 (limited by ghost cells)");
```
markkrumholz, line 68: Number of ghost cells is not necessarily 6 -- this depends on the physics implemented, and is different for different MHD methods.

markkrumholz (commented):
I am a bit concerned that this design is not efficient at scale, and would become a real bottleneck if we attempted to use it in a real problem with large numbers of particles. The issue is that every time we want to do a deposition, for every particle at every time step, we wind up looping over (2N)^3 cells to calculate `inv_norm`. Maybe this doesn't matter because the cost of deposition is negligible even with this extra work, but it is a big increase in work, from 8 cells for CIC, each doing a linear operation, to 216 * 2 exponentials (times 2 for once to compute the normalization, once to compute the actual deposition) for this scheme with N = 3. I worry this becomes prohibitive well before number of particles is approaching number of cells.

I can see two options to address this:

1. Change the PR so that the deposition method is selected at run time, so we have the ability to switch between Gaussian and CIC deposition; use CIC when number of particles is large, and Gaussian only when it is small and thus the performance impact is limited.
2. Pre-compute a table of properly-normalised cell weights based on particle position within the cell, and interpolate on that. That is, we would have a table that would accept as a parameter values of the three components of frac, and would contain pre-tabulated values of `wt` in every cell for a given value of frac. The table would need to be four-dimensional, since it would depend on the values of the three components of frac and sigma, but could be quite low resolution.

BenWibking:
I agree. I think the number of exponential calls is not viable. I think the size of the stencil might actually be okay, if there were no special functions required per cell at all.

Since this is not exactly a Gaussian, since it is truncated, and moreover does not need to be any specific function, I think defining the deposition kernel to be that given by Gaussian-like fastexp function is actually perfectly fine here. That should be fast, as intended, and not require any recomputed tables.

BenWibking:
My other comment is, as I've mentioned before, that there is no need to restrict the particle deposition to the number of ghost cells used by the hydro update. You can just create a new multifab with the number of ghosts needed for particle deposition and use that.

chongchonghe:
> My other comment is, as I've mentioned before, that there is no need to restrict the particle deposition to the number of ghost cells used by the hydro update. You can just create a new multifab with the number of ghosts needed for particle deposition and use that.

This is interesting. I don't think I've heard you said this before; maybe I forgot. We don't need more ghost cells for radiation deposition; 3 is more than enough to get a spherical shape. The `static_assert` is still usable for confine the kernel size to a reasonable range. 

You are right. Since we are using a standalone multifab for particle deposition, we can use any number of ghost cells we want. This is very helpful for the stellar wind and outflow prescription. That means we can use as many as 8 ghost cells for particle deposition (if we go above 8, the blocking factor has to be 16 or higher). We can also do supercycling in stellar feedback, ejecting feedback every N steps instead of every step, to reduce communication cost, if communication becomes a bottleneck.

BenWibking:
Actually, I think it would be preferable to use an SPH-like kernel for this, instead of a truncated Gaussian or Gaussian-like function. These have the property that they are of compact support (so no artificial truncation is needed) and they are purely algebraic (so no special functions are needed to compute them).

There is one such kernel already in Quokka. I originally used it for SN deposition in the RandomBlast problem: https://github.com/quokka-astro/quokka/blob/3fd274ff39a8637414c0e23e510e8272838ed8b3/src/math/quadrature.hpp#L13

markkrumholz:
I think @BenWibking is right -- there's no reason to prefer a true Gaussian to an SPH-like kernel that is designed to be approximately Gaussian, but much cheaper to compute.

markkrumholz:
@BenWibking, just closing the loop on this after @chongchonghe had an in-person discussion of how to proceed with this PR. We're going to try going to an SPH-like deposition but with a kernel size of 2 instead of 3; this will limit the number of cells where the kernel needs to be evaluated to 4^3 instead of 6^3, which is a much smaller number and hopefully will be reasonable in terms of speed, while still being broad enough to avoid quadrilateral artifacts.

src/particles/particle_deposition.hpp
```diff
@@ -49,6 +50,100 @@ struct NearestEight : public Base<NearestEight, amrex::Real> {
...
+	amrex::Real frac[3]{};	// NOLINT fractional cell position per dimension
+	amrex::Real inv_norm{}; // NOLINT 1 / (sum of weights within sphere)
+
+	template <typename P>
+	AMREX_GPU_DEVICE AMREX_FORCE_INLINE WendlandC2(const P &p, amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &plo, // NOLINT
+						       amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const &dxi)
+	{
+		for (int i = 0; i < AMREX_SPACEDIM; ++i) {
+			const amrex::Real l = (p.pos(i) - plo[i]) * dxi[i] + 0.5;
+			index[i] = static_cast<int>(amrex::Math::floor(l)) - N;
```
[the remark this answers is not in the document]
chongchonghe: you are right. mistake fixed.

chongchonghe:
@BenWibking @markkrumholz This is ready for review.

BenWibking:
I haven't been able to review this yet. I will try to review it tomorrow.

BenWibking:
This is what Codex said. It appears to be a valid issue, although perhaps not that important in practice:
```
  - [P2] Renormalize the kernel at physical boundaries — /Users/benwibking/amrex_codes/quokka/src/particles/particle_deposition.hpp:143-145
    When a luminous particle lies within two cells of a non-periodic physical domain boundary, this stencil writes non-zero Wendland
    weights into physical ghost cells, but the later radiation source update only reads the valid indexRange. Because inv_norm was computed
    over the full stencil including those outside-domain cells, part of the particle luminosity is dropped instead of being conserved. This
    needs to clip/renormalize against the domain or otherwise fold physical-boundary ghost deposition back into valid cells.
```

src/particles/particle_deposition.hpp
```diff
@@ -49,6 +50,106 @@ struct NearestEight : public Base<NearestEight, amrex::Real> {
...
+			const amrex::Real dz =
+			    (AMREX_SPACEDIM >= 3) ? static_cast<amrex::Real>(kk - N) + 0.5 - frac[2] : 0.0; // NOLINT(misc-redundant-expression)
+			for (int jj = 0; jj < ny_loop; ++jj) {
+				const amrex::Real dy =
+				    (AMREX_SPACEDIM >= 2) ? static_cast<amrex::Real>(jj - N) + 0.5 - frac[1] : 0.0; // NOLINT(misc-redundant-expression)
+				for (int ii = 0; ii < stencil_width; ++ii) {
+					const amrex::Real dx = static_cast<amrex::Real>(ii - N) + 0.5 - frac[0];
+					const amrex::Real r2 = AMREX_D_TERM(dx * dx, +dy * dy, +dz * dz);
+					if (r2 <= cutoff_r2) {
+						norm_sum += kernel_wendland_c2(std::sqrt(r2) * inv_N);
```
BenWibking, line 114: This will accumulate outside domain boundaries, but I don't think that's an issue in practice.

BenWibking (approved):
LGTM

src/particles/particle_deposition.hpp
chongchonghe: Energy comes from stars and it's not a conserved quantity anyway. What we really care about is probably energy and momentum conservation after radiation is released from stars.

PR #1788, opened 2026-04-08
add debug print to ODEIntegration

BenWibking:
## Summary
- Added a debug print statement to `src/problems/ODEIntegration/testODEIntegration.cpp` to indicate when ODE integration starts.
- Verified the change by building and running the `ODEIntegration` problem using the `quokka` tool.


BenWibking:
Ok, the agent went too far with this. This was just for testing a new LLM.

PR #1791, opened 2026-04-09
Add heating_rate_external runtime parameter

chongchonghe:
### Description
This PR adds a new runtime parameter: `heating_rate_external`.

The parameter is parsed as an AMReX expression with variables `time` and `dt`, is applied as an external heating term with units `erg/s/H` when cooling is enabled, and is documented in `docs/markdown/parameters.md`.

### Related issues
N/A

### Checklist


inputs/TallBoxSf_regression.toml
```diff
@@ -51,12 +52,12 @@ problem.stars_file = ""
...
 problem.rho01 = 4.320441e-24
 problem.sigma1 = 7.0e5
-problem.boost_velocity = [0.0, 0.0, 0.0]  # 0 km/s in x-direction
 
 cooling.enabled = 1
 cooling.cooling_table_type = "resampled"
 cooling.hdf5_data_file = "CloudyData_UVB=HM2012_shielded_resampled_noPE.h5"
-
+# rho=4.320441e-24 g/cm^3, tff = sqrt(3 pi / (32 G rho)), tff = 1.010633461e+15 s = 32.02504187 Myr
+heating_rate_external = "max(min(1.0, 2.0 - time / (32 * Myr)), 0.0024) * 2e-26" # unit: erg/s/H. Gamma_0 = 2e-26
```
BenWibking: What is the meaning of the `0.0024` value here?

src/simulation.hpp
```diff
@@ -1018,6 +1022,39 @@ template <typename problem_t> void AMRSimulation<problem_t>::readParameters()
...
 
+	// optional external heating rate expression (variables: time, dt)
+	heatingRateExternalExpr_.clear();
+	pp.query("heating_rate_external", heatingRateExternalExpr_);
+	useHeatingRateExternalParser_ = !heatingRateExternalExpr_.empty();
+	if (useHeatingRateExternalParser_) {
+		heatingRateExternalParser_.emplace(heatingRateExternalExpr_);
+		auto symbols = heatingRateExternalParser_->symbols();
+		symbols.erase("time");
+		symbols.erase("dt");
```
BenWibking: Why is this `.erase("symbol")`?

From this example (https://github.com/AMReX-Codes/amrex-parser?tab=readme-ov-file#features), it seems like this should be either `.setConstant()` or `.registerVariables()`, but I'm not sure which one is appropriate here, because I don't fully understand the intended usage.

inputs/TallBoxSf_regression.toml
chongchonghe: It the FUV strength from external galactic environment, `0.0024 G_0`

chongchonghe: Number taken from the first TIGRESS paper (and the literature therein)

src/simulation.hpp
chongchonghe: erase("time")/erase("dt") or an equivalent check is needed because I use setConstant to read constants like `Myr` defined previously. This is done in a `for (auto &symbol : symbols)` loop, so we want to skip 'time' and 'dt' because they are runtime parser, not from ParmParse. I changed it to a cleaner solution by doing a skip instead of erase.

BenWibking (approved):

PR #1801, opened 2026-04-14
[WIP] Rewrite hydro integrator to use generic RK class

BenWibking:
### Description
This introduces a generic `quokka::RKIntegrator` class that can be re-used for multiple physics modules, and rewrites the hydro update function `AdvanceHydroAtLevel` to use it.

Compared to the AMReX RK class, `quokka::RKIntegrator` adds the ability to advance multiple multifabs within the integrator, which is required for MHD. It also allows computing a provisional update, then optionally redoing the update with low-order fluxes surrounding bad cells. Once it is stablized in Quokka, this functionality may be migrated to AMReX's RK integrator for maintainability.

The hydro integrator now does FOFC / fail-safe per-stage, rather than across the whole level step:
  * [Balsara et al. (2025)](https://scixplorer.org/abs/2025ApJ...988..134B/abstract) show that this restores stability to high-order hydro schemes in extreme flows, but
  * May produce a lower-quality solution in the FOFC cells (it's no longer a forward Euler time update in those cells)
  * Significantly lower memory high-water mark
  * Significantly reduced complexity
  * Possible to extend hydro to IMEX in the future

Note that when doing FOFC per-stage, it is critical to use a positivity-preserving Riemann solver with a positivity-preserving timestep near the CFL timestep (e.g. LLF). HLLC/HLLD do not work (positivity preserving if wavespeeds are accurate, but timestep can be arbitrarily small). This may be why my original experiments doing this per-stage did not work. It is also critical to use a time integrator with the SSP property.

### Related issues
N/A

### Checklist

### DIFF
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -2174,186 +2174,184 @@ auto QuokkaSimulation<problem_t>::advanceHydroAtLevel(amrex::MultiFab &state_old
 			}
 		}
 
-		if (do_tracers != 0 && final_success) {
+		if (do_tracers != 0 && final_success)
+		{
 			auto const &accum = integrator.getAccumulators();
 			TracerPC->AdvectWithUmac(const_cast<amrex::MultiFab *>(accum.avg_face_vel.data()), lev, dt_lev);
 		}
 
-			return final_success;
-		};
+		return final_success;
+	};
 
-		if (integratorOrder_ == 1) {
-			return run_integrator(std::type_identity<quokka::ForwardEulerScheme>{});
-		}
-
-		return run_integrator(std::type_identity<quokka::SSPRK2Scheme>{});
+	if (integratorOrder_ == 1) {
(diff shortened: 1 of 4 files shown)


BenWibking:
Codex identifies this issue:
```
• - [P2] Intermediate RK stage ghosts are still uninitialized — src/math/RKIntegrator.hpp:65, src/math/RKIntegrator.hpp:293, src/
    hydro/HydroRKPolicy.hpp:425, src/simulation.hpp:3159. CompositeStateData::defineLike() only allocates stage_state_; it never
    setVal(0) or copies a valid state into it. The stage update only overwrites valid cells, so stage ghosts remain undefined.
    That still conflicts with the current hydro path, which checks ghosts in debugAssertFinite(), and with
    fillBoundaryConditions() on refined levels, which asserts the input state including ghosts is finite before filling. In AMR/
    debug/MHD runs this can still trip immediately or feed stale ghost data into the next stage fill.
```

BenWibking:
@codex Review this PR

PR #1806, opened 2026-04-16
Isotropic constant thermal conduction

aditivijayan:
### Description
The changes implemented in this pull request will add thermal conduction capability to the code. 

This PR makes the following changes-

1. Adds a new folder `src/conduction` comprising `ElectronConduction.hpp` which essentially modifies the gas internal energy of the cell. The heat fluxes are calculated at the faces using constant isotropic conductivity. 
2. Modifies `QuokkaSimulation.hpp` to include an initialiser for the conduction routine.
3. Modifies `simulation.hpp` to estimate the conduction time step and compare it against the hydro and particle timesteps. This requires setting up of a conduction cfl which I have set to 0.2 for the test problem.
4. Includes a test problem that evolves 1D gaussian. I have tested the gaussian as well as the step function in 3D. The test also compares the numerical result of the 1D problem against the analytical solution.

Note-

1. This PR implements only constant conductivity. Future PRs will include spitzer and anisotropic conductivities.
2. The thermal conduction module calculates saturated flux but uses only classical. I would prefer to keep it this way since flux limiter is required for future problems.
3. For the 1D and 3D gaussian problems I have verified that the absolute error decreases with increasing resolution.

### Related issues

### Checklist
- [ x] I have added tests for any new physics that this PR adds to the code.


aditivijayan:
### Plots of 1D tests
**Solutions at different times for an initial Gaussian**
image


**Comparison of 1D result with analytical solution**
image

**Error vs Cells**

image

aditivijayan:
### Some plots from 3D runs

**Temperature slice with error, nx=256**
Error = |numerical-analytical|/numerical


image




**Temperature slice with error, nx=512**

image

src/simulation.hpp
BenWibking, line 1202: Because the diffusion timestep scales as $\Delta x^2$ (instead of $\Delta x$), getting this to work with multiple AMR levels will require a substantial rewrite of this function.

However, the current form is fine if you only want to do single-level simulations without AMR.

src/QuokkaSimulation.hpp
```diff
@@ -1002,6 +1022,16 @@ auto QuokkaSimulation<problem_t>::addStrangSplitSourcesWithBuiltin(amrex::MultiF
...
 		td->applyDriving(state, time, dt, cellSizes);
 	}
+
+	if (enableElectronConduction_ == 1) {
+		fillBoundaryConditions(state, state, lev, time, quokka::centering::cc, quokka::direction::na, PreInterpState, PostInterpState);
+		const quokka::conduction::ElectronConductionParams conduction_params{.conductivity_prefactor = electronConductionKappa0_ * C::k_B,
+										     .flux_limiter_phi = electronConductionFluxLimiterPhi_,
+										     .saturation_factor = electronConductionSaturationFactor_,
+										     .min_temperature = electronConductionTempFloor_};
+		quokka::conduction::ElectronConduction<problem_t>::ComputeExplicit(state, state_fc, geom[lev], dt, conduction_params);
```
BenWibking, line 1032: If you want to do AMR, this function must accumulate fluxes and increment the flux registers to ensure conservation at level boundaries.

src/problems/ThermalConduction/testThermalConduction.cpp
BenWibking, line 1: This test (or another similar one) should use the Richardson extrapolation machinery to show the conduction module converges at second order in spatial resolution.

src/simulation.hpp
aditivijayan: I will keep this in mind. We would definitely need to work with AMR in the future. @evan1022

BenWibking: @evan1022 For reference, this is the exact algorithm we currently use to set the timesteps on each level:
```
  function computeTimestepAtLevel(lev):
      # Problem-specific signal speed calculation happens in subclass code.
      computeMaxSignalLocal(lev)

      domain_signal_max    = norm_inf(max_signal_speed_[lev])
      domain_signal_maxloc = index_of_max(max_signal_speed_[lev])

      if signalSpeedAbort_ > 0 and domain_signal_max > signalSpeedAbort_:
          print cell properties at domain_signal_maxloc
          abort

      dx     = geom[lev].CellSizeArray()
      dx_min = min(cell widths on this level)

      hydro_dt = cflNumber_ * dx_min / domain_signal_max

      particle_dt = +infinity

      if AMREX_SPACEDIM == 3 and particleRegister_.HasMassiveParticles():
          max_particle_speed = particleRegister_.computeMaxParticleSpeed(lev)
          assert max_particle_speed is finite and not NaN

          if particleSpeedAbort_ > 0 and max_particle_speed > particleSpeedAbort_:
              abort

          # Avoid division by ~0 / useless particle limit when particles are very slow.
          if max_particle_speed > 1e-5 * (dx_min / hydro_dt):
              particle_dt = particleCflNumber_ * dx_min / max_particle_speed

      return min(hydro_dt, particle_dt)


  function computeTimestep():
      # 1. Compute a local candidate dt on each active level.
      dt_tmp = array(size = finest_level + 1)
      for level = 0 to finest_level:
          dt_tmp[level] = computeTimestepAtLevel(level)

      # 2. Limit how much dt can increase relative to the previous step.
      change_max = 1.1
      for level = 0 to finest_level:
          dt_tmp[level] = min(dt_tmp[level], change_max * dt_[level])

      # 3. Set default subcycling counts from refinement ratios.
      if do_subcycle == 1:
          for lev = 1 to max_level:
              nsubsteps[lev] = MaxRefRatio(lev - 1)

      # 4. Choose the coarse-level timestep dt_0 so that every level can subcycle into it.
      dt_0 = dt_tmp[0]
      level_that_sets_dt_0 = 0
      n_factor = 1

      for level = 0 to finest_level:
          n_factor = n_factor * nsubsteps[level]   # nsubsteps[0] is 1

          old_dt_0 = dt_0
          dt_0 = min(dt_0, n_factor * dt_tmp[level])
          if dt_0 < old_dt_0:
              level_that_sets_dt_0 = level

          dt_0 = min(dt_0, maxDt_)

          if tNew_[level] == 0.0:
              dt_0 = min(dt_0, initDt_)

          if constantDt_ > 0.0:
              dt_0 = constantDt_

      # 5. Optionally shrink the very first coarse step.
      if tNew_[0] == 0.0:
          dt_0 = dt_0 * initShrink_

      # 6. Prevent overshooting stopTime_.
      eps = 1e-3 * dt_0
      if tNew_[0] + dt_0 > stopTime_ - eps:
          dt_0 = stopTime_ - tNew_[0]

      # 7. Assign dt to all active levels.
      dt_[0] = dt_0
      for level = 1 to finest_level:
          dt_[level] = dt_[level - 1] / nsubsteps[level]
```

Regridding has one extra timestep-setting path for newly created fine levels:
```
  after regrid:
      for k = old_finest + 1 to finest_level:
          if do_subcycle != 0:
              dt_[k] = dt_[k - 1] / nsubsteps[k]
          else:
              dt_[k] = dt_[k - 1]
 ```

File: src/conduction/ElectronConduction.hpp [hunk 2]
```diff
@@ -0,0 +1,210 @@
...
+
+			amrex::Real Tgas = NAN;
+			amrex::Real Pgas = NAN;
+			amrex::Real cs_iso = NAN;
+			if constexpr (HydroSystem<problem_t>::nmscalars_ > 0) {
+				amrex::GpuArray<amrex::Real, HydroSystem<problem_t>::nmscalars_> massScalars =
+				    RadSystem<problem_t>::ComputeMassScalars(cons, i, j, k);
+				Tgas = quokka::EOS<problem_t>::ComputeTgasFromEint(rho, Eint, massScalars);
+				Pgas = quokka::EOS<problem_t>::ComputePressure(rho, Eint, massScalars);
+				cs_iso = quokka::EOS<problem_t>::ComputeSoundSpeed(rho, Pgas);
```
BenWibking, line 104: ComputeSoundSpeed returns the adiabatic sound speed, not the isothermal sound speed.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,210 @@
...
+			} else {
+				Tgas = quokka::EOS<problem_t>::ComputeTgasFromEint(rho, Eint);
+				Pgas = quokka::EOS<problem_t>::ComputePressure(rho, Eint);
+				cs_iso = quokka::EOS<problem_t>::ComputeSoundSpeed(rho, Pgas);
```
BenWibking, line 108: This is the adiabatic sound speed, not the isothermal one.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,210 @@
...
+			div_flux += (flux_y_const[bx](i, j + 1, k) - flux_y_const[bx](i, j, k)) / dx[1];
+#endif
+#if AMREX_SPACEDIM == 3
+			div_flux += (flux_z_const[bx](i, j, k + 1) - flux_z_const[bx](i, j, k)) / dx[2];
+#endif
+
+			amrex::Real Eint_new = Eint_old - dt * div_flux;
+
+			state_out[bx](i, j, k, HydroSystem<problem_t>::energy_index) = Eint_new + Ekin;
+			state_out[bx](i, j, k, HydroSystem<problem_t>::internalEnergy_index) = Eint_new;
```
BenWibking, line 194: ~~I think this probably shouldn't overwrite the auxiliary internal energy. It may be necessary to apply the difference in old and new internal energies instead.~~

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,210 @@
...
+#include "hydro/hydro_system.hpp"
+
+namespace quokka::conduction
+{
+
+struct ElectronConductionParams {
+	amrex::Real conductivity_prefactor = 3.e34;
+	amrex::Real flux_limiter_phi = 0.1;
+	amrex::Real saturation_factor = 5.0;
+	amrex::Real min_temperature = 0.0;
```
BenWibking, line 32: I would prefer not to have a separate temperature floor for conduction, and instead use the existing temperature floor that's used elsewhere in the code.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,210 @@
...
+			const amrex::Real rho = cons(i, j, k, HydroSystem<problem_t>::density_index);
+			const amrex::Real Eint = HydroSystem<problem_t>::ComputeInternalEnergy(cons, i, j, k, &local_state_fc);
+
+			amrex::Real Tgas = NAN;
+			amrex::Real Pgas = NAN;
+			amrex::Real cs_iso = NAN;
+			if constexpr (HydroSystem<problem_t>::nmscalars_ > 0) {
+				amrex::GpuArray<amrex::Real, HydroSystem<problem_t>::nmscalars_> massScalars =
+				    RadSystem<problem_t>::ComputeMassScalars(cons, i, j, k);
+				Tgas = quokka::EOS<problem_t>::ComputeTgasFromEint(rho, Eint, massScalars);
```
BenWibking, line 102: Up to now, we have only used the temperature in the cooling function. If you want to do both cooling and conduction, you should use the compute temperature function from the cooling module.

The version from the EOS will not give the same value. The fact that they do not produce the same value is a larger code problem that will require some longer-term code infrastructure changes to address properly.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,210 @@
...
+				Pgas = quokka::EOS<problem_t>::ComputePressure(rho, Eint, massScalars);
+				cs_iso = quokka::EOS<problem_t>::ComputeSoundSpeed(rho, Pgas);
+			} else {
+				Tgas = quokka::EOS<problem_t>::ComputeTgasFromEint(rho, Eint);
```
BenWibking, line 106: Same issue as above. This should probably use the temperature function from the cooling module. The EOS function does not return the same value.

src/simulation.hpp
BenWibking, line 1321: This line will give the wrong timestep when conduction is turned on, because n_factor scales as dx_l / dx_(l-1), instead of the square of this ratio.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,210 @@
...
+	amrex::Real min_temperature = 0.0;
+};
+
+template <typename problem_t> class ElectronConduction
+{
+      public:
+	static void ComputeExplicit(amrex::MultiFab &state, std::array<amrex::MultiFab, AMREX_SPACEDIM> const &state_fc, amrex::Geometry const &geom,
+				    amrex::Real dt, ElectronConductionParams const &params)
+	{
+		static_assert(Physics_Traits<problem_t>::is_hydro_enabled, "Electron conduction requires hydro to be enabled.");
```
BenWibking, line 41: A check should be added here to ensure that the simulation only has a single level and is not using AMR, since it doesn't fully support AMR yet.

src/simulation.hpp
BenWibking, line 1289: This function will need to be rewritten to support AMR. See below for the details.

src/QuokkaSimulation.hpp
```diff
@@ -167,6 +172,10 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 	std::string coolingTableType_;
 	std::string coolingTableFilename_;
 
+	amrex::Real electronConductionFluxLimiterPhi_ = 1.0;
+	amrex::Real electronConductionSaturationFactor_ = 5.0;
+	amrex::Real electronConductionTempFloor_ = 0.0;
```
BenWibking, line 177: I would prefer to re-use the existing temperature floor, rather than create a new one for conduction.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,210 @@
...
+#include "AMReX_REAL.H"
+#include "AMReX_SPACE.H"
+#include "AMReX_Vector.H"
+#include "hydro/hydro_system.hpp"
+
+namespace quokka::conduction
+{
+
+struct ElectronConductionParams {
+	amrex::Real conductivity_prefactor = 3.e34;
```
BenWibking, line 29: What are the units of this value? Can you specify them in a code comment here?

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,210 @@
...
+	amrex::Real flux_limiter_phi = 0.1;
```
BenWibking, line 30: Can you provide a reference for the value you used here?

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,210 @@
...
+	amrex::Real saturation_factor = 5.0;
```
BenWibking, line 31: Can you provide a reference for the value you used here?

src/simulation.hpp
```diff
@@ -226,6 +226,11 @@ template <typename problem_t> class AMRSimulation : public amrex::AmrCore
 	int sn_count_ = 0;	      // number of SN explosions in a step (used for diagnostics)
 	int sn_count_cumulative_ = 0; // cumulative number of SN explosions (used for diagnostics)
 
+	// Conduction parameters
+	amrex::Real electronConductionKappa0_ = 4.17; // default
```
BenWibking, line 230: What are the units of this value? Can you add a code comment specifying them?

BenWibking (changes requested):

BenWibking:
@codex Review this PR.

src/conduction/ElectronConduction.hpp
BenWibking: This ADR describes the underlying issue in more detail: https://github.com/quokka-astro/quokka/pull/1810/changes

src/conduction/ElectronConduction.hpp
BenWibking: I think this is not actually an issue. I hadn't read the dual energy code in a long time, so I think this is actually fine as-is.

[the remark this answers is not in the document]
BenWibking: This is a real issue. The magnetic energy needs to be saved/restored, or the energy update needs to be done as a pure `+=` operation, rather than a replacement.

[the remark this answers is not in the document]
evan1022: Yes I agree.  Especially since our eventual goal is to also allow for anisotropic conduction, conduction and MHD better be compatible.

aditivijayan:
@BenWibking @evan1022 I have noted the comments and will address them!

src/conduction/ElectronConduction.hpp
aditivijayan: Both temperature and cs deteremined using Cooling.

src/conduction/ElectronConduction.hpp
aditivijayan: Done.

src/conduction/ElectronConduction.hpp
aditivijayan: This has been chosen arbitrarily since it is not being used for now. PLUTO manual suggests phi<1.

src/conduction/ElectronConduction.hpp
aditivijayan: This check has been add to QuokkaSimulation.hpp.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,214 @@
...
+		auto const &temp = temperature.const_arrays();
+		auto const &kappa = conductivity.const_arrays();
+		auto const &qsat = saturated_flux.const_arrays();
+		auto flux_x = heat_flux[0].arrays();
+		amrex::ParallelFor(heat_flux[0], [=] AMREX_GPU_DEVICE(int bx, int i, int j, int k) noexcept {
+			const amrex::Real gradT = (temp[bx](i, j, k) - temp[bx](i - 1, j, k)) / dx[0];
+			const amrex::Real kappa_face = 0.5 * (kappa[bx](i, j, k) + kappa[bx](i - 1, j, k));
+			const amrex::Real q_classical = -kappa_face * gradT;
+			const amrex::Real q_sat_face = 0.5 * (qsat[bx](i, j, k) + qsat[bx](i - 1, j, k));
+			const amrex::Real limiter = 1.0 + std::abs(q_classical) / amrex::max(q_sat_face, small);
```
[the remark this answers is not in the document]
aditivijayan: I've re-added the limiter and set it to a large value in the input file in order to obviate its use.

src/problems/ThermalConduction/testThermalConduction.cpp
```diff
@@ -0,0 +1,208 @@
...
+		// const amrex::Real term2 = (sqrt_chit / (2.0 * R * std::sqrt(M_PI))) * (std::exp(-((R0 + R) * (R0 + R)) / chit)
+		//                                                                 - std::exp(-((R0 - R) * (R0 - R)) / chit));
+
+		// const amrex::Real T = Tout + (Tin - Tout) * (term1 + term2);
+		/*-------------------------------------------------*/
+
+		amrex::Real Eint = quokka::EOS<ThermalConductionProblem>::ComputeEintFromTgas(rho, T);
+		state_cc(i, j, k, HydroSystem<ThermalConductionProblem>::density_index) = rho;
+		state_cc(i, j, k, HydroSystem<ThermalConductionProblem>::energy_index) = Eint; // total energy = internal energy + kinetic energy
+		state_cc(i, j, k, HydroSystem<ThermalConductionProblem>::internalEnergy_index) = Eint;
```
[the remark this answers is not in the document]
aditivijayan: Initialized all state vars to zero.

src/conduction/ElectronConduction.hpp
aditivijayan: Done!

src/conduction/ElectronConduction.hpp
aditivijayan: Okay it is not showing my reply to this. Temperature is now estimated using ResampledCooling.

src/conduction/ElectronConduction.hpp
aditivijayan: Done!

src/simulation.hpp
aditivijayan: nfactor is squared if conduction is on.

BenWibking:
@aditivijayan Let me know when this is ready for review again.

chongchonghe:
@claude review

BenWibking:
I think Claude's comments are reasonable to address in this PR.

aditivijayan:
> @aditivijayan Let me know when this is ready for review again.

Apologies! 

I have been essentially trying to get the test working but I am running into a problem which is that the errors, compared to analytical solution, are quite large (>10%). (Error = |numerical solution - analytical solution|/numerical solution)

The reason for this is related to using cooling tables rather than EOS. When I use cooling tables the diffusion constant depends on mu which I guess is not exactly constant for the entire temperature range. 

### Example of solution using EOS


image

### Example of solution using cooling tables


image


In both the cases  I am solving a step-function with T_high ~ 10^4 K and T_low ~ 10 K. But relative errors are quite different.

I am unsure of how to proceed.

markkrumholz:
Just a note for those following the discussion on GitHub that @aditivijayan met in person and worked out how to address this.

aditivijayan:
> Just a note for those following the discussion on GitHub that @aditivijayan met in person and worked out how to address this.

The solution was to use temperatures such that they have the same mmw. This means using a higher than previous temperature. 

I switched back to testing gaussian and chose parameters for it based on the discussion with @markkrumholz. Peak at ~ 1.e8 K plus a constant at 2.e5 K.

The good news is that the errors are small ~ 1%. The bad news is that they don't reduce with Nx and the test fails Richardson. I am unsure why. 

### Errors for different Nx

image

image


image

aditivijayan:
The error is still arising due to incorrect mmw in the analytical solution as the code passes the richardson test for EOS.

Screenshot 2026-04-30 at 2 02 09 pm

In this case, can I do something @markkrumholz had suggested? That is use EOS for testing and not otherwise.

I have implemented an EOS flag as a conductivity parameter. If this is acceptable, I can push the changes.

aditivijayan:
> I think Claude's comments are reasonable to address in this PR.

Will work on them once I've resolved this EOS vs table issue.

markkrumholz:
I'd suggest that for the Richardson test it's fine not to use the tabulated EOS and just use a fixed mean molecular weight, since that is what the analytic solution assumes. The MMW is nearly constant for the temperature range you have chosen, but it is clearly not 100% constant, and this is becoming the dominant source of error, which is why it isn't converging. I think we understand what is going on, so it is fine to move on and not use the tabulated EOS for this test.

BenWibking:
> I'd suggest that for the Richardson test it's fine not to use the tabulated EOS and just use a fixed mean molecular weight, since that is what the analytic solution assumes. The MMW is nearly constant for the temperature range you have chosen, but it is clearly not 100% constant, and this is becoming the dominant source of error, which is why it isn't converging. I think we understand what is going on, so it is fine to move on and not use the tabulated EOS for this test.

That's fine with me.

src/problems/ThermalConduction/testThermalConduction.cpp
aditivijayan: Richardson test has been added.

aditivijayan:
@claude review

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,211 @@
...
+		const amrex::Real t_min = params.min_temperature;
+		const amrex::Real small = std::numeric_limits<amrex::Real>::min();
+		const auto tables_dev = tables.const_tables();
+
+		amrex::MultiFab temperature(state.boxArray(), state.DistributionMap(), 1, state.nGrow());
+		temperature.setVal(0.0);
+		amrex::MultiFab conductivity(state.boxArray(), state.DistributionMap(), 1, state.nGrow());
+		conductivity.setVal(0.0);
+		amrex::MultiFab saturated_flux(state.boxArray(), state.DistributionMap(), 1, state.nGrow());
+		saturated_flux.setVal(0.0);
```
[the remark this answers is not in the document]
BenWibking: It's wrong. This is fine.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,211 @@
...
+			amrex::Real cs = NAN;
+			if (params.eos_flag == 0) {
+				Tgas = quokka::ResampledCooling::ComputeTgasFromEgas(rho, Eint, tables_dev);
+				cs = quokka::ResampledCooling::ComputeSoundSpeedFromRhoEint(rho, Eint, tables_dev);
+			} else if (params.eos_flag == 1) {
+				Tgas = quokka::EOS<problem_t>::ComputeTgasFromEint(rho, Eint);
+				amrex::Real Pgas = quokka::EOS<problem_t>::ComputePressure(rho, Eint);
+				cs = quokka::EOS<problem_t>::ComputeSoundSpeed(rho, Pgas);
+			} else {
+				amrex::Abort("Invalid eos_flag value in ElectronConduction. Must be 0 (resampled cooling) or 1 (quokka::EOS).");
```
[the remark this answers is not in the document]
BenWibking: The AI is wrong in that `amrex::Abort` can actually be called from the GPU, but it is right in that it would be better to do this check outside of the `amrex::ParallelFor` loop.

src/simulation.hpp
```diff
@@ -1297,8 +1319,15 @@ template <typename problem_t> void AMRSimulation<problem_t>::computeTimestep()
 
 	for (int level = 0; level <= finest_level; ++level) {
 		n_factor *= nsubsteps[level];
+
+		auto effective_factor = static_cast<amrex::Real>(n_factor);
+		if (enableElectronConduction_ == 1) {
+			// Conduction timestep scales as dx^2, so we need to use n_factor^2 here instead of n_factor.
+			effective_factor = static_cast<amrex::Real>(n_factor) * static_cast<amrex::Real>(n_factor);
+		}
```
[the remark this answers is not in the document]
BenWibking: I have to think about this a bit more. I think the worst thing that will happen is that the timestep will be too conservative, that is, it will be much smaller than needed if the timestep should be limited by the hydro instead of conduction.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+namespace quokka::conduction
+{
+
+struct ElectronConductionParams {
+	amrex::Real conductivity_prefactor = 3.e34; // units of erg cm^-1 s^-1 K^-1
+	amrex::Real flux_limiter_phi = 0.1;
+	amrex::Real saturation_factor = 5.0; // refer to equation 8 of Cowee & McKee 1977
+	amrex::Real min_temperature = 0.0;   // default value will be overwritten by tempFloor_ during initialization
+	amrex::Real eos_flag = 1;	     // 1 == use quokka::EOS; 0 == use resampled cooling
```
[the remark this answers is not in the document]
BenWibking: This comment is accurate.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,211 @@
...
+				Tgas = quokka::EOS<problem_t>::ComputeTgasFromEint(rho, Eint);
+				amrex::Real Pgas = quokka::EOS<problem_t>::ComputePressure(rho, Eint);
+				cs = quokka::EOS<problem_t>::ComputeSoundSpeed(rho, Pgas);
+			} else {
+				amrex::Abort("Invalid eos_flag value in ElectronConduction. Must be 0 (resampled cooling) or 1 (quokka::EOS).");
+			}
+
+			const amrex::Real Tuse = amrex::max(Tgas, t_min);
+			const amrex::Real kappa = params.conductivity_prefactor;
+			const amrex::Real qsat = amrex::max(saturation_factor * flux_limiter_phi * rho * std::pow(cs, 3), small);
```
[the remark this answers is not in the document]
BenWibking: This comment is accurate.

src/QuokkaSimulation.hpp
```diff
@@ -167,6 +172,10 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 	std::string coolingTableType_;
 	std::string coolingTableFilename_;
 
+	amrex::Real electronConductionFluxLimiterPhi_ = 1.0;
+	amrex::Real electronConductionSaturationFactor_ = 5.0;
+	amrex::Real eosFlagForElectronConduction_ = 1; // 1 == use quokka::EOS; 0 == use resampled cooling
```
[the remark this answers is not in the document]
BenWibking: This comment is accurate.

BenWibking:
@codex review

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,211 @@
...
+		}
+
+		AMREX_ALWAYS_ASSERT_WITH_MESSAGE(state.nGrow() >= 1, "Electron conduction requires at least 1 ghost cell.");
+
+		const auto dx = geom.CellSizeArray();
+		const amrex::Real flux_limiter_phi = params.flux_limiter_phi;
+		const amrex::Real saturation_factor = params.saturation_factor;
+		const amrex::Real t_min = params.min_temperature;
+		const amrex::Real small = std::numeric_limits<amrex::Real>::min();
+		const auto tables_dev = tables.const_tables();
```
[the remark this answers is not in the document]
BenWibking: This is a good point.

src/simulation.hpp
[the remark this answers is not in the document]
BenWibking: Ok, I have written up a general solution to this problem here: https://github.com/quokka-astro/quokka/pull/1863.

For this PR, I propose that when conduction is enabled, it is required to use global timesteps. Then this can be relaxed in a future PR.

aditivijayan:
I am unsure why the tests are failing.

extern/fmt
BenWibking, line 1: This is why the tests are failing. Somehow you checked in this submodule, but it has been removed from Quokka. Can you remove it?

extern/grackle_data_files
BenWibking, line 1: Can you remove this submodule?

extern/cooling/CloudyData_UVB=HM2012_shielded_resampled_noPE_zmet0.5.h5
BenWibking, line 1: Did you mean to include this cooling table? If not, can you remove this file from this PR?

aditivijayan:
The GPU tests are failing because of Out of Memory error-
https://dev.azure.com/quokka-code/quokka-CI/_build/results?buildId=14048&view=logs&j=655dacad-9e02-5f72-e0cf-043ebfb7cecc&t=040ddf27-e411-58f7-ab97-dfe1d2a20f3a&l=8062

@BenWibking any suggestions?

BenWibking:
Since it runs for many steps without running out of memory, this suggests that memory is secularly increasing with time, so something may not be getting deallocated properly. Can you print out the memory usage, say, every 100 timesteps, and see if it's increasing?

You can do that using the method described here: https://github.com/AMReX-Codes/amrex/discussions/4467

BenWibking:
auggie review

aditivijayan:
> Since it runs for many steps without running out of memory, this suggests that memory is secularly increasing with time, so something may not be getting deallocated properly. Can you print out the memory usage, say, every 100 timesteps, and see if it's increasing?
> 
> You can do that using the method described here: [AMReX-Codes/amrex#4467](https://github.com/AMReX-Codes/amrex/discussions/4467)

It runs successfully for 128^3 and 256^3 and crashes only for the 512^3 test. So I am not sure it's an issue with deallocation.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,214 @@
...
+			const amrex::Real Emag = HydroSystem<problem_t>::ComputeMagneticEnergy(i, j, k, &local_state_fc);
+			amrex::Real div_flux = (flux_x_const[bx](i + 1, j, k) - flux_x_const[bx](i, j, k)) / dx[0];
+#if AMREX_SPACEDIM >= 2
+			div_flux += (flux_y_const[bx](i, j + 1, k) - flux_y_const[bx](i, j, k)) / dx[1];
+#endif
+#if AMREX_SPACEDIM == 3
+			div_flux += (flux_z_const[bx](i, j, k + 1) - flux_z_const[bx](i, j, k)) / dx[2];
+#endif
+
+			amrex::Real Eint_new = Eint_old - dt * div_flux;
```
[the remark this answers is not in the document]
aditivijayan: The limits are enforced elsewhere in the code I believe.

BenWibking:
> > Since it runs for many steps without running out of memory, this suggests that memory is secularly increasing with time, so something may not be getting deallocated properly. Can you print out the memory usage, say, every 100 timesteps, and see if it's increasing?
> > You can do that using the method described here: [AMReX-Codes/amrex#4467](https://github.com/AMReX-Codes/amrex/discussions/4467)
> 
> It runs successfully for 128^3 and 256^3 and crashes only for the 512^3 test. So I am not sure it's an issue with deallocation.

Ah, I see. 512^3 is just too big of a simulation to fit on a single GPU. I would recommend starting at a smaller grid size, and then stopping at 256^3.

src/conduction/ElectronConduction.hpp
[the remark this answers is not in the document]
BenWibking: Yeah, I don't think this is a real issue.

aditivijayan:
@BenWibking the tests pass now. Are there more changes you would like to be made?

extern/cooling/CloudyData_UVB=HM2012_resampled_noPE.h5
BenWibking, line 1: Did you remove this file by mistake?

scripts/hpc_profiles/gadi_hopper.profile
BenWibking, line 1: This is a good change to have, but ideally it could go in a separate PR. Would you be able to create a separate PR for it?

src/problems/ThermalConduction/testThermalConduction.cpp
```diff
@@ -0,0 +1,201 @@
...
+}
+
+auto problem_main() -> int
+{
+	/***Richardson Extrapolation ****/
+
+	quokka::richardson::applyQuietDefaults();
+	quokka::richardson::Parameters params{};
+	params.machine_precision_target = 2.0e-9; // limit based on delta_b_magn, smaller values can be used if this is decreased
+	params.nx_initial = 64;
```
BenWibking, line 193: Does this work if you set it to 32?

src/problems/ThermalConduction/testThermalConduction.cpp
```diff
@@ -0,0 +1,201 @@
...
+	params.nx_max = 256;
```
BenWibking, line 194: Does it work if you set this to 128?

src/conduction/ElectronConduction.hpp
BenWibking: This needs to be modified to support chemistry. This is because the EOS can depend on the species abundances.

Take a look at these versions of the functions in src/hydro/EOS.hpp:
```
  ComputeTgasFromEint(rho, Eint, massScalars)
  ComputeEintFromTgas(rho, Tgas, massScalars)
  ComputeEintFromPres(rho, Pressure, massScalars)
  ComputePressure(rho, Eint, massScalars)
  ComputeSoundSpeed(rho, Pressure, massScalars
```

When chemistry is enabled, you need to pass a `massScalars` array of this type:
```
  quokka::optional<amrex::GpuArray<amrex::Real, nmscalars_>> const &massScalars
```

extern/cooling/CloudyData_UVB=HM2012_resampled_noPE.h5
aditivijayan: Yes!

src/problems/ThermalConduction/testThermalConduction.cpp
aditivijayan: The Gaussian is not resolved at 32.

src/problems/ThermalConduction/testThermalConduction.cpp
aditivijayan: It does. But then the extrapolation has just two points. If that's acceptable I can make it 128.

BenWibking: It would be better to have 3 points for the extrapolation.

How does it run on the GPU in this case? Did you disable that test on the GPU? I think that it is okay to only do the Richardson extrapolation for the CPU tests in 1D.

aditivijayan: > It would be better to have 3 points for the extrapolation.
> 
> How does it run on the GPU in this case? Did you disable that test on the GPU? I think that it is okay to only do the Richardson extrapolation for the CPU tests in 1D.

Did you mean the 32^3 test? I didn't run it on GPUs, only on CPUs.

aditivijayan: @BenWibking just to reiterate- the gaussian isn't properly resolved at 32. It passed the richardson test for 64, 128, 256 on GPUs. Is there any other issues that need to be addressed before this can be merged?

BenWibking: Ok, I thought it crashed at 256^3 because it ran out of memory. Is that correct, or am I misremembering this? What was the resolution of that issue?

BenWibking: Ah, sorry. I am misremembering this. It crashed at 512^3. If everything passes at 64^3 through 256^3, then I don't think there's anything left to do.

BenWibking:
I just want to double check that everything passes with the lastest updates.

src/conduction/ElectronConduction.hpp
BenWibking: @aditivijayan This needs to be updated to support the case where there are species abundances.

BenWibking:
@aditivijayan I think the only thing left is to update the EOS calls so they work correctly when chemistry is enabled.

aditivijayan:
> @aditivijayan I think the only thing left is to update the EOS calls so they work correctly when chemistry is enabled.

I am not sure I understand.

BenWibking:
> > @aditivijayan I think the only thing left is to update the EOS calls so they work correctly when chemistry is enabled.
> 
> I am not sure I understand.

I was referring to this comment. If Github is malfunctioning, you may not have seen it:

Screenshot 2026-05-22 at 5 52 25 PM

aditivijayan:
> > > @aditivijayan I think the only thing left is to update the EOS calls so they work correctly when chemistry is enabled.
> > 
> > 
> > I am not sure I understand.
> 
> I was referring to this comment. If Github is malfunctioning, you may not have seen it:
> 
> Screenshot 2026-05-22 at 5 52 25 PM

Github did not show this to me. 
I've added massScalars now.

src/QuokkaSimulation.hpp
BenWibking: The conduction operator should be added to this list, so it can be automatically reversed for the pre-step Strang split call.

src/simulation.hpp
```diff
@@ -1243,6 +1248,21 @@ template <typename problem_t> auto AMRSimulation<problem_t>::computeTimestepAtLe
 		printCellProperties(lev, hydro_dt.index);
 	}
 
+	// compute timestep based on conduction parameters
+	amrex::ValLocPair<amrex::Real, amrex::IntVect> conduction_dt{.value = std::numeric_limits<amrex::Real>::max(),
+								     .index = amrex::IntVect{AMREX_D_DECL(-1, -1, -1)}};
+	if (enableElectronConduction_ == 1) {
+		double c_v = C::k_B / (quokka::EOS_Traits<problem_t>::mean_molecular_weight * (quokka::EOS_Traits<problem_t>::gamma - 1.0));
+		double diffusion_coefficient = electronConductionKappa0_ / (state_new_cc_[lev].min(0) * c_v);
+		conduction_dt.value = conductionCFL * dx_min * dx_min / diffusion_coefficient;
```
BenWibking: This should have a factor of `0.5` in it so that conductionCFL is consistent with the hydro CFL.

As it's currently written, the conduction operator is stable only for less than 0.5 in 1D, less than 0.25 in 2D, and less than 1/6 in 3D. This is confusing to users who expect conductionCFL to be normalized in the same way the hydro CFL is.

BenWibking (changes requested):

chongchonghe:
## PR #1806 Review — Isotropic Constant Thermal Conduction

**Reviewer**: Claude and @chongchonghe 

**Classification:** Feature PR (SPEC → VERIFY → EDGE)

---

### SPEC Checklist

| Claim | Source |
|---|---|
| New `ElectronConduction.hpp` module: explicit flux-limited heat flux, updates total + internal energy | `src/conduction/ElectronConduction.hpp` |
| Conduction CFL: `dt = 0.5 * CFL * dx² / D` where `D = κ₀ / (ρ_min · cᵥ)` | `src/simulation.hpp:1255-1257` |
| 6 TOML parameters under `[conduction]`: `enabled`, `conductivity_prefactor`, `conduction_cfl`, `flux_limiter_phi`, `saturation_factor`, `eos_flag` | `src/QuokkaSimulation.hpp:631-639` |
| 1D Gaussian diffusion test converges at 2nd order (Richardson extrapolation) | `src/problems/ThermalConduction/testThermalConduction.cpp` |
| No AMR support: aborts if `max_level > 0` | `src/QuokkaSimulation.hpp:1063-1065` |

---

### SPEC / VERIFY / EDGE Table

| Phase | Command | Observed |
|---|---|---|
| VERIFY — compile | `quokka build -d 1d ThermalConduction` | Compiled cleanly, no warnings |
| VERIFY — golden path | `quokka run -d 1d ThermalConduction` | Richardson test PASSED; observed rates 2.00 (64→128) and 1.99 (128→256), matching expected 2.0 |
| VERIFY — TOML params | Inspected `src/QuokkaSimulation.hpp:631-639` | All 6 parameters read via `ParmParse`; conduction params correctly forwarded to `ElectronConductionParams` struct |
| VERIFY — timestep | Inspected `src/simulation.hpp:1254-1257` | `D = kappa0 / (rho_min * c_v)` with `c_v = k_B / (mu * (γ-1))`; CFL condition `0.5 * CFL * dx² / D` is correct for explicit diffusion |
| VERIFY — flux limiter | Inspected `ElectronConduction.hpp:144-145` | `q_lim = q / (1 + \|q\| / q_sat)` correctly limits to `q_sat` in the strong-flux regime |
| VERIFY — energy update | Inspected `ElectronConduction.hpp:209-212` | `E_total` and `E_internal` both updated; kinetic + magnetic energy preserved |
| EDGE — regression | `quokka buildrun -d 1d HydroShocktube` | PASSED; hydro unchanged |

---

### Feature Assessment

The implementation correctly delivers constant isotropic thermal conduction using an explicit finite-difference update. The physical algorithm is sound: central-difference gradients, face-averaged conductivity and saturation flux, and the Cowie & McKee flux limiter. The test problem is well-designed — it isolates conduction from hydro by choosing parameters with `t_hydro / t_cond = 10³`, and the Richardson convergence test rigorously confirms 2nd-order accuracy.

---

### PR-Introduced Concerns

**1. Ghost-cell assertion vs. iteration depth mismatch (`ElectronConduction.hpp:54,80`)**

```cpp
// line 54 — checks nGrow() >= 1
AMREX_ALWAYS_ASSERT_WITH_MESSAGE(state.nGrow() >= 1, "Electron conduction requires at least 1 ghost cell.");

// line 80 — iterates over 2 ghost layers
amrex::IntVect ng = amrex::IntVect(AMREX_D_DECL(2, 2, 2));
amrex::ParallelFor(state, ng, [=] AMREX_GPU_DEVICE(...) {
    temperature_arr[bx](i, j, k) = ...;  // writes to ghost cells
});
```

The `temperature`, `conductivity`, and `saturated_flux` MultiFabs are allocated with `state.nGrow()` ghost cells. The `ParallelFor(state, ng=2, ...)` iterates over valid cells plus 2 ghost layers and writes to those MultiFabs. If `state.nGrow() = 1`, this is an out-of-bounds write. The assertion should be `nGrow() >= 2`, or `ng` should be reduced to 1 (one ghost layer of temperature is sufficient for the face-flux kernels that follow). In practice Quokka's `state_new_cc_` is typically allocated with 4 ghost cells, so this does not fail at runtime, but the assertion gives a false safety guarantee.

**2. Copy-paste file header in test (`testThermalConduction.cpp:6-7`)**

```cpp
/// \file testHydroBlast3D.cpp
/// \brief Defines a test problem for a 3D explosion.
```

Should read `testThermalConduction.cpp` and describe the Gaussian diffusion test.

**3. Misleading variable comment (`testThermalConduction.cpp:42`)**

```cpp
const double sigma = 2.410685615625e+17;  // conduction timescale in s
```

`sigma` is the initial Gaussian width (standard deviation, in cm), not a timescale. The actual conduction timescale is `σ²/(2D)`. The comment should read something like `// initial Gaussian half-width, in cm`.

**4. Typo in saturation factor citation (`ElectronConduction.hpp:32`)**

```cpp
amrex::Real saturation_factor = 5.0; // refer to equation 8 of Cowee & McKee 1977
```

The paper is "Cowie & McKee 1977" (not "Cowee").

**5. Interface forces ResampledCooling dependency when `eos_flag = 1`**

`ComputeExplicit` always accepts `const quokka::ResampledCooling::resampled_tables &tables` even when `eos_flag = 1` (the `quokka::EOS` path). This forces problems using pure-EOS conduction to initialize cooling tables, as seen in the test TOML (`cooling.read_tables_even_if_disabled = 1`). For a future-facing design, consider accepting an `std::optional<resampled_tables>` or splitting the two EOS paths into separate specialisations, so that problems without cooling tables can use conduction without modifying their TOML.

**6. No floor on `Eint_new` after conduction update (`ElectronConduction.hpp:209`)**

```cpp
amrex::Real Eint_new = Eint_old - dt * div_flux;
```

For a proper CFL-constrained timestep the internal energy cannot go negative, but if the timestep is externally forced larger than the diffusion limit (e.g. via `constantDt_`), or in the presence of extreme flux-limiter saturation, `Eint_new` could become negative. Applying `Eint_new = amrex::max(Eint_new, Eint_floor)` matches the defensive pattern used elsewhere in the codebase and is cheap to add.

---

### Pre-existing Issues (Out of Scope)

None found during this review.

aditivijayan:
> ## PR #1806 Review — Isotropic Constant Thermal Conduction
> **Reviewer**: Claude and @chongchonghe
> 
> **Classification:** Feature PR (SPEC → VERIFY → EDGE)
> 
> ### SPEC Checklist
> Claim	Source
> New `ElectronConduction.hpp` module: explicit flux-limited heat flux, updates total + internal energy	`src/conduction/ElectronConduction.hpp`
> Conduction CFL: `dt = 0.5 * CFL * dx² / D` where `D = κ₀ / (ρ_min · cᵥ)`	`src/simulation.hpp:1255-1257`
> 6 TOML parameters under `[conduction]`: `enabled`, `conductivity_prefactor`, `conduction_cfl`, `flux_limiter_phi`, `saturation_factor`, `eos_flag`	`src/QuokkaSimulation.hpp:631-639`
> 1D Gaussian diffusion test converges at 2nd order (Richardson extrapolation)	`src/problems/ThermalConduction/testThermalConduction.cpp`
> No AMR support: aborts if `max_level > 0`	`src/QuokkaSimulation.hpp:1063-1065`
> ### SPEC / VERIFY / EDGE Table
> Phase	Command	Observed
> VERIFY — compile	`quokka build -d 1d ThermalConduction`	Compiled cleanly, no warnings
> VERIFY — golden path	`quokka run -d 1d ThermalConduction`	Richardson test PASSED; observed rates 2.00 (64→128) and 1.99 (128→256), matching expected 2.0
> VERIFY — TOML params	Inspected `src/QuokkaSimulation.hpp:631-639`	All 6 parameters read via `ParmParse`; conduction params correctly forwarded to `ElectronConductionParams` struct
> VERIFY — timestep	Inspected `src/simulation.hpp:1254-1257`	`D = kappa0 / (rho_min * c_v)` with `c_v = k_B / (mu * (γ-1))`; CFL condition `0.5 * CFL * dx² / D` is correct for explicit diffusion
> VERIFY — flux limiter	Inspected `ElectronConduction.hpp:144-145`	`q_lim = q / (1 + |q| / q_sat)` correctly limits to `q_sat` in the strong-flux regime
> VERIFY — energy update	Inspected `ElectronConduction.hpp:209-212`	`E_total` and `E_internal` both updated; kinetic + magnetic energy preserved
> EDGE — regression	`quokka buildrun -d 1d HydroShocktube`	PASSED; hydro unchanged
> ### Feature Assessment
> The implementation correctly delivers constant isotropic thermal conduction using an explicit finite-difference update. The physical algorithm is sound: central-difference gradients, face-averaged conductivity and saturation flux, and the Cowie & McKee flux limiter. The test problem is well-designed — it isolates conduction from hydro by choosing parameters with `t_hydro / t_cond = 10³`, and the Richardson convergence test rigorously confirms 2nd-order accuracy.
> 
> ### PR-Introduced Concerns
> **1. Ghost-cell assertion vs. iteration depth mismatch (`ElectronConduction.hpp:54,80`)**
> 
> ```c++
> // line 54 — checks nGrow() >= 1
> AMREX_ALWAYS_ASSERT_WITH_MESSAGE(state.nGrow() >= 1, "Electron conduction requires at least 1 ghost cell.");
> 
> // line 80 — iterates over 2 ghost layers
> amrex::IntVect ng = amrex::IntVect(AMREX_D_DECL(2, 2, 2));
> amrex::ParallelFor(state, ng, [=] AMREX_GPU_DEVICE(...) {
>     temperature_arr[bx](i, j, k) = ...;  // writes to ghost cells
> });
> ```
> 
> The `temperature`, `conductivity`, and `saturated_flux` MultiFabs are allocated with `state.nGrow()` ghost cells. The `ParallelFor(state, ng=2, ...)` iterates over valid cells plus 2 ghost layers and writes to those MultiFabs. If `state.nGrow() = 1`, this is an out-of-bounds write. The assertion should be `nGrow() >= 2`, or `ng` should be reduced to 1 (one ghost layer of temperature is sufficient for the face-flux kernels that follow). In practice Quokka's `state_new_cc_` is typically allocated with 4 ghost cells, so this does not fail at runtime, but the assertion gives a false safety guarantee.
> 
> **2. Copy-paste file header in test (`testThermalConduction.cpp:6-7`)**
> 
> ```c++
> /// \file testHydroBlast3D.cpp
> /// \brief Defines a test problem for a 3D explosion.
> ```
> 
> Should read `testThermalConduction.cpp` and describe the Gaussian diffusion test.
> 
> **3. Misleading variable comment (`testThermalConduction.cpp:42`)**
> 
> ```c++
> const double sigma = 2.410685615625e+17;  // conduction timescale in s
> ```
> 
> `sigma` is the initial Gaussian width (standard deviation, in cm), not a timescale. The actual conduction timescale is `σ²/(2D)`. The comment should read something like `// initial Gaussian half-width, in cm`.
> 
> **4. Typo in saturation factor citation (`ElectronConduction.hpp:32`)**
> 
> ```c++
> amrex::Real saturation_factor = 5.0; // refer to equation 8 of Cowee & McKee 1977
> ```
> 
> The paper is "Cowie & McKee 1977" (not "Cowee").
> 
> **5. Interface forces ResampledCooling dependency when `eos_flag = 1`**
> 
> `ComputeExplicit` always accepts `const quokka::ResampledCooling::resampled_tables &tables` even when `eos_flag = 1` (the `quokka::EOS` path). This forces problems using pure-EOS conduction to initialize cooling tables, as seen in the test TOML (`cooling.read_tables_even_if_disabled = 1`). For a future-facing design, consider accepting an `std::optional<resampled_tables>` or splitting the two EOS paths into separate specialisations, so that problems without cooling tables can use conduction without modifying their TOML.
> 
> **6. No floor on `Eint_new` after conduction update (`ElectronConduction.hpp:209`)**
> 
> ```c++
> amrex::Real Eint_new = Eint_old - dt * div_flux;
> ```
> 
> For a proper CFL-constrained timestep the internal energy cannot go negative, but if the timestep is externally forced larger than the diffusion limit (e.g. via `constantDt_`), or in the presence of extreme flux-limiter saturation, `Eint_new` could become negative. Applying `Eint_new = amrex::max(Eint_new, Eint_floor)` matches the defensive pattern used elsewhere in the codebase and is cheap to add.
> 
> ### Pre-existing Issues (Out of Scope)
> None found during this review.

The floor on Eint is applied right after the Strangsources is called. I am unsure if it needs to be applied here as well.

BenWibking:
> > For a proper CFL-constrained timestep the internal energy cannot go negative, but if the timestep is externally forced larger than the diffusion limit (e.g. via `constantDt_`), or in the presence of extreme flux-limiter saturation, `Eint_new` could become negative. Applying `Eint_new = amrex::max(Eint_new, Eint_floor)` matches the defensive pattern used elsewhere in the codebase and is cheap to add.
> > ### Pre-existing Issues (Out of Scope)
> > None found during this review.
> 
> The floor on Eint is applied right after the Strangsources is called. I am unsure if it needs to be applied here as well.

I don't think it needs to be applied here.

chongchonghe:
OK, looks good.

src/QuokkaSimulation.hpp
```diff
@@ -167,6 +172,10 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 	std::string coolingTableType_;
 	std::string coolingTableFilename_;
 
+	amrex::Real electronConductionFluxLimiterPhi_ = 1.0;
+	amrex::Real electronConductionSaturationFactor_ = 5.0;
+	int eosFlagForElectronConduction_ = 1; // 1 == use quokka::EOS; 0 == use resampled cooling
```
BenWibking: Can you make this an AMREX_ENUM type instead? (See earlier comment.)

src/QuokkaSimulation.hpp
```diff
@@ -618,6 +627,17 @@ template <typename problem_t> void QuokkaSimulation<problem_t>::readParmParse()
...
 
+	// set electron thermal conduction runtime parameters
+	{
+		amrex::ParmParse const hpp("conduction");
+		hpp.query("enabled", enableElectronConduction_);
+		hpp.query("conductivity_prefactor", electronConductionKappa0_);
+		hpp.query("conduction_cfl", conductionCFL);
+		hpp.query("flux_limiter_phi", electronConductionFluxLimiterPhi_);
+		hpp.query("saturation_factor", electronConductionSaturationFactor_);
+		hpp.query("eos_flag", eosFlagForElectronConduction_);
```
BenWibking, line 638: If you change it to an AMREX_ENUM, this line does not need to be changed.

src/conduction/ElectronConduction.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+namespace quokka::conduction
+{
+
+struct ElectronConductionParams {
+	amrex::Real conductivity_prefactor = 3.e34; // units of erg cm^-1 s^-1 K^-1
+	amrex::Real flux_limiter_phi = 0.1;
+	amrex::Real saturation_factor = 5.0; // refer to equation 8 of Cowie & McKee 1977
+	amrex::Real min_temperature = 0.0;   // default value will be overwritten by tempFloor_ during initialization
+	int eos_flag = 1;		     // 1 == use quokka::EOS; 0 == use resampled cooling
```
BenWibking: I think it would be better to make this an `AMREX_ENUM` instead of an `int`. I am trying to migrate the code away from using integer flags with more user-friendly enums. This allows users to specify short text strings like `QuokkaEOS` or `ResampledCoolingEOS` instead of `0` or `1` in the input files.

There is an example of how to declare an `AMREX_ENUM` variable here: https://github.com/quokka-astro/quokka/blob/986814918ba21fd4179d85a960da01952a2a9e98/src/hydro/mhd_system.hpp#L26

BenWibking (changes requested):
Only one minor code cleanup comment:

src/QuokkaSimulation.hpp
```diff
@@ -1038,12 +1058,27 @@ auto QuokkaSimulation<problem_t>::addStrangSplitSourcesWithBuiltin(amrex::MultiF
 		}
 	};
 
+	auto const applyConduction = [&]() {
+		if (enableElectronConduction_ == 1) {
+			if (max_level > 0) {
+				amrex::Abort("Electron conduction not implemented for > 0 levels.");
+			}
+			fillBoundaryConditions(state, state, lev, time, quokka::centering::cc, quokka::direction::na, PreInterpState, PostInterpState);
```
BenWibking, line 1066: Because the face-centered state multifab is also used by your code to subtract the magnetic energy from the total energy, it's necessary to add a call to `fillBoundaryConditions` for `state_fc` as well.

BenWibking (changes requested):
There's one magnetic field-related issue that I missed earlier that needs to be fixed. I didn't see any other issues.

src/QuokkaSimulation.hpp
aditivijayan: I have added the following lines here-
`for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
				auto &mutable_mf = const_cast<amrex::MultiFab &>(state_fc[idim]);
				fillBoundaryConditions(mutable_mf, mutable_mf, lev, time, quokka::centering::fc,
						       static_cast<quokka::direction>(idim), AMRSimulation<problem_t>::InterpHookNone,
						       AMRSimulation<problem_t>::InterpHookNone, FillPatchType::fillpatch_function);
			}`
Using mutable_mf was suggested by Gemini to resolve the followign error-
```
/g/data/jh2/av5889/quokka_movies/quokka/src/simulation.hpp:364:14: note: candidate: ‘template<class PreInterpHook, class PostInterpHook> void AMRSimulation<problem_t>::fillBoundaryConditions(amrex::MultiFab&, amrex::MultiFab&, int, amrex::Real, quokka::centering, quokka::direction, const PreInterpHook&, const PostInterpHook&, FillPatchType) [with PostInterpHook = PreInterpHook; problem_t = ThermalConductionProblem]’
  364 |         void fillBoundaryConditions(amrex::MultiFab &S_filled, amrex::MultiFab &state, int lev, amrex::Real time, quokka::centering cen, quokka::direction dir,
      |              ^~~~~~~~~~~~~~~~~~~~~~
/g/data/jh2/av5889/quokka_movies/quokka/src/simulation.hpp:364:14: note:   template argument deduction/substitution failed:
/g/data/jh2/av5889/quokka_movies/quokka/src/QuokkaSimulation.hpp:1069:55: note:   cannot convert ‘(& state_fc_)->amrex::MultiFab::amrex::FabArray<amrex::FArrayBox>.amrex::FabArray<amrex::FArrayBox>::operator[](idim)’ (type ‘const amrex::FArrayBox’) to type ‘amrex::MultiFab&’
 1069 |                                 fillBoundaryConditions(state_fc_[idim], state_fc_[idim], lev, time, quokka::centering::fc,
      |                                 ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1070 |                                                        static_cast<quokka::direction>(idim), AMRSimulation<problem_t>::InterpHookNone,
      |                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1071 |                                                        AMRSimulation<problem_t>::InterpHookNone, FillPatchType::fillpatch_function);
      |                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
- occurring because `state_fc `is passed as a `const`.

aditivijayan: > Because the face-centered state multifab is also used by your code to subtract the magnetic energy from the total energy, it's necessary to add a call to `fillBoundaryConditions` for `state_fc` as well.

My solution is not compiling. @BenWibking do you have a recommendation?

BenWibking: I would change the function arguments for `addStrangSplitSourcesWithBuiltin` so that `state_fc` is not passed as const. This is the right thing to do, since it actually does need to modify `state_fc` now. The other operator-split physics modules did not need to before.

BenWibking (approved):
Looks good to me!

PR #1810, opened 2026-04-18
[ADR] Consistent gas temperature through EOS interface

BenWibking:
### Description
This ADR describes a proposal to unify the temperature calculations through the `quokka::EOS` interface and eliminate the physical inconsistency in the current codebase.

### Related issues
N/A

### Checklist

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record documenting a unified approach for gas-temperature calculations.
  * Defined a single thermodynamic authority to improve consistency across cooling-table and equation-of-state calculations.
  * Documented rollout considerations and regression testing for CPU and GPU execution.


BenWibking:
cc @markkrumholz @chongchonghe @aditivijayan @evan1022

BenWibking:
@codex review

chongchonghe:
I would prefer Option C, and support two functions: `ComputeIdealGasLawTemperature` to deduce `T` from `e = 1/(gamma-1) (rho / (mu)) k T`, where `mu` is the constant molecular weight in EOS_Traits, and `ComputeChemicalTemperature` (or `ComputePhysicalTemperature`) to compute T from resampled table.

BenWibking:
> I would prefer Option C, and support two functions: `ComputeIdealGasLawTemperature` to deduce `T` from `e = 1/(gamma-1) (rho / (mu)) k T`, where `mu` is the constant molecular weight in EOS_Traits, and `ComputeChemicalTemperature` (or `ComputePhysicalTemperature`) to compute T from resampled table.

@chongchonghe What is the use case for this?

There is really only one real temperature value, and I think it would be better to only have a single source of truth for this value.

chongchonghe:
We obviously don't want to use the resampled-table temperature in radiation and chemistry iterations.

BenWibking:
> We obviously don't want to use the resampled-table temperature in radiation and chemistry iterations.



> We obviously don't want to use the resampled-table temperature in radiation and chemistry iterations.

@chongchonghe I think we do want to use that temperature. Why wouldn't we?

markkrumholz:
Let me propose an architecture that I think handles both @BenWibking and @chongchonghe's goals. This is very close to the approach I take in `criptic` to handle different cosmic ray propagation models, though in the case of `criptic` it is implemented using class-based rather than template-based polymorphism as we'd probably want to do in `quokka`. Here's my basic idea:

1. `quokka::EOS` is an abstract base class provides null implementations for functions to (1) go from conserved to primitive variables, (2) derive temperature, sound speed, pressure, and required thermodynamic derivatives (e.g, gamma = d ln P / d ln rho) individually or in combination from the conserved or primitive variable struct.
2. We then define three standard derived classes that extend this: `quokka::EOSIdeal`, `quokka::EOSTabulated`, and `quokka::EOSMicrophysics`, which provide implementations of these abstract base classes for the three cases we have already defined in the code, i.e., an ideal EOS, and EOS where the conversions are tabulated, and an EOS where we are doing explicit chemistry and thus we compute these conversions based on the local chemical abundances.
3. Individual problems can then defined they own EOS classes, derived from the general `quokka::EOS` implementation. Those that just want to use one of the three standard approaches can just provide one of the three pre-defined classes, but we could also do problem-specific mix-and-match. For example, with photoionization we are going to want to use something like `quokka::EOSMicrophysics` for low-temperature gas and `quokka::EOSTabulated` for high-temperature gas, so we could define a custom EOS for that problem that uses and if to break between the two cases based on the specific internal energy or something like that.

I'd welcome thoughts on this proposal.

BenWibking:
> Let me propose an architecture that I think handles both @BenWibking and @chongchonghe's goals. This is very close to the approach I take in `criptic` to handle different cosmic ray propagation models, though in the case of `criptic` it is implemented using class-based rather than template-based polymorphism as we'd probably want to do in `quokka`. Here's my basic idea:
> 
> 1. `quokka::EOS` is an abstract base class provides null implementations for functions to (1) go from conserved to primitive variables, (2) derive temperature, sound speed, pressure, and required thermodynamic derivatives (e.g, gamma = d ln P / d ln rho) individually or in combination from the conserved or primitive variable struct.
> 2. We then define three standard derived classes that extend this: `quokka::EOSIdeal`, `quokka::EOSTabulated`, and `quokka::EOSMicrophysics`, which provide implementations of these abstract base classes for the three cases we have already defined in the code, i.e., an ideal EOS, and EOS where the conversions are tabulated, and an EOS where we are doing explicit chemistry and thus we compute these conversions based on the local chemical abundances.
> 3. Individual problems can then defined they own EOS classes, derived from the general `quokka::EOS` implementation. Those that just want to use one of the three standard approaches can just provide one of the three pre-defined classes, but we could also do problem-specific mix-and-match. For example, with photoionization we are going to want to use something like `quokka::EOSMicrophysics` for low-temperature gas and `quokka::EOSTabulated` for high-temperature gas, so we could define a custom EOS for that problem that uses and if to break between the two cases based on the specific internal energy or something like that.
> 
> I'd welcome thoughts on this proposal.

Ok, this is basically what I thought I was proposing, but I see I was not being explicit enough.

Microphysics already does this internally with template-based polymorphism. This is how all EOS classes are implemented under the hood in Microphysics. So in principle, we could just write a Microphysics EOS class for a tabulated mean-molecular weight gas. However, this would require Microphysics to know about the tabulating cooling functions, which I think would make things more complex than they currently are.  I agree that it may be the cleanest solution to reimplement template-based polymorphism for the EOS class in Quokka, and delegate to the Microphysics EOS only when doing chemistry.

I do want to avoid having to redefine the EOS for every single problem, though. We only need a small number of types of EOS, so I don't think it helps to define a separate template class for each problem's EOS.

BenWibking:
But if we go ahead and reimplement a compile-time polymorphic EOS classes in Quokka, I think we should just go ahead and copy and paste VODE from Microphysics and put it in Quokka and keep everything internally consistent in this repo.

At that point, I think it's more complicated to delegate things to Microphysics than it is to copy VODE and put it in this repo.

markkrumholz:
> I do want to avoid having to redefine the EOS for every single problem, though. We only need a small number of types of EOS, so I don't think it helps to define a separate template class for each problem's EOS.

Can't we handle this just by a single line that defines `quokka::EOS<myProb>` as an alias for `quokka::EOSIdeal` or something like that?

BenWibking:
> > I do want to avoid having to redefine the EOS for every single problem, though. We only need a small number of types of EOS, so I don't think it helps to define a separate template class for each problem's EOS.
> 
> Can't we handle this just by a single line that defines `quokka::EOS<myProb>` as an alias for `quokka::EOSIdeal` or something like that?

Ah, yes, we could probably do that.

I am not 100% sure you can literally do that, but I'm sure we can find some combination of template and macro magic to do effective the same thing.

markkrumholz:
I do think re-implementing this on the quokka side is going to be the cleanest solution, since we now have lots of different cases that microphysics doesn't know about, as you pointed out: magnetic fields, cosmology, etc. I can see advantages and disadvantages to just copying the microphysics implementation of VODE. The advantage is that we then have full control and can easily do things like, for example, implementing radiation groups to be integrated without needing to worry about interfacing with their infrastructure and doing PRs to their repo. The disadvantage is that the we do not automatically benefit from any upgrades or improvements they make to their solvers.

BenWibking:
> I do think re-implementing this on the quokka side is going to be the cleanest solution, since we now have lots of different cases that microphysics doesn't know about, as you pointed out: magnetic fields, cosmology, etc. I can see advantages and disadvantages to just copying the microphysics implementation of VODE. The advantage is that we then have full control and can easily do things like, for example, implementing radiation groups to be integrated without needing to worry about interfacing with their infrastructure and doing PRs to their repo. The disadvantage is that the we do not automatically benefit from any upgrades or improvements they make to their solvers.

I agree. That is a big downside.

Before now, I thought it was clearly the better choice to keep Microphysics as a submodule, rather than a copy-and-pasted modification, but I'm no longer sure that is the obviously better choice given how different and specialized our code is becoming.

BenWibking:
> > I do think re-implementing this on the quokka side is going to be the cleanest solution, since we now have lots of different cases that microphysics doesn't know about, as you pointed out: magnetic fields, cosmology, etc. I can see advantages and disadvantages to just copying the microphysics implementation of VODE. The advantage is that we then have full control and can easily do things like, for example, implementing radiation groups to be integrated without needing to worry about interfacing with their infrastructure and doing PRs to their repo. The disadvantage is that the we do not automatically benefit from any upgrades or improvements they make to their solvers.
> 
> I agree. That is a big downside.
> 
> Before now, I thought it was clearly the better choice to keep Microphysics as a submodule, rather than a copy-and-pasted modification, but I'm no longer sure that is the obviously better choice given how different and specialized our code is becoming.

@markkrumholz I have tested the YASS integrator from Khoklov with the primordial chemistry network, it is significantly faster than VODE or the classic Backward Euler method at comparable (relatively low) accuracy tolerances.

Since this integrator is so simple (it does not even require a Newton iteration), I think we should implement it in our code independently. I think almost all future improvements to the integrator would be entirely due to the stepsize controller (very few lines of code, easy to port), or to the linear algebra solve (likely to be problem-dependent, or require exploiting sparsity).

markkrumholz:
OK, if you think it would be useful to move away from VODE as our primary integrator, I agree that significantly weakens the case for keeping the dependence on microphysics, and we should therefore more to separate from it. That can be a goal for the next release.

BenWibking:
This can be merged, now that the implementation is done: https://github.com/quokka-astro/quokka/pull/2012.

docs/adrs/ADR-0001-unify-gas-temperature-through-eos.md
```diff
@@ -0,0 +1,115 @@
+# ADR-0001: Unify Gas Temperature Through EOS
+Date: 2026-04-18 • Status: Proposed
```
BenWibking: This should be changed to "adopted", since the implementation of this feature has already been merged.

chongchonghe:
> This can be merged, now that the implementation is done: #2012.

Do you really want to merge an ADR after it's implemented? We've never done this before, and I don't think this is a good strategy in general. 

We can put PRDs or ADRs in the docs folder in the code for the features that we plan to implement, but I think we should remove them after the code is done. If we keep ADR after the coding is complete, there will be a few issues:

1. The user will be confused because an ADR indicates that it is in plan mode and not implemented.
2. It will confuse LLMs. When LLMs are doing code reviews, they will find that the code implementation is different from the ADR, and they will raise noise. 

My recommendation is to keep PRDs or ADRs in a different place, not in the source code. It could be a branch—like a PRDs branch that everyone can push to—or it could be the Discussions tab on GitHub.

BenWibking:
If we put "status: implemented" in the ADR itself, I think that eliminates the confusion.

According to the maintainer guide that we adopted, the policy is to put the adrs in the repo once we've agreed to the design they specify.

BenWibking:
I think this is the first ADR that's been implemented, so that's why this is the first one to merge.

But according to the policy in the maintainer guide, it actually should be merged earlier, when it's accepted and we've agreed on the design it describes: https://quokka-astro.github.io/quokka/adrs.html

chongchonghe (approved):

chongchonghe:
changed to 

`Status: adopted in  [PR2012](https://github.com/quokka-astro/quokka/pull/2012)`

and approved.

BenWibking:
The AI makes a good point. We should fix it so it agrees with the actual implementation details.

chongchonghe:
> The AI makes a good point. We should fix it so it agrees with the actual implementation details.

Right, that's the second concern I raised earlier. Every implementation will not be exactly the same as the ADR, and there will be inconsistency. Then, AI review will see this inconsistency and raise noise. 

I think this is a general problem of this ADR workflow.

My recommendation is to merge an ADR for *long-term* plan only, and remove it upon completion. An ADR that we are going to implement tomorrow doesn't need a separate ADR merged into the main branch; it can simply be part of the PR. Besides, we should remove it from the repo upon merge and replace it with a documentation of the feature. Some of the content in the ADR will be useful for writing the documentation.

@markkrumholz I'd like to hear your thought as well.

BenWibking:
> > The AI makes a good point. We should fix it so it agrees with the actual implementation details.
> 
> Right, that's the second concern I raised earlier. Every implementation will not be exactly the same as the ADR, and there will be inconsistency. Then, AI review will see this inconsistency and raise noise.

I agree it should be consistent. However, I don't think the AI should cause issues if we tell it in a README or AGENTS file that ADRs are frozen historical documents. They should be smart enough now to not get confused if we do that. I haven't seen any evidence that LLMs will be confused if we put a readme or header in each ADR that they are a frozen historical record. If you have tried this and they got confused, I would be interested to hear about this.

> My recommendation is to merge an ADR for _long-term_ plan only, and remove it upon completion. An ADR that we are going to implement tomorrow doesn't need a separate ADR merged into the main branch; it can simply be part of the PR. Besides, we should remove it from the repo upon merge and replace it with a documentation of the feature. Some of the content in the ADR will be useful for writing the documentation.

I think the original purposes I had in mind for ADRs was that they would reflect a *decision record*, not just document the feature. That is, it should explain *why* we agreed that was the correct architecture for that feature, not just a detailed description of the feature. That way we can refer to it in the future if we forget why we implemented it that way, and AI agents can read the discussion to understand the architecture design choices and write new designs that are consistent with what we've historically done. For this purpose, it would be important that we keep it in the repo after it's implemented.

I do agree that for straightforward things, ADRs should not be used. It should really be used to keep a record of design issues that required significant discussion, or nontrivial architectural changes to the code (which the new EOS interface is, in my opinion), and the rationale for the design, which is the part that isn't documented elsewhere.

chongchonghe:
@BenWibking  OK, nice, you convinced me. Then I would recommend merging this ADR as is. There is no point to make it match the implementation; This will be a lot of work if we do this for every ADR. 

The important thing to do is to make it very clear in the ADR doc or in Claude Code/Codex rules to ignore this document when scanning the code.

chongchonghe:
I'll create a PR to update CLAUDE/AGENTS.md for this. 

Feel free to merge it.

PR #1820, opened 2026-04-21
Use mdBook for doc generation

### LINKED ISSUE #1819: Switching to a long-term maintainable docs generator
**Describe the proposal**
MkDocs does not appear to be a viable option to continue using for our HTML documentation generator (long story, see "Additional context" below).

We originallly used Sphinx. We switched to using MkDocs in 2024, and unfortunately, we need to transition to something new _again_. I am not happy with this, and I would prefer finding a stable solution that does not require switching documentation generators every couple of years.

My initial feeling is that we should switch to a documentation generator that has existed for 10+ years and is still being maintained. The only Markdown-based generator that satisfies this requirement is **[mdBook](https://rust-lang.github.io/mdBook/)**. Another option that is almost as long-lasting is **[Docusaurus](https://docusaurus.io/docs)**. Since mdBook is simpler, my feeling is that we should switch to mdBook.

* **Go back to Sphinx.** Unfortunately, Sphinx requires writing documentation in ReStructured Text (RST), which is a significantly less user-friendly format compared to Markdown. I don't think we should do this.
* **Keep using MkDocs 1.x** It has been abandoned. It will probably stop working in the new few years as it fails to receive updates for newer Python versions, etc. The Material for MkDocs theme we are using will not receive updates.
* **Use MkDocs 2.x** There is no open-source license for MkDocs 2.0, and it is backward-incompatible with MkDocs 1.x. It is a complete rewrite of the codebase, so there is no compelling reason to use it. It's basically a completely new tool that happens to have the same name.
* **Switch to Zensical** This is the recommendation of the authors of the theme we use (Material for MkDocs): https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/. However, it is a brand-new project, and it's not clear how long it will last. These projects come and go.
* **Switch to mdBook** This is the documentation generator used by the Rust community. It should last as long as Rust does, which may be many decades. It supports Markdown, and appears to be extremely well supported. It has existed since 2015.
* **Switch to Docusaurus** Created by Facebook, it has existed since 2018. It seems like it will last a long time. However, it requires using React, which is an unnecessary level of complexity for our project.

**Additional context**
MkDocs 1.x has been more or less abandoned. Version 2.0 does not have an open-source license. The PyPI package appears to have been the target of a hostile takeover. For the full story: https://fpgmaas.com/blog/collapse-of-mkdocs/

BenWibking:
### Description
Switches the documentation HTML generator to [mdBook](https://rust-lang.github.io/mdBook/).

The one significant change required to the Markdown documentation itself is the use of different delimiters for LaTeX math mode (https://rust-lang.github.io/mdBook/format/mathjax.html).

This documentation software has existed for over 10 years, and should last as long as Rust does.

### Related issues
Closes https://github.com/quokka-astro/quokka/issues/1819.

### Checklist


BenWibking:
@chongchonghe @markkrumholz This PR is ready to be reviewed.

chongchonghe:
@BenWibking Can you write a script/instructions to install mdbook and dependencies?

docs/markdown/developer_onboarding.md
```diff
@@ -28,10 +28,11 @@
 - **CUDA builds on macOS or Linux.** To build and test CUDA functionality locally, run `./scripts/bash/run-cuda-container.sh`. This script pulls the appropriate Docker image, launches a container, and performs a CUDA build—useful for catching CUDA-specific issues on your development machine.
 - **Writing and building documentation.** The developer is responsible for updating the documentation site. To build and view the documentation site locally:
 ```bash
-pip install -r docs/requirements.txt
-./scripts/bash/build_and_view_docs.sh
+./scripts/bash/install_mdbook.sh
```
BenWibking, line 31: Added a ref to the script here.

BenWibking: @chongchonghe Done.

scripts/bash/install_mdbook.sh
```diff
@@ -0,0 +1,16 @@
...
+
+echo "Install mdBook tooling used by the CI docs jobs"
+
+if ! command -v cargo >/dev/null 2>&1; then
+    echo "Error: cargo is required to install mdBook tooling."
+    echo "Install the Rust toolchain, then rerun this script."
+    exit 1
+fi
+
+cargo install mdbook --version "${MDBOOK_VERSION}" --locked
```
chongchonghe: this '--locked' will cause error if you don't include the .lock file in git. This happended on my macOS. I suggest either remove it, or add the Cargo.lock file you generated to git.

chongchonghe (changes requested):
This looks really good. I like that the citation is MNRAS style and that you can view the index.html file locally without hosting a http server. I left one minor comment about the '--lock' argument causing troubles.

scripts/bash/install_mdbook.sh
BenWibking: Fixed

BenWibking:
@chongchonghe This is ready to go now.

chongchonghe (approved):

PR #1825, opened 2026-04-23
Fix MHD EMF reconstruction boxes

BenWibking:
## Summary

Addresses the same uninitialized intermediate-box read reported in #1748 for the Felker-Stone 2017 EMF path, using a different reconstruction-box strategy.

The current #1748 fix expands the first-stage reconstruction regions so the later `fc->ec` reconstruction reads initialized data. This PR instead tightens the ownership contract in `ReconstructTo`: constant and PLM reconstruction write exactly the requested interface box, while PPM gets explicit scratch space only in the reconstruction direction where its stencil writes one extra interface. The EMF assembly paths then keep the physical edge/quadrant accumulators on the valid edge boxes and allocate larger scratch boxes only where those PPM side-effect writes require them.

A CI test is added that fails in the presence of the original bug, and passes for this PR.

Bug originally discovered by @AstroKriel in #1748. 

Screenshot 2026-04-23 at 5 27 21 PM

## Details

- Fixes the `FelkerStone2017` `cc->fc->ec` reconstruction path so intermediate data read by the edge reconstruction are filled before use.
- Avoids broad expansion of the physical reconstruction/output regions by separating requested output boxes from PPM scratch boxes.
- Applies the same reconstruction-box semantics to the related `Balsara2025` and `Quokka2026` EMF paths.


BenWibking:
@codex review

BenWibking:
I think this is correct and working, but it needs to be cleaned up a bit. I will do that tomorrow.

BenWibking:
@codex review

chongchonghe:
@claude review

chongchonghe (approved):

PR #1830, opened 2026-04-28
Reverse built-in Strang source order for the second half-step

mxchen39:
### Description

The GIRK method requires that, within the Strang split framework  
$$D_{\Delta t/2} H_{\Delta t} D_{\Delta t/2}$$, external forces be included in the hydrodynamics operator $$H_{\Delta t}$$. This ensures that, in the presence of external forces, the error scales as $$\propto \Delta t^{-2}$$ in the limit $$\Delta t \to \infty $$.

This PR updates the built-in Strang-split source term ordering in `advanceHydroAtLevel()`.
The source-term ordering is updated so that dust is applied first in the first source half-step:
`dust -> cooling -> chemistry -> turbulence -> user-defined sources`
The second source half-step applies the same source terms in reverse order:
`user-defined sources -> turbulence -> chemistry -> cooling -> dust`

### Implementation

- Refactor addStrangSplitSourcesWithBuiltin() to select source-term ordering at compile time via SourceOrder and callInOrder()
 - Apply the source terms in forward order for the first half-step and reverse order for the second half-step

### Testing

- Add a new `DustDampingWithExternalForce` regression test.

### Related issues

### Checklist

### DIFF
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -227,6 +227,8 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 	amrex::Long radiationCellUpdates_ = 0; // total number of radiation cell-updates
 	std::unique_ptr<quokka::turbulence::turbulentDriving<problem_t>> td;
 
+	enum class SourceOrder { forward, reverse };
+
 	// member functions
 	explicit QuokkaSimulation(amrex::Vector<amrex::BCRec> &BCs_cc, amrex::Vector<amrex::BCRec> &BCs_fc) : AMRSimulation<problem_t>(BCs_cc, BCs_fc)
 	{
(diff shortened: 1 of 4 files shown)


src/QuokkaSimulation.hpp
```diff
@@ -971,43 +971,69 @@ template <typename problem_t> auto QuokkaSimulation<problem_t>::computeExternalH
...
+		applyChemistry();
+		applyCooling();
+		applyDust();
+	} else {
+		applyDust();
+		applyCooling();
+		applyChemistry();
+		applyTurbulence();
+		applyUserSources();
+	}
```
BenWibking: We can change the order with C++ metaprogramming so it looks like this:
```
// Wrapper chosen by caller at compile time.
template <order Order>
void run_pipeline() {
    call_in_order<Order>(
        [] { std::cout << "step 1\n"; },
        [] { std::cout << "step 2\n"; },
        [] { std::cout << "step 3\n"; }
    );
}
```
Then, to run this, for example:
```
int main() {
    std::cout << "forward:\n";
    run_pipeline<order::forward>();

    std::cout << "\nreverse:\n";
    run_pipeline<order::reverse>();
}
```
Here is the full working example: https://godbolt.org/z/Yn95hEneM

BenWibking (changes requested):

BenWibking (approved):

PR #1863, opened 2026-05-01
[ADR] Optimal AMR subcycling with parabolic operators

BenWibking:
### Description
This proposes a new scheme to select the coarse and level timesteps when doing AMR subcycling that is compatible with using both hyperbolic and parabolic operators in the same simulation.

### Related issues
N/A

### Checklist

### DIFF
--- a/docs/markdown/amr_timestepping.md
+++ b/docs/markdown/amr_timestepping.md
@@ -1,0 +1,240 @@
+# AMR Timestep Selection
+
+This note describes a timestep selection algorithm for AMR simulations that contain both hyperbolic and explicitly integrated parabolic operators. The goal is to choose the coarse timestep and the AMR subcycling factors so that every level uses the largest stable timestep allowed by its own physics, subject to a cap on the number of subcycles between adjacent levels.
+
+The current AMR driver already stores one timestep per level in `dt_` and one integer subcycling factor per level in `nsubsteps`. The recursive advance only requires these subcycling factors to be positive integers. Therefore the timestep policy can choose `nsubsteps[lev]` dynamically instead of fixing it to the spatial refinement ratio.
+
+## Per-level timestep limits
+
+For each active AMR level `l`, compute a single-level stable timestep limit
+
+```text
+tau[l] = min(tau_hyperbolic[l],
+             tau_parabolic[l],
+             tau_particles[l],
+             tau_other_explicit[l],
+             change_max * old_dt[l])
+```
+
+where unavailable physics components report an infinite limit. The hyperbolic limit is the usual CFL condition. For an explicit scalar diffusion operator, a conservative parabolic limit has the form
+
+```text
+tau_parabolic = cfl_parabolic / max_cells(sum_d chi_d / dx_d^2)
+```
+
+where `chi_d` is the directional diffusivity and `cfl_parabolic` is chosen for the actual time integrator and stencil. More general parabolic operators should expose their own timestep estimator and contribute it to the minimum above.
+
+These `tau[l]` values are local stability limits for a single update on level `l`. The AMR scheduler then chooses a synchronized set of level timesteps
+
+```text
+dt[l] = dt[0] / (nsubsteps[1] * ... * nsubsteps[l])
+```
+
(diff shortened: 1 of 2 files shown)


BenWibking:
The one issue not addressed by this ADR is how to handle levels created when regridding during a coarse step. In the current code, they are assigned substep counts equal to their geometric refinement ratios. I am not sure how this should be handled with an additional parabolic operator.

BenWibking:
@markkrumholz @chongchonghe @aditivijayan @evan1022 I would appreciate your comments on this proposal.

docs/markdown/amr_timestepping.md
```diff
@@ -0,0 +1,240 @@
...
+             tau_parabolic[l],
+             tau_particles[l],
+             tau_other_explicit[l],
+             change_max * old_dt[l])
+```
+
+where unavailable physics components report an infinite limit. The hyperbolic limit is the usual CFL condition. For an explicit scalar diffusion operator, a conservative parabolic limit has the form
+
+```text
+tau_parabolic = cfl_parabolic / max_cells(sum_d chi_d / dx_d^2)
```
aditivijayan, line 22: For thermal conduction, `tau_parabolic` can be fixed using the STS.

BenWibking: Can you elaborate? I'm not sure I follow what you mean here.

aditivijayan: Sorry! I was reading the PLUTO manual where they advance using-

Image

So, I was wondering if we could use this for determining the conduction time step. 
But now I realise that would be incorrect. The correct way would be to use the hydro time and then call thermal conduction N times using dt_conduction using the above relation. 

If we do use STS for conduction I guess that would obviate the need to compare hydro and conduction timesteps.

BenWibking: Ok, I understand now. Using STS for the parabolic operator during each update would work if both i) it was stable for nonlinear diffusion problems and ii) we could limit the the number of STS stages $N$. However, neither of these are true in general, so there will (unfortunately) always be cases where we have to fall back to using a single explicit non-STS update per level advance, for the following reasons:

STS is provably stable with constant diffusion coefficients, but there is no proof of stability when the coefficients can change as a function of the solution. In practice, it is often found that is not stable (using the timestep that you wrote above) when the coefficient is not constant.

Even when stable, the number of STS stages may be very large if we want to match the hydro timestep. For Spitzer conduction with realistic cell sizes, this can require several hundred stages per single hydro timestep. In this case, the coarse-fine correction at AMR level boundaries can cause the energy to become negative and the simulation will crash.

AstroKriel (commented):
Hey, this looks good to me. I have a question though.

Previously, floors and dual-energy sync's were applied after the hydro advance and AMR reflux. From what I can tell, this PR pushes these floors until after each Strang half-step. Is this intentional, and could it affect the stability/outputs?

BenWibking:
> Hey, this looks good to me. I have a question though.
> 
> Previously, floors and dual-energy sync's were applied after the hydro advance and AMR reflux. From what I can tell, this PR pushes these floors until after each Strang half-step. Is this intentional, and could it affect the stability/outputs?

@AstroKriel I have no idea what's you're talking about. I don't see anywhere in this proposal where floors or dual-energy is mentioned at all. Can you point me to what you're referring to?

docs/markdown/amr_timestepping.md
aditivijayan: Ah okay. I understand thanks Ben!

AstroKriel:
> > Hey, this looks good to me. I have a question though.
> > Previously, floors and dual-energy sync's were applied after the hydro advance and AMR reflux. From what I can tell, this PR pushes these floors until after each Strang half-step. Is this intentional, and could it affect the stability/outputs?
> 
> @AstroKriel I have no idea what's you're talking about. I don't see anywhere in this proposal where floors or dual-energy is mentioned at all. Can you point me to what you're referring to?

You're right, sorry. Claude led me astray when I tried to think through how this would be implemented. I should've double-checked before commenting.

PR #1872, opened 2026-05-07
[CI] Docs: scope permissions, remove unpinned Rust action, lock Cargo installs

BenWibking:
### Motivation
- A CI supply-chain finding showed docs workflows executed an unpinned third-party Rust action and ran `cargo install` without `--locked` while build steps inherited elevated `pages: write`/`id-token: write` permissions. 
- The intent is to eliminate the attack surface by scoping permissions, avoiding mutable action refs for setup, and making Cargo installs reproducible.

### Description
- Removed workflow-level `pages: write` and `id-token: write` and limited the `build` job to `contents: read` while granting `pages: write` and `id-token: write` only on the `deploy` job in `.github/workflows/docs.yml`. 
- Replaced `uses: dtolnay/rust-toolchain@stable` with explicit runner-local installation of `rustc`/`cargo` via `apt` in both `.github/workflows/docs.yml` and `.github/workflows/docs-toctree.yml` to avoid a mutable third-party action ref. 
- Made mdBook installs reproducible by adding `--locked` to `cargo install` calls in `scripts/bash/install_mdbook.sh`.

### Testing
- Ran `bash -n scripts/bash/install_mdbook.sh` to validate the script syntax and it passed. 
- Attempted YAML parsing validation with `python`/`yaml.safe_load`, but this check failed in the environment due to missing `PyYAML` (the failure was environmental, not a YAML syntax regression).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69fcfc50c9f88321a4838a3f5a2af395)

### DIFF
--- a/.github/workflows/docs-toctree.yml
+++ b/.github/workflows/docs-toctree.yml
@@ -14,7 +14,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Set up Rust
-        uses: dtolnay/rust-toolchain@stable
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends rustc cargo
       - name: Set up Python
         uses: actions/setup-python@v6.2.0
         with:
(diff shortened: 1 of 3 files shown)


BenWibking:
This is probably not worth it.

PR #1886, opened 2026-05-12
Fix resampled cooling internal energy with MHD

### LINKED ISSUE #1885: ResampledCooling does not subtract magnetic energy when computing eint
### Discussed in https://github.com/quokka-astro/quokka/discussions/1883

<div type='discussions-op-text'>

<sup>Originally posted by **YaoguangPei** May 11, 2026</sup>
I found a cooling-related issue when testing an MHD run with resampled cooling in QUOKKA.

My setup is:
- is_hydro_enabled = true
- is_mhd_enabled = true
- is_radiation_enabled = false
- periodic box
- uniform initial magnetic field
```
  [
  \mathbf{B}0 = (5,\mu{\rm G},,0,,0)
  ]
```
The main issue seems to be in the cooling path:

In cooling/ResampledCooling.hpp, cooling updates use
```cpp const Real Eint = RadSystem<problem_t>::ComputeEintFromEgas(rho, x1Mom, x2Mom, x3Mom, Egas); ```

However, in src/radiation/radiation_system.hpp, the implementation is currently

```cpp const double p_sq = X1GasMom * X1GasMom + X2GasMom * X2GasMom + X3GasMom * X3GasMom; 
const double Ekin = p_sq / (2.0 * density); 
const double Eint = Etot - Ekin; 
```
and the inverse function is
```
cpp const double Etot = Eint + Ekin; 
```
This is valid for pure hydro, but not for MHD, because magnetic energy is not included.

For MHD, the gas internal energy should instead be recovered as
```
[
E{\rm int} = E_{\rm tot} - E_{\rm kin} - E_{\rm mag},
\qquad
E_{\rm mag} = \frac{B^2}{8\pi},
]
and similarly the inverse relation should include magnetic energy:
[
E_{\rm tot} = E_{\rm int} + E_{\rm kin} + E_{\rm mag}.
]
```
Because ResampledCooling.hpp uses ComputeEintFromEgas() to obtain the thermal energy before cooling integration, this means that in MHD runs the cooling module may be evolving an incorrect internal energy.

This likely causes inconsistent thermodynamics in MHD + cooling runs.

In my outputs, I observed:
- magnetic field initialization looks correct
- but thermal quantities become problematic after evolution
- inferred gas pressure becomes negative
- plasma beta becomes negative
- temperature can collapse to extremely small values

So the cooling-related concern is:

ResampledCooling.hpp currently relies on RadSystem<problem_t>::ComputeEintFromEgas(), but that function appears to assume pure hydro and does not subtract magnetic energy. Therefore, the cooling update may be using the wrong gas internal energy in MHD runs.`

This suggests that the current resampled cooling path is not MHD-safe unless internal energy recovery is modified to include magnetic energy.</div>

BenWibking:
## Summary
- Pass face-centered MHD state into `ResampledCooling::computeCooling`.
- Recover gas internal energy in the resampled cooling module as total gas energy minus kinetic and magnetic energy.
- Convert `ResampledCoolingTest` to exercise the MHD path in 3D with a uniform magnetic field initialized at plasma beta ~ 1.

Fixes #1885.

## Validation
- `scripts/bash/quokka config -d 3d-debug`
- `cmake --build build/3d-debug --target ResampledCoolingTest`
- `ctest --test-dir build/3d-debug -R ResampledCoolingTest --output-on-failure`

### DIFF
--- a/src/cooling/ResampledCooling.hpp
+++ b/src/cooling/ResampledCooling.hpp
@@ -217,6 +217,7 @@ auto computeCooling(amrex::MultiFab &mf, std::array<amrex::MultiFab, AMREX_SPACE
 		const amrex::Box &indexRange = iter.validbox();
 		auto const &state = mf.array(iter);
 		auto const &nsubsteps = nsubstepsMF.array(iter);
+
 		std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM> state_fc{};
 		if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
 			for (int dir = 0; dir < AMREX_SPACEDIM; ++dir) {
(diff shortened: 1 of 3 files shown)


src/problems/ResampledCoolingTest/testResampledCoolingTest.cpp
```diff
@@ -161,6 +191,86 @@ template <> void QuokkaSimulation<ResampledCoolingTest>::computeAfterTimestep()
 	}
 }
 
+auto checkMagneticFieldCoolingEquivalence(QuokkaSimulation<ResampledCoolingTest> &sim) -> bool
```
BenWibking: This is redundant, since we check against a reference solution computed without magnetic fields enabled.

src/cooling/ResampledCooling.hpp
```diff
@@ -218,18 +217,27 @@ auto computeCooling(amrex::MultiFab &mf, const Real dt_in, resampled_tables &res
...
 			const Real x3Mom = state(i, j, k, HydroSystem<problem_t>::x3Momentum_index);
 			const Real Egas = state(i, j, k, HydroSystem<problem_t>::energy_index);
+			const Real Ekin = (x1Mom * x1Mom + x2Mom * x2Mom + x3Mom * x3Mom) / (2.0 * rho);
+			const Real Emag = HydroSystem<problem_t>::ComputeMagneticEnergy(i, j, k, &state_fc);
 
 			// number density
 			const Real nH = rho * tables.cloudy_H_mass_fraction / C::m_p; // unit: cm^-3
 
-			const Real Eint = RadSystem<problem_t>::ComputeEintFromEgas(rho, x1Mom, x2Mom, x3Mom, Egas);
+			const Real Eint = Egas - Ekin - Emag;
```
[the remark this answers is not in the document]
BenWibking: Fixed.

src/problems/ResampledCoolingTest/testResampledCoolingTest.cpp
BenWibking: Fixed.

markkrumholz (approved):

PR #1892, opened 2026-05-13
Add support for reduced speed of light approximation for photochemistry

James471:
### Description

Add support for reduced speed of light approximation, and test coupling of photochemistry with hydro by a D-Type test.

### Related issues
Dependent on #1719 

### Checklist


James471:
@claude review

BenWibking:
auggie review

James471:
@BenWibking I have changed to the target branch from dev to photoionization. That has reduced the size of the PR. Would augment review work now?

James471:
auggie review

chongchonghe:
# Review: Quokka PR #1892 — Add support for reduced speed of light approximation for photochemistry

Reviewer: Claude Sonnet 4.6 (automated)
Review type: Feature — SPEC → VERIFY → EDGE
Base branch: `photoionization` (PR #1719)

---

## SPEC

**What the PR adds:**

- RSLA (Reduced Speed of Light Approximation) support for the photoionization chemistry. The factor `frac = c_hat / c` is applied to all photoionization rates in `actual_rhs.H`, so when `c_hat < c` the photon absorption rate is correctly proportional to the reduced speed. This allows realistic photon transport timescales while using a coarser radiation timestep.
- A new test problem `StromgrenSphereRSLA` that validates Stromgren sphere expansion with `c_hat = c/100` against an analytical RK4-integrated front-position formula.
- A new test problem `HeatingCoolingTest` that runs a single-zone photochemistry + heating/cooling evolution and writes a temperature time series to CSV (no quantitative pass/fail criterion).
- A new network `H_ionization_TInd` — a temperature-independent variant of the hydrogen ionization network with zero recombination and zero recombination-cooling coefficients.
- A source term correction in `source_terms_single_group.hpp`: the stellar luminosity source `Src` is no longer divided by `cscale = c / c_hat`. This is physically correct: a star emits L erg/s regardless of `c_hat`, so the physical radiation energy density should increase by `S * dt` per step; the RSLA conserved quantity `E_gas + (c/c_hat) * E_rad` then correctly increases by `(c/c_hat) * S * dt`.
- Several physics corrections to `photoionization/actual_rhs.H`:
  - Bug fix: collisional ionization activation energy corrected from 6.4e-12 erg (4 eV) to 2.18e-11 erg (13.6 eV, the correct H ionization threshold).
  - KI cooling second term: coefficient changed from `14` to `1.4e-2`, fixing a factor-of-1000 error relative to Koyama & Inutsuka (2002).
  - KI cooling first term: characteristic temperature changed from 114800 K to 118400 K.
  - Recombination cooling: replaced `k_B * T * alpha_rec * (0.684 - 0.0416 * log(T/1e4))` with `6.1e-10 * k_B * T^0.11`.
  - Complete analytical temperature-derivative functions added to all rate coefficients (used in the Jacobian energy column).
  - Default `recombination_temperature_dependent` changed from 0 → 1.
  - Added `allow_mixed_cell_cooling` parameter to suppress molecular heating/cooling in partially-ionized mixed cells.

---

## VERIFY / EDGE

| Phase | What ran | Result |
|---|---|---|
| VERIFY – StromgrenSphereRSLA | `StromgrenSphereRSLA inputs/StromgrenRSLA.toml` (16³, c_hat=c/100, T-independent recombination) | **PASS** — effective front radius within one cell diagonal of analytical RK4 solution at all times |
| VERIFY – HeatingCoolingTest | `HeatingCoolingTest inputs/HeatingCoolingTest.toml` | **PASS** (always exits 0; see concern below) |
| EDGE – OneZonePhotoionization | `OneZonePhotoionization inputs/OneZonePhotoionization.toml` | **PASS** — L1 norms: species ~1e-12, photons ~1e-14 |
| EDGE – OneZonePhotoionizationWithEnergy | same binary, `inputs/OneZonePhotoionizationWithEnergy.toml` | **PASS** — L1 norms: species ~1e-12, photons ~5e-15, energy ~1e-8 |
| EDGE – StromgrenSphere (no recombination) | `StromgrenSphere inputs/StromgrenNoRecombination.toml` | **PASS** — photon conservation error 3.4e-14 |
| EDGE – StromgrenSphere (T-independent recombination) | `StromgrenSphere inputs/StromgrenTempIndependentRecombination.toml` | **PASS** — all radii within analytical tolerance |

---

## Feature Assessment

The RSLA implementation is physically sound. The `frac = c_hat / c` factor correctly scales the photoionization rate so that the full c-speed radiation source-sink balance (Q ionizations/s) is achieved via the product `c_hat * n_gamma_RSLA = c * n_gamma_physical`. The source term change from `/ cscale` to `/ 1.0` in `source_terms_single_group.hpp` is the matching fix on the injection side: the star injects physical energy L dt regardless of c_hat, and the RSLA conserved energy increases by (c/c_hat) × L dt as expected.

The physics fixes (collisional ionization energy, KI cooling coefficient) are substantive corrections that will change thermochemistry results from the base PR significantly.

---

## Bugs Found

### 1. `testStromgrenSphere.cpp` — accidental code duplication from merging

Three artifacts were introduced at `src/problems/StromgrenSphere/testStromgrenSphere.cpp`:

**Double anonymous namespace (lines 118–173):**
```cpp
namespace   // ← outer (original)
{
namespace   // ← inner (accidentally added by this PR)
{
// ... integrate_radius ...
} // namespace  ← closes inner
} // namespace  ← closes outer
```
`integrate_radius` ends up in a nested unnamed namespace. It still compiles and is callable from the same TU, but the nesting is clearly accidental.

**Redundant nested condition (lines 130–134):**
```cpp
if (dt_target <= 0.0_rt) {
    if (dt_target <= 0.0_rt) {  // ← duplicate check added by this PR
        return R0;
    }
}
```
The inner check is a no-op.

**Duplicate type alias (lines 319–320):**
```cpp
using hist_t = int;
using hist_t = int;  // ← duplicate added by this PR
```
C++ allows redeclaring a typedef with the same type so this compiles, but it is dead and confusing.

All three look like copy-paste artifacts from `testStromgrenSphereRSLA.cpp` being merged into the existing file.

---

## PR-Introduced Concerns

### 2. `HeatingCoolingTest` has no pass/fail criterion

`testHeatingCooling.cpp` always returns `status = 0` regardless of outcome. The test writes `heating_cooling_quokka_output.csv` but never reads it, verifies it, or compares it to a reference. Any regression in heating or cooling will not be detected by CI.

**Recommendation:** Either add a quantitative tolerance check (e.g., verify the equilibrium temperature converges to within X% of the analytical value), or label the test explicitly as a manual-inspection/plotting helper and exclude it from CTest.

### 3. `DTypeFront` test is not registered in CTest

`src/problems/DTypeFront/CMakeLists.txt` has the `add_test` line commented out:
```cmake
# add_test(NAME DTypeFront COMMAND ${JOB_NAME} ..inputs/DTypeFront.toml ...)
```
The binary is built but never run in CI. If this test is intended to be an optional manual run, the file should document that. If it is meant to be a CI test, the `add_test` line should be uncommented (and the typo `..inputs/` → `../inputs/` fixed).

### 4. `src/networks/H_ionization_TInd` — hardcoded VODE indices

The new `H_ionization_TInd/actual_rhs.H` uses hardcoded VODE indices 5 and 6 for the photon equations:
```cpp
ydot(5) = -ionization_term / frac;
ydot(6) = -state.xn[1] * c * sigma_photo * state.rn[1] / frac;
```
This is the same fragility noted in the PR 1719 review for `photoionization/actual_rhs.H`. The same concern applies: these should use `net_ienuc + 1` and `net_ienuc + 2` to remain correct if the species count ever changes.

### 5. `actual_rhs.H` — collisional ionization uses unscaled `n_e` when `assume_sharp_ionization_front = 1`

In the new code, `collisional_ionization_term = n_e * n_HI * sigma_coll` uses the raw `n_e`, not `n_e_eff = n_e * scaling`. For `assume_sharp_ionization_front = 0` (all current tests), `scaling = 1` and the two are identical. When `assume_sharp_ionization_front = 1`, the old code used `n_e_eff * n_HI / scaling = n_e * n_HI` and the new code uses `n_e * n_HI` directly — so the behavior is actually the same. However, the intent of the `n_HI_eff = n_HI * scaling` substitution for photoionization (to account for only the fraction of the cell that is neutral) was to scale `n_HI` for the sharp-front model, but `n_e` is not scaled in either code path. This asymmetry should be documented.

### 6. Rate coefficient numerical changes affect existing test tolerance budgets

The corrections to the KI cooling coefficient (factor of 1000), the collisional ionization threshold (factor of 3.4), and the recombination cooling formula are all silent physics changes relative to the `photoionization` base branch. The existing `OneZonePhotoionization` tests still pass, but only because those tests disable these processes (`collisional_ionization_switch=0`, `KI_*_switch=0`). Any future tests that enable these processes will see different physics from what was intended in the original `photoionization` PR, without an obvious paper trail.

**Recommendation:** Add a brief comment to each changed coefficient citing the source reference (paper + equation number).

---

None.

---

## Minor Items

- `inputs/HeatingCoolingTest.toml`, `inputs/StromgrenRSLA.toml`, and `src/networks/H_ionization_TInd/Make.package` are all missing trailing newlines.
- `src/problems/DTypeFront/CMakeLists.txt`, `src/problems/StromgrenSphereRSLA/CMakeLists.txt`, and `src/problems/StromgenSphereConstTemp/testStromgenSphereConstTemp.cpp` are also missing trailing newlines.
- `src/problems/StromgenSphereConstTemp/` has no `CMakeLists.txt` and is never built; it appears to be an analysis/exploration directory that should either be removed or turned into a proper test.
- `testHeatingCooling.cpp:SetRadEnergySource` reads `stromgen.Q` via ParmParse but the TOML uses `photoionization.Q`, and Q is not used in the function anyway (the radiation energy is set to a hardcoded constant `3.9e-6 * GetChemBandQuanta(0) / 8.0`). The dead `Q` read should be removed.

---

## Summary

All six test scenarios pass: StromgrenSphereRSLA (new), HeatingCoolingTest (new), OneZonePhotoionization (no energy / with energy), and both StromgrenSphere variants. The RSLA implementation is physically correct and the source term fix is well-motivated. The most important items to address before merging are: (1) the three code-duplication artifacts in `testStromgrenSphere.cpp` (Bug #1), (2) the missing pass/fail criterion in `HeatingCoolingTest` (Concern #2), and (3) citations for the numerically changed rate coefficients (Concern #6).

BenWibking:
I think ChongChong's comment in PR https://github.com/quokka-astro/quokka/pull/1932 closed this PR. I've reopened it manually.

James471:
> I think ChongChong's comment in PR #1932 closed this PR. I've reopened it manually.

Thanks Ben!

James471:
@claude review

James471:
@claude review

chongchonghe:
@James471 Note that GPU ci tests need to be manually triggered via `\azp` `run` . After changing PR target to development branch, in order to trigger all tests, you may need a dummy commit and push, but this is not necessary for now since everything is checked in GPU CI anyway.

James471:
> @James471 Note that GPU ci tests need to be manually triggered via `\azp` `run` . After changing PR target to development branch, in order to trigger all tests, you may need a dummy commit and push, but this is not necessary for now since everything is checked in GPU CI anyway.

Thanks ChongChong!

chongchonghe (dismissed):

chongchonghe (dismissed):

BenWibking:
This is the full PR review produced by Claude Fable:

> 
> Summary of the change
> 
> The PR removes the c_hat == c restriction from the photoionization network: photoionization, photon absorption, and photoheating rates are scaled by frac = ĉ/c (Rosdahl & Teyssier-style RSLA), an analytical Jacobian is added, several cooling/heating fits are corrected, and two new 3D test problems (StromgrenSphereRSLA, DTypeFront) are added with chem-band-aware source-term injection in RadSystem.
> 
> Blocking issues
> 
> 1. CI is gutted — must be reverted before merge.
> Both .ci/azure-pipelines.yml:147-152 and .ci/azure-pipelines-amdgpu.yml:85-90 now build only the DTypeFront target and run only ctest -R "DTypeFront". The entire rest of the suite (and the RadMatterCoupling exclusion) is disabled. This was clearly a debugging change (commit 54586e806 "test DType only"), but it matters doubly here because this PR changes physics that existing tests depend on (see `#2`). Restore the original --parallel 16 build and ctest -E "RadMatterCoupling*" lines.
> 
> 2. Behavior of existing tests changes silently, and CI can't catch it.
> Three changes alter answers for pre-existing problems, and none of them are currently exercised by CI because of `#1`:
> - src/networks/photoionization/_parameters:7 flips the default recombination_temperature_dependent from 0 to 1. OneZonePhotoionizationWithEnergy.toml does not pin this parameter (it pins the other switches), so that test now runs with temperature-dependent recombination where it previously didn't.
> - The recombination cooling fit was replaced entirely (actual_rhs.H:69-74), and OneZonePhotoionizationWithEnergy runs with energy_switch = 1 and recombination cooling on.
> - The electron-mass fix in OneZonePhotoionization/CMakeLists.txt and StromgrenSphere/CMakeLists.txt (1.0 amu → 5.4858e-4 amu) changes rho and hence specific energies in those tests — the removed comment said the mass was "never used", but it is used, via spmasses in photochemistry.hpp:60 and the initial conditions. The fix is correct; the point is those tests need re-validation.
> 
> Please re-run the full photoionization test set locally (quokka run --filter 'OneZone|Stromgren') and update expected values if needed before merging.
> 
> 3. Chem-group ordering contradiction in the multi-group source term.
> source_terms_multi_group.hpp:641-646 assumes "the last NChemBands groups are ionizing photon groups" and skips the ĉ/c scaling for g >= nGroups_ - NChemBands. But photochemistry.hpp:62-67 maps chem bands to the first NumChemBands groups (radEnergy_index + numRadVarsPerGroup * nn with nn from 0). Today every chem problem has nGroups == NumChemBands == 1, so the two conventions coincide — but the moment someone builds the mixed thermal+ionizing problem this code claims to support, the scaling will be applied to exactly the wrong groups. Either make photochemistry.hpp read the last groups, or change the source-term condition to g < NChemBands, and add a static_assert (or AMREX_ALWAYS_ASSERT) that the two agree — e.g. assert NChemBands == 0 || NChemBands == nGroups_ until mixed groups are actually supported and tested.
> 
> 4. New problems compile for 2D but index out of bounds in 2D.
> Both new CMakeLists.txt files guard with if(AMReX_SPACEDIM GREATER_EQUAL 2), but:
> - testDTypeFront.cpp:222-225 computes src_k/frac_z from prob_lo[2] and dx[2] unconditionally — out-of-bounds on a GpuArray<Real, 2> in a 2D build (the AMREX_SPACEDIM >= 3 ternaries further down don't help).
> - testStromgrenSphereRSLA.cpp:158-165 loops over k and reads prob_lo[2], dx[2] unconditionally.
> - Both tests are intrinsically 3D anyway (compute_effective_radius uses cbrt and 4π/3; the analytic solutions are spherical).
> 
> Change both guards to if(AMReX_SPACEDIM EQUAL 3), matching the existing StromgrenSphere/CMakeLists.txt.
> 
> 5. RSLA test compares against the non-RSLA analytic solution.
> testStromgrenSphereRSLA.cpp:329 passes C::c_light into integrate_radius, whose denominator Q/c_light + 4πR²n is the photon-buildup term that limits early front propagation. In an RSLA run the simulated photon density is Q/(4πR²ĉ), so this term should be Q/ĉ — and with ĉ = c/100 the difference is material: the buildup term dominates out to R ≈ 1.2e21 cm and is still a ~5% front-speed correction at r_trunc ≈ 4.9e21 cm where checking begins. Testing that the front is correctly limited by ĉ is the whole point of an RSLA test. Pass the file's own c_hat constant instead (and rename the parameter accordingly).
> 
> Physics I verified as correct
> 
> - RSLA scaling (actual_rhs.H): ionization rate ĉσn_HI·n_γ, photon destruction, flux attenuation, and photoheating all carry frac = ĉ/c, while collisional ionization and recombination don't — consistent with the RT13 convention. The unscaled chem-band injection in source_terms_single_group.hpp:56 is the matching half of this convention (steady state gives N = Q/(4πr²ĉ), so ĉσN reproduces the physical rate).
> - Analytical Jacobian: I re-derived all entries. dRecombination_coefficient_dT, dCollisional_ionization_coefficient_dT, dKI_cooling_coefficient_dT, dRecombination_cooling_coefficient_dT, d_ion_ff_cooling_coefficient_dT, and every jac(i,j) including the dTde = 1/dedT chain-rule entries check out. Terms that look "missing" (∂σ_photo/∂T, ∂Γ_KI/∂T) are legitimately zero since those coefficients are constants.
> - Coefficient fixes: collisional ionization threshold 6.4e-12 erg (4 eV — wrong) → 2.18e-11 erg (13.6 eV — correct); KI cooling 14√T → 1.4e-2√T and 114800 → 118400 matches the corrected Koyama & Inutsuka fit per Vázquez-Semadeni et al. (2007).
> - Corner-source bookkeeping: in both tests the source sits at the domain corner with reflecting BCs; the kernel normalization deposits exactly Q/8 per octant and compute_effective_radius multiplies the ionized volume by 8. Consistent (though note DTypeFront folds the 1/8 into the kernel normalization while StromgrenSphereRSLA divides L_star by 8 explicitly — worth a comment, since the asymmetry looks like a bug at first read).
> - D-type analytics: Spitzer solution, t_s = r_s/c_i with μ = 0.5, and α_B(T_eq) with the −0.7 exponent all agree with the network settings in DTypeFront.toml.
> 
> GPU safety
> 
> No violations found. Device lambdas capture only GpuArrays, Array4/MultiArray4, and scalars by value; spmasses and the network parameters are GPU-managed globals (same pattern as existing tests); the host-side temperature check in testDTypeFront.cpp:451-453 correctly stages device data through a pinned-arena FArrayBox with a sync.
> 
> Minor / style
> 
> - _parameters:30: allow_mixed_cell_cooling actually gates KI heating and cooling (molecular_process_switch multiplies both, actual_rhs.H:125-126), and the comment "Switch off cooling in mixed cells" is backwards (the switch allows cooling when 1). Rename to something like mixed_cell_molecular_processes or fix the comment.
> - The mixed-cell switch block is duplicated verbatim in actual_rhs and actual_jac — factor into a small helper to keep the two from drifting.
> - actual_rhs.H:59: "// Fix KI cooling function based on…" — temporal comment; make it evergreen, e.g. "// KI cooling function per Vázquez-Semadeni et al. (2007), corrected from Koyama & Inutsuka (2002)".
> - Λ_rec's hard cutoff at T < 100 K (actual_rhs.H:70) puts a discontinuity in the RHS and Jacobian, which stiff integrators handle poorly near the threshold; consider a smooth taper.
> - Both new test files have wrong \file doc comments — testDTypeFront.cpp:6-7 describes "the static Stromgren sphere", and testStromgrenSphereRSLA.cpp:6 names the wrong file.
> - testDTypeFront.cpp:479 calls the 0.01 < x_HII < 0.99 window "cavity", but that window selects the *mixed front* cells and excludes the fully ionized interior (equilibrium x_HII ≈ 0.999 here). The check still works because the equilibrium temperature is independent of x_HII when KI processes are off, but the name and the comparison against T_ion_eq are misleading — rename to "front/mixed region" or widen to x_HII > 0.01. Also note the 0.01/0.99 bounds are hard-coded duplicates of mixed_cell_lb/ub.
> - testStromgrenSphereRSLA.cpp:147: pp2.query("n_cell", n) reads a scalar from the array-valued amr.n_cell — fragile ParmParse usage; derive the cell count from dx and the domain extent you already have.
> - SimulationData<DTypeFront>::recombination_switch is declared but never used.
> - The AMREX_SPACEDIM >= 3 inline ternaries in testDTypeFront.cpp:228-250 need // NOLINT(misc-redundant-expression) per project convention (and the always-constant RadSystem_NChemBands<problem_t>::value > 0 operand in source_terms_multi_group.hpp:644 may trip the same check when NChemBands is 0) — clang-tidy hasn't run on this branch since CI was narrowed.
> - Signed/unsigned loop comparisons (int i < t_vec_.size()) in both test files' check loops.
> 
> Test completeness
> 
> Both new problems are complete per the checklist: test*.cpp + CMakeLists.txt with JOB_NAME and add_test, plus inputs/*.toml (problem dirs are auto-registered via the GLOB in src/problems/CMakeLists.txt). One naming nit: the input is StromgrenRSLA.toml while everything else is named StromgrenSphereRSLA — consider renaming for discoverability. The new Jacobian path is exercised (integrator.jacobian = 1 in both tomls), which is good.
> 
> Bottom line: the physics core of this PR is solid — the RSLA scaling and the analytic Jacobian both check out exactly. But items 1–5 (CI restoration, re-validating existing tests against the changed defaults/fits, the chem-group ordering contradiction, the 2D out-of-bounds guards, and the analytic ĉ in the RSLA test) should all be addressed before merge.

BenWibking:
I think the only major issue it identified is this:

* Chem-group ordering contradiction in the multi-group source term:
  `source_terms_multi_group.hpp:641-646` assumes "the last NChemBands groups are ionizing photon groups" and skips the ĉ/c scaling for g >= nGroups_ - NChemBands. But photochemistry.hpp:62-67 maps chem bands to the first NumChemBands groups (radEnergy_index + numRadVarsPerGroup * nn with nn from 0). Today every chem problem has nGroups == NumChemBands == 1, so the two conventions coincide — but the moment someone builds the mixed thermal+ionizing problem this code claims to support, the scaling will be applied to exactly the wrong groups. Either make photochemistry.hpp read the last groups, or change the source-term condition to g < NChemBands, and add a static_assert (or AMREX_ALWAYS_ASSERT) that the two agree — e.g. assert NChemBands == 0 || NChemBands == nGroups_ until mixed groups are actually supported and tested.

BenWibking:
However, it does think there is a better way to compute the I-front position:

----

The current volume-integral estimator is actually a defensible choice — it's photon-weighted, uses every cell, and is insensitive to angular noise. But it has real weaknesses, and there are a few standard alternatives from the RT comparison literature that would make these tests both sharper and more diagnostic. Here's my assessment.

Weaknesses of the current approach

1. It conflates front position with everything else. compute_effective_radius integrates x_HII over the whole domain, so residual neutral fraction in the cavity (~1e-3 here), diffusive photon leakage ahead of the front (an M1/flux-limiter artifact), and the actual front location all fold into one number. A compensating pair of errors passes silently.
2. It assumes sphericity. R-type fronts on Cartesian grids develop grid-aligned anisotropy that depends on the closure/flux limiter. The volume estimator averages this away — which is exactly what you don't want a test to do.
3. DTypeFront checks only the endpoint. A single final-time comparison can pass by coincidence if the numerical and analytic R(t) curves cross. The StromgrenSphereRSLA test checks the whole series, which is strictly stronger; DTypeFront should too.
4. Cell-based tolerances conflate discretization error with model error. For the D-type test in particular, the Spitzer solution R(t) = r_s(1 + 7t/4t_s)^{4/7} is itself only an approximation — STARBENCH (Bisbas et al. 2015) showed codes correctly track between Spitzer and Hosokawa–Inutsuka, with the early phase closer to HI. A tolerance in "cells" silently absorbs that model error and won't tighten under resolution refinement.

What I'd recommend, in order of value

1. Add a photon-conservation check (highest value, and cheap). Verify cumulative injected photons balance the ionization/recombination budget:

Q·t ≈ N_HII(t) − N_HII(0) + ∫₀ᵗ dt ∫ α_B n_e n_HII dV

evaluated discretely from the state you already have at each computeAfterTimestep. This is a tight check (percent-level, not cell-level) and it directly catches the class of bug this PR is about — a missing or doubled ĉ/c factor shifts the photon budget immediately, even when the late-time Strömgren radius looks fine because the equilibrium radius is independent of ĉ. The radius check alone can't distinguish "correct RSLA transient" from "wrong transient that settles to the right equilibrium"; this can.

2. Measure the front from the radial profile, not the volume integral. Bin x_HI into spherical shells around the source (or extract along the x-axis and the body diagonal) and locate the x_HI = 0.5 crossing by linear interpolation between shells. This is the standard I-front definition from the RT comparison project (Iliev et al. 2006/2009), it gives sub-cell accuracy, and it's independent of the cavity's residual neutral fraction. Measuring along axis vs. diagonal separately also gives you an anisotropy diagnostic for free — worth asserting on, since that's where M1/FLD artifacts show up first.

3. Check the front width too. Record the radii where x_HI = 0.1 and 0.9. The width should be a few cells (resolution-limited here, since the physical mean free path is under-resolved); a front that suddenly smears over many cells is a regression the position check won't see until it's large.

4. Check the full time series in DTypeFront, as StromgrenSphereRSLA already does — e.g., max of |r_eff − r_ana| over all outputs past the initial Strömgren-establishment phase, or bracket the numerical curve between Spitzer (lower) and Hosokawa–Inutsuka (upper) per STARBENCH rather than testing distance to Spitzer alone. The bracket is the more physically honest assertion and lets you tighten the tolerance.

5. For the RSLA test specifically, test the early phase — it's the whole point. The RSLA signature is the photon-buildup-limited transient at small R, but the test currently skips everything inside r_trunc because the Gaussian source is enormous (σ = 1.5/16 of the box). If you swap in the compact Wendland kernel from DTypeFront (2-cell support), you can start checking R(t) almost immediately, where the Q/ĉ term dominates and where a ĉ-scaling bug is maximally visible. Right now the test only inspects the regime where the RSLA correction has decayed to ~5% — i.e., it's weakest exactly where it should be strongest. (This pairs with the C::c_light → c_hat fix from my review; with the analytic solution corrected, the early phase becomes a genuinely discriminating test.)

If you only do two of these, I'd pick the photon-conservation check and the profile-based front position: together they decouple "are photons being counted right" from "is the front in the right place," which is precisely the separation this PR needs to validate the ĉ/c factors independently of the equilibrium solution.

BenWibking:
The papers it cites do appear to be apt:
* Bisbas et al. (2015): https://scixplorer.org/abs/2015MNRAS.453.1324B/abstract
* Iliev et al. (2006): https://scixplorer.org/abs/2006MNRAS.371.1057I/abstract

chongchonghe:
We discussed and planned the more robust I-front calculation, including I-front width calculation. We will do these in post-processing for the paper; It's not necessary to do it in the problem file for the CI tests. 

> Add a photon-conservation check (highest value, and cheap). Verify cumulative injected photons balance the ionization/recombination budget:

I've thought about this, but it is unnecessary, and it's subject to the uncertainty of time integration. Photoionization conservation is already checked in one-zone tests and in the R-type test.

chongchonghe:
The current volume-based calculation gives an error of about three cells, and the test passed, which means it's good enough for CI tests.

chongchonghe:
> I think the only major issue it identified is this:
> 
> * Chem-group ordering contradiction in the multi-group source term:
>   `source_terms_multi_group.hpp:641-646` assumes "the last NChemBands groups are ionizing photon groups" and skips the ĉ/c scaling for g >= nGroups_ - NChemBands. But photochemistry.hpp:62-67 maps chem bands to the first NumChemBands groups (radEnergy_index + numRadVarsPerGroup * nn with nn from 0). Today every chem problem has nGroups == NumChemBands == 1, so the two conventions coincide — but the moment someone builds the mixed thermal+ionizing problem this code claims to support, the scaling will be applied to exactly the wrong groups. Either make photochemistry.hpp read the last groups, or change the source-term condition to g < NChemBands, and add a static_assert (or AMREX_ALWAYS_ASSERT) that the two agree — e.g. assert NChemBands == 0 || NChemBands == nGroups_ until mixed groups are actually supported and tested.

This is indeed a problem, and it's the only useful comment from the Claude Fable review. 

@James471 Can you do this? We should set the chemical groups to the **last** Nchem bands.

James471:
@chongchonghe Sure, I will do it.

chongchonghe (approved):

PR #1904, opened 2026-05-23
Fix MHD wave convergence tests: enable MPI splits and add TOML parameters

AstroKriel:
### Description

The three MHD wave convergence tests (`AlfvenWaveLinearConvergence`, `FastWaveConvergence`, `SlowWaveConvergence`) had two issues. First, `blocking_factor` and `max_grid_size` were unset, which caused AMReX to create a single box covering the entire domain. With a single box, all MPI ranks beyond the first are silently idle; the test ran sequentially regardless of rank count. Setting `blocking_factor_x = 16` and `max_grid_size = 128` allows AMReX to decompose a 512-cell domain into up to 32 boxes, enabling proper MPI parallelism. Second, the Richardson convergence framework only refines `nx`; oblique wave modes (`num_modes_y` or `num_modes_z` nonzero) will not converge under this strategy and previously produced incorrect results silently. An abort guard is added to catch this at startup.

`setup.nx_max` and `setup.machine_precision_target` are also now queryable from the input TOML, so sweep scripts can control the resolution ceiling and early-exit threshold without modifying source code.

### Related issues

N/A

### Checklist

### DIFF
--- a/src/problems/AlfvenWaveLinearConvergence/testAlfvenWaveLinearConvergence.cpp
+++ b/src/problems/AlfvenWaveLinearConvergence/testAlfvenWaveLinearConvergence.cpp
@@ -485,7 +485,11 @@ auto problem_main() -> int
 	quokka::richardson::Parameters params{};
 	params.machine_precision_target = 2.0e-10; // limit based on delta_b_magn, smaller values can be used if this is decreased
 	params.nx_initial = 16;
-	params.nx_max = 128; // cap at 256 for quick tests. otherwise, it can take ~1-2 hours for 2048
+	params.nx_max = 128; // default; override with setup.nx_max in the input file.
+	{
+		amrex::ParmParse const pp("setup");
+		pp.query("nx_max", params.nx_max);
+	}
 	params.expected_rate = 2.0;
 	params.tolerance = 0.3;
 	params.test_name = "Alfven Wave";
--- a/src/problems/AlfvenWaveLinearConvergence/testAlfvenWaveLinearConvergence.cpp
+++ b/src/problems/AlfvenWaveLinearConvergence/testAlfvenWaveLinearConvergence.cpp
@@ -393,7 +393,7 @@ auto runWaveTest(int nx) -> double
 	}
 
 	if (num_modes_y != 0 || num_modes_z != 0) {
-		amrex::Abort("oblique modes are not supported: Richardson only refines nx, so transverse resolution is fixed and oblique waves will not converge.");
+		amrex::Abort("Oblique modes are not supported: Richardson only refines nx, so transverse resolution is fixed and oblique waves will not converge.");
 	}
 
 
--- a/src/problems/AlfvenWaveLinearConvergence/testAlfvenWaveLinearConvergence.cpp
+++ b/src/problems/AlfvenWaveLinearConvergence/testAlfvenWaveLinearConvergence.cpp
@@ -483,11 +483,12 @@ auto problem_main() -> int
 	quokka::richardson::applyQuietDefaults();
 
 	quokka::richardson::Parameters params{};
-	params.machine_precision_target = 2.0e-10; // limit based on delta_b_magn, smaller values can be used if this is decreased
+	params.machine_precision_target = 2.0e-10; // default; set to 0 via setup.machine_precision_target to disable early exit.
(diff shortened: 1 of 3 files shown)


chongchonghe:
# PR 1904 Review: Fix MHD wave convergence tests

**Reviewer**: Claude and @chongchonghe 

## Summary

The PR fixes two real bugs in `AlfvenWaveLinearConvergence`, `FastWaveConvergence`, and `SlowWaveConvergence`, and adds a useful TOML enhancement. Both bugs are confirmed via RED/GREEN testing.

---

## RED / GREEN / Edge-case Table

| Phase | What you ran | What you observed |
|---|---|---|
| RED (development) – Bug 2 | `AlfvenWaveLinearConvergence ... setup.num_modes_y=1 setup.nx_max=64` | Ran silently, convergence rate 0.09 (expected 2.0), failed but no oblique-mode diagnostic |
| GREEN (PR branch) – Bug 2 | Same binary + flags on PR branch | Immediate abort: "Oblique modes are not supported: Richardson only refines nx..." |
| GREEN (PR branch) – default | `quokka run -d 3d AlfvenWaveLinearConvergence` | PASSED, convergence rate 2.01 ✓ |
| GREEN (PR branch) – nx_max param | `... setup.nx_max=32` | Swept only to nx=32, PASSED in 2.2s — TOML param works ✓ |
| Edge: FastWaveConvergence | `quokka run -d 3d FastWaveConvergence` (default) | PASSED ✓ |
| Edge: SlowWaveConvergence | `quokka run -d 3d SlowWaveConvergence` (default) | PASSED ✓ |
| Edge: oblique guard (Fast+Slow) | Each with `setup.num_modes_y=1` | Both abort with the correct message ✓ |

---

## Root Cause Analysis

### Bug 1 — Single-box MPI parallelism

**Old code** (all three tests):
```cpp
const int blocking_x = std::max(16, nx);
if (!pp.contains("blocking_factor_x")) {
    pp.add("blocking_factor_x", blocking_x);
}
const int max_grid_x = nx;
if (!pp.contains("max_grid_size")) {
    pp.add("max_grid_size", max_grid_x);
}
```

For large `nx` (e.g. 512), this sets `blocking_factor_x = 512` and `max_grid_size = 512`, so AMReX creates exactly 1 box. All MPI ranks beyond rank 0 are silently idle.

**New code**: `blocking_factor_x = 16` and `max_grid_size = 128` (fixed). For `nx = 512`: produces 4 boxes of 128 cells each, enabling MPI parallelism.

**Important nuance**: The CI input files already have `amr.max_grid_size = 128`, so the `max_grid_size` half of the bug only matters when there's no input file or when the value is missing. The `blocking_factor_x = max(16, nx)` part is the main active issue, and it only affects runs with `nx > 128` (i.e., custom sweep scripts). The default CI test (nx_max=128, max_grid_size=128 from TOML) always produced 1 box regardless, so this is a performance-for-large-sweeps fix, not a CI-visible correctness fix.

**AlfvenWave-specific**: The old code also had redundant `pp.query` calls that the PR correctly removes — they updated local variables that were then unused (the `!pp.contains` guard prevented them being added anyway).

### Bug 2 — Silent convergence failure with oblique modes

The Richardson framework only refines `nx` each iteration; `ny = nz = 8` cells are fixed. An oblique wave (`num_modes_y != 0 || num_modes_z != 0`) requires adequate transverse resolution, so error norms plateau rather than converge. With the old code, this produced convergence rate ~0.09 (expected 2.0) with no diagnostic. The fix adds an early abort.

### Enhancement — TOML-queryable parameters

`setup.nx_max` and `setup.machine_precision_target` are now queryable via `amrex::ParmParse`. Confirmed working: `setup.nx_max=32` limits the sweep correctly.

---

## Concerns / Minor Issues

1. **PR description box-count inaccuracy**: The description claims "up to 32 boxes" for a 512-cell domain. With `max_grid_size = 128`, it's actually 4^3=64 boxes. This is a documentation issue only, not a code issue.

---

## Verdict

**Approve**. Both bugs are real and correctly fixed. Tests pass on all three convergence problems. The abort guard for oblique modes is the most impactful fix (prevents silent wrong results). The blocking-factor fix and TOML parameters are correct enhancements for high-resolution sweep workflows.

chongchonghe (approved):

PR #1905, opened 2026-05-24
Add dust runtime parameters: grain radius and grain density

mxchen39:
### Description
This PR adds optional runtime parameters `dust.grain_radius` and `dust.grain_density` for dust problems that use the Kwok stopping-time helper. These parameters are read through a small dust runtime helper and allow problem setups to override grain properties from the input file without hard-coding them in the source. If they are not provided, each problem continues to use its local default values. The affected dust test problems were updated to support this interface, and `DustDampingIteration` is kept as an example of the default-in-code usage.

### Related issues

### Checklist

### DIFF
--- a/src/problems/DustDampingIterationMHDZeroCharge/testDustDampingIterationMHDZeroCharge.cpp
+++ b/src/problems/DustDampingIterationMHDZeroCharge/testDustDampingIterationMHDZeroCharge.cpp
@@ -27,7 +27,7 @@ constexpr double B_Z = 0.5;
 constexpr double MAGNETIC_ENERGY = 0.5 * (B_X * B_X + B_Y * B_Y + B_Z * B_Z);
 
 AMREX_GPU_MANAGED amrex::GpuArray<amrex::Real, 2> g_dust_grain_radius = {0.02, 0.01}; // NOLINT
-AMREX_GPU_MANAGED amrex::GpuArray<amrex::Real, 2> g_dust_grain_density = {1.0, 1.0}; // NOLINT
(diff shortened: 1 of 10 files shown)


src/problems/DustDampingIterationMHDZeroCharge/testDustDampingIterationMHDZeroCharge.cpp
```diff
@@ -274,6 +275,7 @@ auto run_mhd_zero_charge_simulation() -> SimulationData<DustDampingMHDZeroCharge
 	sim.cflNumber_ = 0.3;
 	sim.constantDt_ = -1.0;
 	sim.enableIterDustStoptime_ = 1;
+	sim.dust_omega_res_ = 1.0;
```
chongchonghe: This sets `omega_rk_residual` from its default 0.0 to 1.0 for the MHD zero-charge run only — the reference simulation it is compared against still uses 0.0. Is this intended?

chongchonghe:
Here is my review:

(1) Unexplained behavioral change — `sim.dust_omega_res_ = 1.0` 

See inline comments. 

(2) No CI coverage for the new feature [should address]

All problems that actually read grain params from input (`DustDampedGyromotion`, `DustDampingIterationMHDZeroCharge`, `DustMagnetizedRDI`) have `ADD_TEST OFF`. The only CI-covered test (`DustDampingIteration`) uses hardcoded defaults — its input file was left unchanged. The PR checklist item "I have added tests for any new physics" is unchecked. At minimum, add the default grain param values to `inputs/DustDampingIteration.toml` so CI exercises the parameter-reading code path.

chongchonghe (approved):

PR #1910, opened 2026-05-25
Add SlowWave: slow MHD wave correctness test

AstroKriel:
### Description

While Quokka has standalone correctness tests for Alfven and fast linear MHD waves, the slow MHD wave only had a convergence test, so this PR adds `SlowWave` that mirrors the existing pattern. The test runs the slow magnetosonic eigenmode and reports the L1 error against the analytic solution.

mag-profiles
vel-profiles
rho-profiles
pressure-profiles

For one wave period (where the wavevector k points along x_0, and the background magnetic field lies at 45 degrees relative to k in the x_0-x_1 plane), b_0 stays constant, b_2 remains exactly zero (there's no out-of-plane / Alfvenic component), and b_1 oscillates. Density, pressure, and the in-plane velocity components oscillate in step with b_1, as expected.

### Related issues

- #1912: this test exposes the Quokka2026 EMF scheme instability reported there.

### Checklist


src/problems/SlowWave/test_slow_wave.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef TEST_SLOW_WAVE_HPP_ // NOLINT
```
chongchonghe: this hpp file is not necessary and should be removed.

chongchonghe:
The number of the test problems is exploding. I think we should try to avoid repeating test problems. This test is just repeating the convergence test of the same problem. We can run the Richardson convergence test and the answer test in one single problem. It should be straightforward to do: just call the runWaveTest one more time in problem_main.

chongchonghe (changes requested):

AstroKriel:
> The number of the test problems is exploding. I think we should try to avoid repeating test problems. This test is just repeating the convergence test of the same problem. We can run the Richardson convergence test and the answer test in one single problem. It should be straightforward to do: just call the runWaveTest one more time in problem_main.

Fair point on test count, and I think a test audit is worth doing. That said, `SlowWave` is pulling its weight: it exposes a numerical instability in the `Quokka2026` EMF scheme (see #1912), and it's the validation case I'm using in the resistivity branch to show the instability is damped away. It's doing work, and doesn't just duplicate the convergence sweep.

I do agree with you that a broader audit may be necessary soon: looking at the pure MHD tests, `MHDBlast` and `MHDQuirk` aren't doing as much for us, so they might be candidates for removal. Are there other tests you'd flag? Also, do you want to put this on the agenda for the telecom next week?

BenWibking:
> The number of the test problems is exploding. I think we should try to avoid repeating test problems. This test is just repeating the convergence test of the same problem. We can run the Richardson convergence test and the answer test in one single problem. It should be straightforward to do: just call the runWaveTest one more time in problem_main.

I strongly agree with this.

AstroKriel:
> > The number of the test problems is exploding. I think we should try to avoid repeating test problems. This test is just repeating the convergence test of the same problem. We can run the Richardson convergence test and the answer test in one single problem. It should be straightforward to do: just call the runWaveTest one more time in problem_main.
> 
> I strongly agree with this.

@chongchonghe, @BenWibking, the change isn't straightforward, and while interesting, it would require an architectural change that would also need to be applied across all five convergence tests in the repo (`AlfvenWaveLinearConvergence`, `EntropyWaveConvergence`, `FastWaveConvergence`, `HydroWaveConvergence`, `SlowWaveConvergence`), otherwise it brings about inconsistency/asymmetry. I would advocate that one more test is not enough to justify a whole redesign of the convergence-test workflow. @markkrumholz, what do you think?

chongchonghe:
@AstroKriel Can you take a look at #1914? If that recommended change makes sense to you, you can selectively merge in the changes. Once this PR is merged, you can create another PR to clean up other similar problem sets using this design.

AstroKriel:
Following the feedback here, I've put together a broader proposal in #1916 that addresses the underlying concern about  the growing test proliferation. Rather than having separate test binaries for convergence and correctness tests, the idea is to consolidate both into a single binary where convergence is the default pathway, and correctness can be optionally tested at a desired resolution.

If there's agreement on that direction, I'll rework this PR to implement it for `SlowWave`/`SlowWaveConvergence` as the first instance.

@chongchonghe @markkrumholz @BenWibking

AstroKriel:
> Following the feedback here, I've put together a broader proposal in #1916 that addresses the underlying concern about the growing test proliferation. Rather than having separate test binaries for convergence and correctness tests, the idea is to consolidate both into a single binary where convergence is the default pathway, and correctness can be optionally tested at a desired resolution.
> 
> If there's agreement on that direction, I'll rework this PR to implement it for `SlowWave`/`SlowWaveConvergence` as the first instance.
> 
> @chongchonghe @markkrumholz @BenWibking

After discussing this with @markkrumholz, I'd like to ask to defer this change until after the physical resistivity PR (#1913) lands. That PR depends directly on this one, and I'd like to get resistivity through, so I can move on to artificial resistivity. I'm committed to doing the refactor (outlined in #1916), but only once the science capabilities are in.

Could we merge this PR in the interim, please?

@chongchonghe @BenWibking

BenWibking:
> After discussing this with @markkrumholz, I'd like to ask to defer this change until after the physical resistivity PR (#1913) lands. That PR depends directly on this one, and I'd like to get resistivity through, so I can move on to artificial resistivity. I'm committed to doing the refactor (outlined in #1916), but only once the science capabilities are in.
> 
> Could we merge this PR in the interim, please?
> 
> @chongchonghe @BenWibking

I don't think #1913 should depend on this PR. Why do you think it does?

BenWibking:
You can fully verify the correctness of #1913 with the Alfven wave decay test. This is just an extra test problem to show it stabilizes the Quokka2026 method, which I don't think is strictly necessary to merge that PR.

AstroKriel:
> > After discussing this with @markkrumholz, I'd like to ask to defer this change until after the physical resistivity PR (#1913) lands. That PR depends directly on this one, and I'd like to get resistivity through, so I can move on to artificial resistivity. I'm committed to doing the refactor (outlined in #1916), but only once the science capabilities are in.
> > Could we merge this PR in the interim, please?
> > @chongchonghe @BenWibking
> 
> I don't think #1913 should depend on this PR. Why do you think it does?

I added the standalone ideal-MHD test, here, and then extended it in #1913 to also support eta>0 solutions. I made it its own PR to keep PRs small and managable.

AstroKriel:
> You can fully verify the correctness of #1913 with the Alfven wave decay test. This is just an extra test problem to show it stabilizes the Quokka2026 method, which I don't think is strictly necessary to merge that PR.

That is true for physical resistivity, but the slow wave will be more directly useful for artificial resistivity, since the Quokka2026 scheme goes unstable for it at high resolutions (see #1912), and the slow wave has an analytic solution to confirm correctness. The Orszag-Tang vortex is also a good stability test, but it is more expensive to run and lacks an analytic solution.

BenWibking:
> > > After discussing this with @markkrumholz, I'd like to ask to defer this change until after the physical resistivity PR (#1913) lands. That PR depends directly on this one, and I'd like to get resistivity through, so I can move on to artificial resistivity. I'm committed to doing the refactor (outlined in #1916), but only once the science capabilities are in.
> > > Could we merge this PR in the interim, please?
> > > @chongchonghe @BenWibking
> > 
> > 
> > I don't think #1913 should depend on this PR. Why do you think it does?
> 
> I added the standalone ideal-MHD test, here, and then extended it in #1913 to also support eta>0 solutions. I made it its own PR to keep PRs small and managable.

Ok, I think you have added an unnecessarily complicated dependency structure.

The way to solve this is to create a separate, third PR for the eta > 0 modification that can go in after both this PR and #1913 are merged.

AstroKriel:
> > > > After discussing this with @markkrumholz, I'd like to ask to defer this change until after the physical resistivity PR (#1913) lands. That PR depends directly on this one, and I'd like to get resistivity through, so I can move on to artificial resistivity. I'm committed to doing the refactor (outlined in #1916), but only once the science capabilities are in.
> > > > Could we merge this PR in the interim, please?
> > > > @chongchonghe @BenWibking
> > > 
> > > 
> > > I don't think #1913 should depend on this PR. Why do you think it does?
> > 
> > 
> > I added the standalone ideal-MHD test, here, and then extended it in #1913 to also support eta>0 solutions. I made it its own PR to keep PRs small and managable.
> 
> Ok, I think you have added an unnecessarily complicated dependency structure.
> 
> The way to solve this is to create a separate, third PR for the eta > 0 modification that can go in after both this PR and #1913 are merged.

Done. I've rebased #1913 onto `development`, so the SlowWave no longer appear in that PR's diff; these two PRs are now entirely independent.

AstroKriel:
@chongchonghe @BenWibking, just following up on this one. Since the discussion last week, I think everything is resolved on my end:

- I've responded to all your feedback and requested changes.
- Per @BenWibking's suggestion, I rebased #1913 onto `development` so the SlowWave test no longer appears in that diff. The two PRs are now fully independent, so this is no longer a dependency for the resistivity work.
- On the test-proliferation concern: I'm committed to the consolidation (the #1914 direction, convergence + correctness in one binary), but as a dedicated follow-up that I can apply consistently across all the convergence tests, once the resistivity work is in.

Given that, could we get this re-reviewed and merged, please?

chongchonghe (approved):
Assuming GPU tests pass, LGTM.

Issue #1912, opened 2026-05-26
The Quokka2026 EMF scheme shows instabilities for slow waves at high resolution


AstroKriel:
**Describe the bug**

The `Quokka2026` EMF compute scheme is prone to produce a numerical instability at high resolution with oblique or complex magnetic field geometries. This was first seen in the Orszag-Tang vortex (runs stalled at t ~ 0.75 at 256^2), and has now been seen again in the slow magnetosonic wave at high resolution. The likely cause is that `Quokka2026` reconstructs face-centered HLLD Riemann velocities (a nonlinear quantity) to cell edges, whereas `FelkerStone2017` reconstructs cell-centered velocities that remain well-behaved for linear perturbations. The error with `Quokka2026` diverges sharply from the expected 2nd-order convergence trend at high resolution, where `FelkerStone2017` and `Balsara2025` remain well behaved. The instability is insensitive to CFL reduction.

<img width="3329" height="2070" alt="Image" src="https://github.com/user-attachments/assets/acdf517f-6809-4d8d-95c7-4881f302ad99" />

**To Reproduce**

1. Build the `SlowWave` problem (added in PR #1910).
2. Add the following overrides to `inputs/SlowWave.toml`:
```toml
amr.n_cell = [512, 8, 8]
amr.max_grid_size_x = 128
amr.blocking_factor_x = 16
mhd.emf_compute_scheme = "Quokka2026"
mhd.emf_averaging_scheme = "Balsara2025"
```
3. Run: `./SlowWave ../inputs/SlowWave.toml`
4. See that the L1 error is orders of magnitude larger than the expected converged errorr.

**Additional context**

- The instability is specific to the `Quokka2026` EMF compute path; it is not present for `FelkerStone2017` or `Balsara2025` at the same resolution.
- The EMF averaging scheme (`LondrilloDelZanna2004` vs `Balsara2025`) does not affect whether the instability appears, only the compute scheme does.
- Physical resistivity suppresses the instability. This is the proposed fix, to be implemented in a follow-up PR.
`

PR #1913, opened 2026-05-26
Add resistivity: Ohmic diffusion in the MHD induction update

AstroKriel:
### Description

I've added constant physical (Ohmic) resistivity to the constrained-transport MHD induction update. At each cell edge, the current density is estimated as a centred difference of the surrounding face-centred magnetic fields (`j_edge = curl(B_face)`), and the resistive contribution `eta * j_edge` is subtracted from the ideal EMF before the magnetic field is updated:

```
emf_total = emf_ideal - eta * j_edge
```

A new parameter `mhd.resistivity` is introduced (it defaults to 0, and must be >= 0), alongside an explicit parabolic timestep limit `dt <= cfl * dx^2 / (6*eta)` (this only works in 3D). Startup guards also abort if incompatible settings are used: neither AMR subcycling nor the dual energy formalism (Ohmic heating is implicit in single-energy mode via conservation, but is not yet added to the auxiliary energy equation) are supported.

**Validation.** The resistive Alfvén wave test (`AlfvenWaveLinear`) is extended with an exact analytic reference solution for the resistive case. The decay rate `gamma = eta * k^2 / 2` is measured from the amplitude envelope at `t = 1`, and converges to the analytic value at 2nd order.

I've also confirmed that the measured decay rate `gamma` matches the analytic prediction `gamma = eta * k^2 / 2` across the full range of `eta` tested with the damped linear Alfven wave run with 256 cells in the `x_0` direction.

alfven-decay-rate-vs-eta

**Energy conservation.** A new `MHDResistiveDiffusion` test validates that magnetic energy lost to Ohmic diffusion is correctly transferred to gas energy (relevant fotlr adiabatic systems). The test sets up a sinusoidal `B_z` field and checks that it decays as `b_z = b_amp * exp(-eta*k^2*t) * sin(kx)`; the resistive Poynting flux (`AddResistiveEnergyFlux`) should heat the gas uniformly as the field decays. Two checks are applied: the combined `B_z` error norm must fall below 0.01, confirming the decay rate, and the `E_tot` relative error must fall below 0.02 against a reference encoding the expected Joule heating mean. Disabling `AddResistiveEnergyFlux` leaves the total energy with the wrong spatial distribution, raising the relative error to 6.5%, a factor of ~3 above threshold.

### Related issues

- #1912: this PR incidentally stabilises the Quokka2026 EMF instability reported there.

### Checklist


markkrumholz (dismissed):

src/hydro/mhd_system.hpp
```diff
@@ -1001,4 +1030,19 @@ void MHDSystem<problem_t>::SolveInductionEqn(std::array<amrex::MultiFab, AMREX_S
...
+					  amrex::Real dx_w1, amrex::Real resistivity, amrex::Box const &box_ec)
+{
+	const BL_PROFILE("MHDSystem::AddResistivity()");
+	// EMF_total = EMF_ideal - eta * J_edge (Quokka stores EMF = -E, so resistive Ohm's law gives E += eta*J -> EMF -= eta*J)
+	// J at edge = dB_w1/dx_w0 - dB_w0/dx_w1, computed via backward differences placed exactly at the edge
+	amrex::ParallelFor(box_ec, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+		const amrex::Real J_edge = (B_w1(i, j, k) - B_w1(i - vec_w0[0], j - vec_w0[1], k - vec_w0[2])) / dx_w0 -
+					   (B_w0(i, j, k) - B_w0(i - vec_w1[0], j - vec_w1[1], k - vec_w1[2])) / dx_w1;
+		E_edge(i, j, k) -= resistivity * J_edge;
+	});
```
BenWibking: This is a very small kernel, and it will almost certainly be more efficient on the GPU to call this from within the existing EMF kernel as a device function, rather than creating a new kernel and incurrent the extra kernel launch latency overhead.

src/problems/SlowWave/testSlowWave.cpp
BenWibking, line 1: I think this should be moved to https://github.com/quokka-astro/quokka/pull/1910.

src/QuokkaSimulation.hpp
```diff
@@ -796,6 +809,22 @@ template <typename problem_t> void QuokkaSimulation<problem_t>::computeMaxSignal
 				     "compute a time step.");
 		}
 	}
+
+	// diffusive CFL constraint for Ohmic resistivity: dt <= cfl * dx^2 / (2*NDIM*eta)
+	if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
+		if (mhdResistivity_ != 0.0) {
+			const auto &dx = geom[level].CellSizeArray();
+			const amrex::Real dx_min = std::min({AMREX_D_DECL(dx[0], dx[1], dx[2])});
+			const amrex::Real resistive_signal = 2.0 * AMREX_SPACEDIM * mhdResistivity_ / dx_min;
```
BenWibking: We don't include AMREX_SPACEDIM in the existing hydro CFL, so I think this is confusing. I think it would be better to be consistent and *not* include the `AMREX_SPACEDIM` factor. That means that a CFL ~ 0.3 will be stable in 3D, consistent with both hydro and the thermal conduction implementation PR.

BenWibking (changes requested):
There is an inconsistency in the way the CFL is defined here as compared to hydro (and as compared to Aditi's thermal conduction PR) that should be fixed.

The other comments are minor.

src/QuokkaSimulation.hpp
AstroKriel: Good point. I've removed the `AMREX_SPACEDIM` factor to be consistent with hydro and thermal conduction, and also added a comment noting that in N dimensions the true stability limit requires `cfl < 1/N`, so anyone running 3D problems should keep `cfl < 1/3`.

src/problems/SlowWave/testSlowWave.cpp
AstroKriel: I have rebased this repo so its no longer stacked on top of the slow wave PR, so this file is now gone.

src/hydro/mhd_system.hpp
AstroKriel: Done. The resistive EMF contribution is now inlined into the EMF averaging kernel.

AstroKriel:
@BenWibking following up on this one. I've fixed the CFL inconsistency you flagged and replied to your minor comments. I've also just merged `development` in, so it's up to date and mergeable. Could you take another look when you get a chance, please?

src/hydro/mhd_system.hpp
```diff
@@ -66,36 +66,49 @@ template <typename problem_t> class MHDSystem : public HyperbolicSystem<problem_
 	static void ComputeEMF(std::array<amrex::MultiFab, AMREX_SPACEDIM> &ec_mf_emf_components, amrex::MultiFab const &cc_mf_cVars,
 			       std::array<amrex::MultiFab, AMREX_SPACEDIM> const &fcx_mf_vel, std::array<amrex::MultiFab, AMREX_SPACEDIM> const &fcx_mf_cVars,
 			       std::array<amrex::MultiFab, AMREX_SPACEDIM> const &fcx_mf_fspds, int reconstructionOrder, EMFAvgScheme emf_avg_scheme,
-			       SlopeLimiter plmLimiter, EMFComputeScheme emf_compute_scheme);
+			       SlopeLimiter plmLimiter, EMFComputeScheme emf_compute_scheme, amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dx,
+			       amrex::Real resistivity = 0.0);
```
BenWibking, line 70: I think it would be better to treat the resistivity analogous to the way the opacity is treated in the radiation solver. Instead of using a runtime variable, it should be computed corner-by-corner via a call to a template function that's specialized for each problem.

This is needed for any future extension to a non-constant resistivity, but also is needed today to avoid making the MHD solver slower for simulations that don't use resistivity. (Adding runtime variables to device kernels will always make it slower, even if the value is zero at runtime.)

BenWibking (changes requested):

src/hydro/mhd_system.hpp
AstroKriel: It looks like opacity isn't configurable at run time (via the input params TOML file), with its value instead hardcoded in each problem. If resistivity follows the same pattern, I wouldn't be able to scan `eta` between small-scale dynamo simulation runs, without changing the code and recompiling for each simulation. This isn't really practical for the parameter sweeps I need.

If your concern is runtime overhead for simulations that don't use resistivity, I can implement a `ResistivityModel` enum in `Physics_Traits`, similar to how `OpacityModel` already works:

```cpp
enum class ResistivityModel { none, constant, spatially_varying };
// default:
static constexpr ResistivityModel resistivity_model = ResistivityModel::none;
```

With an `if constexpr` on `resistivity_model` in the kernel:
- `none` : the resistive block is eliminated entirely (no overhead)
- `constant`: keeps `mhd.resistivity` as a TOML parameter (read at runtime)
- `spatially_varying`: calls a per-cell `computeResistivity(i, j, k, state_arr)` specialisation (I can implement the empty fn that would need to be overwritten by a test probem)

What do you think?

markkrumholz: @AstroKriel, that looks like a very reasonable compromise that manages to cover both your and @BenWibking's intended use cases and concerns.

markkrumholz: Another option that achieves the same effect would be to just have resistivity treated like opacity in that there is always a call to a `computeResistivity` function, but then implement the part about reading a TOML parameter within the problem setup file. That is, the way you would handle the case of wanting to be able to change the resistivity at runtime without recompiling is to set a problem where the during initialization a constant resistivity value is read from the input file and copied to the GPU, so that it is accessible to the `computeResistivity` function when that is called. I don't have a strong preference between this solution and the one you propose; I'm just mentioning it as another option.

AstroKriel: That's an interesting suggestion, @markkrumholz. I have a slight preference for the `Physics_Traits` approach I described, but I'm happy to do it this way if it's what @BenWibking agrees is the better way.

BenWibking: I think your `ResistivityModel` approach is probably better from a code design perspective, but I think Mark's suggestion will be simpler to implement.

AstroKriel:
@BenWibking and @markkrumholz, I've implemented the compile-time `ResistivityModel` we discussed above. `Physics_Traits` now carries a `resistivity_model` enum with three values: `none` (the default option which carries no overhead), `constant` (uniform eta read from `mhd.resistivity` in the TOML), and `problem_defined` (spatially varying eta returned by a per-problem `computeResistivity` device function). The `problem_defined` path has a default template that throws a compile-time error if the user sets the model and forgets to implement an overwrite of `computeResistivity` for their problem.

src/hydro/mhd_system.hpp
```diff
@@ -768,6 +807,12 @@ void MHDSystem<problem_t>::EMFAverage_LondrilloDelZanna2004(amrex::Array4<amrex:
 		const double term2 = ((a1_m * a1_p) / (a1_m + a1_p)) * (B0_p_ - B0_m_) + ((a0_m * a0_p) / (a0_m + a0_p)) * (B1_m_ - B1_p_);
 
 		E2_ave(i, j, k) = (numerator / denominator) + term2;
+		if constexpr (Physics_Traits<problem_t>::resistivity_model == ResistivityModel::constant) {
+			E2_ave(i, j, k) -= computeResistiveEMF(B_w0, B_w1, i, j, k, delta_w0, delta_w1, dx_w0, dx_w1, resistivity);
```
BenWibking: In principle, this should work. However, we should check that the CUDA and ROCm compilers optimize out `resistivity` when this is not used. Since nothing else is changed in this function, you should be able to do this by compiling both the development branch and this PR and extracting the GPU kernel properties for this ParallelFor. The register usage should be the same if the compiler is performing the optimization we want it to.

AstroKriel: I've checked and confirmed that the dead code is removed by the compiler. I compiled the `OrszagTang` problem on both `development` and this branch (where `OrszagTang` has `resistivity_model = none`) and compared register counts for the two EMF averaging kernels where `applyResistiveCorrection` is called (`EMFAverage_LondrilloDelZanna2004` and `EMFAverage_Balsara2025`) using `cuobjdump --dump-sass`. Both kernels have identical register counts on branches (46 and 78 respectively), so NVCC has fully eliminated the resistivity code for problems where resistivity is turned off (default = none).

BenWibking: Perfect! Thanks.

BenWibking (changes requested):
I asked Codex to review this, and it says that for resistive MHD, there should be a resistive term in the total energy flux. I think it is right.

This is the resistive flux term given by the PLUTO paper (Equation 5 of https://arxiv.org/abs/1110.0740):

$\nabla \cdot \Pi_{\mathcal{E}} = \nabla \cdot ( -\eta \cdot J \times B )$

added on the right hand side of the total energy equation. So this should probably be computed as $\eta J \times B$ on each face and added to the face-centered total energy flux prior to the RK update in each stage.

BenWibking:
This is Codex's explanation of how to compute it using the edge-centered $\eta J$ and face-centered $B$ components. This is what I would do:

> Let
> 
>   R = eta J
> 
>   be edge-centered, using exactly the same discrete curl as computeResistiveEMF in src/hydro/mhd_system.hpp:1055. For a face normal to
>   direction d, with cyclic tangential directions
> 
>   a = (d + 1) % 3;
>   b = (d + 2) % 3;
> 
>   the resistive energy flux is
> 
>   F_eta[d] = R_a B_b - R_b B_a
> 
>   computed on that d-face by averaging the two bounding edges of each tangential direction.
> 
>   For an x-face, this is the concrete stencil:
> 
>   F_eta_x(i,j,k) =
>     1/4 * [
>         R_y(i,j,k)   * (B_z(i,j,k)   + B_z(i-1,j,k))
>       + R_y(i,j,k+1) * (B_z(i,j,k+1) + B_z(i-1,j,k+1))
>       - R_z(i,j,k)   * (B_y(i,j,k)   + B_y(i-1,j,k))
>       - R_z(i,j+1,k) * (B_y(i,j+1,k) + B_y(i-1,j+1,k))
>     ]
> 
>   Then cycle indices for y and z faces. Add this to fluxArrays[d](..., energy_index) before ComputeRhsFromFluxes; its sign convention will then
>   apply the correct -div(F_eta) update in src/hydro/hydro_system.hpp:757.
> 
>   One sign caveat: the formula above assumes R = eta J. If you instead reuse Quokka’s resistive contribution to the CT EMF, that stored
>   contribution is -eta J, because the code subtracts it from v x B. In that case the energy flux is -(EMF_res x B).

src/problems/ResistiveMHDEnergyBalance/testResistiveMHDEnergyBalance.cpp
```diff
@@ -0,0 +1,199 @@
...
+	const amrex::Real norm = std::sqrt(std::max(l2_norm, std::numeric_limits<amrex::Real>::min()));
+	return {std::sqrt(l2_correct) / norm, std::sqrt(l2_missing_flux) / norm};
+}
+
+auto problem_main() -> int
+{
+	QuokkaSimulation<ResistiveMHDEnergyBalance> sim;
+	sim.setInitialConditions();
+	sim.evolve();
+
```
BenWibking: The new setup initializes a periodic 1D magnetic-pressure balance with

  - `rho = 1`
  - `v = 0`
  - `B_y = A sin(kx)`
  - `p + B_y^2 / 2 = const`
  - `gamma = 2`
  - constant Ohmic resistivity

For this state, the correct one-hydro-step pressure response is local Ohmic heating, `eta J^2` (prop. to `cos^2(kx)`), while today the code instead follows the profile expected when the resistive `eta J x B` total-energy flux is omitted (prop. to `sin^2(kx)`).

(There is nothing physically significant about `gamma = 2`. It just makes the algebra simpler.)

Image

AstroKriel:
with #1949 the number of files changed goes down from 98 to 6 :-)

src/problems/ResistiveMHDEnergyBalance/testResistiveMHDEnergyBalance.cpp
AstroKriel: Hey, I have added the resistive energy flux term via `AddResistiveEnergyFlux` (in `mhd_system.hpp`), which computes `eta * J x B` on each face and adds it to the total energy flux before the CT update. The old `ResistiveMHDEnergyBalance` test problem has been replaced by `MHDResistiveDiffusion`, which now validates that this term is working properly: the total energy `E_tot` for a resistive-regulated decay matches the expected Joule heating mean to within 2%. Without `AddResistiveEnergyFlux`, the relative error rises to ~6.5%, which is well above the threshold I had calibrated. Could you take another look when you get the chance, please?

markkrumholz (approved):

BenWibking (approved):

PR #1914, opened 2026-05-27
combine SlowWave answer test and convergence test into a single binary

chongchonghe:
Merges testSlowWave.cpp and testSlowWaveConvergence.cpp into a single file. problem_main() now runs both tests sequentially, prints the status of each, and returns 0 only when both pass.

The convergence test reference solution uses tNew_[0] (same as the answer test); this is correct because the simulation stops at exactly one wave period, at which point the analytic solution equals the initial state by periodicity.

### Related issues
#1910 

### Checklist


BenWibking:
I think this is the right solution. Thanks for simplifying these tests.

PR #1915, opened 2026-05-27
SNII WR AGB metal yields

Meowwish:
### Description
Add table-driven SNII, WR, and AGB metal yields.

Yield data are preprocessed into Quokka DataTable CSV files. At runtime, the simulation reads the user-requested channels and isotopes, then builds compact selected-isotope DataTables for GPU interpolation during yield feedback. The current yield tables use fixed metallicity `Z = 0.02`.

The validation tests print simulated/table yield ratios so reviewers can directly inspect agreement between deposited scalar masses and the lookup tables.

### Related issues
No.

### Validation
- `cmake --build build --target test_SNII_Yields test_WR_AGB_yields -j 8`
- `ctest --test-dir build --output-on-failure -R 'test_(SNII_Yields|WR_AGB_yields)'`

### Checklist


src/particles/particle_types.hpp
```diff
@@ -241,6 +277,22 @@ constexpr int StochasticStellarPopParticleMassAtBirthIdx = static_cast<int>(Stoc
 constexpr int StochasticStellarPopParticleLumIdx = static_cast<int>(StochasticStellarPopParticleRealIdx::luminosity); // Base index for luminosity components
 constexpr int StochasticStellarPopParticleStageIdx = static_cast<int>(StochasticStellarPopParticleIntIdx::evolution_stage);
 
+template <typename problem_t> constexpr int StochasticStellarPopParticleChemistryBlockSize() { return Physics_Traits<problem_t>::numPassiveScalars; }
+
+template <typename problem_t> constexpr int StochasticStellarPopParticleChemistryBaseIdx()
+{
+	if constexpr (Physics_Traits<problem_t>::is_hydro_enabled || Physics_Traits<problem_t>::is_radiation_enabled) {
+		return 14 + Physics_Traits<problem_t>::nGroups;
```
BenWibking, line 271: `14` should not be hard-coded here. This should be obtained from Physics_Indices.

src/particles/particle_chemical_yield.hpp
```diff
@@ -0,0 +1,518 @@
...
+
+namespace quokka
+{
+
+namespace ChemicalYieldLookup
+{
+
+constexpr int kMaxTrackedIsotopes = 32;
+constexpr int kMaxTrackedChannels = 3;
+constexpr int kMaxChemicalYieldEntries = 4096;
```
BenWibking: These should follow the existing naming convention. `k` should not be used as a variable prefix.

src/particles/particle_chemical_yield.hpp
```diff
@@ -0,0 +1,518 @@
...
+
+AMREX_GPU_MANAGED inline int g_num_entries = 0;		 // NOLINT
+AMREX_GPU_MANAGED inline bool g_loaded = false;		 // NOLINT
+AMREX_GPU_MANAGED inline int g_num_tracked_isotopes = 0; // NOLINT
+AMREX_GPU_MANAGED inline int g_num_tracked_channels = 0; // NOLINT
```
BenWibking: These should also follow the existing naming conventions.

src/particles/particle_chemical_yield.hpp
```diff
@@ -0,0 +1,518 @@
...
+		if (std::filesystem::exists(alt)) {
+			return alt;
+		}
+	}
+
+	return candidate;
+}
+
+inline auto loadTable(const std::string &filename, const std::vector<std::string> &tracked_isotopes, const std::vector<std::string> & /*tracked_channels*/)
+    -> bool
```
BenWibking: Instead of writing a custom C++ parser, you should put them in the existing `quokka::DataTable` table format and using the existing parser.

src/particles/particle_chemical_yield.hpp
```diff
@@ -0,0 +1,518 @@
...
+		const amrex::Real dm = std::abs(std::log10(amrex::max<amrex::Real>(g_mass_msun[i], 1.0e-12)) - log_mass);
+		const amrex::Real dz = std::abs(std::log10(amrex::max<amrex::Real>(g_metallicity[i], 1.0e-12)) - log_z);
+		const amrex::Real dist2 = dm * dm + dz * dz;
+		if (dist2 < exact_tol) {
+			return frac_i;
+		}
+		const amrex::Real w = 1.0 / (dist2 + eps);
+		wsum += w;
+		frac_sum += w * frac_i;
+	}
```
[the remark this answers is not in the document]
BenWibking: This is an important point. You should never do a linear scan over a table. It should either be indexed with fixed bin widths, or for irregular bin widths, you should do a binary search over the table.

src/particles/particle_deposition.hpp
```diff
@@ -367,9 +314,11 @@ depositThermalSNR(amrex::Array4<amrex::Real> const &local_buffer, const int ix,
 				// Deposit passive scalar if enabled
 				// TODO(chongchonghe): Add support for multiple passive scalars (currently only deposits to scalar0)
```
BenWibking: Should this comment be updated?

src/particles/particle_deposition.hpp
```diff
@@ -525,8 +474,10 @@ depositThermalKineticMomentumSNR(amrex::Array4<amrex::Real> const &local_state,
...
```
BenWibking: Should this comment be updated?

src/particles/particle_deposition.hpp
```diff
@@ -984,6 +935,205 @@ void updateEvolutionStage(ContainerType *container, int lev_min, amrex::Real ste
...
+							const amrex::Real dz =
+							    (AMREX_SPACEDIM >= 3) ? static_cast<amrex::Real>(kk - W_stencil_N) + 0.5 - interp.frac[2] : 0.0;
+							for (int jj = 0; jj < ny_loop; ++jj) {
+								const amrex::Real dy =
+								    (AMREX_SPACEDIM >= 2) ? static_cast<amrex::Real>(jj - W_stencil_N) + 0.5 - interp.frac[1] : 0.0;
+								for (int ii = 0; ii < W_stencil_width; ++ii) {
+									const amrex::Real dx = static_cast<amrex::Real>(ii - W_stencil_N) + 0.5 - interp.frac[0];
+									const amrex::Real r2 = AMREX_D_TERM(dx * dx, +dy * dy, +dz * dz);
+									if (r2 <= W_cutoff_r2) {
+										const amrex::Real wt = kernel_wendland_c2(std::sqrt(r2) * W_inv_N) * interp.inv_norm;
```
BenWibking: It appears that different kernels are being used for different injections. Is there a reason to do this? If not, can you make them consistent?

src/particles/particle_deposition.hpp
```diff
@@ -984,6 +935,205 @@ void updateEvolutionStage(ContainerType *container, int lev_min, amrex::Real ste
...
+				if (store_channel_fields && enable_SNII_metal) {
+					const int sn_comp = HydroSystem<problem_t>::scalar0_index + scalar_offset + nchem + n;
+					if (sn_comp < HydroSystem<problem_t>::scalar0_index + nPassive) {
+						for (int ii = -SN_stencil_size; ii <= SN_stencil_size; ++ii) {
+							for (int jj = -SN_stencil_size; jj <= SN_stencil_size; ++jj) {
+								for (int kk = -SN_stencil_size; kk <= SN_stencil_size; ++kk) {
+									const int iii = std::abs(ii);
+									const int jjj = std::abs(jj);
+									const int kkk = std::abs(kk);
+									const amrex::Real kernel_times_vol_inverse = stencil_weights_gpu[iii][jjj][kkk] * vol_inverse;
```
BenWibking: This is not the same injection kernel. Is there a reason to use a different kernel?

src/particles/particle_deposition.hpp
```diff
@@ -984,6 +935,205 @@ void updateEvolutionStage(ContainerType *container, int lev_min, amrex::Real ste
...
+				      {0.00884198143074, 0.00861063982859, 0.00400459528385, 0.00000136166514},
+				      {0.00262865918549, 0.00119306623841, 0.00000136166514, 0.00000000000000}}},
+				    {{{0.00884198143074, 0.00884198143074, 0.00596795726055, 0.00005052308190},
+				      {0.00884198143074, 0.00861063982859, 0.00400459528385, 0.00000136166514},
+				      {0.00596795726055, 0.00400459528385, 0.00045652034325, 0.00000000000000},
+				      {0.00005052308190, 0.00000136166514, 0.00000000000000, 0.00000000000000}}},
+				    {{{0.00416240696843, 0.00262865918549, 0.00005052308190, 0.00000000000000},
+				      {0.00262865918549, 0.00119306623841, 0.00000136166514, 0.00000000000000},
+				      {0.00005052308190, 0.00000136166514, 0.00000000000000, 0.00000000000000},
+				      {0.00000000000000, 0.00000000000000, 0.00000000000000, 0.00000000000000}}}}};
```
BenWibking: How were these weights computed?

src/particles/particle_deposition.hpp
```diff
@@ -984,6 +935,205 @@ void updateEvolutionStage(ContainerType *container, int lev_min, amrex::Real ste
...
+		const auto dxi = geom.InvCellSizeArray();
+		const amrex::Real vol_inverse = AMREX_D_TERM(dxi[0], *dxi[1], *dxi[2]);
+		const int chem_block_size = StochasticStellarPopParticleChemistryBlockSize<problem_t>();
+		const int chem_base = StochasticStellarPopParticleChemistryBaseIdx<problem_t>();
+		constexpr int W_stencil_N = 2;
+		constexpr int W_stencil_width = 2 * W_stencil_N + 1;
+		constexpr amrex::Real W_cutoff_r2 = static_cast<amrex::Real>(W_stencil_N * W_stencil_N);
+		constexpr amrex::Real W_inv_N = 1.0 / static_cast<amrex::Real>(W_stencil_N);
+
+		amrex::ParallelFor(np, [=] AMREX_GPU_DEVICE(int64_t idx) {
```
BenWibking, line 1150: This is a very complicated function. I think it would be better to refactor it to call a helper function for each injection channel, rather than putting it all in a single block of code. Each injection channel should call the same generic helper with different arguments specific to the injection channel.

src/particles/particle_types.hpp
```diff
@@ -68,6 +75,35 @@ template <typename problem_t> constexpr void verify_particle_switch_type()
...
+	std::vector<std::string> items;
+	std::stringstream stream(input);
+	std::string token;
+	while (std::getline(stream, token, ',')) {
+		token.erase(std::remove_if(token.begin(), token.end(), [](unsigned char c) { return std::isspace(c) != 0; }), token.end());
+		if (!token.empty()) {
+			items.push_back(token);
+		}
+	}
+	return items;
```
BenWibking: We should never write custom parsers in Quokka. This can be done using AMReX's existing ParmParse machinery instead.

src/problems/TallBoxSf/testTallBoxSf.cpp
```diff
@@ -78,7 +78,7 @@ template <> struct Physics_Traits<TheProblem> {
 	static constexpr bool is_dust_enabled = false;
 	static constexpr int nDustGroups = 1;			     // number of dust groups
 	static constexpr int numMassScalars = 0;		     // number of mass scalars
-	static constexpr int numPassiveScalars = numMassScalars + 1; // number of passive scalars
+	static constexpr int numPassiveScalars = 30;		     // number of passive scalars
```
BenWibking: This should not be hard-coded if possible. Can you query this value or set it to an existing variable?

BenWibking (changes requested):
Thanks for submitting a much smaller PR.

There are a number of places where the code can be simplified, or existing code (such as file parsers) can be re-used, rather than inventing new ones. I've added detailed comemnts below.

src/particles/particle_update.hpp
```diff
@@ -103,6 +120,132 @@ template <> struct ParticlePropertyUpdateTraits<ParticleType::StochasticStellarP
...
+		}
+
+		const int nPassive = Physics_Traits<problem_t>::numPassiveScalars;
+		const int scalar_offset = std::max(0, chemical_scalar_offset);
+		const int nchem = std::max(0, std::min(chemical_num_scalars, nPassive - scalar_offset));
+		if (nchem <= 0) {
+			return;
+		}
+
+		amrex::MultiFab state_buffer(state.boxArray(), state.DistributionMap(), state.nComp(), state.nGrow());
```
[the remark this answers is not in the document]
Meowwish: This is a real bug, I’ll change this buffer to `state.nComp() + 1` to match the other deposition paths.

src/particles/particle_chemical_yield.hpp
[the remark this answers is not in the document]
Meowwish: Agreed. I’m going to restructure the yield lookup, and the device-side query uses direct indexing, not an O(N) scan.

src/particles/particle_chemical_yield.hpp
```diff
@@ -0,0 +1,518 @@
...
+		return -1.0;
+	}
+	const std::string digits = name.substr(1, pos - 1);
+	if (digits.empty()) {
+		return -1.0;
+	}
+	const int numer = std::stoi(digits);
+	const amrex::Real denom = std::pow(10.0, static_cast<int>(digits.size()));
+	return static_cast<amrex::Real>(numer) / denom;
+}
```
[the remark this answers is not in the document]
Meowwish: Agreed.

src/particles/particle_deposition.hpp
Meowwish: Yes, the different kernels here are intentional. The metal deposition is tied to the parent feedback channel rather than forced to use a single kernel for all channels: SNII metals follow the SN injection kernel, while WR/AGB metals follow the stellar-wind injection kernel. My intent was to keep the chemical deposition spatially consistent with the underlying mass/energy feedback channel.
I’ll make that coupling explicit in the comments.

src/particles/particle_deposition.hpp
Meowwish: These have been copied from around line 500 of the same code file particle_depostion.hpp, I do this to ensure weights consistency between metal injection and SN feedback.

src/particles/particle_deposition.hpp
Meowwish: Agreed, let me refactor these.

src/particles/particle_types.hpp
Meowwish: Got it.

src/particles/particle_deposition.hpp
Meowwish: For SNII and AGB, I have opted for a single injection at the time of death. The corresponding code is written in particle_deposition. For WR, I have chosen to inject continuously in particle_updates, this allows for better temporal separation between WR and following SNII, thereby performs the destructive impact of SNII on WR better. As AGB are too numerous and have excessively long lifetimes, and as the space-time distribution of their injection is less correlated with the other channels, it's safe to regard it as a single injection.

To be consistent, SNII follows the SN kernel, whilst WR and AGB follow the stellar wind kernel.

Currently, WR injection employs a method of averaging yields over the entire lifetime; this is a temporary solution. I will subsequently create a new PR to implement a more realistic continuous mass loss model, replacing the current average injection scheme.

src/particles/particle_chemical_yield.hpp
Meowwish: I will process the original yield tables into SNII/WR/AGB DataTables locally.

src/particles/particle_types.hpp
chongchonghe: Actually, this is the number of particle components, not state variables, and it's independent of this PR. However, you raise a good point. I will create a pull request to fix this. We can use the value of the last index to replace the hard-coded 14 here.

src/particles/particle_chemical_yield.hpp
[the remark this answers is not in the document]
Meowwish: Changed to Quokka DattaTable.

src/particles/particle_types.hpp
Meowwish: Leave this for Chongchong.

src/particles/particle_chemical_yield.hpp
Meowwish: Modifed.

src/particles/particle_chemical_yield.hpp
Meowwish: Modified.

src/particles/particle_chemical_yield.hpp
Meowwish: Preprocessed the original tables locally, only submitted generated Quokka DataTables.

src/particles/particle_deposition.hpp
Meowwish: Refactored.

src/problems/TallBoxSf/testTallBoxSf.cpp
Meowwish: Tallbox stuff is not necessarily related to this PR, I plan to custermise it at local. All TallBox modification has been removed from this PR.

src/particles/particle_deposition.hpp
Meowwish: Updated. This comment now describes the legacy passive-scalar SN yield path, which is only used when table-driven chemical feedback is disabled.

scripts/validate_chemical_yields.py
```diff
@@ -0,0 +1,227 @@
...
+        print(f"  {isotope:4s} scalar_{i}: simulated={measured[i]:.8e} table={expected:.8e} sim/table={ratio:.8f}")
+    return max_error
+
+
+def validate_wr_agb(args):
+    isotopes = ["C12", "O16", "Fe56"]
+    tables = load_yield_tables(args.yield_root)
+    ds = yt.load(latest_plotfile(args.plotdir))
+    data = ds.all_data()
+    stages = data[("StochasticStellarPop_particles", "particle_evolution_stage")].to_value()
```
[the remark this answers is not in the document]
Meowwish: Removed the unused `stages` variable from the validation script.

chongchonghe:
# PR #1915 Review: SNII WR AGB Metal Yields

**PR**: https://github.com/quokka-astro/quokka/pull/1915
**Author**: Meowwish (Chuhan Zhang)
**Type**: Feature / Enhancement
**Review cycle**: SPEC → VERIFY → EDGE

---

## SPEC (Extracted Contract)

### New Inputs (TOML `particles.*` parameters)

| Parameter | Type | Default | Description |
|---|---|---|---|
| `enable_chemical_feedback` | bool | false | Master switch for chemical feedback |
| `enable_SNII_metal` | bool | true | Enable SNII channel |
| `enable_WR_metal` | bool | true | Enable WR channel |
| `enable_AGB_metal` | bool | true | Enable AGB channel |
| `store_channel_fields` | bool | true | Store per-channel scalars separately |
| `chemical_scalar_offset` | int | 0 | Offset in passive scalar array |
| `chemical_num_scalars` | int | 1 (auto-sizes) | Number of chemistry scalars |
| `use_table_driven_chemical_yield` | bool | true | Use lookup tables vs. parameterized |
| `chemical_yield_table_file` | string | `"yields"` | Path to yield table directory |
| `chemical_tracked_isotopes` | [string] | `["C12","N14","O16"]` | Isotopes to track |
| `chemical_tracked_channels` | [string] | `["SNII","WR","AGB"]` | Channels to enable |
| `snii_metal_yield_fraction` | real | 0.1 | Fallback SNII yield fraction |
| `wr_metal_yield_rate_per_mass` | real | 0.0 | Fallback WR yield rate [1/s] |
| `agb_metal_yield_rate_per_mass` | real | 0.0 | Fallback AGB yield rate [1/s] |
| `stellar_metallicity_fraction` | real | 0.014 | Metallicity for table lookup |
| `wr_age_start/end` | real | 0.0 | WR window for fallback mode |
| `agb_age_start/end` | real | 0.0 | AGB window for fallback mode |

### Outputs/Effects

- **SNII**: Isotopes deposited at supernova time via 7³-cell SN stencil (same kernel as thermal SN feedback)
- **WR**: Isotopes deposited continuously during WR phase via Wendland C2 kernel; time distribution controlled by 2D cumulative-fraction table (age × birth mass)
- **AGB**: Isotopes deposited at death time (stage `HighMassNonExploding`, ≤8 M☉) via Wendland C2 kernel; particle marked `Removed` after deposition

### Bundle

- Preprocessed yield tables: `extern/yields/quokka_yield_tables.tar.gz` (55MB)
- Two test problems: `test_SNII_Yields`, `test_WR_AGB_yields`
- Validation script: `scripts/validate_chemical_yields.py`
- CMake extract script: `scripts/cmake/extract_chemical_yields.cmake`

---

## VERIFY

### Build

| What | Result |
|---|---|
| `quokka build -d 3d test_SNII_Yields test_WR_AGB_yields` | ✅ SUCCESS (195/195 compile units) |
| Yield table extraction from tar.gz | ✅ 4 CSV files extracted correctly |
| Table metadata validation | ✅ SNII: 1D, 1662 mass pts, 389 isotopes. WR_dist: 2D, 2048×512. Correct format. |

### Runtime

| Phase | What was run | Observation |
|---|---|---|
| VERIFY — SNII yields | `test_SNII_Yields` with table-driven yields, 32³ grid | ⚠️ Hung >10 minutes on CPU (see Concerns) |
| VERIFY — SNII yields (reduced) | 8³ grid, max_timesteps=1, tables disabled | ⚠️ Still hung >2 min in Coarse STEP 1 |
| VERIFY — WR/AGB yields | `test_WR_AGB_yields` with 32³ grid | ⚠️ Hung >10 minutes on CPU |
| VERIFY — baseline | `Advection` 3D test on same build | ✅ Completes normally (~100 steps in <2 min) |

The tests **do not complete** on a CPU-only node (arm64, 1 MPI rank). The hang is not in the chemical yield code — it persists even with `use_table_driven_chemical_yield=0`. It occurs during the first hydro timestep (`Coarse STEP 1`), suggesting the SN scheme `SN_thermal_or_thermal_momentum` or particle initialization is causing dt to shrink to near-zero.

**The baseline Advection test works**, confirming the 3D build environment is functional.

---

## EDGE

Could not run edge-case tests due to the runtime hang. The following were attempted:
- Reduced grid (8³)
- Reduced timesteps (max_timesteps=1)
- Table-driven yields disabled
- Increase initial_dt to 1e14

None resolved the hang.

---

## Code Review Findings

### Architecture

The implementation is well-structured with clean separation:
- `particle_chemical_yield.hpp` — table loading, GPU-safe interpolation queries
- `ChemicalFeedbackDeposition` in `particle_deposition.hpp` — SNII + AGB event-driven deposition
- `updateChemicalFeedback` in `particle_update.hpp` — WR continuous deposition
- `PhysicsParticles.hpp` — wiring through the particle descriptor/registry pattern

The table-driven approach correctly:
1. Loads full 389-isotope tables from CSV
2. Extracts only user-requested isotopes into compact `SelectedChemicalYieldDataTable`s
3. Copies GPU-compatible const tables for device kernels
4. Gracefully falls back to parameterized yields if tables fail to load

### GPU Safety ✅

- `AMREX_GPU_MANAGED` correctly applied to all global state accessed from device
- `GpuArray` used for channel flags
- `ChemicalYieldGpuConstTables` passed by value into device lambdas
- No raw host pointers captured in `ParallelFor` lambdas

### Thread Safety ✅

- `mutableTables()` uses simple nullptr check (not thread-safe for first call, but only called from `particleParmParse()` during initialization — single-threaded)
- `constTables()` copies data by value — safe for concurrent reads

### DataTable Enhancement ✅

The new `FromFlatData` static method on `DataTable` is properly implemented and does not affect existing functionality. The CSV comment typo fixes (× → x) are cosmetic.

---

## PR-Introduced Concerns

### 1. Test performance makes verification impossible on CPU (HIGH)

The 3D hydro test problems with 32³ cells and SN feedback are impractically slow on CPU-only machines. They fail to complete even a single timestep within 2 minutes on 8³ cells. This means:
- Contributors cannot run tests locally without GPUs
- CI costs will be high (the PR has not yet triggered `/azp run` for GPU tests)
- The test design couples yield validation to full 3D hydro, which is unnecessarily expensive for testing scalar mass deposition

**Recommendation**: Decouple yield validation from hydro. Use `do_hydro = 0` or a single-cell test that only exercises the particle→grid deposition paths. The current validation checks total deposited mass against table-predicted mass — this does not require solving hydro.

### 2. Validation tests print results but don't assert (MEDIUM)

The Python validation script `validate_chemical_yields.py` prints sim/table ratios with `rtol=1e-10`, but the test problems themselves merely print "completed" with no pass/fail signal in the simulation output. The validation is only done in a separate ctest step. If someone runs the test binary directly (without ctest), they get no validation.

**Recommendation**: Consider adding a brief summary line to the simulation output showing the key ratios.

### 3. Yield table loading is ~55MB of CSV parsing at startup (LOW)

The full 389-isotope tables are parsed from CSV at every simulation startup. The WR mass-loss distribution table alone is 2048×512 = ~1M data points. For production runs with many restarts, this overhead could become significant.

**Recommendation**: Consider a binary cache format (e.g., write compact binary after first CSV parse, check mtime on subsequent runs). Not blocking for this PR.

### 4. Particle scalar initialization could be race-prone (LOW)

In `particle_creation.hpp`, chemistry scalars are initialized from the grid state:
```cpp
p.rdata(chem_base + nn) = std::max<amrex::Real>(0.0, state_arr(i, j, k, total_scalar) / cell_density);
```
This reads from `state_arr` which is the *current* grid state. If particles are created mid-simulation (not just at t=0), the grid scalars may already contain previously deposited yields, causing double-counting of birth abundances. For the current test problems this is not an issue (particles created at t=0 with zero scalars), but it should be documented.

---

None identified that are new to this PR.

---

## Summary

The PR adds a substantial, well-architected table-driven chemical yield system for SNII, WR, and AGB channels. The code quality is good — GPU safety is handled correctly, the DataTable integration is clean, and the fallback-to-parameterized design is robust.

**However, the tests cannot be verified** because they are impractically slow on CPU hardware. The hang is not in the new chemical yield code (it persists with tables disabled) but in the hydro/SN feedback path of the test setup. The test design needs to be simplified before this can be properly validated — decoupling yield deposition validation from 3D hydro would both fix the performance issue and make the tests more focused.

chongchonghe:
# PR #1915 Review: SNII WR AGB metal yields

**PR:** https://github.com/quokka-astro/quokka/pull/1915  
**Branch:** `SN-WR-AGB-metal-yields-minimal`  
**Review type:** Feature (SPEC → VERIFY → EDGE)  
**Reviewer:** Claude Code (automated)

---

## SPEC — Contract

This PR adds table-driven SNII, WR, and AGB metal yield deposition to `StochasticStellarPop` particles. The stated contract:

| Claim | Inputs | Expected output |
|---|---|---|
| SNII yields deposited at SN event | `enable_SNII_metal=1`, yield table | Scalar mass ≈ `f(mass) × M_birth` |
| WR yields deposited continuously | `enable_WR_metal=1`, WR distribution table | Cumulative scalar ≈ `f(mass) × M_birth × CDF(age)` |
| AGB yields deposited at particle death | `enable_AGB_metal=1`, yield table | Scalar mass ≈ `f(mass) × M_birth` |
| Table-driven lookup via `chemical_yield_table_file` | CSV tables in `extern/yields/` | 1:1 agreement with table |
| Two new validation tests | `test_SNII_Yields`, `test_WR_AGB_yields` | Ratio simulated/table ≈ 1.0 (rtol 1e-10) |

---

## VERIFY — SPEC Checklist

### CPU build (3D, no GPU): PASSED

```
quokka build -d 3d test_SNII_Yields test_WR_AGB_yields
test_SNII_Yields SUCCESS
test_WR_AGB_yields SUCCESS
```

Both targets compiled without errors on CPU.

### CPU test runs: FAILED (timeout — test design bug, see Finding 2)

Both `test_SNII_Yields` and `test_WR_AGB_yields` timed out at 1500 s (ctest default) without completing more than one timestep. Diagnosis via `gdb` on the stuck process shows the hang is inside `createParticlesFromState` → `RandomPoisson`:

```
#3  amrex::RandomPoisson(double)
#4  quokka::ParticleCreationTraits<StochasticStellarPop>::ParticleChecker::operator()(...)
#5  quokka::ParticleCreationTraits<StochasticStellarPop>::createParticles(...)
#6  PhysicsParticleDescriptor::createParticlesFromState(...)
#7  AMRSimulation::particleMeshInteraction(...)
```

The high-density ambient gas (`n0 = 1e5 cm^-3` with `eps_ff = 0.5`) causes the expected number of new star particles per timestep to be enormous, making `RandomPoisson` effectively non-terminating. The tests never exit STEP 1 and never exercise the yield validation logic.

### GPU build (3D-HIP, ROCm 6.4): FAILED — 9 compile errors

```
particle_deposition.hpp:388: error: reference to __host__ variable 'enable_chemical_feedback' in __device__ function
particle_deposition.hpp:547: error: reference to __host__ variable 'enable_chemical_feedback' in __device__ function
particle_update.hpp:178:  error: reference to __host__ variable 'stellar_metallicity_fraction' in __device__ function
particle_update.hpp:181:  error: reference to __host__ variable 'enable_WR_metal' in __device__ function
particle_update.hpp:201:  error: reference to __host__ variable 'wr_age_start' in __device__ function
particle_update.hpp:202:  error: reference to __host__ variable 'wr_age_start' in __device__ function
particle_update.hpp:205:  error: reference to __host__ variable 'wr_metal_yield_rate_per_mass' in __device__ function
particle_update.hpp:238:  error: reference to __host__ variable 'store_channel_fields' in __device__ function
particle_update.hpp:238:  error: reference to __host__ variable 'enable_WR_metal' in __device__ function
```

The PR does not build with any GPU backend. This blocks the quick CI pipeline and means GPU correctness has not been established.

---

## PR-Introduced Concerns

### Finding 1 — GPU BUILD BLOCKER: host variables accessed in device code (critical)

**Files:** `src/particles/particle_deposition.hpp:388,547`; `src/particles/particle_update.hpp:178,181,201,202,205,238`

The PR introduces direct access to `inline` host-only global variables (`enable_chemical_feedback`, `stellar_metallicity_fraction`, `enable_WR_metal`, `wr_age_start`, `wr_metal_yield_rate_per_mass`, `store_channel_fields`) from inside `AMREX_GPU_DEVICE` lambdas and device functions. On any GPU backend (CUDA or HIP) these are compile errors.

`depositThermalSNR` and `depositThermalKineticMomentumSNR` are `AMREX_GPU_DEVICE AMREX_FORCE_INLINE` functions; they cannot read host memory. The PR modified these functions at lines 387-393 and 545-551 to add `if (!enable_chemical_feedback)` guards, but `enable_chemical_feedback` is an `inline bool` host variable declared in `particle_types.hpp`. Similarly, the `ParallelFor` lambda in `updateChemicalFeedback` reads six host-side global variables directly without capturing local copies first.

**Fix:** Copy each affected host variable to a local named copy before the `ParallelFor` (as the author already did for `enable_snii_metal`, `enable_agb_metal`, `store_channel_fields_local`, etc. in `ChemicalFeedbackDeposition` at lines 1226-1231). For the device functions `depositThermalSNR`/`depositThermalKineticMomentumSNR`, pass `enable_chemical_feedback` as a `bool` parameter from the calling context where it can be captured.

---

### Finding 2 — TEST CORRECTNESS BUG: tests never exit step 1 (critical)

**Files:** `inputs/test_SNII_Yields.toml`, `inputs/test_WR_AGB_yields.toml`

Both tests use `n0 = 1e5 cm⁻³` (dense ISM) with `eps_ff = 0.5` but do not disable particle creation from the ISM. At this density, the free-fall time is ~2.9×10¹² s, shorter than `initial_dt = 1e13 s`. Each call to `createParticlesFromState` computes a Poisson rate proportional to `eps_ff × dt/t_ff × M_cell/M_sampling_mass`, which gives an enormous expected particle count. The `RandomPoisson(lambda)` call in `ParticleChecker` then samples from a Poisson distribution with a huge parameter, and the Kinderman-Ramage algorithm uses a rejection loop around `std::normal_distribution` that effectively never terminates.

Both tests were run independently (no competing processes) and both exceeded 25 minutes on one step of a 32³ grid with 2 particles.

**Fix:** Add an input parameter to disable particle creation for these tests (e.g. `particles.star_formation_enabled = 0`, or set `eps_ff = 0.0`, or use a gas density well below the Jeans/density threshold so no cells qualify for star formation). Alternatively, disable the `createParticlesFromState` call path entirely when `enable_chemical_feedback = 1` and the test purpose is to validate yield accounting for pre-existing particles.

---

### Finding 3 — GPU HOST/DEVICE CONFUSION in table loading (secondary GPU bug)

**File:** `src/particles/particle_chemical_yield.hpp:799`

`loadChannelTable` reads yield table data on the host via:
```cpp
const auto full_const = full_table.const_tables();  // returns device pointers
const auto data = full_const.dataViewArrays[out_idx];
for (int i = 0; i < num_entries; ++i) {
    flat_data[...] = data(i);  // host read from GPU device pointer
}
```

`const_tables()` returns a `DataTableGpuConst` backed by `data_d_` (device memory, not accessible from host). The `DataTable` class provides `const_tables_host()` for this use case, with documentation explicitly saying: "Use this for host-side initialization code that reads table values directly." This bug would cause a segfault or incorrect values on GPU builds (invisible on CPU builds since `data_d_` is host memory there).

**Fix:** Replace `full_table.const_tables()` with `full_table.const_tables_host()` on line 799 of `particle_chemical_yield.hpp`.

---

### Finding 4 — Dead code: commented-out return (minor)

**File:** `src/particles/particle_types.hpp:1397`

A dead commented-out return statement remains after the new particle component name generation code:
```cpp
return names;
// return expandEnumNames<StochasticStellarPopParticleRealIdx, ...>();
```

**Fix:** Remove the commented line.

---

- **NaN particle position in timestep output**: The test output shows `particle timestep limited at position (nan, nan, nan)` when `max particle velocity = 0`. This appears to be a pre-existing issue in the SN timestep estimator (division by zero when velocity is zero), unrelated to this PR.

---

## Summary

The PR implements a well-structured system for table-driven SNII/WR/AGB metal yield deposition with a clean separation between table loading, GPU interpolation, and deposition kernels. However, it cannot be merged in its current state due to two critical blockers:

1. **GPU build fails** with 9 host-in-device errors across `particle_deposition.hpp` and `particle_update.hpp` — the PR will break all GPU CI pipelines.
2. **Both test cases hang** indefinitely on step 1 due to runaway star formation from the high-density ambient gas — the yield validation logic has never been exercised.

A secondary GPU bug (Finding 3) would produce incorrect behavior when loading tables on GPU builds once Finding 1 is fixed.

chongchonghe:
# PR #1915 Review v2: SNII WR AGB metal yields

**PR:** https://github.com/quokka-astro/quokka/pull/1915  
**Branch:** `SN-WR-AGB-metal-yields-minimal` (Meowwish/quokka_metal)  
**Review type:** Feature (SPEC → VERIFY → EDGE), follow-up to [v1 review](pr1915-snii-wr-agb-yields-REVIEW.md)  
**Reviewer:** Claude Code (automated)  
**Date:** 2026-07-20

---

## Previous Findings: Status Update

Since the initial review, the author has added 14+ commits addressing the reported issues, including "fix GPU chemical yield captures", "fix CUDA runtime parameter captures", "fix chemical yield validation issues", "move chemical yield validation into C++ tests", and "fix host yield table validation."

### Finding 1 (GPU BUILD BLOCKER) — ✅ FIXED

The host variables that were previously accessed directly inside `AMREX_GPU_DEVICE` lambdas (`enable_chemical_feedback`, `stellar_metallicity_fraction`, `enable_WR_metal`, `wr_age_start`, `wr_metal_yield_rate_per_mass`, `store_channel_fields`) are now properly captured into local `_d` copies before the `ParallelFor` calls. `enable_chemical_feedback` is passed as a parameter to the device functions `depositThermalSNR` / `depositThermalKineticMomentumSNR`.

**GPU build verification:**
```
quokka build -d 3d-cuda SNIIYields WRAGBYields
SNIIYields SUCCESS
WRAGBYields SUCCESS
```
Both targets compile cleanly with the CUDA backend (nvcc, SM 7.5). Zero compile errors.

### Finding 3 (GPU HOST/DEVICE CONFUSION) — ✅ FIXED

`particle_chemical_yield.hpp:209` now uses `full_table.const_tables_host()` (with `_host` suffix) instead of `const_tables()`. The `constTablesHost()` helper at line 310–312 also correctly delegates to the `_host` variants. Verified by code inspection; also consistent with successful GPU build.

### Finding 4 (Dead code) — ✅ FIXED

The commented-out `// return expandEnumNames<...>()` line is no longer present. The enum name generation code has been restructured and cleaned up.

---

## Finding 2 (TEST HANG) — ❌ NOT FULLY FIXED (v2 root cause identified)

### Previous fix attempt

The `eps_ff = 0.0` setting was added to both test input files (`SNIIYields.toml`, `WRAGBYields.toml`). This correctly zeroes out the nominal star formation probability path:

```cpp
// particle_creation.hpp:452
const amrex::Real nominal_prob_star_formation = (eps_ff_ / eps_star) * (dt / t_ff);
// = (0.0 / 0.5) * (dt / t_ff) = 0.0  ← correct
```

### Why it doesn't work

The problem is the `J_truncate` forced star formation path at `particle_creation.hpp:454`:

```cpp
// particle_creation.hpp:453-454
// force P_sf to 1 if we are very far below the Jeans length
const amrex::Real actual_prob_star_formation =
    (LambdaJ < (J_truncate * dx[0])) ? 1.0 : nominal_prob_star_formation;
```

This path forces `actual_prob_star_formation = 1.0` whenever the cell's Jeans length is very small compared to the grid scale, **regardless of `eps_ff`**. With the test parameters (`n0 = 1e5` as mass density, `Tamb = 10 K`, domain ~8e19 cm, 32³ grid):

| Parameter | Value |
|---|---|
| Sound speed `cs` | 4.8e-8 cm/s |
| Jeans length `LambdaJ` | 5.9e-7 cm |
| `J_truncate * dx` | 1.2e16 cm |
| `J * dx` | 1.2e18 cm |
| `LambdaJ < J_truncate * dx` | **true** |
| `LambdaJ < J * dx` | **true** |

Both Jeans conditions are satisfied for every cell. The code enters the particle creation block and computes:

- `particle_mass = cell_density × cell_volume × eps_star = 1e5 × 1.5e55 × 0.5 ≈ 7.8e59 g`
- `num_high_mass_stars_exp ≈ 4e24`
- `RandomPoisson(4e24)` → **hangs** (Kinderman-Ramage algorithm cannot handle lambdas this large)

### Evidence

**Test with original `n0 = 1e5`** (via ctest, max_timesteps=1):  
Step 1 hydro advance completes ("Advanced 32768 cells"), then hangs indefinitely inside `particleMeshInteraction` → `createParticlesFromState` → `RandomPoisson`.

**Test with `n0 = 1e-24`** (via ctest, max_timesteps=2):  
Completes in 0.63 seconds but creates 32,769 particles (one per cell) — confirming the Jeans forced path is triggered even with `eps_ff = 0.0`:

```
>>>Particle creation:
	Time: 0 - Created 32769 particles at level 0
```

Validation fails because the SN has not occurred by `t = 2e13` (death_time = 1e14), but the critical point is that **particle creation is not disabled**.

### Fix Recommendation

The cleanest fix is an early-exit guard in `ParticleChecker::operator()` (`particle_creation.hpp:439`):

```cpp
AMREX_GPU_DEVICE auto operator()(...) const -> int
{
    if (eps_ff_ == 0.0) return 0;  // ← ADD: respect zero efficiency
    // ... existing Jeans / probability logic ...
}
```

This is consistent with the intent of `eps_ff = 0.0` and protects ALL paths (normal + Jeans-forced). A more targeted fix would be to multiply `actual_prob_star_formation` by a factor that vanishes when `eps_ff_ == 0.0`, but the early exit is simpler and less error-prone.

**Important:** This fix belongs in `particle_creation.hpp`, which is shared infrastructure — it's not test-specific. However, since the bug is only reachable with extreme parameter combinations that no existing production test hits, and the fix is a single well-defined line, it is appropriate to include in this PR.

---

## Evidence Table

| Phase | What you ran | What you observed |
|---|---|---|
| VERIFY – CPU build | `quokka build -d 3d SNIIYields WRAGBYields` | Both compile SUCCESS |
| VERIFY – CPU run | `ctest -R 'SNIIYields\|WRAGBYields' --timeout 600` | Both timeout after step 1 hydro advance (stuck in `createParticlesFromState`) |
| VERIFY – GPU build | `quokka build -d 3d-cuda SNIIYields WRAGBYields` | Both compile SUCCESS on CUDA (zero errors) |
| EDGE – low-density run | `n0 = 1e-24`, max_timesteps=2 | Completes in 0.63s but creates particles in every cell; validation fails (SN not triggered yet) |

---

## PR-Introduced Concerns

### Finding 2-v2 — TEST HANG: Jeans forced path bypasses `eps_ff = 0.0` (critical)

**File:** `src/particles/particle_creation.hpp:454`

The `J_truncate` forced star formation path sets `actual_prob_star_formation = 1.0` unconditionally when cells are below the Jeans length, bypassing the `eps_ff = 0.0` guard. The test's extreme parameters (density = 1e5 g/cm³, cold gas) trigger this path in every cell, leading to `RandomPoisson` calls with enormous lambda values that never return.

**Fix:** Add `if (eps_ff_ == 0.0) return 0;` at the top of `ParticleChecker::operator()` to provide a clean "no particle creation" switch that covers all code paths.

---

- **NaN particle position in timestep estimator**: Both tests show `particle timestep limited at position (nan, nan, nan)` when `max particle velocity = 0.0`. This is a pre-existing issue in the SN timestep estimator (division by zero when velocity is zero), unrelated to this PR.

---

## Summary

The GPU-related findings (1, 3, 4) from the previous review are all **fixed** — the PR now compiles cleanly on both CPU and CUDA backends, host variables are properly captured, and host/device table access is correct. The dead code has been removed.

However, the test hang (Finding 2) persists in a new form. The `eps_ff = 0.0` fix correctly disables the normal star formation path but does not prevent particle creation through the `J_truncate` Jeans-length forced path, which is triggered by the test's extreme density parameters. The fix is a single-line early-exit guard in `ParticleChecker::operator()`.

Once Finding 2-v2 is addressed, both validation tests should run and exercise their yield comparison logic for the first time.

chongchonghe:
**Verdict:** 1 blocking — `WRAGBYields` aborts 100% of the time on CPU *and* GPU; the yield physics itself validates exactly. The v2 `eps_ff`/Jeans blocker is fixed.
**Checked:** `3d` + `3d-hip` (ROCm, real device), both targets build with 0 errors. `SNIIYields` passes on both, all 3 isotopes at `sim/expected=1` (rtol 1e-10). `WRAGBYields` exits 8; with FPE traps off all 12 channel checks are exactly 1.

**Blocking**
1. **`assertClose` divides by `expected`, which is `0.0` for the SNII-channel checks** — `testWRAGBYields.cpp:81`. GCC -O3 speculates the divide above the `expected > 0.0` guard (faulting `divsd` at `0x2a0ad7` sits *before* the `ja`), and the test's own `amrex.fpe_trap_invalid=1` turns `0.0/0.0` into SIGFPE. Rewriting the ternary as an `if` does **not** help (still if-converted); use `|sim - exp| <= tol * exp` and never divide — verified green with traps on. Same construct at `testSNIIYields.cpp:80` is latent only because every expected value there is > 0.

**Non-blocking**
- **Particles grow by `4 * numPassiveScalars` reals unconditionally** — `particle_types.hpp:291`. Applies to DiskGalaxy/RandomBlast/TallBoxSf/ParticleSF even though `enable_chemical_feedback` defaults to `false`, and it breaks restart from existing checkpoints.
- **`14 + nGroups` hard-coded twice** — `particle_types.hpp:271` duplicates line 286. Adding one real component silently shifts every `p.rdata(chem_base + n)` read; derive it from the enum instead.
- **SNII stencil deposits when the yield is zero** — `particle_deposition.hpp:1176`. 343 atomic adds of 0.0 per isotope per particle per step; the AGB call two lines down already guards on `> 0.0`.
- **Archive ships 5 macOS `._*` files** — `extern/yields/quokka_yield_tables.tar.gz`; they land untracked in `git status` after the prepare-yields fixture. Repack with `COPYFILE_DISABLE=1`. (The 10 MB blob in git history is a separate maintainer call.)

Pre-existing, not this PR: all 4 `ParticleSF` variants SIGFPE on the first ADVANCE — reproduced identically at merge base `16e5b96f`.

| Phase | Ran | Observed |
|---|---|---|
| Build CPU | `quokka build -d 3d SNIIYields WRAGBYields` | both `SUCCESS` |
| Build GPU | `quokka build -d 3d-hip SNIIYields WRAGBYields` | both `SUCCESS`, 0 errors |
| Test CPU | `ctest -R 'SNIIYields\|WRAGBYields'` | `SNIIYields` Passed 2.96s; `WRAGBYields` **Failed** 3.92s |
| Test GPU | both binaries, real AMD device | `SNIIYields` exit 0; `WRAGBYields` exit 8, same abort |
| v2 regression | `particle_creation.hpp:441` | `if (eps_ff_ <= 0.0) { return 0; }` present; no hang, no spurious particles |
| Edge / base | `ParticleSF` at merge base `16e5b96f` | fails identically → pre-existing |

```
SNIIYields (GPU):
  C12 scalar_0: simulated=4.8556969e+32   expected=4.8556969e+32   sim/expected=1
  N14 scalar_1: simulated=8.11502849e+31  expected=8.11502849e+31  sim/expected=1
  O16 scalar_2: simulated=4.112635623e+33 expected=4.112635623e+33 sim/expected=1

WRAGBYields (traps on):
  C12 AGB scalar_9: simulated=5.747871126e+31 expected=5.747871126e+31 sim/expected=1
Abort(8) on node 0 (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 8)
Backtrace.0 frame 3: (anonymous namespace)::assertClose(...) at testWRAGBYields.cpp

2a0ac9:  ucomisd %xmm2,%xmm1     ; test expected > 0.0
2a0ad7:  divsd   %xmm1,%xmm2     ; <-- faults here, BEFORE the branch
2a0ae7:  ja      2a0af5          ; the guard, taken after the divide

WRAGBYields (traps off) — all 12 checks sim/expected=1, including
  Fe56 WR scalar_8: 2.703016916e+32 / 2.703016916e+32
  Fe56 SNII scalar_5: 0 / 0
```

</details>

src/problems/SNIIYields/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+if(AMReX_SPACEDIM EQUAL 3)
+    set(JOB_NAME SNIIYields)
+
+    quokka_add_problem(JOB_NAME ${JOB_NAME})
+    add_test(
+        NAME ${JOB_NAME}_zero_yield
+        COMMAND ${JOB_NAME} ${CMAKE_SOURCE_DIR}/inputs/${JOB_NAME}.toml
+            problem.test_zero_yield_only=1
+            amrex.fpe_trap_invalid=1 amrex.fpe_trap_overflow=1 amrex.fpe_trap_zero=1)
+    add_test(
```
chongchonghe: This causes the problem to rely on setting system env in order to run the test problem, which is not a good way to set up the test problem. It also requires one to run the `_prepare_yields` test before running the SNIIYields test, which adds an unnecessary hurdle. Should follow up the example in DiskGalaxy to use the `file` command to extract the yields table.

src/problems/WRAGBYields/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+if(AMReX_SPACEDIM EQUAL 3)
+    set(JOB_NAME WRAGBYields)
+
+    quokka_add_problem(JOB_NAME ${JOB_NAME})
+    add_test(
+        NAME ${JOB_NAME}_zero_yield
+        COMMAND ${JOB_NAME} ${CMAKE_SOURCE_DIR}/inputs/${JOB_NAME}.toml
+            problem.test_zero_yield_only=1
+            amrex.fpe_trap_invalid=1 amrex.fpe_trap_overflow=1 amrex.fpe_trap_zero=1)
+    add_test(
```
chongchonghe: similarly, should use `file` command. See my other comments.

extern/yields/.gitignore
```diff
@@ -0,0 +1,12 @@
+# Raw yield data and extracted Quokka datatables are generated artifacts.
+AGB_Karakas16/
+SNII_Kobayashi0611/
+SNII_Sukhbold16/
+superAGB_Doherty14/
+chuhan_channel_yields.dat
```
chongchonghe: you can ignore all `*.csv` files here, and remove all irrelevant files from gitignore

src/problems/SNIIYields/testSNIIYields.cpp
```diff
@@ -0,0 +1,231 @@
...
+{
+	const volatile amrex::Real zero_yield = 0.0;
+	assertClose("zero yield", zero_yield, zero_yield);
+	const auto legacy_names = quokka::getParticleRealCompNames<quokka::ParticleType::StochasticStellarPop, YieldStorageDisabled>();
+	AMREX_ALWAYS_ASSERT(legacy_names.size() == quokka::StochasticStellarPopParticleRealComps<YieldStorageDisabled>);
+	const auto chemistry_names = quokka::getParticleRealCompNames<quokka::ParticleType::StochasticStellarPop, test_SNII_Yields>();
+	AMREX_ALWAYS_ASSERT(chemistry_names.size() == quokka::StochasticStellarPopParticleRealComps<test_SNII_Yields>);
+	bool test_zero_yield_only = false;
+	amrex::ParmParse("problem").query("test_zero_yield_only", test_zero_yield_only);
+	if (test_zero_yield_only) {
```
chongchonghe: I seems to me `test_zero_yield_only` is not doing anything. The only difference it makes is that the `SNIIYields_zero_yield` test returns here, but the `SNIIYields` test also runs the code above plus the rest of the code, so `SNIIYields` includes what `SNIIYields_zero_yield` does. You could remove this variable and remove the `SNIIYields_zero_yield` test.

src/problems/SNIIYields/testSNIIYields.cpp
```diff
@@ -0,0 +1,231 @@
...
+namespace
+{
+
+constexpr amrex::Real yield_validation_rtol = 1.0e-10;
+
+struct InitialParticleRecord {
+	std::vector<amrex::Real> rdata;
+};
+
+auto readInitialParticleRecords(const std::string &filename, int nreal) -> std::vector<InitialParticleRecord>
```
chongchonghe: Optional: you can use the `getParticleDescriptor` function to read the initial particle information from `sim` so that you don't need to write your own particle reader. See examples in testBinaryOrbitCIC.cpp (code below)

```
	const auto &real_data = sim.particleRegister_.getParticleDescriptor(quokka::ParticleType::CIC)->getParticleDataAtLevel(0).first;
```

src/problems/WRAGBYields/testWRAGBYields.cpp
```diff
@@ -0,0 +1,264 @@
...
+namespace
+{
+
+constexpr amrex::Real yield_validation_rtol = 1.0e-10;
+
+struct InitialParticleRecord {
+	std::vector<amrex::Real> rdata;
+};
+
+auto readInitialParticleRecords(const std::string &filename, int nreal) -> std::vector<InitialParticleRecord>
```
chongchonghe: see my other similar comment

src/problems/WRAGBYields/testWRAGBYields.cpp
```diff
@@ -0,0 +1,264 @@
...
+template <> struct Particle_Traits<WRAGBYields> : DefaultParticleTraits {
+	static constexpr bool enable_chemical_feedback = true;
+	static constexpr ParticleSwitch particle_switch = ParticleSwitch::StochasticStellarPop;
+};
+
+template <> struct HydroSystem_Traits<WRAGBYields> {
+	static constexpr bool reconstruct_eint = true;
+};
+
+template <> struct Physics_Traits<WRAGBYields> : DefaultPhysicsTraits {
```
chongchonghe, line 119: Most of these variables are assigned to their default value, so you can skip them. See examples in any other test problems.

src/problems/WRAGBYields/testWRAGBYields.cpp
```diff
@@ -0,0 +1,264 @@
...
+	});
+}
+
+auto problem_main() -> int
+{
+	const volatile amrex::Real zero_yield = 0.0;
+	assertClose("zero yield", zero_yield, zero_yield);
+	bool test_zero_yield_only = false;
+	amrex::ParmParse("problem").query("test_zero_yield_only", test_zero_yield_only);
+	if (test_zero_yield_only) {
```
chongchonghe: see my other similar comment

src/particles/particle_types.hpp
```diff
@@ -249,16 +263,29 @@ constexpr int StochasticStellarPopParticleMassAtBirthIdx = static_cast<int>(Stoc
...
+
+template <typename problem_t> constexpr auto StochasticStellarPopParticleChemistryBlockBaseIdx(int blockIndex) -> int
+{
+	return StochasticStellarPopParticleChemistryBaseIdx<problem_t>() + blockIndex * StochasticStellarPopParticleChemistryBlockSize<problem_t>();
+}
+
+// Number of real components for StochasticStellarPop_particles, mass + 3 velocity components + times + positions + death density + luminosity
+template <typename problem_t>
+constexpr int StochasticStellarPopParticleRealComps =
+    StochasticStellarPopParticleChemistryBaseIdx<problem_t>() + 4 * StochasticStellarPopParticleChemistryBlockSize<problem_t>();
```
chongchonghe: The number 4 is a magic number here; it's the number of channels + 1. I would make this programmatic: . So you can set `constexpr int numOfChannels = 3` . And then, use `numOfChannels+1` here.

src/particles/particle_types.hpp
```diff
@@ -408,7 +435,36 @@ template <ParticleType particleType, typename problem_t> auto getParticleRealCom
...
+		}
+		const auto &isotopes = chemicalTrackedIsotopeList();
+		const auto isotope_count = static_cast<int>(isotopes.size());
+		const auto isotopeName = [&](int idx) -> std::string {
+			if (idx >= 0 && idx < isotope_count) {
+				return isotopes[static_cast<std::size_t>(idx)];
+			}
+			return "unused_" + std::to_string(idx);
+		};
+		const std::array<std::string, 3> channel_names = {"SNII", "WR", "AGB"};
```
chongchonghe: use `numOfChannels` here and other places.

src/particles/particle_types.hpp
```diff
@@ -249,16 +263,29 @@ constexpr int StochasticStellarPopParticleMassAtBirthIdx = static_cast<int>(Stoc
 constexpr int StochasticStellarPopParticleLumIdx = static_cast<int>(StochasticStellarPopParticleRealIdx::luminosity); // Base index for luminosity components
 constexpr int StochasticStellarPopParticleStageIdx = static_cast<int>(StochasticStellarPopParticleIntIdx::evolution_stage);
 
-// Number of real components for StochasticStellarPop_particles, mass + 3 velocity components + times + positions + death density + luminosity
-template <typename problem_t>
-constexpr int StochasticStellarPopParticleRealComps = []() constexpr {
+template <typename problem_t> constexpr auto StochasticStellarPopParticleChemistryBlockSize() -> int
```
chongchonghe: A name like `numOfYieldSpecies` is more readable than `ChemistryBlockSize`.

src/particles/particle_types.hpp
```diff
@@ -408,7 +435,36 @@ template <ParticleType particleType, typename problem_t> auto getParticleRealCom
 	} else if constexpr (particleType == ParticleType::CICRad) {
 		return expandEnumNames<CICRadParticleRealIdx, CICRadParticleRealComps<problem_t>, true>();
 	} else if constexpr (particleType == ParticleType::StochasticStellarPop) {
-		return expandEnumNames<StochasticStellarPopParticleRealIdx, StochasticStellarPopParticleRealComps<problem_t>, true>();
+		amrex::Vector<std::string> names;
```
chongchonghe, line 439: Since the construction of component names here is different from other particle types, it's helpful to document what you do here. You could explain how the component names are constructed with explicit examples.

src/particles/particle_update.hpp
```diff
@@ -98,6 +115,154 @@ template <> struct ParticlePropertyUpdateTraits<ParticleType::StochasticStellarP
...
+		ChemicalYieldLookup::ChemicalYieldGpuConstTables yield_tables{};
+		if (use_chemical_tables) {
+			yield_tables = ChemicalYieldLookup::constTables();
+		}
+
+		amrex::MultiFab state_buffer(state.boxArray(), state.DistributionMap(), state.nComp() + 1, state.nGrow());
+		state_buffer.setVal(0.0);
+
+		for (typename ContainerType::ParIterType pti(*container, lev); pti.isValid(); ++pti) {
+			auto &particles = pti.GetArrayOfStructs();
```
chongchonghe, line 167: Most of the variables between lines 151 and 170 could be moved outside of this for-loop. A basic principle is that if a variable does not change with the for loop, define it outside.

src/simulation.hpp
```diff
@@ -1564,6 +1564,7 @@ template <typename problem_t> void AMRSimulation<problem_t>::evolve()
 			// Stellar evolution and SN deposition; only apply to star particles
 			// Update particle properties (e.g., luminosity) before particle-mesh interaction
 			particleRegister_.updateParticleProperties(cur_time, dt_[0]);
+			particleRegister_.updateChemicalFeedback(state_new_cc_[finest_level], finest_level, cur_time, dt_[0]);
```
chongchonghe: Shouldn't this be inside `particleMeshInteraction`? Both your current `updateChemicalFeedback` and `particleMeshInteraction` will execute every time step, so it will handle both continuous deposition and a one-time deposition.

chongchonghe (changes requested):

src/problems/SNIIYields/CMakeLists.txt
Meowwish: Fixed locally by moving extraction to configure time using file(ARCHIVE_EXTRACT), following the DiskGalaxy approach. Both yield tests now share the same preparation logic.

src/problems/WRAGBYields/CMakeLists.txt
Meowwish: fixed

extern/yields/.gitignore
Meowwish: fixed

src/problems/SNIIYields/testSNIIYields.cpp
Meowwish: Removed the runtime switch and the separate SNII zero-yield test. I preserved the regression coverage in a dedicated YieldValidation test that exercises the same shared assertion helper with floating-point traps explicitly enabled. This retains isolated zero-yield coverage without adding a special execution mode to the simulation.

src/problems/SNIIYields/testSNIIYields.cpp
Meowwish: Fixed by using the StochasticStellarPop descriptor's getParticleDataAtLevel(0) instead of a separate ASCII parser. The snapshot is taken after initialization and before evolution, preserving the initial properties needed for the reference calculation. The helper also accounts for position components and broadcasts the gathered data to the other MPI ranks.

src/problems/WRAGBYields/testWRAGBYields.cpp
Meowwish: Fixed

src/problems/WRAGBYields/testWRAGBYields.cpp
Meowwish: fixed. Now it retains only the overrides that differ from `DefaultPhysicsTraits : is_hydro_enabled = true` and `numPassiveScalars = 12`

src/problems/WRAGBYields/testWRAGBYields.cpp
Meowwish: Romoved.

src/particles/particle_types.hpp
Meowwish: fixed

src/particles/particle_types.hpp
Meowwish: fixed

src/particles/particle_types.hpp
Meowwish: Renamed to `StochasticStellarPopParticleChemistryBlockCapacity`. I chose “capacity” rather than `numOfYieldSpecies` because this value represents the storage stride, not the actual number of tracked isotopes. For example, WRAGBYields reserves 12 slots per block while tracking three isotopes.

src/particles/particle_update.hpp
Meowwish: fixed

src/simulation.hpp
Meowwish: Fixed locally by moving continuous WR feedback to the beginning of `particleMeshInteraction`, before state fixup, accretion, and particle creation.

src/particles/particle_types.hpp
Meowwish: Added documentation describing the total/SNII/WR/AGB block order and names such as `chem_birth_total_C12`, `chem_birth_SNII_C12`, and `chem_birth_AGB_C12`. It also explains reserved slots such as `chem_birth_total_unused_3` and clarifies that these components store birth abundances.

PR #1921, opened 2026-05-28
[WIP] Add artificial resistivity: stabiliser for Quokka2026

### LINKED ISSUE #1912: The Quokka2026 EMF scheme shows instabilities for slow waves at high resolution

AstroKriel:
### Description

I have added artificial resistivity to the constrained-transport MHD induction update as a numerical stabiliser for ideal MHD runs using the `Quokka2026` scheme. At each cell edge, a resistive coefficient `eta_dynamic` is computed from the local Alfven speed and grid spacing, and the term `eta_dynamic * j_edge` is subtracted from the ideal EMF:

    emf_total = emf_ideal - eta_dynamic * j_edge

    eta_dynamic = alpha * |b|_perp / sqrt(rho_avg) * sqrt(dx_1 * dx_2)

where `|b|_perp = sqrt(b_w0_avg^2 + b_w1_avg^2)` is the magnitude of the two face-centred field components perpendicular to the edge (each reconstructed at the cell edge by averaging the components on neighbouring faces), and `rho_avg` is the density averaged over the four surrounding cell centres. Here a new parameter `mhd.artificial_resistivity_coefficient` (off by default) regulates this local artificial resistivity.

**Motivation.** The `Quokka2026` EMF scheme goes unstable for the slow magnetosonic wave at N >= 512 cells per wavelength (#1912). The instability grows in `b_0` (the field component along the propagation axis) from machine-noise amplitudes. A current-based formula (`eta_dynamic` proportional to `|j|`) fails here because `|j| ~ 0`, even in regions where the instability is forming; that is, the instability mode grows before any current builds up. However, the `c_A * dx` scaling is non-zero wherever `|b| > 0`, so when `alpha > 0`, it provides background damping at all amplitudes from the first timestep. This is analogous to artificial bulk viscosity.

I have validated this scheme in a number of ways, and confirmed it brings the stability of `Quokka2026` to be in line with both `FelkerStone2017` and `Balsara2025`.

**Linear wave convergence** (`Quokka2026`, alpha = 0.05). `eta_dynamic` damps wave amplitudes at a rate gamma ~ alpha * c_A * k^2 * dx, which limits convergence to p ~ 0.5 for all wave families. Without artificial resistivity, the slow-wave L2 error increases sharply at N = 512 as the instability forms; with alpha = 0.05 this is suppressed and the error decreases monotonically to N = 512. The fast wave follows 2nd-order convergence at low resolution before transitioning to the same 0.5-order plateau as `eta_dynamic` becomes the dominant error source. The same O(dx) damping costs ~ 1% Alfven wave amplitude per orbit at N = 256 with alpha = 0.025, halving with each resolution doubling.

(The top left axis shows the behaviour when there is no artificial resisitivity, and the remaining axis show when it is turned on)
art_eta_convergence

| `Quokka2026` with alpha | b_z amplitude loss per orbit |
|---|---|
| alpha = 0.025 | ~ 1% |
| alpha = 0.05 | ~ 2% |

**Orszag-Tang vortex** (512 x 512 x 8, `Quokka2026`). alpha = 0.05 triggers a CFL crash at t ~ 0.7 from concentrated current sheets. When turning on alpha = 0.1, the simulation runs to completion, and performs comparably to `FelkerStone2017`.

**Recommended coefficient.** `alpha` is a tunable knob. For 1D wave tests where amplitude accuracy matters, alpha = 0.025 gives ~ 1% Alfven wave damping per orbit at N = 256. For problems with strong current sheets (Orszag-Tang strength), alpha ~ 0.1 is needed for stability.

### Related issues

Stacked on #1913 (physical resistivity), which should merge first. This PR's diff is against the `nkriel/add/resistivity` branch.

Closes #1912 (`Quokka2026` slow-wave instability at N >= 512).

### Checklist


BenWibking:
This is encouraging that it stabilizes the Orzag-Tang problem.

However, I am concerned something is not quite right. A linear artificial viscosity for hydro of the form $\alpha (\Delta x) c_s$ limits the order of convergence to 1. This should do the same. I think it should really converge at order 1, not order 0.5, if it's designed and implemented correctly.

markkrumholz:
I have a major concern with this approach: since the magnitude of the resistivity doesn’t depend on the gradient of anything, it will just be applied everywhere. That is, in the linear wave tests the Alfven speed is basically constant everywhere, so the is approach just reduces to using a constant resistivity everywhere. In that sense it is not at all analogous to an artificial viscosity, which only kicks in where there is a large compression. So I don’t think this is a particularly good approach, because it’s not clear why you should ever use this and not just set the resistivity to a constant.

markkrumholz:
Following up: you said that the approach of taking resistivity proportional to j fails. How large a coefficient did you try?

BenWibking:
> I have a major concern with this approach: since the magnitude of the resistivity doesn’t depend on the gradient of anything, it will just be applied everywhere. That is, in the linear wave tests the Alfven speed is basically constant everywhere, so the is approach just reduces to using a constant resistivity everywhere. In that sense it is not at all analogous to an artificial viscosity, which only kicks in where there is a large compression. So I don’t think this is a particularly good approach, because it’s not clear why you should ever use this and not just set the resistivity to a constant.

Ok, I basically agree with this.

I think this is actually the kind of artificial viscosity that Lagrangian codes often use (proportional to the cell size times the sound speed), but it's very, very diffusive. If this were required to make Quokka2026 stable, I would not recommend using it in practice.

AstroKriel:
> This is encouraging that it stabilizes the Orzag-Tang problem.
> 
> However, I am concerned something is not quite right. A linear artificial viscosity for hydro of the form α ( Δ x ) c s limits the order of convergence to 1. This should do the same. I think it should really converge at order 1, not order 0.5, if it's designed and implemented correctly.

This p = 1/2 is actually expected as a consequence of how the convergence tests are set up.

The Richardson tests only refine `n_0` (cells in `x_0`), so the grid is always `n_0 x 8 x 8`. The artificial resistivity acting on an edge that damps the wave along `x_0` is then `eta_dynamic = alpha * c_A * sqrt(dx_0 * dx_1)`. Since `dx_1 = L/8` is fixed (where `L = 1`; similarly `dx_2 = L/8`), `eta_dynamic` scales as `sqrt(dx_0)`, giving p = 1/2 on the log-log axis shown. p = 1 is correct for isotropic refinement.

Note that the p = 2 scaling at coarse resolution (visible in the fast wave panel) is because the truncation error for a 1D wave depends only on `dx_0` (fields are uniform in `x_1` and `x_2`), and this error dominates at low resolutions. The `eta_dynamic` error differs because both perpendicular edge spacings are important.

AstroKriel:
> I have a major concern with this approach: since the magnitude of the resistivity doesn’t depend on the gradient of anything, it will just be applied everywhere. That is, in the linear wave tests the Alfven speed is basically constant everywhere, so the is approach just reduces to using a constant resistivity everywhere. In that sense it is not at all analogous to an artificial viscosity, which only kicks in where there is a large compression. So I don’t think this is a particularly good approach, because it’s not clear why you should ever use this and not just set the resistivity to a constant.

You're right, this term is always on, like with constant resistivity, but the difference is that `eta_dynamic = alpha * c_A * dx` goes to zero as resolution increases (`dx` becomes smaller). Physical `eta` is fixed and its effect on the wave does not change with resolution.

AstroKriel:
> Following up: you said that the approach of taking resistivity proportional to j fails. How large a coefficient did you try?

I tested up to alpha = 1. The instability in the slow wave continued to grow over three wave periods.

markkrumholz:
> > Following up: you said that the approach of taking resistivity proportional to j fails. How large a coefficient did you try?
> 
> I tested up to alpha = 1. The instability in the slow wave continued to grow over three wave periods.

OK, then I think we want to look a bit more closely at the nature of the instability, i.e., what does the current look like in the cells where it is growing. The current solution is just too diffusive to be practical to use in a simulation where we don't want there to be physical resistivity everywhere, so we need to find something more narrowly tailored.

PR #1933, opened 2026-06-03
add AMR to thermal conduction module

aditivijayan:
### Description
Adding AMR capabilities to the Thermal Conduction routine. 

Major change is in the definition of addStrangSplitSourceswithBuiltin which now need flux registers to be passes are arguments.

### Related issues

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Increased AMR refinement depth for thermal conduction runs to allow more levels in the refinement hierarchy.

* **Bug Fixes**
  * Improved electron conduction operator splitting to compute and reflux heat flux consistently across refinement boundaries (including higher AMR levels).
  * Added a validation to require subcycling to be off when electron conduction is enabled.
  * Reduced/disabled component residual output during runs.

* **Tests**
  * Updated the thermal conduction test parameters and GPU refinement tagging, switched to a Richardson-extrapolation workflow for verification.


aditivijayan:
@BenWibking - not sure why the DiskGalaxy is failing.

BenWibking:
> @BenWibking - not sure why the DiskGalaxy is failing.

I think it was just a Github hiccup. I've triggered a re-run for that action.

aditivijayan:
> > @BenWibking - not sure why the DiskGalaxy is failing.
> 
> I think it was just a Github hiccup. I've triggered a re-run for that action.

Great thanks! Please do let me know if there are comments/issues I need to address.

aditivijayan:
@markkrumholz  @BenWibking - can this be merged if there are no other comments/issues?

BenWibking:
> @markkrumholz @BenWibking - can this be merged if there are no other comments/issues?

I have not had a chance to review this in detail. I will try to get to in the next couple of days.

BenWibking:
@chongchonghe Do you know why `/azp run quick` didn't work on this PR?

chongchonghe:
now quick and rocm-quick is working

chongchonghe:
@aditivijayan The PR claims to add AMR support to ThermalConduction, but the test problem has not turned on AMR. Can you turn it on?

chongchonghe:
you also need to add a refinement criteria to the problem

chongchonghe:
I think this PR is not ready for review yet. You need to test the feature you added.

src/QuokkaSimulation.hpp
```diff
@@ -1115,6 +1116,8 @@ auto QuokkaSimulation<problem_t>::addStrangSplitSourcesWithBuiltin(amrex::MultiF
 				amrex::Abort("Electron conduction not implemented for > 0 levels.");
```
BenWibking: I don't see how this PR actually works. It looks like it should abort if more than 1 level exists in the test problem. Can you run this PR with an AMR test problem and include a plot of the results?

aditivijayan:
> you also need to add a refinement criteria to the problem

The refinement criterion has been added at - https://github.com/quokka-astro/quokka/pull/1933/changes#diff-40536436be20f3760e4de5cf5cd1320896802865fcb5639712d8e2cdd8bb391cR91

The input has been updated.

inputs/ThermalConduction.toml
```diff
@@ -14,7 +14,7 @@ amr.v = 1      # verbosity in Amr
 # Resolution and refinement
 # *****************************************************************
 amr.n_cell = [128, 128, 128]
-amr.max_level = 0  # number of levels = max_level + 1
+amr.max_level = 1  # number of levels = max_level + 1
```
BenWibking, line 18: This parameter is overridden by the Richardson test machinery, so this problem does not actually test AMR.

aditivijayan: Yep, that was it. Currently testing the problem.

aditivijayan:
Okay, so I checked-- the evolution seems to be okay.

Level 0:
image

Level 1:
image

The width of the Gaussian plotted against time is also a straight line and matches the expectation.
image

Following our discussion in the meeting as @markkrumholz suggested I'll try and figure out why exactly the test is failing.

aditivijayan:
image

@markkrumholz @BenWibking I did a bunch of experiments with changing resolution and AMR on/off. It seems that relative error increases when AMR is on. Is there any other way to estimate error with AMR?

aditivijayan:
So, I did the experiment with plotting out the relative errors. It seems most of the error is coming from the region being refined rather than out. 

image

The error seems to be increasing at the coarse fine boundary (shown in vertical lines).

markkrumholz:
That's interesting. If you add just 64 with no AMR to that plot, how does it compare?

aditivijayan:
> That's interesting. If you add just 64 with no AMR to that plot, how does it compare?

image

Shape of 64 (no AMR) is similar to 128 (no AMR) and it matches 64+AMR at large distances from the coarse-fine boundary.

BenWibking:
This looks consistent with error being generated at the coarse-fine boundary, although I don't see an obvious bug that would cause this.

BenWibking:
Did you run this test with AMR subcycling turned on?

BenWibking:
Ok, Codex found the issue: when subcycling is on (it is by default), the conduction module does not actually take a smaller timestep on refined levels compared to hydro (which it must, for stability):

```
• ## Review findings

  - P1 — AMR conduction violates its own stability limit. This PR enables max_level = 1 and removes the multilevel abort,
    exposing an existing timestep error. src/simulation.hpp:1332 takes only r fine substeps but multiplies the fine
    stability limit by r²; the assigned fine timestep is still dt₀/r at src/simulation.hpp:1387. For refinement ratio 2,
    fine conduction advances at twice its computed stable timestep. This is likely the numerical failure that looks like bad
    refluxing.

  - P1 — Rejected hydro attempts contaminate the flux register. Conduction immediately calls incrementFluxRegisters() at
    src/QuokkaSimulation.hpp:1155, before the CFL/burn acceptance check at src/QuokkaSimulation.hpp:2607. When an attempt
    fails, the retry loop restores only the state at src/QuokkaSimulation.hpp:2201; it does not undo the conduction
    CrseInit(ADD)/FineAdd. Reflux therefore includes fluxes from rejected attempts. Conduction contributions should be
    accumulated temporarily and committed only when final_success is true, like the hydro fluxes.

  - P1 — Valid no-op conduction configurations dereference undefined fluxes. ComputeExplicit() returns without defining
    heat_flux for dt <= 0, nonpositive conductivity, or an isothermal EOS at src/conduction/ElectronConduction.hpp:50. The
    reflux block nevertheless calls boxArray() and DistributionMap() on those default-constructed MultiFabs at src/
    QuokkaSimulation.hpp:1158. Either always define zero fluxes or return whether fluxes were produced.
```

The other two P1 issues need to be fixed before merging.

aditivijayan:
> Ok, Codex found the issue: when subcycling is on (it is by default), the conduction module does not actually take a smaller timestep on refined levels compared to hydro (which it must, for stability):
> 
> ```
> • ## Review findings
> 
>   - P1 — AMR conduction violates its own stability limit. This PR enables max_level = 1 and removes the multilevel abort,
>     exposing an existing timestep error. src/simulation.hpp:1332 takes only r fine substeps but multiplies the fine
>     stability limit by r²; the assigned fine timestep is still dt₀/r at src/simulation.hpp:1387. For refinement ratio 2,
>     fine conduction advances at twice its computed stable timestep. This is likely the numerical failure that looks like bad
>     refluxing.
> 
>   - P1 — Rejected hydro attempts contaminate the flux register. Conduction immediately calls incrementFluxRegisters() at
>     src/QuokkaSimulation.hpp:1155, before the CFL/burn acceptance check at src/QuokkaSimulation.hpp:2607. When an attempt
>     fails, the retry loop restores only the state at src/QuokkaSimulation.hpp:2201; it does not undo the conduction
>     CrseInit(ADD)/FineAdd. Reflux therefore includes fluxes from rejected attempts. Conduction contributions should be
>     accumulated temporarily and committed only when final_success is true, like the hydro fluxes.
> 
>   - P1 — Valid no-op conduction configurations dereference undefined fluxes. ComputeExplicit() returns without defining
>     heat_flux for dt <= 0, nonpositive conductivity, or an isothermal EOS at src/conduction/ElectronConduction.hpp:50. The
>     reflux block nevertheless calls boxArray() and DistributionMap() on those default-constructed MultiFabs at src/
>     QuokkaSimulation.hpp:1158. Either always define zero fluxes or return whether fluxes were produced.
> ```
> 
> The other two P1 issues need to be fixed before merging.

Right. How do you recommend I proceed with the sub stepping? One way would be run the ComputeExplicit loop nsubsteps times in QuokkaSimulation.hpp.

BenWibking:
> Right. How do you recommend I proceed with the sub stepping? One way would be run the ComputeExplicit loop nsubsteps times in QuokkaSimulation.hpp.

I would just have it abort if both conduction and `do_subcycle=1` are set, and require using a global timestep when conduction is enabled for now.

aditivijayan:
> > Right. How do you recommend I proceed with the sub stepping? One way would be run the ComputeExplicit loop nsubsteps times in QuokkaSimulation.hpp.
> 
> I would just have it abort if both conduction and `do_subcycle=1` are set, and require using a global timestep when conduction is enabled for now.

image

Turning off subcycling changes nothing, effectively.

BenWibking:
I realized another way error could be introduced at the coarse-fine boundary is if the coarse-level AMR interpolation is interpolated at the wrong time in the fine-level boundary condition call. For Strang splitting, the boundary conditions live at different times. If both the first-half and the second-half Strang split update pass the same time to the boundary conditions function, it will cause the order of accuracy to drop to first order.

BenWibking:
So I would redo the convergence test without subcycling, and plot L2, L1, and L_inf error norms. L2 should be close to second order, and L_inf should be close to zeroth order.

aditivijayan:
> So I would redo the convergence test without subcycling, and plot L2, L1, and L_inf error norms. L2 should be close to second order, and L_inf should be close to zeroth order.

image

It isn't monotonic for these resolutions. But if I were to remove 256:

image

aditivijayan:
> P1 — Valid no-op conduction configurations dereference undefined fluxes. ComputeExplicit() returns without defining
>     heat_flux for dt <= 0, nonpositive conductivity, or an isothermal EOS at src/conduction/ElectronConduction.hpp:50. The
>     reflux block nevertheless calls boxArray() and DistributionMap() on those default-constructed MultiFabs at src/
>     QuokkaSimulation.hpp:1158. Either always define zero fluxes or return whether fluxes were produced.

My commits this week have addressed these issues. 

Commit fixing updating flux registers, for issue 2, also solved issues 3.

aditivijayan:
@markkrumholz - this is ready for comments.

markkrumholz:
Do you have a convergence test result now showing that you get the expected order of convergence across resolution when you turn off level-by-level time stepping?

inputs/ThermalConduction.toml
markkrumholz: Is this resolved? If so, please close.

src/QuokkaSimulation.hpp
```diff
@@ -389,8 +393,8 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 
 	void addStrangSplitSources(amrex::MultiFab &state, int lev, amrex::Real time, amrex::Real dt_lev);
 	template <SourceOrder Order>
-	auto addStrangSplitSourcesWithBuiltin(amrex::MultiFab &state, std::array<amrex::MultiFab, AMREX_SPACEDIM> &state_fc, int lev, amrex::Real time,
-					      amrex::Real dt_lev) -> bool;
```
markkrumholz, line 407: This is a fairly major code infrastructure change, since you're changing the call signature of a routine that's used in many places. What is the reason for making this change, and have you verified that making this API change doesn't break other problems that use `addStringSplitSourcesWithBuiltin`?

src/QuokkaSimulation.hpp
```diff
@@ -1064,6 +1065,13 @@ auto QuokkaSimulation<problem_t>::addStrangSplitSourcesWithBuiltin(amrex::MultiF
...
 			}
 			fillBoundaryConditions(state, state, lev, time, quokka::centering::cc, quokka::direction::na, PreInterpState, PostInterpState);
+			std::array<amrex::MultiFab, AMREX_SPACEDIM> heat_flux;
+
+			for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+				auto ba_face = amrex::convert(state.boxArray(), amrex::IntVect::TheDimensionVector(idim));
+				heat_flux[idim].define(ba_face, state.DistributionMap(), state.nComp(), 0);
+				heat_flux[idim].setVal(0.0);
+			}
 			if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
```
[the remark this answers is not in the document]
markkrumholz: Has this comment been addressed?

src/QuokkaSimulation.hpp
```diff
@@ -1140,22 +1145,36 @@ auto QuokkaSimulation<problem_t>::addStrangSplitSourcesWithBuiltin(amrex::MultiF
...
+			std::array<amrex::MultiFab, AMREX_SPACEDIM> heat_flux;
+
 			if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
 				for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
 					fillBoundaryConditions(state_fc[idim], state_fc[idim], lev, time, quokka::centering::fc,
 							       static_cast<quokka::direction>(idim), AMRSimulation<problem_t>::InterpHookNone,
 							       AMRSimulation<problem_t>::InterpHookNone, FillPatchType::fillpatch_function);
 				}
 			}
+			// for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
```
markkrumholz: If this is dead code, please remove

src/QuokkaSimulation.hpp
```diff
@@ -1140,22 +1145,36 @@ auto QuokkaSimulation<problem_t>::addStrangSplitSourcesWithBuiltin(amrex::MultiF
...
+			// 	heat_flux[idim].setVal(0.0);
+			// }
 			const quokka::conduction::ElectronConductionParams conduction_params{.conductivity_prefactor = electronConductionKappa0_,
 											     .flux_limiter_phi = electronConductionFluxLimiterPhi_,
 											     .saturation_factor = electronConductionSaturationFactor_,
 											     .min_temperature = tempFloor_};
-			quokka::conduction::ElectronConduction<problem_t>::ComputeExplicit(state, state_fc, geom[lev], dt, conduction_params);
+			quokka::conduction::ElectronConduction<problem_t>::ComputeExplicit(state, state_fc, geom[lev], dt, conduction_params, heat_flux);
+			if (do_reflux) {
+				// std::array<amrex::MultiFab, AMREX_SPACEDIM> recal_fluxes;
```
markkrumholz: Again, if these commented-out lines (here and below the for loop) are dead code, please remove.

src/QuokkaSimulation.hpp
```diff
@@ -1325,6 +1344,7 @@ auto QuokkaSimulation<problem_t>::computeComponentErrors() -> std::vector<std::t
 	amrex::MultiFab residual(state_ref_level0.boxArray(), state_ref_level0.DistributionMap(), ncomp, 0);
 	amrex::MultiFab::Copy(residual, state_ref_level0, 0, 0, ncomp, 0);
 	amrex::MultiFab::Saxpy(residual, -1., state_new_cc_[0], 0, 0, ncomp, 0);
+	// amrex::WriteSingleLevelPlotfile("point_wise_error", residual, componentNames_cc_, geom[0], /*time=*/0.0, /*level_step=*/0);
```
markkrumholz: Remove dead code.

src/QuokkaSimulation.hpp
markkrumholz: Never mind, I see this is addressed in the PR description.

markkrumholz:
@aditivijayan, I have left some relatively minor cleanup comments. The main question is, do you have a plot demonstrating that with adaptive time stepping turned off, you achieve the expected convergence rate?

aditivijayan:
> @aditivijayan, I have left some relatively minor cleanup comments. The main question is, do you have a plot demonstrating that with adaptive time stepping turned off, you achieve the expected convergence rate?

It passes the Richardson test so I didn't plot anything. Did you have a recommendation?

markkrumholz:
> > @aditivijayan, I have left some relatively minor cleanup comments. The main question is, do you have a plot demonstrating that with adaptive time stepping turned off, you achieve the expected convergence rate?
> 
> It passes the Richardson test so I didn't plot anything. Did you have a recommendation?

Since there is a plot in the PR history showing a failed test, I think one showing a successful test with your changes is useful to include, just to complete the record. Just make a quick plot of the Richardson test numerical output.

aditivijayan:
> > > @aditivijayan, I have left some relatively minor cleanup comments. The main question is, do you have a plot demonstrating that with adaptive time stepping turned off, you achieve the expected convergence rate?
> > 
> > 
> > It passes the Richardson test so I didn't plot anything. Did you have a recommendation?
> 
> Since there is a plot in the PR history showing a failed test, I think one showing a successful test with your changes is useful to include, just to complete the record. Just make a quick plot of the Richardson test numerical output.

The plot below shows the relative error in internal energy across the domain. The vertical lines indicate the AMR region. 

image

markkrumholz:
OK, looks good now. @aditivijayan, can you please make sure you have addressed and closed all the open comments and conversations, and then I can approve?

src/QuokkaSimulation.hpp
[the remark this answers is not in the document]
aditivijayan: Yes, this comment has been addressed. In the earlier versions `heat_flux` was defined in `ComputeElectronConduction`. That reference has been removed and `heat_flux` is now passed as an argument.

aditivijayan:
> OK, looks good now. @aditivijayan, can you please make sure you have addressed and closed all the open comments and conversations, and then I can approve?

Thanks! I've replied to the unresolved comment. This PR may be approved once it passes tests.

markkrumholz (dismissed):

chongchonghe:
The GPU tests failed.

chongchonghe:
# PR #1933 Review — "add AMR to thermal conduction module"

**Focus of this review:** diagnose the failing GPU CI jobs and recommend fixes.

**Headline:** The failing CI job is **not a GPU problem**. `ThermalConduction` fails identically on a
plain x86 CPU build. The real defect is that the new AMR conduction path is **correct in 1D and 2D but
wrong in 3D** — and 3D is the only configuration CI actually executes, so a 3D bug presents itself as a
"GPU failure".

---

## 1. What CI is actually reporting

Both failing jobs fail on the same single test, with **bit-identical** error norms:

| Azure build | Pipeline | Result |
|---|---|---|
| [14744](https://dev.azure.com/quokka-code/13c66f5b-4db2-4bf3-b04e-10fa5e417a1d/_build/results?buildId=14744) | `release [GPU, 3D] (ROCm, amdclang)` | 130/131 pass; `ThermalConduction` **Failed** |
| [14746](https://dev.azure.com/quokka-code/13c66f5b-4db2-4bf3-b04e-10fa5e417a1d/_build/results?buildId=14746) | `rocm-quick` | 0/1 pass; `ThermalConduction` **Failed** |

```
Resolution   Error Norm
        32   3.223677e-03
        64   1.488040e-03
       128   2.476436e-03      <-- error INCREASES

  32 ->  64        1.12   (expected 2.0)
  64 -> 128       -0.73   (expected 2.0)
Overall convergence rate: 0.19
✗ Richardson convergence test FAILED
```

Every compile-only job (CUDA 13.0, HIP ROCm 7.14, all `build` jobs) passes, so this is a **runtime
numerical failure**, not a device-code or lambda-capture problem.

---

## 2. Evidence table

All local runs are on the PR head (`5f49a35f`), CPU only, this box (no GPU device present).
Commands used:

```bash
quokka config -d <1d|2d|3d> --delete --source ~/superpowers/bin/quokka.rc --root $(pwd) -DQUOKKA_PYTHON=OFF
quokka build  -d <preset> ThermalConduction --source ~/superpowers/bin/quokka.rc --root $(pwd)
./build/<preset>/src/problems/ThermalConduction/ThermalConduction inputs/ThermalConduction.toml
```

| # | Configuration | nx=32 | nx=64 | nx=128 | Verdict |
|---|---|---|---|---|---|
| 1 | **3D GPU (ROCm CI)**, AMR on | 3.223677e-03 | 1.488040e-03 | 2.476436e-03 | **FAILED** |
| 2 | **3D CPU**, AMR on | 3.223677e-03 | 1.488040e-03 | not run | reproduces CI **bit-for-bit** at nx=32 and nx=64 |
| 3 | **1D CPU**, AMR on | 2.139514e-03 | 5.332087e-04 | 1.168682e-04 | **PASSED** (rates 2.00 / 2.19) |
| 4 | **2D CPU**, AMR on | 2.139514e-03 | 5.332087e-04 | 1.168682e-04 | **PASSED** — identical to 1D bit-for-bit |
| 5 | **1D CPU**, AMR off | 3.780588e-03 | 9.576027e-04 | 2.389567e-04 | PASSED (rates 1.98 / 2.00) |
| 6 | **3D CPU**, AMR off | 3.780588e-03 | 9.576027e-04 | — | **identical to 1D bit-for-bit** |
| 7 | **1D CPU**, AMR on, `do_reflux=0` | 2.925017e-03 | 7.490420e-04 | 1.905456e-04 | PASSED (rates 1.97 / 1.98) |
| 8 | **3D CPU**, AMR on, `do_reflux=0` | 2.826764e-03 | — | — | within 3.4% of 1D |

Rows 5–6 are the critical control: **with AMR disabled, 3D and 1D agree to the last bit.** The
dimensional discrepancy appears *only* once `max_level = 1` is active.

Row 2 is the critical CI finding: a ROCm GPU and an x86 CPU produce the *same digits*. There is nothing
GPU-specific to fix.

### Reproducing without a GPU

The failure reproduces in ~25 min on a 48-core CPU box, and the cheap resolutions (nx=32) reproduce in
about a minute. This is a far better debugging loop than Azure. 1D/2D builds run the identical physics
(the problem is uniform in y and z) in seconds — but note they **pass**, so only a 3D build exposes the bug.

---

## 3. Analysis

### 3.1 The problem is exactly 1-dimensional, yet 3D disagrees

The initial condition, the analytic reference solution, and the refinement criterion
(`std::abs(x) < 2*sigma`, `testThermalConduction.cpp:121`) all depend on **x only**. The refined region is
a slab spanning the full y–z extent of the domain — verified from the level-1 cell counts (nx=32: 3D
level 1 = 131072 = 32×64×64; 1D level 1 = 32; nx=128 CI: 6291456 = 96×256×256, 1D: 96). The grid geometry
is therefore identical across dimensionalities, and `y-GasMomentum` / `z-GasMomentum` errors are
**exactly** `0.0000e+00` in the 3D run — the state really does stay uniform in y and z.

It follows that in exact arithmetic 3D must reproduce 1D. Rows 5–6 confirm the code does exactly that
**when AMR is off**. So the 3D-only discrepancy is introduced by the AMR path this PR enables.

### 3.2 The conduction reflux helps in 1D and hurts in 3D

Comparing rows 3/7 (1D) and 2/8 (3D) at nx=32:

| | reflux **off** | reflux **on** | effect of reflux |
|---|---|---|---|
| 1D | 2.925017e-03 | 2.139514e-03 | **−27%** (improves, as intended) |
| 3D | 2.826764e-03 | 3.223677e-03 | **+14%** (degrades) |

With reflux disabled, 1D and 3D agree to 3.4%. With reflux enabled, they diverge by 51%, and the 3D
convergence rate collapses (row 1). **The new conduction contribution to the flux registers is the
dominant source of the 3D error**, and it is being applied incorrectly in 3D. The residual 3.4% gap in
row 8 indicates a second, smaller 3D/AMR discrepancy that also deserves a look.

This points squarely at the block added in `src/QuokkaSimulation.hpp:1165-1177` and its companion call at
`src/QuokkaSimulation.hpp:2642`.

Worth noting while debugging: the flux-register scaling itself is dimension-invariant by construction —
`incrementFluxRegisters` (`src/simulation.hpp:2345-2357`) multiplies by `face_area` (1, `dx[1]`,
`dx[1]*dx[2]` in 1/2/3D) and `FluxRegister::Reflux` (`src/simulation.hpp:2291`) divides by cell volume,
so the ratio is `1/dx` in every dimensionality. A naive area/volume factor error would therefore break 2D
too, and 2D is bit-identical to 1D (row 4). The bug is something that only engages with two transverse
directions.

### 3.3 Why CI did not catch this earlier

No GitHub Actions workflow runs `ctest` — every CPU job is build-only (verified: `grep -l ctest
.github/workflows/*.yml` returns nothing). The Azure GPU pipelines are the **only** place Quokka tests
actually execute, and they build 3D. Combined with the fact that this bug is 3D-only, the sole signal
available is a "GPU CI failure".

Also note the last `development` GPU build that ran the full suite before this PR ([14078], May 2026) had
**114** tests and did not include `ThermalConduction` at all; the current suite has 131. The relevant
green baseline is the merge-queue build
[14752](https://dev.azure.com/quokka-code/13c66f5b-4db2-4bf3-b04e-10fa5e417a1d/_build/results?buildId=14752)
(2026-08-01, 131/131 pass) — i.e. `ThermalConduction` passes on `development` with `max_level = 0`, and
fails only with this PR's `max_level = 1`.

---

## 4. Recommended fixes

**Do not chase the GPU, and do not relax the tolerance.** The test is correctly reporting a real defect.

1. **Debug in 3D on CPU.** It reproduces bit-for-bit and iterates far faster than Azure. Use nx=32 (about
   a minute) for the inner loop.

2. **Instrument the transverse heat fluxes first.** Since the state is provably uniform in y and z,
   `heat_flux[1]` and `heat_flux[2]` must be identically zero on *both* levels at every step. Print
   `heat_flux[idim].norm0()` in `ElectronConduction::ComputeExplicit` after the flux loops.
   - If they are non-zero → the fine-level ghost fill at the coarse–fine boundary
     (`fillBoundaryConditions`, `QuokkaSimulation.hpp:1150`) is contaminating transverse gradients.
   - If they are zero → the fault is in how the x-register is accumulated/applied in 3D; compare the
     level-0 reflux correction against the analytically expected
     `dt * (F_fine_avg - F_coarse) / dx` at the two coarse–fine interfaces.

3. **Consider folding the conduction flux into the existing hydro flux array** rather than making a
   second `incrementFluxRegisters` pass. Adding the (weighted) heat flux into `flux_rk2`'s energy
   components before the single existing call removes an entire duplicated code path and guarantees the
   conduction flux gets exactly the same scaling, timing and `final_success` gating as hydro. This is the
   change most likely to make the 3D behaviour fall into line with 1D/2D.

4. **Fix the still-open latent crash** (Codex P1 #3 — reported as fixed, but it is not).
   `ElectronConduction::ComputeExplicit` returns early at `ElectronConduction.hpp:46-53` for
   `dt <= 0`, `conductivity_prefactor <= 0`, or an isothermal EOS — **before** `heat_flux` is defined at
   `ElectronConduction.hpp:119-123`. The caller then unconditionally does
   `amrex::MultiFab::Saxpy(..., heat_flux[idim], ...)` at `QuokkaSimulation.hpp:1172-1175`, operating on
   default-constructed (undefined) MultiFabs. Setting `conduction.enabled = 1` with
   `conduction.conductivity_prefactor = 0` and `max_level > 0` will crash. Either always define and zero
   the flux MultiFabs before the early returns, or have `ComputeExplicit` return a `bool` the caller checks.

5. **Justify or revert the `sigma` change.** `testThermalConduction.cpp:43` changes the Gaussian width
   from `2.410685615625e+17` to `2.41016e+17` (commit `f9cccf4a`, "remove test for debugging"), a 0.02%
   change with no stated rationale. The old value is exactly `5*dx` at nx=64, matching the documented
   recipe in the file header ("Width of the gaussian = sigma = 5 * dx"); the new value is not. Changing a
   test's physical parameters in the same PR that changes the physics the test validates makes the result
   hard to interpret — please state why, or restore the original.

6. **Get 3D tests running somewhere cheaper.** Because the only executing CI is the GPU pipeline, any
   3D-only regression will keep surfacing as an opaque GPU failure. A CPU `ctest` job (even a small subset
   in 3D) would have caught this directly. Note a 1D/2D CPU job would *not* have helped here — both pass.

### Minor / non-blocking

7. `refineGrid` (`testThermalConduction.cpp:126-133`) loops over
   `{x0,x1} × {y0,y1} × {z0,z1}` and calls `tagIfPointInRegion` 8 times, but the lambda ignores its `y`
   and `z` arguments and tests only `std::abs(x) < refine_Lmax`. Six of the eight calls are redundant.
   Collapsing to a loop over `{x0, x1}` would say what is meant. (Copied from another problem's tagging
   routine, where the y/z checks are real.)

---

## 5. PR-Introduced Concerns

1. **The 3D AMR conduction result is wrong** (§3). Convergence collapses to rate 0.19; error *grows* from
   nx=64 to nx=128. Blocking.
2. **Latent crash on undefined `heat_flux` MultiFabs** when `ComputeExplicit` returns early with reflux
   and AMR active (`ElectronConduction.hpp:46-53` vs `QuokkaSimulation.hpp:1172-1175`). Reported as fixed
   in the PR discussion, but the code path still exists.
3. **Unexplained change to the test's physical parameter `sigma`** (`testThermalConduction.cpp:43`).

Codex's earlier P1 #2 (rejected hydro attempts contaminating the flux register) **is** genuinely fixed:
`recal_fluxes` is now a local of `advanceHydroAtLevel` (`QuokkaSimulation.hpp:2290`), freshly zeroed per
attempt, and is only committed under `if (do_reflux == 1 && final_success)` (`QuokkaSimulation.hpp:2640`).
The retry loop lives in `advanceHydroAtLevelWithRetries` and re-enters `advanceHydroAtLevel` per attempt,
so no accumulation leaks across attempts.

The new `do_subcycle == 0` assertion (`QuokkaSimulation.hpp:292-295`) is safe for the existing test
suite: `inputs/ThermalConduction.toml` is the only input file that enables conduction, and it sets
`do_subcycle = 0`.

---

## 6. Pre-existing Issues (Out of Scope)

1. **The conduction timestep has no dimensional factor.** `src/simulation.hpp:1275` uses
   `0.5 * conductionCFL * dx_min^2 / D` regardless of `AMREX_SPACEDIM`. The stability limit for an
   explicit isotropic diffusion operator tightens as `1/(2*NDIM)`, so in 3D this sits at roughly 90% of
   the linear stability boundary rather than the ~30% the CFL number implies. It is not the cause of this
   failure (this problem has zero transverse gradients, so the effective operator is 1D, and 3D without
   AMR is exact — rows 5–6), but a genuinely 3D conduction problem would be far closer to the edge than
   the input file suggests. Pre-existing; does not block this PR; worth a separate issue.

---

## Notes on method

- A temporary, uncommitted diagnostic hook (`QK_MAX_LEVEL` env var around the hardcoded
  `pp.add("max_level", …)`) was used locally to produce rows 5–8; it is **not** part of any proposed
  change and was reverted.
- Rows 2, 6 and 8 at the highest resolutions were still completing when this report was written; the
  numbers shown are those confirmed at the time. The conclusions rest on rows 1–7, all of which are
  complete for the resolutions listed.
- The 3D CPU build used `-DAMReX_OMP=ON` for speed. Results were identical at 4, 6, 16 and 40 threads and
  identical to the ROCm GPU, so OpenMP is not a factor.

aditivijayan:
Alrighty! The main issue with this PR was the failing Richardson test. After a lot of debugging the error was narrowed down to the AMR interpolator falling to properly interpolate at the fine-coarse boundary since it was limiting. The fix has been to force it not limit (see simulation.hpp). 

image

aditivijayan:
Some of the tests were cancelled. I am unsure why.

chongchonghe:
> Some of the tests were cancelled. I am unsure why.

it's because they've reached time limit. Either a server bug, or some code issue that causes the quokka to freeze and never finish.

aditivijayan:
The GPU tests are failing because of timeout issues.

markkrumholz:
The quick test failures are not timeouts, they're real failures of the conduction test. I just pulled the logs, and the quick tests ran the thermal conduction test, but it reported failure.

markkrumholz:
I take that back -- it is a timeout, but if the test can't run in 25 minutes then it's too long, and should be shortened.

aditivijayan:
> I take that back -- it is a timeout, but if the test can't run in 25 minutes then it's too long, and should be shortened.

The GPU tests are 3D and have AMR over base grids of 32**3, 64**3, 128**3. The test takes 450 s on gadi hopper queue. I've reduced the duration over which the solution is evolved.

aditivijayan:
Following a conversation with @markkrumholz the test has been modified to compare with result from a full convergence test. This is to fix the GPU test fails which were failing due to the tests taking too long to evolve 3D problem with the three sets of resolutions. Instead now the test only evolves nx=32 and compares with the result from the full convergence test.

For posterity's sake, I am showing the results of the full convergence test.

### nx=32 
Screenshot 2026-08-24 at 1 45 58 pm


### nx=64

Screenshot 2026-08-24 at 1 46 25 pm


### nx=128

Screenshot 2026-08-24 at 1 46 47 pm


### Rate estimation-

Screenshot 2026-08-24 at 1 47 16 pm

aditivijayan:
The csv file from the full convergence test.

[thermal_conduction_convergence.csv](https://github.com/user-attachments/files/31360249/thermal_conduction_convergence.csv)

aditivijayan:
@markkrumholz - this is ready for review.

markkrumholz (approved):

markkrumholz:
@aditivijayan, I have approved. To merge there is just one outstanding AI review comment that I think you have already addressed, but please just confirm and if so resolve the conversation. Once that is done, you should be able to merge.

src/QuokkaSimulation.hpp
```diff
@@ -1076,7 +1084,11 @@ auto QuokkaSimulation<problem_t>::addStrangSplitSourcesWithBuiltin(amrex::MultiF
 											     .saturation_factor = electronConductionSaturationFactor_,
 											     .min_temperature = tempFloor_,
 											     .eos_flag = eosFlagForElectronConduction_};
-			quokka::conduction::ElectronConduction<problem_t>::ComputeExplicit(state, state_fc, geom[lev], dt, conduction_params, resampledTables_);
+			quokka::conduction::ElectronConduction<problem_t>::ComputeExplicit(state, state_fc, geom[lev], dt, conduction_params, resampledTables_,
+											   heat_flux);
+			if (do_reflux) {
+				incrementFluxRegisters(fr_as_crse, fr_as_fine, heat_flux, lev, dt);
+			}
 		}
```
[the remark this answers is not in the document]
aditivijayan: @markkrumholz - Yes, this comment has been addressed. `recal_fluxes` is now defined as a mf [here](https://github.com/quokka-astro/quokka/pull/1933/changes#diff-6b93dd1680d502d10d17da95078058219ea0277d3952e8dfeec0057f5c8e2fa0R2344). The internal energy and total energy components of `recal_fluxes` are updated [here](https://github.com/quokka-astro/quokka/pull/1933/changes#diff-6b93dd1680d502d10d17da95078058219ea0277d3952e8dfeec0057f5c8e2fa0R1218). Finally, the flux registers are updated only after the final success [here](https://github.com/quokka-astro/quokka/pull/1933/changes#diff-6b93dd1680d502d10d17da95078058219ea0277d3952e8dfeec0057f5c8e2fa0R2694).

PR #1939, opened 2026-06-08
Fix FOFC EMF fallback to use donor-cell reconstruction

AstroKriel:
### Description

The first-order flux correction (FOFC) for MHD was computing the fallback EMF at the user-requested reconstruction order (e.g. order 5 for PPM-EP) rather than dropping to first order. This is inconsistent with the hydro fallback path, which uses donor-cell (piecewise-constant) reconstruction and the LLF Riemann solver.

The fix passes `reconstructionOrder=1` to `ComputeEMF` for the FOFC fallback, so it uses the existing donor-cell (`ReconstructStatesConstant`) path — matching what `hydroFOFluxFunction` does on the hydro side.

### Related issues

Identified while investigating the Quokka2026 EMF instability. The order-1 reconstruction path already exists in `MHDSystem::ReconstructTo`; it was simply not being used for FOFC.

### Checklist


AstroKriel:
Closing to revise — submitted prematurely.

PR #1956, opened 2026-06-15
fix(photochem): exclude flux from VODE convergence; hard-code atol for DTypeFront; add docs

chongchonghe:
### Description

Add hard-coded VODE absolute tolerances for the DTypeFront photoionization problem, and exclude radiation flux from VODE convergence checks. `docs/markdown/photoionization.md` now has **complete documentation** for the photoionization module: governing equations (M1 radiative transfer, hydrogen thermochemistry, on-the-spot approximation), numerical scheme (operator splitting with VODE), and the full VODE tolerance design rationale. `docs/markdown/parameters.md` has the complete `integrator.*` parameter reference.

The `SetAtolFromPhysics` machinery (automatic atol derivation from physical scales) has been moved to PR #1980.

### Summary of changes

| File | Change |
|---|---|
| `src/problems/DTypeFront/testDTypeFront.cpp` | `Erad_floor = a_rad * (0.01 K)^4` with documentation of the relationship to `atol_rad_num` |
| `inputs/DTypeFront.toml` | Hard-coded `integrator.atol_*` values with documentation |
| `src/networks/photoionization/actual_rhs.H` | Clamp `N_gamma >= 0` in RHS and Jacobian to prevent sign reversal when photons are fully absorbed |
| `docs/markdown/photoionization.md` | New: full photoionization module documentation — governing equations, numerical scheme, VODE tolerance design |
| `docs/markdown/parameters.md` | New Integrator section documenting all `integrator.*` parameters |
| `extern/Microphysics/.../vode_dvode.H` | Zero flux error weight — excludes it from all convergence norms |
| `extern/Microphysics/.../vode_dvnlsd.H` | Skip flux indices in corrector DEL and accumulated error ACNRM |
| `extern/Microphysics/.../vode_dvstep.H` | Remove flux change-factor step rejection check |
| `extern/Microphysics/.../vode_dvhin.H` | Do not constrain initial step by flux tolerance |

### Key design decisions

- **Hard-coded tolerances for DTypeFront** — `atol_spec = 1.4e-3 cm^-3`, `atol_enuc = 1.24e8 erg/g`, `atol_rad_num = 5e-2 cm^-3`, `radiation_failure_tolerance = 0.5 cm^-3`. `atol_spec` and `atol_enuc` correspond to a negligibility floor of 1e-5 × n_H and a 1 K temperature accuracy at n_H = 139.9 cm^-3. `atol_rad_num = 5e-2` is the performance plateau for this problem (see Performance below).
- **`atol_rad_num` and `radiation_failure_tolerance` are coupled** — `radiation_failure_tolerance` must exceed the maximum negative excursion that the BDF predictor-corrector produces at the ionization front before converging (~0.05–0.2 cm^-3 for DTypeFront). Setting `radiation_failure_tolerance = 0.5` provides a safe margin. Note: Microphysics commit `47696e49` introduced a 1.5× final-state factor for **species** (`vode_final_state_species_failure_tolerance_factor`), which allows `species_failure_tolerance = atol_spec` without manual inflation — but no equivalent factor exists for radiation, so `radiation_failure_tolerance` requires its own margin.
- **`Erad_floor` is separate from `atol_rad_num`** — the former is a compile-time `constexpr` in `RadSystem_Traits` controlling the M1 hyperbolic floor; the latter is the VODE tolerance. They must satisfy `N_gamma_floor << atol_rad_num` so dark cells return in one BDF step.
- **Flux is excluded from VODE convergence checks** — flux is a passive scalar (one-way coupled: `dF/dt` depends on `n_HI` but flux does not feed back into species, energy, or N_gamma). N_gamma has an additional isotropic source term (stellar emission) that flux lacks, so flux is not analytically determined by N_gamma — but it remains a diagnostic quantity whose convergence should not drive the timestep.

### Performance

All benchmarks: ROCm GPU (64^3, AMD GPU on moth.anu.edu.au).

**vs. `development` branch (64^3, `stop_time = 2.895e12` ≈ 3 t_s):**

| Branch | Total (s) | PhotoChem (s) | PhotoChem % |
|---|---|---|---|
| `development` | 439.8 | 436.1 | 99.1 % |
| this PR | 14.2 | 10.5 | 73.9 % |
| **speedup** | **31.0×** | **41.6×** | |

**Effect of `atol_rad_num` tuning within this PR (64^3, `stop_time = 1.93e13` = 20 t_s):**

`atol_rad_num` was swept from 1.25e-6 to 0.2 cm^-3. Performance plateaus at ~5e-2 cm^-3: below this VODE spends extra steps in dark cells; above it the step-rejection loop (N_gamma overshooting near −`radiation_failure_tolerance` at the ionization front) erodes the gain.

| `atol_rad_num` | Total (s) | PhotoChem (s) | PhotoChem % |
|---|---|---|---|
| 1.25e-6 (initial) | 218.7 | 194.3 | 88.8 % |
| 5e-2 (final) | 98.3 | 73.9 | 75.2 % |
| **speedup** | **2.22×** | **2.63×** | |

### Related issues
N/A

### Checklist


BenWibking:
This looks like a good approach. However, this might take me a few days to review in detail.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,113 @@
...
+### Derived atol values
+
+Let `T_min ≡ typical_minimal_radiation_T` for brevity.
+`SetAtolFromPhysics` computes:
+
+| Variable                      | Formula                                      | Rationale                                                                                                              |
+| ----------------------------- | -------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- |
+| `atol_spec`                   | `spec_abundance_tol × typical_n_H`           | Species below this fraction of `n_H` are negligible                                                                    |
+| `atol_enuc`                   | `c_v × desired_accuracy_on_T_at_typical_n_H` | Converts temperature accuracy to internal energy tolerance (`c_v = 3/2 × k_B / m_p`)                                   |
+| `atol_rad_num`                | `1e-6 × a_rad × T_min⁴ / E_photon`           | One millionth of the blackbody photon density at `T_min` — radiation below this is negligible                         |
```
BenWibking: This seems less justifiable than the other thresholds. For photoionization, it might make sense for the absolute photon number density error to be comparable to the absolute electron number density error.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,113 @@
...
+  below the physically meaningful radiation level at the minimum temperature.
+- Cells with radiation below this threshold are considered "dark" and VODE
+  returns in a single BDF step.
+
+### radiation_failure_tolerance
+
+This is a **physical guard**, not a numerical tolerance. It defines the maximum allowed
+negative photon number density (cm⁻³) before a burn is declared failed. The default of
+0.01 cm⁻³ means that up to 0.01 photons per cm³ can be "created from nothing" by VODE's
+Newton overshoot — negligible in any astrophysical context. This parameter does not scale
```
BenWibking: I think this is not negligible if we are considering photoionization in the CGM. Then this is comparable to the electron number density in the CGM.

chongchonghe: addressed

docs/markdown/photoionization.md
chongchonghe: note that we set the failure tolerance (= the electron number density error) to 10⁻². 

Here is what I found from testing: if we set atol_rad_num to the same value, it will cause VODE to choose a large time step, which results in large overshoot, bigger than the failure tolerance, and the solver will fail. 

So, I believe the current choice is a better solution.

BenWibking:
What error does VODE give when this happens? Is it a corrector failure, or too many substeps failure? This matters because it may have to be tuned again if we switch to Rosenbrock, which is fine.

chongchonghe:
@BenWibking it causes the photon number density to drop below `-radiation_failure_tolerance`. When this happens, VODE claims the solve is a failure. James wrote this logic, but it's probably a standard implementation of the VODE solver (I haven't checked).

BenWibking:
> @BenWibking it causes the photon number density to drop below `-radiation_failure_tolerance`. When this happens, VODE claims the solve is a failure. James wrote this logic, but it's probably a standard implementation of the VODE solver (I haven't checked).

I don't remember this being a feature in the existing solver. I need to look at this more carefully.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,292 @@
...
+term (stellar emission $\dot{N}$). Isotropic sources add photons uniformly in all
+directions — they contribute to $N_\gamma$ but produce no net flux. Flux must be
+integrated to track the attenuation of the directional radiation field across the
+timestep.
+
+## 3. VODE Tolerances
+
+### 3.1 Overview
+
+Quokka uses CVODE/VODE (via Microphysics) to integrate the chemistry and internal energy
```
BenWibking: CVODE is not used. It is a different code. Only the Microphysics version of VODE is used.

James471:
> @BenWibking it causes the photon number density to drop below `-radiation_failure_tolerance`. When this happens, VODE claims the solve is a failure. James wrote this logic, but it's probably a standard implementation of the VODE solver (I haven't checked).

This is the radiation equivalent of the `species_failure_tolerance` parameter.

docs/markdown/photoionization.md
chongchonghe: Okay, another AI hallucination.

docs/markdown/parameters.md
```diff
@@ -113,6 +113,57 @@ These parameters are read in the `QuokkaSimulation<problem_t>::readParmParse()`
...
+| `integrator.species_failure_tolerance` | Float | `0.01` | Maximum violation of `0 ≤ X ≤ 1` before the burn is declared failed. |
+
+### Other integrator parameters
+
+| Parameter Name | Type | Default | Description |
+|---|---|---|---|
+| `integrator.jacobian` | Integer | `1` | Jacobian type: `1` = analytical, `2` = numerical. |
+| `integrator.ode_max_steps` | Integer | `150000` | Maximum number of VODE internal steps per burn call. |
+| `integrator.ode_max_dt` | Float | `1.e30` | Maximum internal timestep for VODE. |
+| `integrator.use_number_densities` | Boolean (0/1) | `0` | If 1, evolve species as number densities instead of mass fractions. |
```
BenWibking: This must be set to `1` for it to work correctly for Quokka.

docs/markdown/parameters.md
```diff
@@ -113,6 +113,57 @@ These parameters are read in the `QuokkaSimulation<problem_t>::readParmParse()`
...
+| `integrator.subtract_internal_energy` | Boolean (0/1) | `1` | If 1, subtract internal energy before integration. Must be `0` for Quokka. |
+| `integrator.renormalize_abundances` | Boolean (0/1) | `0` | If 1, renormalize mass fractions to sum to unity at each step. |
+| `integrator.do_species_clip` | Boolean (0/1) | `0` | If 1, clip species to [0,1] during cleanup. |
```
BenWibking: This does not work with number densities. We should probably remove this from the docs.

docs/markdown/parameters.md
```diff
@@ -113,6 +113,57 @@ These parameters are read in the `QuokkaSimulation<problem_t>::readParmParse()`
...
+### Other integrator parameters
+
+| Parameter Name | Type | Default | Description |
+|---|---|---|---|
+| `integrator.jacobian` | Integer | `1` | Jacobian type: `1` = analytical, `2` = numerical. |
+| `integrator.ode_max_steps` | Integer | `150000` | Maximum number of VODE internal steps per burn call. |
+| `integrator.ode_max_dt` | Float | `1.e30` | Maximum internal timestep for VODE. |
+| `integrator.use_number_densities` | Boolean (0/1) | `0` | If 1, evolve species as number densities instead of mass fractions. |
+| `integrator.subtract_internal_energy` | Boolean (0/1) | `1` | If 1, subtract internal energy before integration. Must be `0` for Quokka. |
+| `integrator.renormalize_abundances` | Boolean (0/1) | `0` | If 1, renormalize mass fractions to sum to unity at each step. |
```
BenWibking: This causes VODE to fail in Piyush and I's earlier tests. I think we should remove it from the docs.

James471:
https://github.com/AMReX-Astro/Microphysics/blob/e3e47125bfc18a4f8fc8aae4c03c7611c542a372/integration/VODE/vode_dvjac.H#L118
This is a division by zero after setting ewt(flux)=0 in dvode. However, this code block is not being used anywhere in the existing tests, since we don't use an analytical jacobian.

https://github.com/AMReX-Astro/Microphysics/blob/e3e47125bfc18a4f8fc8aae4c03c7611c542a372/integration/VODE/vode_dvnlsd.H#L138
https://github.com/AMReX-Astro/Microphysics/blob/e3e47125bfc18a4f8fc8aae4c03c7611c542a372/integration/VODE/vode_dvnlsd.H#L242
These two may not be necessary since ewt(flux) is already set to 0 in dvode.

https://github.com/AMReX-Astro/Microphysics/blob/e3e47125bfc18a4f8fc8aae4c03c7611c542a372/integration/VODE/vode_dvhin.H#L57
Instead of setting the flux atol to 1e30 here, you can also do it in https://github.com/AMReX-Astro/Microphysics/blob/e3e47125bfc18a4f8fc8aae4c03c7611c542a372/integration/integrator_setup_strang.H#L32 or inject it in extern_parameters()

If we make these suggestions, the changes made to the VODE integrator will be reduced, since we don't really need to touch dvnlsd and dvhin.

docs/markdown/parameters.md
```diff
@@ -113,6 +113,57 @@ These parameters are read in the `QuokkaSimulation<problem_t>::readParmParse()`
...
+| `integrator.do_species_clip` | Boolean (0/1) | `0` | If 1, clip species to [0,1] during cleanup. |
+| `integrator.call_eos_in_rhs` | Boolean (0/1) | `1` | If 1, call EOS in the RHS to update temperature from internal energy. |
+| `integrator.integrate_energy` | Boolean (0/1) | `1` | If 1, enable energy integration; if 0, freeze energy. |
+| `integrator.scale_system` | Boolean (0/1) | `0` | If 1, scale the ODE system to be O(1). |
+| `integrator.use_burn_retry` | Boolean (0/1) | `0` | If 1, retry failed burns with swapped Jacobian or relaxed tolerances. |
+| `integrator.retry_swap_jacobian` | Boolean (0/1) | `1` | If 1, swap Jacobian type (analytic ↔ numerical) on retry. |
+| `integrator.burner_verbose` | Boolean (0/1) | `0` | If 1, print diagnostic output after each burn. |
+| `integrator.SMALL_X_SAFE` | Float | `1.e-30` | Species floor to prevent underflow in the integrator. |
+| `integrator.X_reject_buffer` | Float | `1.0` | Buffer factor for the species rejection threshold. |
```
James471: Piyush has always set it to 1e100. I have done the same.

docs/markdown/parameters.md
```diff
@@ -113,6 +113,57 @@ These parameters are read in the `QuokkaSimulation<problem_t>::readParmParse()`
...
+
+| Parameter Name | Type | Default | Description |
+|---|---|---|---|
+| `integrator.atol_spec` | Float | `1.e-10` | Absolute tolerance for species number densities (cm⁻³). |
+| `integrator.rtol_spec` | Float | `1.e-10` | Relative tolerance for species number densities. |
+| `integrator.atol_enuc` | Float | `1.e-25` | Absolute tolerance for internal energy (erg g⁻¹). |
+| `integrator.rtol_enuc` | Float | `1.e-10` | Relative tolerance for internal energy. |
+| `integrator.atol_rad_num` | Float | `1.e-10` | Absolute tolerance for radiation number density (cm⁻³). |
+| `integrator.rtol_rad_num` | Float | `1.e-10` | Relative tolerance for radiation number density. |
+| `integrator.radiation_failure_tolerance` | Float | `0.05` | Maximum allowed negative photon number density (cm⁻³) before the burn is declared failed. Physical guard, not a numerical tolerance. |
```
James471: This is equivalent to species_failure_tolerance, but for radiation number density.

docs/markdown/parameters.md
```diff
@@ -113,6 +113,57 @@ These parameters are read in the `QuokkaSimulation<problem_t>::readParmParse()`
...
+| `integrator.species_failure_tolerance` | Float | `0.01` | Maximum violation of `0 ≤ X ≤ 1` before the burn is declared failed. |
```
James471: We don't run quokka using mass fractions, so we don't actually make use of the 0<=X<=1 condition. This parameter is the maximum allowed negative number density.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,292 @@
...
+The source term $n_e n_{\rm H^+}(\alpha_A - \alpha_B)$ represents diffuse recombination
+radiation — photons re-emitted when H recombines directly to the ground state (case A
+minus case B correction).
+
+### 1.2 Hydrogen Thermochemistry
+
+The neutral hydrogen fraction evolves as:
+
+$$
+\frac{D n_{\rm H^0}}{Dt} = \alpha_A n_e n_{\rm H^+} - \beta n_e n_{\rm H^0} - \Gamma_{\gamma {\rm H}^0} n_{\rm H^0},
```
James471, line 46: beta is not defined in the table above

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,292 @@
...
+$$
+
+where $\hat{c}$ is the reduced speed of light. Only one directional flux component is
+integrated (normalized to 1.0 before the burn); the other two are scaled proportionally
+after the ODE solve. The state vector has 6 components for a single chemical band:
+$(n_e, n_{\rm H^0}, n_{\rm H^+}, e, N_\gamma, F_\gamma)$.
+
+Note that $n_{\rm H^+} = n_{\rm H} - n_{\rm H^0}$ and $n_e = n_{\rm H^+}$ by
+construction. Although only one of $n_{\rm H^0}$ or $n_{\rm H^+}$ is an independent
+variable, both are integrated for symmetry.
```
James471, line 135: We are also forced to carry them for doing the eos correctly

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,292 @@
...
+
+Under OTSA, VODE integrates the following system over the implicit timestep $\Delta t$:
+
+$$
+\begin{aligned}
+\frac{d N_\gamma}{d t} &= - n_{\rm H^0} \hat{c} \sigma_\gamma N_\gamma + \dot{N}, \\[2pt]
+\frac{d F_{\gamma,i}}{d t} &= - n_{\rm H^0} \hat{c} \sigma_\gamma F_{\gamma,i} \quad (i = x, y, z), \\[2pt]
+\frac{d n_{\rm H^0}}{d t} &= \alpha_B n_e n_{\rm H^+} - \beta n_e n_{\rm H^0} - \hat{c} \sigma_\gamma N_\gamma n_{\rm H^0}, \\[2pt]
+\frac{d n_{\rm H^+}}{d t} &= -\alpha_B n_e n_{\rm H^+} + \beta n_e n_{\rm H^0} + \hat{c} \sigma_\gamma N_\gamma n_{\rm H^0}, \\[2pt]
+\frac{d e}{d t} &= n_{\rm H^0} \hat{c} \sigma_\gamma \epsilon_\gamma N_\gamma - \text{[cooling terms]}.
```
James471, line 124: Electron number density equation

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,292 @@
...
+
+$$
+\rho \frac{D}{Dt}\!\left(\frac{e}{\rho}\right) = \mathcal{H}_{\rm photo} - \mathcal{L},
+$$
+
+where $\mathcal{H}_{\rm photo} = n_{\rm H^0} c \sigma_\gamma \epsilon_\gamma N_\gamma$ is
+the photoheating rate and $\epsilon_\gamma = h(\bar{\nu} - \nu_{{\rm H}^0})$ is the mean
+excess photon energy above the ionization threshold (29.65 eV for a $10^5$ K blackbody).
+The cooling rate $\mathcal{L}$ includes case A recombination cooling, collisional
+excitation of H, collisional ionization cooling, and Bremsstrahlung. Rate coefficients
```
James471: We use KI heating and cooling function. But this will anyways change when we implement a JAFF network.

BenWibking (changes requested):
`inputs/StromgrenRSLA.toml` and `inputs/StromgrenTempIndependentRecombination.toml` are missing tolerances.

I am still a bit bothered by how the radiation energy atol is set in this PR. I don't think it's a good idea to work around the integrator failure by choosing a different atol. It really should be the value below which we don't care about the answer, not a tuned value to make the integrator take a reasonable timestep. I think we need to solve the underlying timestepping issue in VODE, or else just give up on VODE and use Rosenbrock, rather than tune the radiation energy atol to a value that doesn't really make physical sense.

chongchonghe's reply: I have deferred this feature to #1980 , so we'll continue this discussion there.

src/networks/photoionization/actual_rhs.H
```diff
@@ -174,7 +175,8 @@ template <class MatrixType> AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void actual
 	const amrex::Real n_e = state.xn[0];
 	const amrex::Real n_HI = state.xn[1];
 	const amrex::Real n_HII = state.xn[2];
```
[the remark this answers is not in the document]
BenWibking: I think it's right about this. The Jacobian needs to be consistent with the clamping done in the rhs.

src/networks/photoionization/actual_rhs.H
```diff
@@ -174,7 +175,8 @@ template <class MatrixType> AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void actual
...
-	const amrex::Real n_gamma = state.rn[0];
+	const amrex::Real n_gamma =
+	    amrex::max(state.rn[0], 0.0_rt); // clamp at zero: prevents sign reversal if VODE steps rn[0] slightly negative within tolerance
```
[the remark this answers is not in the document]
BenWibking: I think it's right about this.

docs/markdown/parameters.md
BenWibking: Ok, I see. I was confused because Piyush never set this for the primordial_chem network. In Rosenbrock, this is ignored and instead the integrator rejects steps if it any species is  less than `-atol_spec`.

The description here is not correct though. It is a numerical tolerance when VODE is used. VODE will reject an internal substep if a species is more negative than this, and will reduce the timestep by a factor of 4 and retry the internal substep.

docs/markdown/parameters.md
BenWibking: Ok, I think this should actually be set equal to atol_spec. This is what Rosenbrock does internally and it works well.

docs/markdown/parameters.md
BenWibking: This parameter does not make sense for number densities. It only make sense for mass fractions. So setting it to 1e100 makes it do nothing.

BenWibking:
@chongchonghe Can you update this to the latest development branch of Microphysics? My PR to adjust the failure threshold post-update has been merged, and I think it will avoid the need to set atol in the fine-tuned way you do here.

docs/markdown/parameters.md
chongchonghe: OK

docs/markdown/parameters.md
chongchonghe: OK.

docs/markdown/parameters.md
chongchonghe: Since we don't use mass fraction, this parameter does not apply (there is no limit on species number density). I'll remove this from doc.

docs/markdown/photoionization.md
chongchonghe: Okay, then I think that's a code architecture limitation, not a requirement.

docs/markdown/photoionization.md
chongchonghe: done

docs/markdown/photoionization.md
chongchonghe: done

docs/markdown/photoionization.md
chongchonghe: done

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,299 @@
...
+the photoheating rate and $\epsilon_\gamma = h(\bar{\nu} - \nu_{{\rm H}^0})$ is the mean
+excess photon energy above the ionization threshold (29.65 eV for a $10^5$ K blackbody).
+For heating and cooling that are not directly due to hydrogen photoionization, we
+follow the optically thin prescription of Krumholz (2007). In molecular gas, the
+approximate cooling and heating functions of Koyama & Inutsuka (2002) are used.
+In partially ionized gas, the cooling rate is computed following Osterbrock (1989),
+which includes cooling by ion-electron collisions involving the first and second
+ionized states of O, N, and Ne — the dominant coolants in H II regions at solar
+metallicity. A future PR will extend the cooling model to use the RIGEL prescription
+(Deng et al. 2024).
```
BenWibking: I think it would be good to explicitly state that when photoionization is turned on, the standalone ResampledCooling cooling module should not be used. This incompatibility should also be added to the compatibility matrix in https://github.com/quokka-astro/quokka/blob/development/docs/markdown/components.md#physics-module-compatibility-matrix

BenWibking:
This is what Codex says:
```
  - [P2] Gate photon-number Jacobian entries on the clamp — /Users/benwibking/amrex_codes/quokka/src/networks/photoionization/
    actual_rhs.H:178-179
    When state.rn[0] is negative, this clamp makes the RHS independent of photon number, but the analytic Jacobian below still
    reports nonzero derivatives with respect to rn[0] (jac(1,5), jac(net_ienuc,5), jac(5,5), etc.). Since the updated inputs use
    integrator.jacobian = 1 and explicitly allow small negative photon densities, VODE will use an inconsistent Newton matrix in the
    exact states this change is meant to tolerate; those derivatives should be zeroed whenever the unclamped photon number is below
    the clamp.

  - [P2] Remove or implement the documented tolerance API — /Users/benwibking/amrex_codes/quokka/docs/markdown/parameters.md:122-122
    This documents integrator.typical_* as the recommended configuration path, but the current code does not read these keys and
    SetAtolFromPhysics()/src/radiation/photochem_atol.H do not exist in the tree. Users following this new section will have those
    parameters ignored rather than deriving atol_* or aborting, so the documentation should either match the implemented raw
    tolerance parameters or land with the code that implements this API.
```

docs/markdown/photoionization.md
chongchonghe: good point. done.

src/networks/photoionization/actual_rhs.H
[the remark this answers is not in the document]
chongchonghe: looks good. done.

chongchonghe:
> This is what Codex says:
> 
> ```
>   - [P2] Gate photon-number Jacobian entries on the clamp — /Users/benwibking/amrex_codes/quokka/src/networks/photoionization/
>     actual_rhs.H:178-179
>     When state.rn[0] is negative, this clamp makes the RHS independent of photon number, but the analytic Jacobian below still
>     reports nonzero derivatives with respect to rn[0] (jac(1,5), jac(net_ienuc,5), jac(5,5), etc.). Since the updated inputs use
>     integrator.jacobian = 1 and explicitly allow small negative photon densities, VODE will use an inconsistent Newton matrix in the
>     exact states this change is meant to tolerate; those derivatives should be zeroed whenever the unclamped photon number is below
>     the clamp.

Looks good. Addressed in [this commit](https://github.com/quokka-astro/quokka/pull/1956/commits/9cd4b9f546d2880f2d883825c7f230d353d1f879) 

>   - [P2] Remove or implement the documented tolerance API — /Users/benwibking/amrex_codes/quokka/docs/markdown/parameters.md:122-122
>     This documents integrator.typical_* as the recommended configuration path, but the current code does not read these keys and
>     SetAtolFromPhysics()/src/radiation/photochem_atol.H do not exist in the tree. Users following this new section will have those
>     parameters ignored rather than deriving atol_* or aborting, so the documentation should either match the implemented raw
>     tolerance parameters or land with the code that implements this API.
> ```

This feature is deferred to another PR #1980 . I also updated the documentation to reflect this.

chongchonghe:
@BenWibking I addressed all comments. Ready for review.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,275 @@
...
+| `a_rad` | `7.5657e-15 erg cm^-3 K^-4`             | Radiation constant (from `fundamental_constants.H`)              |
+| `k_B`   | `1.380649e-16 erg K^-1`                 | Boltzmann constant                                               |
+| `m_p`   | `1.67262192e-24 g`                      | Proton mass                                                      |
+| `c_v`   | `3/2 x k_B / m_p ~ 1.24e8 erg g^-1 K^-1` | Specific heat of monatomic hydrogen gas                           |
+
+### 3.5 radiation_failure_tolerance
+
+This is a **physical guard**, not a numerical tolerance. It defines the maximum allowed
+negative photon number density (cm^-3) before a burn is declared failed — at most this
+many spurious photons can be "created from nothing" by VODE's Newton overshoot.
```
BenWibking: What is the "newton overshoot" you are referring to here?

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,275 @@
...
+The default of 0.05 cm^-3 is appropriate for galactic disk or GMC environments, where the
+typical ionized gas density is ~10^2–10^4 cm^-3 (ratio <= 5x10^-4). For low-density environments
+such as the CGM or IGM, where the ionized gas density can be ~10^-4–10^-3 cm^-3, this
+default competes with the physical ionization equilibrium — override it in the input file.
+
+**Rule of thumb:** set `radiation_failure_tolerance` to at least two orders of magnitude
+below the **typical density of ionized gas** in the problem. This ensures that spurious
+photon creation cannot measurably affect the ionization fraction. The value does not
+scale with `atol_rad_num` because the Newton overshoot in the stiff radiation-chemistry
+system has a floor independent of the tolerance.
```
BenWibking: I don't understand this sentence.

I would that that `radiation_failure_tolerance` should scale with `atol_rad_num`. This approach seems to work for the species failure tolerance.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,275 @@
...
+\frac{\partial N_\gamma}{\partial t} + \nabla \cdot \mathbf{F}_\gamma =
+- n_{\rm H^0} c \sigma_\gamma N_\gamma
++ n_e n_{\rm H^+} (\alpha_A - \alpha_B)
++ \dot{N}^*_\gamma,
+$$
+
+$$
+\frac{\partial \mathbf{F}_\gamma}{\partial t} + c^2 \nabla \cdot \mathsf{P}_\gamma =
+- n_{\rm H^0} c \sigma_\gamma \mathbf{F}_\gamma.
+$$
```
BenWibking, line 23: Are $\mathcal{O}(v/c)$ terms included in the equations solved for photoionization?

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,275 @@
...
+
+$$
+\rho \frac{D}{Dt}\!\left(\frac{e}{\rho}\right) = \mathcal{H}_{\rm photo} - \mathcal{L},
+$$
+
+where $\mathcal{H}_{\rm photo} = n_{\rm H^0} c \sigma_\gamma \epsilon_\gamma N_\gamma$ is
+the photoheating rate and $\epsilon_\gamma = h(\bar{\nu} - \nu_{{\rm H}^0})$ is the mean
+excess photon energy above the ionization threshold (29.65 eV for a $10^5$ K blackbody).
+For heating and cooling that are not directly due to hydrogen photoionization, we
+follow the optically thin prescription of Krumholz (2007). In molecular gas, the
```
BenWibking: If this is a description of the Krumholz (2007) prescription, this period should be replaced by a colon, i.e.: "we follow the optically thin prescription of Krumholz (2007): in molecular gas, the..."

docs/markdown/photoionization.md
BenWibking, line 1: Can you change the citation format to use `mdbook-bib`: https://francisco-perez-sorrosal.github.io/mdbook-bib/config.html?

inputs/StromgrenNoRecombination.toml
```diff
@@ -34,15 +34,15 @@ integrator.renormalize_abundances = 0
...
-integrator.atol_rad_flux = 1e-14
-integrator.atol_enuc = 1e-14
-
-integrator.rtol_spec = 1e-14
-integrator.rtol_rad_num = 1e-14
-integrator.rtol_rad_flux = 1e-14
-integrator.rtol_enuc = 1e-14
+integrator.atol_spec                     = 1.0e-2   # cm^-3; spec_abundance_tol (1e-5) × n_H (1000)
+integrator.atol_enuc                     = 1.24e8   # erg/g; c_v × 1 K temperature accuracy
+integrator.atol_rad_num                  = 1.25e-6  # cm^-3; 1e-6 × a_rad × T_min_rad^4 / E_photon
```
BenWibking: This value seems too low to me. Why is it not of the order of `intergrator.atol_spec`?

src/problems/DTypeFront/testDTypeFront.cpp
```diff
@@ -53,7 +53,18 @@ template <> struct Physics_Traits<DTypeFront> : DefaultPhysicsTraits {
 
 template <> struct RadSystem_Traits<DTypeFront> {
 	static constexpr double c_hat_over_c = c_hat / C::c_light;
-	static constexpr double Erad_floor = 1e-99;
+	// Erad_floor sets the M1 radiation energy density floor (erg cm^-3), defined here as a
+	// blackbody at T=0.01 K.  The corresponding photon number density floor is
+	//   N_gamma_floor = Erad_floor / E_photon ~ 1.25e-10 cm^-3.
+	//
+	// SetAtolFromPhysics() derives atol_rad_num = 1e-6 * a_rad * T_min^4 / E_photon, where
```
BenWibking, line 60: Was this comment was copied from your other PR? I don't think it makes sense here.

docs/markdown/photoionization.md
BenWibking: Is there a reason not to treat `radiation_failure_tolerance` the same way as `species_failure_tolerance` and set it automatically to 1.5 times `atol_rad_num`?

BenWibking (changes requested):

docs/markdown/photoionization.md
chongchonghe: No, but you've raised a good and important point. 

The radiation pressure from FUV photons is often comparable to thermal pressure in ionized gas, so I think it's important to include it. I would choose a simple approach: do a single Forward Euler update to include the work term AFTER photochemistry solve. I'll do it in a separate PR.

inputs/StromgrenNoRecombination.toml
chongchonghe: In DTypeFront, `integrator.atol_rad_num = 5e-2`, which is the same order of `intergrator.atol_spec` . I will also try this value in this StromgrenNoRecombination test.

src/problems/DTypeFront/testDTypeFront.cpp
chongchonghe: I forgot to move this to the other PR, #1980.

docs/markdown/photoionization.md
chongchonghe: ok

chongchonghe:
@BenWibking Thanks for the detailed review! I've addressed your comments:

**Microphysics (new branch `chong/claude/radiation-failure-tolerance-factor` on fork):**
- Added `vode_final_state_radiation_failure_tolerance_factor = 1.5` in `vode_type.H`, mirroring the species_failure_tolerance design — internal VODE step checks use `radiation_failure_tolerance` directly; the final interpolated state check uses 1.5x to account for non-monotonic interpolation.
- Updated `actual_integrator.H`, `actual_integrator_sdc.H`, and both `integrator_cleanup` functions.

**`parameters.md`:**
- Replaced `typical_*` sections with a single tolerance parameter table (the `SetAtolFromPhysics` machinery has moved to #1980).
- Updated `radiation_failure_tolerance` description: it's a numerical tolerance at internal nodes, should equal `atol_rad_num`; final state relaxed to 1.5x.
- Added notes that `integrate_energy`, `scale_system`, and `X_reject_buffer` don't work with number densities.
- Noted `subtract_internal_energy` and `use_number_densities` must be set correctly for Quokka.

**`photoionization.md`:**
- Added citations `[@Krumholz2007]`, `[@Koyama2002]`, `[@Osterbrock1989]`.
- Removed all `SetAtolFromPhysics` / `typical_*` content (moved to #1980).
- Clarified `radiation_failure_tolerance` and `species_failure_tolerance` design.

Could you take another look when you have a moment?

Generated by Claude Code

docs/markdown/photoionization.md
BenWibking: I think that's fine, but I think it should be a *backward* Euler update. That is required for it to be asymptotic-preserving. For case B it is never in the asymptotic diffusion limit, but in general, it could be.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,275 @@
...
+
+$$
+\rho \frac{D}{Dt}\!\left(\frac{e}{\rho}\right) = \mathcal{H}_{\rm photo} - \mathcal{L},
+$$
+
+where $\mathcal{H}_{\rm photo} = n_{\rm H^0} c \sigma_\gamma \epsilon_\gamma N_\gamma$ is
+the photoheating rate and $\epsilon_\gamma = h(\bar{\nu} - \nu_{{\rm H}^0})$ is the mean
+excess photon energy above the ionization threshold (29.65 eV for a $10^5$ K blackbody).
+For heating and cooling that are not directly due to hydrogen photoionization, we
+follow the optically thin prescription of Krumholz (2007) [@Krumholz2007]. In molecular
```
BenWibking: I think without a colon here, it's ambiguous as to whether you are explaining the Krumholz prescription or describing cooling and heating terms that you added in addition to the Krumholz prescription.

So if it's the former, can you change this to a colon, e.g.: "we
follow the optically thin prescription of Krumholz (2007) [@Krumholz2007]: in molecular gas, ...".

If it's the latter, can you add an introductory clause that clarifies that these are additional terms, e.g.: "Additionally, in molecular gas, the approximate cooling and heating...".

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,275 @@
...
+$\texttt{atol\_rad\_num} \gg N_{\gamma,{\rm floor}}$. A ratio of $\geq 10^4$ is sufficient:
+
+$$
+\frac{\texttt{atol\_rad\_num}}{N_{\gamma,{\rm floor}}} \geq 10^4.
+$$
+
+For typical `Erad_floor` values corresponding to $T_{\rm floor} = 0.01$–$1$ K, an
+`atol_rad_num` on the order of $10^{-6}$–$10^{-2}$ cm^-3 satisfies this constraint.
+
+### 3.7 species_failure_tolerance
```
BenWibking, line 246: I think it's important to clarify that the primary use of this variable by VODE is to determine whether or not to reject an internal substep if a value becomes more negative than this.

The secondary use is to determine whether or not to accept the final state (after interpolating to the final time) based on whether it becomes more negative than 1.5 times this tolerance.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,275 @@
...
+
+| Parameter                               | Description                                                       |
+| --------------------------------------- | ----------------------------------------------------------------- |
+| `integrator.atol_spec`                  | Absolute tolerance for chemical species (cm^-3)                   |
+| `integrator.atol_enuc`                  | Absolute tolerance for gas internal energy (erg g^-1)             |
+| `integrator.atol_rad_num`               | Absolute tolerance for photon number density (cm^-3)              |
+| `integrator.rtol_spec`                  | Relative tolerance for chemical species                           |
+| `integrator.rtol_enuc`                  | Relative tolerance for gas internal energy                        |
+| `integrator.rtol_rad_num`               | Relative tolerance for photon number density                      |
+| `integrator.species_failure_tolerance`  | Maximum allowed negative species number density (cm^-3)           |
```
BenWibking: I think "maximum allowed" is somewhat confusing. This is used in two different ways: see my comment below.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,275 @@
...
+| `integrator.radiation_failure_tolerance`| Maximum allowed negative photon number density before burn failure (cm^-3, see SS 3.5) |
```
BenWibking: I think this is somewhat confusing way to describe this parameter. See my comment below.

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,275 @@
...
+| Symbol  | Value                                   | Description                                                      |
+| ------- | --------------------------------------- | ---------------------------------------------------------------- |
+| `a_rad` | `7.5657e-15 erg cm^-3 K^-4`             | Radiation constant (from `fundamental_constants.H`)              |
+| `k_B`   | `1.380649e-16 erg K^-1`                 | Boltzmann constant                                               |
+| `m_p`   | `1.67262192e-24 g`                      | Proton mass                                                      |
+| `c_v`   | `3/2 x k_B / m_p ~ 1.24e8 erg g^-1 K^-1` | Specific heat of monatomic hydrogen gas                           |
+
+### 3.5 radiation_failure_tolerance
+
+This is a **physical guard**, not a numerical tolerance. It defines the maximum allowed
```
BenWibking: I think this statement is wrong. See my comments below.

BenWibking (changes requested):

docs/markdown/photoionization.md
BenWibking: I think this is still confusing. See my new comment about this below.

docs/markdown/photoionization.md
BenWibking: @chongchonghe Can you clarify this?

docs/markdown/photoionization.md
chongchonghe: OK, done.

docs/markdown/photoionization.md
chongchonghe: done. please take a look again.

docs/markdown/photoionization.md
chongchonghe: done

docs/markdown/photoionization.md
chongchonghe: done

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,285 @@
...
+
+$$
+\rho \frac{D}{Dt}\!\left(\frac{e}{\rho}\right) = \mathcal{H}_{\rm photo} - \mathcal{L},
+$$
+
+where $\mathcal{H}_{\rm photo} = n_{\rm H^0} c \sigma_\gamma \epsilon_\gamma N_\gamma$ is
+the photoheating rate and $\epsilon_\gamma = h(\bar{\nu} - \nu_{{\rm H}^0})$ is the mean
+excess photon energy above the ionization threshold (29.65 eV for a $10^5$ K blackbody).
+For heating and cooling that are not directly due to hydrogen photoionization, we
+follow the optically thin prescription of [@Krumholz2007]: In molecular
```
BenWibking: `prescription of [@Krumholz2007]: In molecular` -> `prescription of [@Krumholz2007]: in molecular`

docs/markdown/photoionization.md
```diff
@@ -0,0 +1,285 @@
...
+gas, the approximate cooling and heating functions of [@Koyama2002] are used. 
+In partially ionized gas, the cooling rate is computed
```
BenWibking: `In partially ionized gas,` -> `, while in partially ionized gas,`

BenWibking (changes requested):
I think your description of the cooling is still grammatically confusing. I have suggested two rewordings below.

docs/markdown/photoionization.md
BenWibking: This combines the two sentences in a way that indicates that both the molecular and partially ionized gas cooling terms are part of the Krumholz (2007) prescription.

docs/markdown/photoionization.md
BenWibking: No capitalization needed for `in`.

chongchonghe:
OK, comments addressed. Ready for review again. @BenWibking

BenWibking (dismissed):
LGTM.

docs/markdown/photoionization.md
chongchonghe: No, we don't do backward Euler on Flux. It's unnecessary. For the RHD module, we do a Picard iteration on the work term. We can do something similar for FUV, but I'll test and see if it is necessary.

docs/markdown/photoionization.md
chongchonghe: A negative solution on a positive-definite variable is a result of a overshoot, likely from the Newton-Raphson step in VODE.

BenWibking (approved):

PR #1961, opened 2026-06-17
fix: ensure sink particles stay on finest AMR level in all cases

### LINKED ISSUE #1957: sink particles at coarse levels cause problems
**Describe the bug**

The bug: A run with sink particles and `amr.max_level > 0` and `setForceFinestLevel(true)` may still have some sink particles not on the finest level. This is not expected to happen, and it also causes non-conservation of mass.

To enforce sink particles at finest level and fix this bug, we should enforce `setForceFinestLevel(true)` on sink particles and also set `amr.n_error_buf = 4` and `amr.grid_eff = 0` .

chongchonghe:
### Description

Fixes #1957: sink particles drop off the finest AMR level under subcycling (`do_subcycle=1`) when `max_level >= 2`, causing mass non-conservation. The fix makes `do_subcycle=1` work correctly for sink particles — the old `do_subcycle=0` restriction is lifted.

**Root cause:** A subcycled level-L regrid holds level L fixed and can only rebuild level L+1. If a ForceFinestLevel particle sits outside level L's safe interior (e.g. near the patch boundary), no properly-nested level-(L+1) box can be placed around it. The particle never reaches the finest level.

**Fix (three changes):**

1. `timeStepWithSubcycling` forces a level-0 regrid at the start of each coarse step when ForceFinestLevel particles exist. This expands level-L patches to cover all particles before any advance, and before any subcycled regrid would lock the level structure.

2. `tagCellsAroundParticles(lev)` projects particles from ALL AMR levels onto level-lev (not just particles stored at level lev). After collecting projected cells on each rank, `amrex::AllGatherBoxes` broadcasts them to all MPI ranks so every rank can tag the cells it owns — fixing a silent tag loss when a particle's projected level-lev cell lives on a different rank than the particle itself.

3. `registerParticleType` for Sink particles calls `setForceFinestLevel(true)` so that `refineGridsAroundParticles` is active automatically — callers can no longer forget to set it.

**Why both (1) and (2) are necessary:** In AMReX's `MakeNewGrids`, `ErrorEst(lev)` always operates on the *current* `grids[lev]`, never on newly-built grids from the same regrid call (the loop iterates downward from finest to lbase). So during a level-0 regrid, `ErrorEst(1)` runs *before* `MakeNewGrids(1)` expands level 1. The force-regrid (1) expands level 1 in the first regrid pass; the cross-level tagging (2) ensures `ErrorEst(1)` can still tag cells for level-2 refinement even when the particle hasn't been redistributed to level 1 yet. Without (2), the particle becomes invisible to `ErrorEst(L)` once it moves up a level, causing the refinement hierarchy to collapse behind it — we verified this fails on a test branch that used current-level-only tagging.

**Test:** New `ParticleSinkSubcycle` regression test reproduces the exact bug geometry: a sink particle outside the initial level-1 patch, with `max_level=2`, `do_subcycle=1`, `blocking_factor=8`. Two ctest variants verify both `do_subcycle=0` and `do_subcycle=1` reach the finest level and stay there across multiple coarse steps.

### Related issues
Closes #1957

### Checklist
- [x] I have added a link to any related issues (see above).


src/particles/PhysicsParticles.hpp
```diff
@@ -1170,6 +1171,17 @@ template <typename problem_t> class PhysicsParticleRegister
...
 	}
 
+	// Returns true if any registered particle type requires all particles to reside on the finest level
+	[[nodiscard]] auto anyParticleRequiresFinestLevel() const -> bool
+	{
+		for (const auto &[type, descriptor] : particleRegistry_) {
+			if (descriptor->getForceFinestLevel()) {
+				return true;
+			}
+		}
```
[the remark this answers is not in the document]
BenWibking: I don't think this is necessary.

BenWibking:
This is what Codex says:
```
  Review comment:

  - [P2] Validate ForceFinestLevel after user configuration — /Users/benwibking/amrex_codes/quokka/src/simulation.hpp:3673-3673
    This check runs inside InitPhyParticles(), which is called by sim.setInitialConditions(), but several existing problems enable
    ForceFinestLevel only afterward (for example ParticleCreation sets it on Test particles after setInitialConditions(), and
    BinaryOrbitCIC does the same for CIC particles). In those cases, a run with amr.max_level >= 2 and default do_subcycle = 1 still
    bypasses the new abort and can hit the mass non-conservation case this patch is trying to prevent; the validation needs to happen
    after these problem-level setters or in setForceFinestLevel()/before evolve().
```

chongchonghe:
@BenWibking Can you check if this fix looks right?

src/particles/PhysicsParticles.hpp
```diff
@@ -735,28 +735,67 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
+				}
+			}
+		}
+		// Tag collected cells with a separate MFIter (no ParIter active, no nesting)
+		if (!cross_level_cells.empty()) {
+			for (amrex::MFIter mfi(tags); mfi.isValid(); ++mfi) {
+				const amrex::Box &valid_box = mfi.validbox();
+				for (const auto &cell : cross_level_cells) {
+					if (valid_box.contains(cell)) {
+						tags[mfi](cell) = amrex::TagBox::SET;
```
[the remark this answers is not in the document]
BenWibking: @chongchonghe @markkrumholz I think Codex is right about this. I think a concrete case where this can happen is when `amr.max_level=2` and a particle lives on a level 0 grid. This particle should be tagged for refinement on both level 0 and level 1 in order to create nested refinement grids on levels 1 and 2, respectively.

However, if the level 1 FAB onto which the particle's position would be projected isn't on the same MPI rank as the level 0 FAB it currently lives on, then the tag is silently dropped and the cells that lie on top of this particle aren't fully refined to max_level as they should be, because the cells surrounding this particle are not tagged on every level (as they should be). When the refinement tags are not consistent across levels, AMReX may fail to propagate the refinement down the AMR hierarchy beyond the level at which the tags are (incorrectly) missing.

This can cause refinement to fail even when `do_subcycle=0`, and may be the cause of the issue in https://github.com/quokka-astro/quokka/discussions/1917. ORION didn't have to worry about this because all particles were replicated on all MPI ranks.

BenWibking (changes requested):
I think there's still a logic bug that will in general prevent the intended refinement when there are multiple MPI ranks, even when subcycling is off.

See my comment underneath the comment from Codex.

BenWibking:
For a small number of sink particles, this can be done by broadcasting the cell indices that need to be tagged to all MPI ranks with `amrex::AllGatherBoxes`:

```
  amrex::Vector<amrex::Box> cross_level_tag_cells;

  // for each valid particle at l != lev:
  amrex::IntVect cell(AMREX_D_DECL(ix, iy, iz));
  if (geom_lev.Domain().contains(cell)) {
      cross_level_tag_cells.emplace_back(cell, cell);
  }

  // Example:
  //
  // before
  // rank 0: bxs = {A, B}
  // rank 1: bxs = {C}
  //
  // after amrex::AllGatherBoxes(bxs)
  // rank 0: bxs = {A, B, C}
  // rank 1: bxs = {A, B, C}

  amrex::AllGatherBoxes(cross_level_tag_cells);

  for (amrex::MFIter mfi(tags); mfi.isValid(); ++mfi) {
      const amrex::Box &valid_box = mfi.validbox();
      for (const auto &cell_box : cross_level_tag_cells) {
          const amrex::IntVect cell = cell_box.smallEnd();
          if (valid_box.contains(cell)) {
              tags[mfi](cell) = amrex::TagBox::SET;
          }
      }
  }
```

Of course, this will not scale to large numbers of sink particles. If you ever need, say, millions of sinks, then this can be revisited. The communication required to make this arbitrarily scalable will be complicated and tricky to debug, so I would instead follow the all-to-all communication pattern shown above, which should work up to several thousand sink particles.

chongchonghe:
After projecting particles from all levels into a level and broadcasting them to all MPI ranks using `AllGatherBoxes`, the particles are always at the finest level, whether `do_subcycle` is 0 or 1. 

@BenWibking  The PR is ready for review.

src/particles/PhysicsParticles.hpp
```diff
@@ -735,28 +735,66 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
 				tag(ix, iy, iz) = amrex::TagBox::SET;
 			});
 		}
+
+		// Cross-level: project particles from ALL other levels to level lev.
+		// A ForceFinestLevel particle may live at a level != lev; project its
+		// physical position to level-lev coordinates so the cell gets tagged.
+		// amrex::AllGatherBoxes guarantees every rank sees every tag cell —
+		// a particle on rank A may project to a level-lev cell owned by rank B.
+		amrex::Gpu::synchronize();
```
BenWibking: Can you make this `amrex::Gpu::streamSynchronize()`? That should be slightly faster. This is only a performance issue, not a correctness issue.

src/particles/PhysicsParticles.hpp
```diff
@@ -735,28 +735,66 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
+				continue;
+			}
+			for (typename ContainerType::ParIterType pti(*container_, l); pti.isValid(); ++pti) {
+				const auto &soa = pti.GetArrayOfStructs();
+				const amrex::Long np = pti.numParticles();
+				for (amrex::Long idx = 0; idx < np; ++idx) {
+					if (soa()[idx].id() <= 0) { // NOLINT(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+						continue;
+					}
+					const auto &p = soa()[idx]; // NOLINT(cppcoreguidelines-pro-bounds-pointer-arithmetic)
```
BenWibking: This is a major bug. This is not GPU-safe and will cause crashes because this is a device pointer when GPU support is turned on. It will work on Setonix/Frontier due to hardware support for cross-memory access, but will fail on NVIDIA GPUs.

src/particles/PhysicsParticles.hpp
```diff
@@ -735,28 +735,66 @@ template <typename ContainerType, typename problem_t, ParticleType particleType>
...
+									       static_cast<int>(amrex::Math::floor((p.pos(1) - plo[1]) * dxi[1])),
+									       static_cast<int>(amrex::Math::floor((p.pos(2) - plo[2]) * dxi[2]))));
+					if (geom_lev.Domain().contains(cell)) {
+						cross_level_boxes.emplace_back(cell, cell);
+					}
+				}
+			}
+		}
+		if (!cross_level_boxes.empty()) {
+			amrex::AllGatherBoxes(cross_level_boxes);
```
BenWibking: This is a major bug. This cannot be skipped on MPI ranks with zero boxes. It will cause an MPI deadlock if you do this.

BenWibking (changes requested):
There are two major issues with the implementation. See inline comments.

src/particles/PhysicsParticles.hpp
BenWibking: @chongchonghe I think this is causing the crash in the test problem in the GPU tests.

chongchonghe: OK, I see. I am checking it now.

src/particles/PhysicsParticles.hpp
chongchonghe: Done — changed to `amrex::Gpu::streamSynchronize()` in commit 871e010f.

src/particles/PhysicsParticles.hpp
chongchonghe: Fixed in commit 871e010f: replaced `soa()[idx]` (device pointer dereference from host) with `amrex::Gpu::copy(deviceToHost, ...)` into a host-side `amrex::Vector`, following the pattern in `particle_IO.hpp`.

src/particles/PhysicsParticles.hpp
chongchonghe: Fixed in commit 871e010f: moved `amrex::AllGatherBoxes(cross_level_boxes)` outside the `if (!empty())` guard so it is called unconditionally on all MPI ranks.

src/simulation.hpp
```diff
@@ -2141,8 +2142,12 @@ template <typename problem_t> void AMRSimulation<problem_t>::particleMeshInterac
 	// Sink accretion, stage 2: update the particle states -- compute scale_down, apply to particle, apply to cells
 	particleRegister_.applySinkAccretion(state_new_cc_[lev], accretion_rate_at_level, state_fc_ptr, geom[lev], lev, time, dt);
 
-	// We allow particle formation at the finest level only to avoid duplicate particle creation from multiple levels at the same location.
-	particleRegister_.createParticlesFromState(state_new_cc_[lev], accretion_rate_at_level, lev, time, dt, state_fc_ptr, verbose);
+	// Only create particles when the AMR hierarchy has fully refined to max_level.
+	// Creating a ForceFinestLevel particle at a sub-max level violates the invariant
+	// that sink particles always reside on the finest level of the AMR hierarchy.
+	if (finest_level == max_level) {
```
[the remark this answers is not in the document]
BenWibking: This is obviously not a useful comment, so I'll resolve it.

BenWibking, line 2148: For sinks this is fine, but as written, I think this will prevent creating star particles as well, which I think is not what we want.

src/simulation.hpp
```diff
@@ -2178,12 +2188,16 @@ template <typename problem_t> void AMRSimulation<problem_t>::timeStepWithSubcycl
 		// regrid changes level "lev+1" so we don't regrid on max_level
 		// also make sure we don't regrid fine levels again if
 		// it was taken care of during a coarser regrid
-		if (lev < max_level && istep[lev] > last_regrid_step[lev]) {
-			if (istep[lev] % regrid_int == 0) {
+		bool force_finest_regrid = false;
+#if AMREX_SPACEDIM == 3
+		force_finest_regrid = (lev == 0 && lev < max_level && istep[lev] == 0 && particleRegister_.anyParticleRequiresFinestLevel());
```
BenWibking, line 2197: This should be documented, so users aren't confused by the fact that `regrid_interval` is ignored when sinks are turned on.

BenWibking (changes requested):

src/simulation.hpp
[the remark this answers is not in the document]
chongchonghe: this is the right physics. Truelove refinement condition should refine to max level before the first sink can form (will be added in another PR).

chongchonghe: This is probably fine; it makes no difference to the sink or star particles.

There are two situations to consider:

1. For static grids/SMR like your DiskGalaxy or my TallBox run, star formation is dictated by the condition `lev = finest_level`. This means stars only form at the **current** finest level.
2. In an AMR run, the same routine decides that particles can only form at the current finest level. Then, the question is: do we want to form stars when the current finest level has not reached the desired amr.max_level?
The answer is probably no. For all situations I can think of, we should set a refinement criteria so that the cells must refine to the maximum level before forming a star.

If we ever want to form stars at coarser levels, we should handle that in a separate PR by changing the pre-existing condition`lev = finest_level` .

BenWibking: > The answer is probably no. For all situations I can think of, we should set a refinement criteria so that the cells must refine to the maximum level before forming a star.

I think the answer is yes. One example I can immediately think of is in a cosmological simulation, where you may want to refine to a maximum physical resolution and form stars at that resolution. Due to expansion, this will require refining to higher levels at lower redshifts.

chongchonghe:
@BenWibking I've addressed both of your remaining comments:

1. **Star particle creation guard** — As discussed, `particleMeshInteraction()` already creates particles only at `finest_level` (line 2106), so the `finest_level == max_level` guard is correct for all particle types. With proper refinement criteria, cells refine to max_level before reaching star-forming conditions.

2. **regrid_interval documentation** — Added a comment in commit f61a8917 explaining that when ForceFinestLevel particles exist, a level-0 regrid is forced at every coarse step regardless of `regrid_interval`.

Ready for re-review.

BenWibking (dismissed):

chongchonghe:
@BenWibking Need your approval again after reverting CI tigger.

BenWibking (approved):

src/simulation.hpp
chongchonghe: I'll leave an issue to remind myself to do this later when we really want this.

chongchonghe: #1979

PR #1962, opened 2026-06-17
Modular stellar-evolution framework for Star particles

chongchonghe:
### Description
Add a modular, compile-time-selectable stellar-evolution framework for `Star` particles, with a stateless analytic toy model as the default, plus a validation test that checks radius and luminosity against closed-form solutions and asserts the numerical Bondi accretion rate.

### Motivation
Currently Quokka has no `Star` particle type with stellar-evolution physics. This PR adds the framework, a toy model, and a validating test problem so that more sophisticated stateful models (e.g., ORION model or tabulated stellar tracks) can be plugged in later without changing the dispatch machinery.

### What changed

**Framework (`src/particles/stellar_models.hpp`, `src/particles/starparticle_radiation.hpp`)**
- `ToyStellarModel` — stateless analytic laws: R ∝ M^0.4, L_star ∝ M^3.5, L_acc = GMṁ/R. Pure GPU device functions.
- `Model::evolve(Real *rdata, int n_groups, Real dt)` — takes the full particle real-data array so the model can read and modify any component (mass, radius, luminosity groups, velocity, etc.). The dispatcher simply passes `&p.rdata(0)` — the model owns all field access.

**Shared index header (`src/particles/star_particle_indices.H` — new)**
- Extracted `StarParticleDataIdx` enum and all index constants (`StarParticleMassIdx`, `StarParticleMdotIdx`, `StarParticleRadiusIdx`, `StarParticleLumIdx`, etc.) into a standalone header that both `stellar_models.hpp` and `particle_types.hpp` include. Avoids circular dependencies while letting models use named indices.

**Star particle type (`src/particles/particle_types.hpp`, `src/particles/PhysicsParticles.hpp`)**
- New `ParticleSwitch::Star` and `ParticleType::Star`.
- Star real-component layout: `mass, vx, vy, vz, birth_time, mdot, radius, lum` with `lum` expandable to `nGroups` slots. Component count is auto-composed from `nGroups` + the model's `nExtraReal`/`nExtraInt`.
- Star descriptor registered with `allows_accretion=true` and `mdotIndex_=StarParticleMdotIdx`.

**`DefaultParticleTraits` refactoring (`particle_types.hpp` + 15 test problems)**
- Introduced `DefaultParticleTraits` as a base struct holding `particle_switch = None` and `using stellar_model = quokka::ToyStellarModel`. `Particle_Traits<problem_t>` now inherits from it. This avoids having to touch every test problem when new particle-trait parameters are added.
- All 15 existing `Particle_Traits` specializations updated to `: DefaultParticleTraits`.
- Removed the standalone `StellarModel_Traits` — model selection now goes through `Particle_Traits::stellar_model`.

**Simulation wiring (`src/simulation.hpp`, `src/QuokkaSimulation.hpp`, `src/linear_advection/AdvectionSimulation.hpp`)**
- `StarParticles` container member, `createInitialStarParticles()` pure virtual + default empty definition, Star block in `InitPhyParticles`.
- Star particles are correctly loaded from checkpoint on restart (via `initializeParticleContainerFromCheckpoint`) matching the pattern used by all other particle types.

**Validation test (`src/problems/ParticleStarEvolution/`)**
- Places one Star particle at the domain centre in a uniform, isothermal medium and lets it accrete via the grid Bondi module.
- r_B is set ~10× smaller than dx so the accretion is in the sub-grid Bondi regime.
- Validates each coarse step: R(M) and L(M,ṁ) match the toy model closed forms within 2%, accounting for the one-step lag between accretion and the stellar update.
- Validates the mean numerical dM/dt matches the analytic Bondi rate within 10%.
- B₀ is set to 1×10⁻¹¹ so that β ≫ 1 and the magnetised Bondi rate converges to the unmagnetised formula.

**Documentation (`docs/markdown/particles.md`)**
- New "Star Particle Type" section covering attributes, model selection, the toy model, and the validation test.

### Related issues
N/A

### Checklist


src/particles/PhysicsParticles.hpp
```diff
@@ -918,6 +927,11 @@ template <typename problem_t> class PhysicsParticleRegister
 		} else if constexpr (particleType == ParticleType::Sink) {
 			descriptor = std::make_unique<PhysicsParticleDescriptor<ContainerType, problem_t, ParticleType::Sink>>(
 			    container, SinkParticleMassIdx, -1, -1, -1, true, false, -1, true, -1, SinkParticleMdotIdx, SinkParticleLxIdx);
+		} else if constexpr (particleType == ParticleType::Star) {
+			constexpr int lum_idx = (Physics_Traits<problem_t>::nGroups > 0) ? StarParticleLumIdx : -1;
+			descriptor = std::make_unique<PhysicsParticleDescriptor<ContainerType, problem_t, ParticleType::Star>>(
+			    container, StarParticleMassIdx, lum_idx, StarParticleBirthTimeIdx, -1, /*allows_creation=*/false,
+			    /*allows_destruction=*/false, /*evolution_stage_idx=*/-1, /*allows_accretion=*/true, /*mass_at_birth_idx=*/-1, StarParticleMdotIdx);
```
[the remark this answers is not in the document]
markkrumholz: This does seem like a genuine issue, but since physically I have trouble seeing where we would ever use both accreting sink and star particles in the same simulation, the solution is probably just to explicitly forbid this combination.

src/particles/starparticle_radiation.hpp
```diff
@@ -0,0 +1,35 @@
...
+// modify any component (mass, radius, luminosity groups, etc.).
+class StellarUpdate
+{
+      public:
+	template <typename problem_t, typename ParticleType, int Nout>
+	AMREX_GPU_DEVICE AMREX_FORCE_INLINE static void updateStellarProperties(ParticleType &p, amrex::Real /*current_time*/, amrex::Real dt,
+										LuminosityGpuConstTables<Nout> const & /*gpu_tables*/) noexcept
+	{
+		using Model = typename Particle_Traits<problem_t>::stellar_model;
+		Model::evolve(&p.rdata(0), Nout, dt);
```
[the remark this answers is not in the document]
markkrumholz: The integer components are going to be needed for the Orion stellar evolution module, because they are used to keep track of the stage of stellar evolution.

markkrumholz:
Looks like there is a CI fail, and the AI-indicated issues look real to me (though easy enough to fix).

src/particles/PhysicsParticles.hpp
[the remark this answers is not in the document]
chongchonghe: yes, I'll just abort the code when the user turn on both star and sink since there is no point having both particle types in one simulation. 

However, for future reference in case we want to revisit this, having both sink and star particles (and more accreting particle types) is possible. We just have to do the buffer phase from all particle types first, compute an overall accretion rate and limiting factor, and then apply accretion on all particle types with the same limiting factor.

src/particles/starparticle_radiation.hpp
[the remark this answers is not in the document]
chongchonghe: resolved

chongchonghe:
@markkrumholz  Ready for review again.

src/particles/particle_types.hpp
```diff
@@ -407,6 +432,8 @@ template <ParticleType particleType, typename problem_t> auto getParticleIntComp
 								     // No integer components
 	} else if constexpr (particleType == ParticleType::CICRad) { // NOLINT
 								     // No integer components
+	} else if constexpr (particleType == ParticleType::Star) {   // NOLINT
```
markkrumholz, line 436: I think you should put
`expandEnumNames<StarParticleDataIdx, StarParticleIntegerComps<problem_t>, true>();`
here. This can default to zero for the type you have implemented, but will provide a hook for star particle types that want to have integer components. Again, the use-case I'm thinking of here is Orion-style star particles, which have an integer component.

src/particles/particle_update.hpp
```diff
@@ -105,6 +106,30 @@ template <> struct ParticlePropertyUpdateTraits<ParticleType::StochasticStellarP
 	}
 };
 
+#if AMREX_SPACEDIM == 3
+// Specialization for Star particles: dispatches to the modular stellar-evolution framework.
+template <> struct ParticlePropertyUpdateTraits<ParticleType::Star> : ParticlePropertyUpdateBase<ParticleType::Star> {
+	template <typename problem_t, typename ParticleType, int Nout>
+	AMREX_GPU_DEVICE AMREX_FORCE_INLINE static void updateProperties(ParticleType &p, amrex::Real current_time, amrex::Real dt,
+									 LuminosityGpuConstTables<Nout> const &gpu_tables) noexcept
```
markkrumholz: Is the design idea here that you're passing a data table that contains luminosity in each band as a function of stellar properties? If so, is that something that it makes sense to pass here, or should any tables needed simply hang off the relevant star particle class. That is, rather than having an empty table that gets passed for some stars and not others, why not have the design be that star particle classes can have their own internal tables as class members, which they are free to use in the evolution function?

src/particles/particle_update.hpp
```diff
@@ -105,6 +106,30 @@ template <> struct ParticlePropertyUpdateTraits<ParticleType::StochasticStellarP
...
+
+	template <typename problem_t, typename ContainerType>
+	static void updateParticleProperties(ContainerType *container, amrex::Real current_time, amrex::Real dt) noexcept
+	{
+		const BL_PROFILE("ParticlePropertyUpdateTraits<Star>::updateParticleProperties()");
+		if (container == nullptr) {
+			return;
+		}
+		constexpr int nGroups = Physics_Traits<problem_t>::nGroups;
+		LuminosityGpuConstTables<nGroups> const gpu_tables{}; // unused by the stellar model, passed for signature parity
```
markkrumholz: See my previous comment about gpu_tables -- don't see why it needs to be passed at all.

markkrumholz (changes requested):
There's a merge conflict that needs to be resolved, but on top of that, I'm confused by the design of passing an array of gpu_constants data to the evolve routine. Why do that? Why not just have the evolve routine implemented by each star particle class have access to the internal class members, some of which can be data tables that live on GPU? Since these tables are presumably read once during startup and then stored and not altered, I don't see the reason to pass them around rather than just have them be class members that are set on initialization.

markkrumholz:
@chongchonghe, comments added.

chongchonghe:
@markkrumholz All comments addressed. Ready for review.

src/particles/particle_types.hpp
chongchonghe: good point. done

src/particles/particle_update.hpp
chongchonghe: yes, you are right. I have moved the gpu_tables to inside each stellar update modules, e.g. luminosity table to inside `updateLuminosity`. Now, the code looks much cleaner.

src/particles/particle_update.hpp
chongchonghe: Similarly, done.

markkrumholz (approved):

PR #1964, opened 2026-06-19
Fix EMF averaging: gather edge wave speeds from straddling faces

AstroKriel:
### Description

Both `EMFAverage_LondrilloDelZanna2004` and `EMFAverage_Balsara2025` gather edge-centred wave speeds by taking the max of the face speed at the edge node and a neighbouring face. In both routines the offset pointed in the wrong direction: `ld04` offset along the parallel axis and `b25` along the transverse axis but with the wrong sign. The fix reads the face on the opposite side of the edge node in the transverse direction in both cases, which is the correct straddle and is covariant under point reflection.

I confirmed that this fix does not affect the correctness of any MHD behaviour: the Alfvén wave converges for both `q26+ld04` and `q26+b25`, the Brio-Wu shock tube passes, and most importantly, the Orszag-Tang problem now produces point-symmetric solutions to floating-point precision for both EMF averaging schemes; I tested this for `q26+b25` and `fs17+ld04` at 1024×1024×8 on 4 A100 GPUs. Below I show slices of the current density at t=1, alongside a second version of the same plot with the numerical difference between the domain and its point-reflected counterpart overlaid.

(left panel is `fs17+ld04` and the right panel is `q26+b25`)
ot-fix-cur
ot-fix

### Related issues

None.

### Checklist

### DIFF
--- a/src/hydro/mhd_system.hpp
+++ b/src/hydro/mhd_system.hpp
@@ -735,10 +735,10 @@ void MHDSystem<problem_t>::EMFAverage_LondrilloDelZanna2004(amrex::Array4<amrex:
 
 	amrex::ParallelFor(box_ec, [=] AMREX_GPU_DEVICE(int i, int j, int k) {
 		// 	// LondrilloDelZanna2004 scheme:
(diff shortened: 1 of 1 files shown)


markkrumholz (approved):
Good catch!

PR #1990, opened 2026-06-23
Add wide-density ISRF shielding cooling table

YaoguangPei:
### Description
Current Quokka shielding cooling  tables, 'CloudyData_UVB=HM2012_shielded_resampled_noPE.h5' or 'CloudyData_UVB=HM2012_shielded_resampled.h5' , only have very narrow density range, the maximum density is $10^4 cm^{-3}$, which is very limited in the simulation with self-gravity. Thus I created a new table using Cloudy with a more wide density range and shielding.

Below, I will give a brief explanation to the setup of my table,
1)The range of density, $10^{-6}\leq n \leq10^9 \ \mathrm{cm^{-3}}$, step is 0.1 in log.
2)The range of temperature, $10\leq T \leq 10^9 \ \mathrm{K}$,step is 0.05 in log.
3)Using ISRF
4)Turn on CR
5)Turn on grains(i.e. dust) for PE, but turn off when the temp reached the $T_{max}$ we set in the sim.
6)Using solar abundance.
7)With iteration
8)Set a thickness to simulate the self-shielding. And below I will explain how to calculate the stop thickness in my sim.

Here I briefly explain how the effective shielding length / thickness is calculated in my simulation.
First, we calculate the Jeans length as

$$
L_{\rm Jeans} = \sqrt{\frac{\pi c_s^2}{G\rho}},
$$

where

$$
\rho = \mu m_p n_{\rm H},
$$

and

$$
c_s^2 = \frac{k_{\rm B}T}{\mu m_p}.
$$

We also set a target column density, $N_{\rm H}^{\rm target}$. The corresponding shielding length is

$$
L_{\rm shield} = \frac{N_{\rm H}^{\rm target}}{n_{\rm H}}.
$$

Then we take the smaller value between the Jeans length and the shielding length:

$$
L_{\rm raw} = \min \left( L_{\rm Jeans}, L_{\rm shield} \right).
$$

Finally, we compare this value with a minimum allowed length, $L_{\min}$, and define the effective shielding length as

$$
L_{\rm eff} = \max \left( L_{\rm raw}, L_{\min} \right).
$$

The reason for introducing $L_{\rm shield}$ is that if the Jeans length is used directly, the estimated column density is

$$
N_{\rm H} = n_{\rm H} L_{\rm Jeans} \propto \sqrt{n_{\rm H}T}.
$$

Therefore, $N_{\rm H}$ keeps increasing as the density increases, which may overestimate the shielding effect at high densities. The shielding length $L_{\rm shield}$ limits the effective column density to approximately $N_{\rm H}^{\rm target}$.
The minimum length $L_{\min}$ is introduced to prevent the shielding length from becoming unrealistically small at very high densities. In practice, this relaxes the strict restriction imposed by $L_{\rm shield}$ in the high-density regime.

This figure shows the estimated Av vs density.
image
Below is the cooling results and heating results,
image
image

I also found a theoretical result for cooling,
image
The figure shows the theoretical cooling results for solar metallicity with $n = 1 , \mathrm{cm^{-3}}$.

Note that SD93, SB97, GS07, and the solid line show cooling results under non-equilibrium conditions, while S09 shows the CIE cooling result. The S09 result is consistent with our cooling table result.

But for heating, I do not find a standard results, or in other words, I do not find a theoretical results as  same as mine. That may due to the reasons I mentioned above. But for conference, I still provide my results, as you can see . The discontinuity points at $T = 1\times 10^4\ \mathrm{K}$ and $T=400 \ \mathrm{K}$ is  due to the artificial switching model, turn off dust and turn on CO, separately. And for $n_H=10^7 \mathrm{cm^{-3}}$, we can see a much higher plateau, which is due to the dust-gas collision.
### Related issues

### Checklist


YaoguangPei:
@chongchonghe

BenWibking:
In practice, the reason the cooling table included in Quokka stops at around ~1e4 cm^-3 is because that's how far the Grackle table from which it was converted extends. Although not always applicable to every problem, it is now a fairly standard cooling table in the literature, so we have included it for comparison to other works.

I think the number of possible assumptions one can make about shielding and dust grains is very large, and will be highly problem-dependent. If a table extending to these densities is provided with Quokka, I am concerned that a lot of users will just use it in their simulations without thinking whether these assumptions are applicable to their problem.

At minimum, this needs to include a README file explaining how this table was created (including the Cloudy version and input file used). This is because in general users of Quokka will not look up the PR description for the PR that added this table, and will just see the table in the repo without any context.

Perhaps we could create a separate repository for additional cooling tables, or add a list of links to contributed tables in the documentation that includes a detailed description of their physics assumptions, and ideally a complete script to regenerate the table from scratch that includes a precise specification of all physics assumptions.

YaoguangPei:
> In practice, the reason the cooling table included in Quokka stops at around ~1e4 cm^-3 is because that's how far the Grackle table from which it was converted extends. Although not always applicable to every problem, it is now a fairly standard cooling table in the literature, so we have included it for comparison to other works.
> 
> I think the number of possible assumptions one can make about shielding and dust grains is very large, and will be highly problem-dependent. If a table extending to these densities is provided with Quokka, I am concerned that a lot of users will just use it in their simulations without thinking whether these assumptions are applicable to their problem.
> 
> At minimum, this needs to include a README file explaining how this table was created (including the Cloudy version and input file used). This is because in general users of Quokka will not look up the PR description for the PR that added this table, and will just see the table in the repo without any context.
> 
> Perhaps we could create a separate repository for additional cooling tables, or add a list of links to contributed tables in the documentation that includes a detailed description of their physics assumptions, and ideally a complete script to regenerate the table from scratch that includes a precise specification of all physics assumptions.

Thanks for your reply, I will add a README file to explain the detail of the table, including Cloudy version and input files.

YaoguangPei:
@BenWibking, I have updated the `README.md` and added the Cloudy input-generation script to the cooling directory. With this script, users can reproduce the Cloudy input grid and, if needed, regenerate a cooling table with their own Cloudy installation.

chongchonghe:
@YaoguangPei I'll turn this into draft mode since we are working on including another cooling table for high density gas (> 1e5 cm-3). Once you finished the new cooling table, please update this PR and make it ready for review, or create another PR.

PR #1993, opened 2026-06-24
feat(radiation): support MHD + radiation coupling

### LINKED ISSUE #1992: Support radiation + MHD
**Describe the proposal**

Currently, `MHD + radiation` is explicitly disabled in `src/QuokkaSimulation.hpp` with `static_assert(!(is_mhd_enabled && is_radiation_enabled), "MHD + Radiation is not supported yet.")`. The reason is the potential erroneous internal energy calculation in the radiation module when MHD is enabled. 

Solution:
- Move `ComputeEgasFromEint` to `hydro_system.hpp`, and add docs to make it clear that it computes gas total energy (not including B field) from gas internal energy
- Write a function `ComputeEintFromEmhdtot` and `ComputeEmhdtotFromEint` to compute `Eint` and `Emhdtot` (the state total energy, including gas and B field) from float numbers, similar to `ComputeEgasFromEint`
- When computing Eint at the beginning of a radiation-matter coupling step, need to exclude the B field energy. 
- Check all occurrence of `ComputeEgasFromEint`, and make sure they all have accounted for B field when they should. If inside a test problem without B field, add a comment to remind the user that B components is explicitly ignored in the calculation. 
- Check the whole database to make sure the code handles gas internal energy correctly when both MHD and radiation are turned on.

chongchonghe:
### Description
Enables MHD + radiation coupling by making all energy conversion functions MHD-aware, so magnetic energy is correctly excluded from gas internal energy throughout the radiation solver.

Previously, `ComputeEintFromEgas` (total → internal energy) and `ComputeEgasFromEint` (internal → total energy) lived in `RadSystem` and silently ignored magnetic energy. A `static_assert(!is_mhd_enabled)` blocked MHD + radiation to prevent incorrect results. This PR removes that restriction and ensures `Eint` is computed correctly everywhere.

### Design

**Energy conversion (`src/hydro/EOS.hpp`):**  
`ComputeEintFromEgas` and `ComputeEgasFromEint` are moved from `RadSystem` to `quokka::EOS` and now take an explicit `magnetic_energy` scalar:
```cpp
auto ComputeEintFromEgas(double rho, double mx, double my, double mz, double Etot, double magnetic_energy) -> double;
```
When MHD is disabled, callers pass `0.0`. A `static_assert` fires at every call site that passes `0.0` when MHD is enabled, making it impossible to silently compute wrong temperatures.

**Cell-centred B² (`src/physics_info.hpp`, `src/hydro/hydro_system.hpp`):**  
`ComputeCellCenteredMagneticEnergy<problem_t>(i, j, k, cons_fc)` averages adjacent face-centred B components to produce the magnetic energy density at cell centre. When MHD is disabled the function is a no-op returning 0.

The function cannot live in `hydro_system.hpp` even though it logically belongs to `HydroSystem`. The include chain `hydro_system.hpp` → `radiation_system.hpp` → `source_terms_*.hpp` means those bottom-included files cannot `#include "hydro/hydro_system.hpp"` — the include guard is already set (the header is mid-compilation) so `HydroSystem` would be undefined when they try to use it, a classic circular-include deadlock. Instead, the device function is defined in `physics_info.hpp`, a leaf header included early by both `hydro_system.hpp` and `radiation_system.hpp`. `HydroSystem::ComputeCellCenteredMagneticEnergy` then delegates to it as a thin static wrapper for callers outside the radiation subsystem.

**Radiation source terms — by-value `cons_fc` (`src/radiation/source_terms_*.hpp`):**  
`AddSourceTermsSingleGroup` and `AddSourceTermsMultiGroup` now accept `cons_fc` by value (`std::array<Array4<const Real>, AMREX_SPACEDIM>`) instead of by pointer. `Array4` is a trivially-copyable struct, so capture-by-value inside `AMREX_GPU_DEVICE` lambdas is safe on both CUDA and HIP. The old pointer-to-host-stack approach was undefined behaviour.

**Optical depth with MHD (`src/radiation/radiation_system.hpp`):**  
`ComputeCellOpticalDepth` now accepts physical indices and `cons_fc` so it can compute `Emag` for the left and right cells and pass it to `ComputeEintFromEgas`. `cons_fc` is threaded through `ComputeFluxes` → `fluxFunction` → `computeRadiationFluxes` with a default-constructed `{}` for non-MHD cases.

**Photochemistry with MHD (`src/radiation/photochemistry.hpp`):**  
`computePhotoChemistry` now accepts `std::array<amrex::MultiFab const*, AMREX_SPACEDIM> const &fc_mfs` alongside the existing cell-centred `MultiFab`. The face-centred arrays are resolved per-FAB via `fc_mfs[d]->const_array(iter)` (the same `MFIter` already used for the cell-centred loop), then captured by value in the `AMREX_GPU_DEVICE` lambda. `Emag` is computed per cell with `ComputeCellCenteredMagneticEnergy` and passed to `ComputeEintFromEgas`, replacing the previous hardcoded `0.0`. The `static_assert(!is_mhd_enabled)` is removed.

**Global-namespace prefix `::quokka::` (`src/` throughout):**  
Every call site that references `quokka::EOS`, `quokka::EOS_Traits`, etc. uses the `::quokka::` spelling with a leading `::`. The `::` anchors name lookup to the global namespace root. Without it, HIP's clang-based frontend can mis-resolve the name inside device-code templates that are themselves compiled in a nested namespace: it applies *relative* lookup and finds (or invents) `quokka::quokka::EOS` instead of `::quokka::EOS`. GCC and NVCC handle the ambiguous case correctly, but the ROCm compiler does not, so the explicit global-namespace anchor is required for correctness on all three backends.

### Changes
- **`src/hydro/EOS.hpp`**: `ComputeEintFromEgas` / `ComputeEgasFromEint` moved here from `RadSystem`; take explicit `magnetic_energy` scalar
- **`src/physics_info.hpp`**: add `ComputeCellCenteredMagneticEnergy<problem_t>` free function
- **`src/hydro/hydro_system.hpp`**: expose `HydroSystem::ComputeCellCenteredMagneticEnergy` delegating to the above
- **`src/radiation/radiation_system.hpp`**: remove `static_assert(!is_mhd_enabled)`; add `cons_fc` to `ComputeFluxes` and `ComputeCellOpticalDepth`; update source-term declarations
- **`src/radiation/source_terms_single_group.hpp`**: `cons_fc` by-value; use `ComputeCellCenteredMagneticEnergy`
- **`src/radiation/source_terms_multi_group.hpp`**: same
- **`src/radiation/photochemistry.hpp`**: add `fc_mfs` parameter; MHD-aware `Emag` via `ComputeCellCenteredMagneticEnergy`; `static_assert` removed
- **`src/QuokkaSimulation.hpp`**: remove static_assert; thread `cons_fc_arr` through radiation flux and source-term calls; MHD-aware Shu-Osher combination
- **16 test problem files + assorted headers**: updated namespace to `::quokka::EOS`, explicit `0.0` at former ambiguous call sites; `::` prefix required for ROCm correctness (see above)

### Related issues
Closes #1992

### Checklist


chongchonghe:
Fixed another CI issue found while monitoring:

**Root cause**: Commit `9d8c5929` added `::quokka::` prefix to function-body calls (e.g., `::quokka::EOS<X>::ComputeEintFromTgas(...)`) in addition to template specialization declarations. The prior fix (`5232f518`) only reverted the specialization declarations. The function-body occurrences remained, causing 83 problem files to appear as "changed" vs `development` in the `git diff`.

**Failure**: The `rocm-quick` pipeline auto-detects changed problems and tries to build them in 3D. Several of those 83 files belong to 1D-only problems (e.g., `RadhydroPulse`, `RadhydroPulseDyn`, `RadhydroPulseGrey`, etc.) that have no target in the 3D build → `gmake: *** No rule to make target 'RadhydroPulse'. Stop.`

**Fix** (`64d5c96f`, `3761f025e`): Reverted remaining `::quokka::` → `quokka::` in function bodies across all problem files. Now only 29 problem directories (all 3D-capable) appear as changed vs `development`, so the pipeline can build them all.

src/radiation/radiation_system.hpp
```diff
@@ -135,6 +135,25 @@ template <typename problem_t> struct FluxUpdateResult {
 	return 0.5 * (sgn(a) + sgn(b)) * std::min(std::abs(a), std::abs(b));
 }
 
+// Helper to compute cell-centered magnetic energy density (0.5 * B^2) from face-centered data.
+// Defined as a separate device function (rather than inline in a ParallelFor lambda)
+// to work around an NVCC limitation that disallows first-capturing variables in
+// constexpr-if contexts inside extended device lambdas.
+template <typename problem_t>
+AMREX_GPU_DEVICE auto ComputeCellCenteredMagneticEnergy(int i, int j, int k, std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM> const &fc) -> double
```
BenWibking: Can you make this `AMREX_FORCE_INLINE`?

Can this go in HydroSystem instead of RadSystem?

src/QuokkaSimulation.hpp
```diff
@@ -3154,6 +3162,17 @@ void QuokkaSimulation<problem_t>::subcycleRadiationAtLevel(int lev, amrex::Real
 				const amrex::Box &indexRange = iter.validbox();
 				auto const &stateNew = state_new_cc_[lev].array(iter);
 				auto const &stateTmp = state_tmp1_cc.const_array(iter);
+
+				// Build face-centered array for MHD-aware energy conversion
+				auto cons_fc_shu = std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM>{};
```
BenWibking: Why `_shu`?

chongchonghe: because this is in the block of code about Shu-Osher formalism. Claude probably found this interesting and added a _shu suffix. Doesn't hurt.

BenWibking: No, but it doesn't really add useful information.

src/radiation/radiation_system.hpp
chongchonghe: There will be circular import if i put it in `HydroSystem`. Claude found a compromising solution: put it in physics_info.hpp . I think this is fine.

src/QuokkaSimulation.hpp
chongchonghe: ok, changed.

chongchonghe:
@BenWibking All three review comments have been addressed:

1. **`ComputeCellOpticalDepth` MHD support** — `cons_fc` (face-centered B fields) is now threaded through `computeRadiationFluxes` → `fluxFunction` → `ComputeFluxes` → `ComputeCellOpticalDepth`, so magnetic energy is correctly subtracted when computing gas temperature for the optical depth estimate. The two `static_assert(!is_mhd_enabled)` guards are removed.

2. **GPU pointer capture bug** — `cons_fc` is now passed by value (`std::array<Array4>`) everywhere instead of as a pointer-to-host-stack. `Array4` is a small trivially-copyable struct, so capture-by-value in GPU lambdas is safe on both CUDA and HIP.

3. **Photochemistry + MHD crash** — `computePhotoChemistry` no longer takes a `cons_fc` pointer parameter (which was structurally broken for multi-FAB loops). A `static_assert(!is_mhd_enabled)` with a descriptive message blocks the combination until the function is redesigned to accept face-centered `MultiFab`s.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -758,7 +764,9 @@ template <> auto QuokkaSimulation<DiskGalaxy>::ComputeStatistics() -> std::map<s
 		const Real x2Mom = state(i, j, k, HydroSystem<DiskGalaxy>::x2Momentum_index);
 		const Real x3Mom = state(i, j, k, HydroSystem<DiskGalaxy>::x3Momentum_index);
 		const Real Egas = state(i, j, k, HydroSystem<DiskGalaxy>::energy_index);
-		const Real Eint = RadSystem<DiskGalaxy>::ComputeEintFromEgas(rho, x1Mom, x2Mom, x3Mom, Egas);
+		// computeVolumeIntegral does not provide face-centred data, so Emag is not available.
```
BenWibking, line 767: I can fix this in a separate PR. It should compute it accurately. It's easy to fix.

BenWibking (commented):
This looks like a reasonable approach. However, it's different from the PR description, where you say you will use a variadic argument. In this code, it looks like `Emag` is always required, but you set it to `0.0` for pure hydro problems. I think that's fine, but the PR description should be updated to match the code.

Why did you add the global namespace qualifier `::quokka::EOS` in front of all the EOS calls? Was there a compiler issue? It would be slightly preferable if we could solve the compiler issue in a way that avoids making the syntax uglier.

I have a few possible minor nits about possible code style improvements, but I will add those tomorrow.

src/QuokkaSimulation.hpp
```diff
@@ -3200,21 +3232,35 @@ void QuokkaSimulation<problem_t>::subcycleRadiationAtLevel(int lev, amrex::Real
...
 									       dustGasInteractionCoeff_, rad_tol, rad_tol_rel, tempFloor, p_iteration_counter,
-									       p_iteration_failure_counter);
+									       p_iteration_failure_counter, &cons_fc_arr);
 			}
 		}
 #ifdef PHOTOCHEMISTRY
 		if (enablePhotoChemistry_ == 1) {
 			// compute photo-chemistry
+			// Build cons_fc for photochemistry
 			quokka::photochemistry::computePhotoChemistry<problem_t>(state_new_cc_[lev], dt_radiation, 1, max_density_allowed, min_density_allowed);
```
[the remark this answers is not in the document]
BenWibking: This appears to be a real issue, as far as I can tell.

chongchonghe:
Right. Passing the B field components as a vector is not necessary, because only magnetic pressured is used. So, by explicitly required passing 'B = ' to the function, the programmer is reminded to handle B field properly. In hydro-only test problems, it's not necessary to construct an empty B field vector and pass it. My solution is to `static_assert` MHD is off and pass 'B=0'.

chongchonghe:
PR description updated.

src/radiation/source_terms_single_group.hpp
```diff
@@ -42,6 +43,20 @@ void RadSystem<problem_t>::AddSourceTermsSingleGroup(array_t &consVar, arraycons
 		const double Egastot0 = consPrev(i, j, k, gasEnergy_index);
 		auto massScalars = RadSystem<problem_t>::ComputeMassScalars(consPrev, i, j, k);
 
+		// Compute cell-centered magnetic field, if MHD is enabled
+		const auto B_cc = [&]() -> std::array<amrex::Real, 3> {
+			if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
+				const amrex::Real bx = 0.5 * ((*cons_fc)[0](i, j, k, Physics_Indices<problem_t>::mhdFirstIndex) +
+							      (*cons_fc)[0](i + 1, j, k, Physics_Indices<problem_t>::mhdFirstIndex));
```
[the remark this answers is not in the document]
chongchonghe: done.

src/QuokkaSimulation.hpp
```diff
@@ -3313,7 +3344,15 @@ void QuokkaSimulation<problem_t>::advanceRadiationForwardEuler(int lev, amrex::R
 		const amrex::Box &indexRange = iter.validbox();
 		auto const &stateOld_cc = state_old_cc_[lev].const_array(iter);
 		auto const &stateNew_cc = state_out.array(iter);
-		auto [fluxArrays, fluxDiffusiveArrays] = computeRadiationFluxes(stateOld_cc, indexRange, ncompHyperbolic_, dx);
+		std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM> cons_fc_arr;
+		if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
+			cons_fc_arr[0] = state_new_fc_[lev][0].const_array(iter);
+			cons_fc_arr[1] = state_new_fc_[lev][1].const_array(iter);
+#if (AMREX_SPACEDIM == 3)
```
BenWibking: **Temporal mismatch in the radiation flux optical depth** (low–medium severity)

`cons_fc_arr` is sourced from `state_new_fc_[lev]` — the post-hydro B field at **time n+1** — but the accompanying flux is computed from `stateOld_cc`, i.e. the gas at **time n**. So inside `ComputeCellOpticalDepth`:

```
Eint_L = Etot_old - Ekin_old - Emag_new
```

mixes time levels n (gas) and n+1 (B). The source-term call sites just below are consistent (`state_new_cc` paired with `state_new_fc_`), but the flux is paired mismatched.

For consistency with the pre-hydro gas, the flux should source Emag from `state_old_fc_[lev]` (B^n), which is still valid and ghost-filled at this point in the timestep.

Practical impact is limited — `use_wavespeed_correction_` defaults to `false` (`QuokkaSimulation.hpp:218`), and the error is O(ΔB·dt), largest in low-beta plasma where Emag ≳ Eint. This is not a conservation issue: Emag here only feeds the wavespeed-correction optical depth, not the energy bookkeeping.

BenWibking (commented):
## Overview

Reviewed the MHD + radiation coupling logic. The energy-conversion math in `ComputeCellOpticalDepth`, the `ComputeCellCenteredMagneticEnergy` face-averaging, and the `static_assert(!is_mhd_enabled)` guards at the `0.0` call sites are correct, as is the `FluxDir` index mapping in each of the three branches. A few items below for consideration.

---

### ⚠️ No test covers the new MHD+radiation path

No problem under `src/problems/` enables both `is_mhd_enabled = true` and `is_radiation_enabled = true` (DiskGalaxy is MHD-only). This PR removes the `static_assert` that previously blocked this combination and adds the coupling logic, but no regression test exercises it — so the PR checklist item ("I have added tests for any new physics") isn't satisfied.

Per `AGENTS.md`, a CTest target and a matching `inputs/` file should be added for the MHD+radiation combination so that any future regression in this code path isn't silent.

src/QuokkaSimulation.hpp
```diff
@@ -3365,8 +3404,16 @@ void QuokkaSimulation<problem_t>::advanceRadiationMidpointRK2(int lev, amrex::Re
 		auto const &stateOld_cc = state_old_cc_[lev].const_array(iter);
 		auto const &stateInter_cc = state_inter.const_array(iter);
 		auto const &stateNew_cc = state_new_cc_[lev].array(iter);
-		auto [fluxArraysOld, fluxDiffusiveArraysOld] = computeRadiationFluxes(stateOld_cc, indexRange, ncompHyperbolic_, dx);
-		auto [fluxArrays, fluxDiffusiveArrays] = computeRadiationFluxes(stateInter_cc, indexRange, ncompHyperbolic_, dx);
+		std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM> cons_fc_arr;
+		if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
+			cons_fc_arr[0] = state_new_fc_[lev][0].const_array(iter);
+			cons_fc_arr[1] = state_new_fc_[lev][1].const_array(iter);
+#if (AMREX_SPACEDIM == 3)
```
BenWibking: **Same temporal-mismatch issue** as in `advanceRadiationForwardEuler` above — this is the RK2 second call site, also mixing `stateOld_cc` (gas^n) with `state_new_fc_` (B^{n+1}) in `cons_fc_arr`. Should similarly use `state_old_fc_[lev]` if you decide to fix the first occurrence.

src/radiation/source_terms_single_group.hpp
```diff
@@ -486,7 +490,8 @@ void RadSystem<problem_t>::AddSourceTermsSingleGroup(array_t &consVar, arraycons
 			// 3. Deal with the work term.
 			if constexpr ((gamma_ != 1.0) && (beta_order_ != 0)) {
 				// compute difference in gas kinetic energy before and after momentum update
-				amrex::Real const Egastot1 = ComputeEgasFromEint(rho, x1GasMom1, x2GasMom1, x3GasMom1, Egas_guess);
+				amrex::Real Egastot1 = NAN;
+				Egastot1 = ::quokka::EOS<problem_t>::ComputeEgasFromEint(rho, x1GasMom1, x2GasMom1, x3GasMom1, Egas_guess, Emag);
```
BenWibking: **Minor: `const` dropped and redundant `NAN` init**

The multi-group equivalent keeps `auto const Egastot1 = ::quokka::EOS<problem_t>::ComputeEgasFromEint(...)`. Here it was split into

```cpp
amrex::Real Egastot1 = NAN;
Egastot1 = ::quokka::EOS<problem_t>::ComputeEgasFromEint(...);
```

Looks like a debugging leftover — `Egastot1` should be `const` (it is never reassigned after this), and the `NAN` initialization is redundant. Suggest restoring the single `const` form for consistency with the multi-group file.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
BenWibking, line 767: **Documented, low priority:** `Eint = Egas - Ekin` here (without Emag) overestimates the internal energy for this MHD problem because `computeVolumeIntegral` doesn't expose face-centred data. The comment documents the limitation; the impact is on `mass_T_lt_1e4` — overestimated T → undercounted cold mass, worst in low-beta regions.

Acceptable as documented. Worth threading `state_fc` into `computeVolumeIntegral` if that statistic matters for this scenario.

BenWibking:
@chongchonghe The review comments are from GLM-5.2. I have reviewed them in detail and I agree with them.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
BenWibking: I think this can be done in subsequent PR.

BenWibking:
It might make sense to add a constant magnetic field to one of the 1D radiation tests. That should leave the solution unchanged.

src/QuokkaSimulation.hpp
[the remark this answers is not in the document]
chongchonghe: resolved by [8cd6f89](https://github.com/quokka-astro/quokka/pull/1993/commits/8cd6f89b30f3b77514d5c4328494266ec4e5ca21)

chongchonghe:
@BenWibking Three more comments addressed:

1. **Temporal mismatch (both call sites)** — both `advanceRadiationForwardEuler` and `advanceRadiationMidpointRK2` now source `cons_fc_arr` from `state_old_fc_[lev]` (B^n) instead of `state_new_fc_[lev]` (B^{n+1}), matching the time level of `stateOld_cc` that the radiation fluxes are computed from.

2. **`const` / redundant `NAN`** — `Egastot1` in `AddSourceTermsSingleGroup` is now a single-line `const` declaration, consistent with the multi-group file.

src/QuokkaSimulation.hpp
chongchonghe: addressed

src/QuokkaSimulation.hpp
chongchonghe: addressed

src/radiation/source_terms_single_group.hpp
chongchonghe: addressed

chongchonghe:
TODO: add MHD to one of the radiation test problem

chongchonghe:
> Why did you add the global namespace qualifier `::quokka::EOS` in front of all the EOS calls? Was there a compiler issue? It would be slightly preferable if we could solve the compiler issue in a way that avoids making the syntax uglier.

explained in the description.

BenWibking (approved):
LGTM

BenWibking (approved):

PR #2012, opened 2026-06-28
Unify gas temperature through EOS backend hierarchy

chongchonghe:
### Description
Introduce a compile-time EOS backend hierarchy (EOSIdeal, EOSMicrophysics,
EOSTabulated) selected via EOS_Traits, making quokka::EOS the single authority
for gas temperature. All 17 direct ResampledCooling table-temperature call
sites now route through EOS. The temperature branch of EOSFlagforConduction
in electron conduction is removed.

### Design summary
Full design doc: `docs/superpowers/specs/2026-06-28-unified-eos-temperature-design.md`

**Three backends:**
- `EOSIdeal<problem_t>` — gamma-law ideal gas. Always available.
- `EOSMicrophysics<problem_t>` — only compiled under `CHEMISTRY`/`PHOTOCHEMISTRY`;
  uses the Microphysics `eos_t` state and `eos()` free function.
- `EOSTabulated<problem_t>` — temperature and entropy methods read the resampled
  table via a managed-memory global registry. All other methods (pressure, sound
  speed, isothermal sound speed, pressure derivatives) delegate to `EOSIdeal`.

**Trait selection:**
- `EOS_Traits<problem_t>` is a plain struct with gamma-law defaults and
  `using EOSBackend = EOSIdeal<problem_t>` (or `EOSMicrophysics` under
  `CHEMISTRY`/`PHOTOCHEMISTRY`).
- `detail::EOSBackendHelper` uses SFINAE: if `EOS_Traits<T>::EOSBackend`
  exists, use it; otherwise fall back to the preprocessor-appropriate default.
  This preserves backward compatibility with all existing full specializations.
- `EOS<problem_t>` forwards method calls to the resolved backend (not
  inheritance), so existing explicit member-function specializations
  (RadMarshak, RadSuOlson, RadMatterCoupling, etc.) remain valid.

**Registry:**
- `EOSTabulatedRegistry` lives in `EOSTabulatedRegistry.hpp` alongside the
  consolidated `resampledGpuConstTables` struct and the `ComputeTgasFromEgas` /
  `ComputeEgasFromTgas` / `ComputeEntropyFromRhoEint` table-access functions.
- The registry container is allocated in `The_Managed_Arena`; the pointer
  is `AMREX_GPU_MANAGED`. EOSTabulated methods use `AMREX_IF_ON_DEVICE` /
  `AMREX_IF_ON_HOST` to pick the correct sub-handle.
- Registration happens in `QuokkaSimulation.hpp` immediately after
  `readResampledData()`. A guard aborts if `cooling.enabled=1` but the
  EOS backend is not `EOSTabulated`.
- `ComputeEintTempDerivative` is optimized to 1 root-find + 2 cheap table
  lookups instead of 2 root-finds.

**Scope boundaries (out of scope):**
- Pressure / sound speed unification through the table — would change the
  hydro closure and wave speeds.
- The EOS ↔ VODE / Microphysics fork and integrator replacement.
- Per-problem custom mixed EOS.

### Files changed
| File | Change |
|------|--------|
| `src/hydro/EOS.hpp` | Backend structs, `EOS_Traits`, `EOSBackendHelper`, `EOS<problem_t>` forwarding |
| `src/cooling/EOSTabulatedRegistry.hpp` | **New.** Registry, `resampledGpuConstTables`, table-access functions |
| `src/cooling/ResampledCooling.hpp` | Removed duplicated struct/functions (now in registry header) |
| `src/cooling/ResampledCooling.cpp` | `g_eos_tabulated_registry` + `registerEOSTabulated()` |
| `src/QuokkaSimulation.hpp` | Registration + guard for cooling without EOSTabulated |
| `src/conduction/ElectronConduction.hpp` | Temperature always from EOS; sound speed remains table call |
| 5 problem files | `EOSBackend = EOSTabulated` + call site migration + dead code removal |

### Related issues
N/A (tracks issue #1810)

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced unified EOS temperature handling with tabulated backends for gas temperature and entropy (driven by an EOS backend hierarchy).
  * Updated temperature/entropy-related diagnostics to derive results via the EOS path using internal energy.
  * When cooling is enabled, unsupported non-tabulated EOS configurations are now rejected with clear guidance.

* **Bug Fixes**
  * Ensured consistent `Tgas` computation across CPU/GPU by removing resampled-cooling table branching from temperature in electron conduction.

* **Tests**
  * Updated multiple simulation tests/diagnostics to use the new EOS temperature/entropy computation.

* **Documentation**
  * Refined local bootstrap behavior (auto-installs `quokka.rc` when present) and updated the formatting/CI hook command.


src/hydro/EOS.hpp
```diff
@@ -28,442 +30,587 @@
...
-	const double Eint = Etot - Ekin - magnetic_energy;
-	AMREX_ASSERT_WITH_MESSAGE(Eint > 0., "Gas internal energy is not positive!");
-	return Eint;
-}
-
-template <typename problem_t>
-AMREX_GPU_HOST_DEVICE auto EOS<problem_t>::ComputeEgasFromEint(const double rho, const double mx, const double my, const double mz, const double Eint,
-							       const double magnetic_energy) -> double
+template <typename T, typename = void> struct EOSBackendHelper {
+	using type = EOSIdeal<T>;
```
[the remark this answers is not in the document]
chongchonghe: done

src/conduction/ElectronConduction.hpp
```diff
@@ -104,19 +104,20 @@ template <typename problem_t> class ElectronConduction
...
 			const amrex::Real rho = cons(i, j, k, HydroSystem<problem_t>::density_index);
 			const amrex::Real Eint = HydroSystem<problem_t>::ComputeInternalEnergy(cons, i, j, k, &local_state_fc);
-			amrex::Real Tgas = NAN;
+			// Temperature always from EOS
+			const int nmscalars_ = Physics_Traits<problem_t>::numMassScalars;
+			quokka::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars = {};
+			if constexpr (nmscalars_ > 0) {
+				massScalars = RadSystem<problem_t>::ComputeMassScalars(cons, i, j, k);
+			}
+			const amrex::Real Tgas = ::quokka::EOS<problem_t>::ComputeTgasFromEint(rho, Eint, massScalars);
```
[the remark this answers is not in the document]
chongchonghe: We should add a guard to abort the code if `cooling.enabled = 1` and `EOS` is not `EOSTabulated` .

chongchonghe (commented):
Addressed all review feedback:
- **ComputeEintTempDerivative**: optimized from 2 root-finds to 1 + 2 table lookups
- **ComputeTgasFromEgas / ComputeEgasFromTgas**: added rho <= 0.0 guards
- **SFINAE fallback**: now respects CHEMISTRY/PHOTOCHEMISTRY preprocessor defines
- **ElectronConduction guard**: QuokkaSimulation.hpp aborts if cooling.enabled=1 but EOS is not EOSTabulated

chongchonghe:
# PR #2012 Review: Unify gas temperature through EOS backend hierarchy

**Review type:** Feature (SPEC → VERIFY → EDGE)

## SPEC: Feature Contract

**What the feature does:** Introduce a compile-time EOS backend hierarchy (`EOSIdeal`, `EOSMicrophysics`, `EOSTabulated`) selected via `EOS_Traits<problem_t>::EOSBackend`, with a `DefaultEOSTraits`-style base that preserves backward compatibility. `EOSTabulated` routes temperature queries through the resampled cooling table via a managed-memory global registry. All 17 direct `ResampledCooling` table-temperature call sites migrate to `quokka::EOS::ComputeTgasFromEint`.

**Inputs:** `EOS_Traits::EOSBackend` type alias (opt-in); existing `cooling_table_type = "resampled"` and `cooling.enabled = 1` in TOML input files. Registration happens automatically after `readResampledData()` in `QuokkaSimulation.hpp`.

**Outputs:** `quokka::EOS::ComputeTgasFromEint` returns the table-backed temperature for tabulated problems; `ComputeEintFromTgas` returns table-backed internal energy; `ComputeEintTempDerivative` returns a finite-difference approximation. All non-temperature methods delegate to `EOSIdeal`.

**Stated boundaries:**
- Temperature-only scope — pressure/sound speed/entropy remain ideal-gas
- Strict compile-time selection: a problem compiled with `EOSTabulated` requires registered tables at runtime
- The 5 resampled-cooling problems (DiskGalaxy, TallBoxSf, ShockCloud, RandomBlast, ResampledCoolingTest) opt in; all ~70 other problems inherit `EOSIdeal` or `EOSMicrophysics` defaults unchanged
- Sound speed in `ElectronConduction` remains a direct `ResampledCooling::ComputeSoundSpeedFromRhoEint` call (cs is out of scope)

## SPEC/VERIFY/EDGE Results

|---|---|---|
| VERIFY – golden path | `quokka buildrun -d 1d ResampledCoolingTest` | PASSED. Relative L1 error norm = 0.0122 vs reference (tolerance 0.02). Cooling solution converged correctly through EOS table path. |
| VERIFY – backward compat | `quokka buildrun -d 1d PrimordialChem` | PASSED. Non-cooling Microphysics problem works unchanged through `EOSMicrophysics` backend. |
| EDGE – adjacent feature | `quokka buildrun -d 1d RadMarshak` | PASSED. Radiation Marshak wave (uses `EOS::ComputeTgasFromEint` for opacity/temperature coupling) runs correctly. Relative L1 error norm = 0.0182. |

All three builds and runs completed without error. The `quokka::EOS<ResampledCoolingTest>::ComputeTgasFromEint` successfully reads the resampled table via the registry and produces results bitwise-equivalent to the old `ResampledCooling::ComputeTgasFromEgas` path (same tables, same interpolation code, now routed through EOS).

## Design Assessment

The implementation faithfully follows the design doc. Key structural decisions are sound:

- **Trait-based selection** with SFINAE fallback (`detail::EOSBackendHelper`) preserves backward compatibility — all pre-existing `EOS_Traits` specializations that don't define `EOSBackend` continue to use `EOSIdeal` or `EOSMicrophysics` based on the `#if defined(CHEMISTRY/PHOTOCHEMISTRY)` preprocessor default.
- **Managed memory registry** (`The_Managed_Arena`) with separate host/device sub-handles and `AMREX_IF_ON_DEVICE`/`AMREX_IF_ON_HOST` dispatch correctly handles the split `DataTable` storage introduced in this PR.
- **Temperature-only scope** is the right choice for this PR — changing pressure/sound-speed to the table path would alter the Riemann solver wave speeds and require separate validation.
- **DataTable dual-storage** (host pinned + device, via `const_tables()` / `const_tables_host()`, with `sync_tables_to_device()` called automatically after `fillDataTables`) is clean and prevents accidental host dereference of device pointers.

## PR-Introduced Concerns

1. **Hard-coded finite-difference epsilon in `EOSTabulated::ComputeEintTempDerivative`** (`src/hydro/EOS.hpp` line ~1353, `eps = 1.0e-6`). Near the table's `eint_min` or `eint_max` boundaries, the perturbation `Eint ± dEint` may fall outside the interpolation domain, causing the root-find or table lookup to hit boundary clamping. The current logic (`dT > 0.0 ? ... : 0.0`) guards against division by zero but not against clamped/bogus derivatives at domain edges. Consider a follow-up PR that uses a one-sided difference near boundaries or validates the derivative against a known tolerance.

2. **`resampledGpuConstTables` struct duplication.** The same struct is defined in both `src/cooling/ResampledCooling.hpp` and `src/cooling/EOSTabulatedRegistry.hpp`. The diff shows it was *moved* from `ResampledCooling.hpp` to `EOSTabulatedRegistry.hpp`, but both headers are included from `EOS.hpp`. If any code path includes `ResampledCooling.hpp` without `EOSTabulatedRegistry.hpp`, it would fail to compile. Verify all consumers include the right header. (Non-blocking — CI build passed.)

- `EOSMicrophysics::ComputeIsothermalSoundSpeed` has `static_assert(gamma_ == 1.0)`, which blocks compilation for non-isothermal Microphysics problems. This is pre-existing behavior, not introduced by this PR.
- The `EOSFlagforConduction` enum retains a value for `EOS` even though the temperature branch was removed — the enum is now used only for sound speed routing. Minor cleanup opportunity for a separate PR.

## Verdict

The PR is well-designed and correctly implemented. The golden path (ResampledCoolingTest table interpolation), backward compatibility (PrimordialChem Microphysics path), and an adjacent feature (RadMarshak radiation coupling) all pass. The temperature migration touches all 17 call sites across the 5 tabulated problems in a mechanical, value-preserving way. Recommend proceeding after GPU CI passes (manual `/azp run` pending).

🤖 Generated with [Claude Code](https://claude.com/claude-code) + DeepSeek-v4

BenWibking:
> Hard-coded finite-difference epsilon in EOSTabulated::ComputeEintTempDerivative (src/hydro/EOS.hpp line ~1353, eps = 1.0e-6). Near the table's eint_min or eint_max boundaries, the perturbation Eint ± dEint may fall outside the interpolation domain, causing the root-find or table lookup to hit boundary clamping. The current logic (dT > 0.0 ? ... : 0.0) guards against division by zero but not against clamped/bogus derivatives at domain edges. Consider a follow-up PR that uses a one-sided difference near boundaries or validates the derivative against a known tolerance.

I think I have a better way to do this - I can write it up shortly.

chongchonghe:
# PR #2012 Review: Unify gas temperature through EOS backend hierarchy

**PR type:** Feature / Enhancement (SPEC → VERIFY → EDGE cycle)

**Branch:** `chong/new-eos` → `development`

---

## SPEC: Feature contract (derived from diff and PR description)

**New inputs / configuration:**
- `EOS_Traits<P>::EOSBackend` — optional `using` alias selecting the compile-time backend; SFINAE `detail::EOSBackendHelper` falls back to `EOSIdeal<P>` (or `EOSMicrophysics<P>` under `CHEMISTRY`) if absent, preserving backward compatibility for all ~75 existing specializations.
- Five problems opt in: DiskGalaxy, TallBoxSf, ShockCloud, RandomBlast, ResampledCoolingTest — each adds `using EOSBackend = quokka::EOSTabulated<P>;` to their `EOS_Traits` specialization.
- `quokka::ResampledCooling::registerEOSTabulated(host_tables, device_tables)` — called from `QuokkaSimulation::readParmParse()` immediately after `readResampledData()`.

**New outputs / effects:**
- `EOS<P>::ComputeTgasFromEint` for tabulated problems → reads from `EOSTabulatedRegistry`, selecting device sub-handle on GPU, host sub-handle on CPU via `AMREX_IF_ON_DEVICE/HOST`.
- `EOS<P>::ComputeEintFromTgas` → root-find (toms748, up to 32 iters) on the temperature table.
- `EOS<P>::ComputeEintTempDerivative` → centered finite difference via two table lookups.
- All other EOS methods (pressure, sound speed, etc.) delegate to `EOSIdeal` — wave speeds unchanged.
- 17 direct `ResampledCooling::ComputeTgasFromEgas` call sites across 5 problem files migrated to `EOS<P>::ComputeTgasFromEint`.
- `ElectronConduction`: temperature now always from `EOS::ComputeTgasFromEint`; sound speed still from `ResampledCooling::ComputeSoundSpeedFromRhoEint` for `ResampledCooling` flag.
- `computePhotoelectricHeating`: switched from `const_tables()` to `const_tables_host()` — correctness fix for CUDA builds where device-memory tables cannot be dereferenced on the host.

**Stated boundary / error conditions:**
- `EOSTabulated` with null or inactive registry → `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` (device-safe abort).
- `enableCooling_ == 1` without `EOSTabulated` backend → runtime assertion via `EOS<P>::is_tabulated`.
- `rho <= 0` in `ComputeTgasFromEgas` → returns 0.0 (new guard; old code divided by zero / returned NaN).

---

## SPEC / VERIFY / EDGE table

| Phase | Command | Observed |
|---|---|---|
| VERIFY — golden path | `quokka run -d 1d ResampledCoolingTest` on `chong/new-eos` | **SUCCESS**: L1 error 0.0122 < tolerance 0.02; solution matches grackle reference |
| VERIFY — CI inputs | `grep cooling.enabled,cooling_table_type inputs/` for all 5 problems | All 5 CI inputs (DiskGalaxy, ShockCloud, TallBoxSf, RandomBlast, ResampledCoolingTest) have `cooling.enabled = 1` and `cooling_table_type = "resampled"`; registry will always be populated |
| EDGE | `quokka buildrun -d 1d HydroShocktube` on `chong/new-eos` | **SUCCESS**: non-tabulated ideal-gas problem unaffected; SFINAE default fallback (`EOSIdeal`) is working |

---

## Feature assessment

The architecture is clean and correct:

- **SFINAE dispatch** (`detail::EOSBackendHelper`) correctly falls back for the ~75 existing problem specializations, so no existing problems need to be touched.
- **Managed-memory registry** (`AMREX_GPU_MANAGED EOSTabulatedRegistry *`) delivers runtime table data to `HOST_DEVICE` static EOS methods without any call-site signature changes. The split `host`/`device` sub-handles (pinned vs. device-backed `DataTableGpuConst`) selected via `AMREX_IF_ON_DEVICE/HOST` is the correct pattern for mixed host-GPU code.
- **`DataTable` host/device split** (`coords_h_`/`coords_d_`, `const_tables_host()`) is a well-scoped addition that only changes the storage strategy; `fillDataTables` calls `sync_tables_to_device()` automatically so users don't need to think about the sync.
- **`computePhotoelectricHeating` fix** (device → host tables) is a genuine correctness improvement for CUDA builds bundled with this PR — the old code read device-only memory from host code, which is undefined behaviour on CUDA.
- **`ElectronConduction` simplification** (temperature always from EOS, sound speed stays per-flag) is consistent with the stated temperature-only scope and matches the design doc.

---

## PR-Introduced Concerns

### 1. Dead code in `ShockCloud::ComputeStatistics()` — incomplete call-site migration

`src/problems/ShockCloud/testShockCloud.cpp`, lines 535 and 579:

```cpp
auto tables = resampledTables_.const_tables();   // line 535 — dead
// … 5 lambdas each doing:
ComputeCellTempResampled(i, j, k, state, HydroSystem<ShockCloud>::gamma_, tables);

auto tables_orig = resampledTables_.const_tables();  // line 579 — dead
// … 5 more lambdas passing tables_orig
```

`ComputeCellTempResampled` (line 484) now ignores both `tables` and `/*gamma*/` in its body; it calls `EOS<ShockCloud>::ComputeTgasFromEint` directly. The function signature still carries `resampledGpuConstTables const &tables` as an unused parameter, and both `tables` / `tables_orig` are unnecessary. Two dead `const_tables()` calls are made and their results captured into ten GPU lambdas for no effect.

Not a correctness issue (ShockCloud is 3D-only and CI would catch breakage), but it is incomplete follow-through on this specific call site — other ShockCloud lambdas (e.g. `c_s`, `nH*`, `ComputeCoolingLength`) legitimately still use `const_tables()` for sound speed and density, so they are not dead.

**Suggested fix:** remove the `tables` / `tables_orig` locals from `ComputeStatistics()`, remove `resampledGpuConstTables const &tables` from the `ComputeCellTempResampled` signature, and remove `gamma_` from all its call sites.

### 2. No standalone unit tests for `EOSTabulated`

The design doc specified:
- EOS/table agreement: `EOS::ComputeTgasFromEint` matches `ResampledCooling::ComputeTgasFromEgas` across the table domain.
- Round-trip: `Eint → T → Eint` via tabulated backend.
- Temperature floor: `temperature_floor` with tabulated backend yields expected `T`.
- `ComputeEintTempDerivative` correctness.

The PR relies entirely on the existing `ResampledCoolingTest` regression (L1 error vs grackle reference). That test exercises `ComputeTgasFromEint` in a time-integration loop and confirms agreement with the reference, but it does not test `ComputeEintFromTgas` invertibility or `ComputeEintTempDerivative`. For the current temperature-only scope this is a low risk (none of the 5 migrated problems use radiation + tabulated EOS, which is the main consumer of `ComputeEintTempDerivative`), but the stated design tests are absent.

---

None identified.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) with Sonnet 4.6

chongchonghe:
@coderabbitai review

BenWibking:
I think CodeRabbit won't review draft PRs.

BenWibking:
@chongchonghe Let me know when this is ready for review.

chongchonghe:
@BenWibking It is ready for review. I'll clean up the design docs and some changes from #2034 later.

BenWibking:
Codex:
```
  Review comment:

  - [P2] Keep existing cooling inputs runnable — /Users/benwibking/amrex_codes/quokka/src/QuokkaSimulation.hpp:662-666
    When cooling.enabled = 1 is used by existing problems that were not migrated to EOSTabulated (for example ParticleSF with inputs/
    ParticleSF.toml or SN with inputs/SN.toml), this new unconditional assertion aborts during parameter parsing before the table is read.
    Either opt all existing cooling-enabled problem traits into the new backend or narrow this check so those inputs do not become unusable.
```

BenWibking:
Another LLM review:

## Potential Bugs

1. **Dropped gamma_ == 1.0 isothermal short-circuit (src/hydro/EOS.hpp)**
- EOSMicrophysics::ComputePressure and ComputeSoundSpeed no longer short-circuit for isothermal (gamma=1.0) cases
- Currently dormant (no chemistry problem uses gamma=1.0), but EOS_Traits is user-specializable
- Fix: re-add the if constexpr (gamma_ == 1.0) guard, or assert Microphysics problems must use gamma != 1.0
2. **EOSTabulated::ComputeEintTempDerivative returns 0.0 at table boundaries (src/hydro/EOS.hpp:476-494)**
- Callers divide by this value → inf/NaN source terms at clamp boundaries
- Currently dormant (no tabulated+radiation problem exists yet), but will trip the first such problem
- Fix: clamp Eint ± dEint to valid range and use one-sided difference at boundaries
3. **ElectronConduction.hpp:104-120 behavior change for ResampledCooling conduction without EOSTabulated**
- Tgas now always comes from EOS<problem_t>::ComputeTgasFromEint (gamma-law) instead of the table
- The new guard only asserts cooling.enabled == 1 ⇒ EOSTabulated, not the conduction analog
- Fix: add conduction.eos_flag=ResampledCooling ⇒ EOSTabulated assertion, or restore table lookup

src/QuokkaSimulation.hpp
```diff
@@ -659,11 +659,18 @@ template <typename problem_t> void QuokkaSimulation<problem_t>::readParmParse()
 		}
 #endif
 		if ((enableCooling_ == 1) || (alwaysReadTables == 1)) {
+			if (enableCooling_ == 1) {
+				AMREX_ALWAYS_ASSERT_WITH_MESSAGE(quokka::EOS<problem_t>::is_tabulated,
```
BenWibking: **[correctness — confirmed]** This assert aborts existing, unmigrated tests: `inputs/SN.toml:30` and `inputs/ParticleSF.toml:46` set `cooling.enabled = 1`, but the `EOS_Traits` specializations in `testSN.cpp:56` and `testParticleSF.cpp:35` define no `EOSBackend`, so `is_tabulated` is false and the registered CTests (SN, ParticleSF, ParticleSF2, ParticleSF_restart_cap*) abort at startup. These problems need to be migrated in this PR, or the guard relaxed.

src/QuokkaSimulation.hpp
```diff
@@ -659,11 +659,18 @@ template <typename problem_t> void QuokkaSimulation<problem_t>::readParmParse()
...
+								 "cooling.enabled = 1 requires the EOSTabulated EOS backend. "
+								 "Add 'using EOSBackend = quokka::EOSTabulated<YourProblem>;' "
+								 "to your EOS_Traits specialization.");
+			}
 			hpp.query("hdf5_data_file", coolingTableFilename_);
 			if (coolingTableType_ == "resampled") {
 				// read resampled cooling tables
 				amrex::Print() << "Reading resampled cooling tables...\n";
 				cooling_table_include_pe = quokka::ResampledCooling::readResampledData(coolingTableFilename_, resampledTables_);
+				quokka::ResampledCooling::registerEOSTabulated(resampledTables_.const_tables_host(), resampledTables_.const_tables());
```
BenWibking: **[correctness — confirmed]** The backend/cooling check is one-directional: registration only happens inside `if ((enableCooling_ == 1) || (alwaysReadTables == 1))`. A tabulated problem run adiabatically (`cooling.enabled = 0`, e.g. RandomBlast for a Sedov comparison) passes `readParmParse` cleanly, then `setInitialConditionsOnGrid` (`testRandomBlast.cpp:75`) calls `EOSTabulated::ComputeEintFromTgas` inside a GPU lambda and dies on "registry not registered before use!" as a cryptic in-kernel device abort. Consider adding the reverse check (`is_tabulated` requires tables read/registered) here at setup time.

src/cooling/EOSTabulatedRegistry.hpp
```diff
@@ -0,0 +1,96 @@
...
+	amrex::Real cloudy_H_mass_fraction;
+
+	// Physical specific-energy bounds (erg/g) -- used as root-finding bounds in ComputeEgasFromTgas
+	amrex::Real eint_min;
+	amrex::Real eint_max;
+};
+
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE auto ComputeTgasFromEgas(Real const rho, Real const Eint, resampledGpuConstTables const &tables) -> Real
+{
+	if (rho <= 0.0) {
```
BenWibking, line 38: **[behavior change — confirmed]** These `rho <= 0.0 -> return 0.0` guards (also lines 47 and 74) are new — the pre-PR `ResampledCooling.hpp` had no such guard and produced a loud NaN via `Eint/rho`. A vacuum/corrupted cell now silently reads T = 0, which passes every `T < threshold` cut (e.g. the ShockCloud/DiskGalaxy cold-mass statistics accumulate the garbage cell), and FPE trapping no longer catches the bad state. Prefer NaN (matching `EOSIdeal` semantics) or an assert, and let the density-floor machinery own the rho <= 0 case.

src/conduction/ElectronConduction.hpp
BenWibking, line 104: **[correctness — confirmed, config-gated]** Temperature now always comes from `quokka::EOS`, but the sound speed two lines below still uses the runtime `eos_flag` with a direct table call. On a problem without `EOSBackend = EOSTabulated`, the config `cooling.enabled = 0` + `cooling.read_tables_even_if_disabled = 1` + `conduction.eos_flag = "ResampledCooling"` passes the new guard yet silently computes T from the fixed-mu ideal gas while cs comes from the table — physically inconsistent conduction fluxes and a silent behavior change vs pre-PR. Nothing cross-checks `eos_flag` against the compile-time backend; either add that check or route the sound speed through the backend too.

src/hydro/EOS.hpp
```diff
@@ -28,442 +30,621 @@
...
+	{
+		// Compute dEint/dT with one root-find + two cheap table lookups
+		// instead of two root-finds (toms748, up to 32 iterations each).
+		const amrex::Real Eint = ComputeEintFromTgas(rho, Tgas, massScalars);
+		constexpr amrex::Real eps = 1.0e-6;
+		const amrex::Real dEint = amrex::max(eps * Eint, eps);
+		const amrex::Real T_plus = ComputeTgasFromEint(rho, Eint + dEint, massScalars);
+		const amrex::Real T_minus = ComputeTgasFromEint(rho, Eint - dEint, massScalars);
+		const amrex::Real dT = T_plus - T_minus;
+		return (dT > 0.0) ? (2.0 * dEint) / dT : 0.0;
```
BenWibking: **[correctness — latent]** Two failure modes here (only radiation currently calls this, and no migrated problem couples radiation yet — but worth fixing before someone does): (a) at/beyond table edges the clamped lookups give `dT == 0` and this returns c_v = 0, which the radiation Newton solvers divide by (`source_terms_single_group.hpp:294/296`, `radiation_dust_system.hpp:451/819`), producing inf/NaN; (b) `dEint = max(eps*Eint, eps)` has an absolute floor of 1e-6 in energy-density units — at ISM densities (Eint ~ 1e-14 erg/cm^3) the floor dwarfs Eint by ~8 orders of magnitude, both probes clamp to opposite table edges, and the returned derivative is garbage. Also: the callers already hold Eint, so an Eint-based overload would skip the `ComputeEintFromTgas` root-find entirely.

src/hydro/EOS.hpp
```diff
@@ -28,442 +30,621 @@
...
+
+	// Temperature methods — use the resampled table via the global registry
+	[[nodiscard]] AMREX_FORCE_INLINE AMREX_GPU_HOST_DEVICE static auto
+	ComputeTgasFromEint(amrex::Real rho, amrex::Real Eint, quokka::optional<amrex::GpuArray<amrex::Real, nmscalars_>> const &massScalars = {})
+	    -> amrex::Real
+	{
 		amrex::ignore_unused(massScalars);
-		cs = EOS_Traits<problem_t>::cs_isothermal;
-		return cs;
+		auto *reg = ResampledCooling::getEOSTabulatedRegistry();
```
BenWibking: **[efficiency — confirmed]** Every per-cell EOS temperature call in GPU kernels now dereferences the `AMREX_GPU_MANAGED` global registry pointer and executes two `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` checks that are *not* compiled out in Release builds; `ComputeEintTempDerivative` does this 3x per evaluation. The pre-PR pattern captured the `resampledGpuConstTables` handle by value into the kernel closure (as `ElectronConduction.hpp` still does with `*tables_dev`), keeping table metadata in kernel parameters instead of per-cell managed-memory traffic. Consider copying `reg->device` into a local at method entry and demoting the per-call asserts to debug-only `AMREX_ASSERT` — the registration invariant is already checked at setup.

src/cooling/EOSTabulatedRegistry.hpp
```diff
@@ -0,0 +1,96 @@
...
+		return 0.0;
+	}
+	const Real eint = Eint / rho;
+	std::array<amrex::Real, 2> const point = {rho, eint};
+	return tables.all_tables.interpolate_single(point, ENTROPY_IDX);
+}
+
+struct EOSTabulatedRegistry {
+	bool active = false;
```
BenWibking: **[simplification]** `active` is redundant state: `registerEOSTabulated` (`ResampledCooling.cpp:124`) is the only writer and unconditionally sets it true on the only path that makes the global pointer non-null, so `active == (reg != nullptr)` always. Dropping it removes one of the two per-call asserts in each `EOSTabulated` method. Relatedly, the registry-fetch block (getter + asserts + `AMREX_IF_ON_DEVICE/HOST` sub-handle pick) is copy-pasted verbatim three times in `EOS.hpp` (lines 454-459, 467-472, 527-532) — a single private `tables()` helper collapses ~18 duplicated lines.

src/hydro/EOS.hpp
```diff
@@ -28,442 +30,621 @@
...
+};
+
+// ==================== EOS backend selection ====================
+// If EOS_Traits<problem_t> defines EOSBackend, use it; otherwise default to EOSIdeal.
+// This preserves backward compatibility with existing full specializations that
+// predate the EOSBackend mechanism.
+
+namespace detail
+{
+template <typename T, typename = void> struct EOSBackendHelper {
```
BenWibking: **[simplification]** The CHEMISTRY/PHOTOCHEMISTRY-conditional default-backend selection is written twice: as the `EOSBackend` alias in the primary `EOS_Traits` template (lines 50-54) and again here in the helper's fallback. Both copies are live (the alias serves problems inheriting the primary trait; this fallback serves full specializations without `EOSBackend`), so they must stay in sync forever — a future default-policy change edited in only one place diverges silently. A single `DefaultEOSBackend<T>` alias referenced by both would remove the drift risk.

src/problems/RandomBlast/testRandomBlast.cpp
```diff
@@ -36,6 +36,7 @@ template <> struct Physics_Traits<RandomBlast> : DefaultPhysicsTraits {
 template <> struct quokka::EOS_Traits<RandomBlast> {
 	static constexpr double gamma = 5. / 3.;
 	static constexpr double mean_molecular_weight = C::m_u;
+	using EOSBackend = quokka::EOSTabulated<RandomBlast>;
```
BenWibking, line 43: **[behavior change — confirmed]** Adding this backend silently changes the initial conditions: the IC's `ComputeEintFromTgas(rho, Tgas0)` at line 75 used to resolve to the ideal-gas inverse with the fixed mu from `EOS_Traits<RandomBlast>`; it now root-finds through the table, changing the initial Eint/pressure field by the table-vs-fixed-mu ratio and silently clamping to `rho * eint_min` if `Tgas0` is below the table's minimum at `rho0`. Likely intended, but it changes regression answers and deserves an explicit note; the silent edge-clamp applies to any IC temperature outside the table.

src/cooling/ResampledCooling.hpp
```diff
@@ -25,23 +25,6 @@ namespace quokka::ResampledCooling
 {
```
BenWibking, line 26: **[build robustness]** This header still uses `resampledGpuConstTables` (line 36+) and `ComputeEgasFromTgas` (line 147) but no longer includes their defining header — it compiles only via the transitive chain `hydro_system.hpp -> EOS.hpp -> EOSTabulatedRegistry.hpp`. Any TU that includes `ResampledCooling.hpp` without that chain breaks. One-line fix: add a direct `#include "cooling/EOSTabulatedRegistry.hpp"`.

BenWibking (commented):
## Fable Code review (13 findings: 8 inline + 3 below)

**Overview:** The backend hierarchy design is sound and the SFINAE backward-compatibility mechanism works as advertised (verified that the explicit member specializations in RadMarshak et al. do require the forwarding wrappers in `EOS<problem_t>`). The findings below could not be anchored inline because their lines are outside the diff:

**1. `src/hydro/hydro_system.hpp:1100` — EnforceLimits temperature floor changes semantics for migrated problems (plausible).** A cell whose Eint undershoots below `rho * eint_min` interpolates a clamped `Tgas = T(eint_min)`; if that edge temperature is >= `tempFloor`, the floor branch never fires and the unphysically small Eint survives into the Riemann solver as near-zero pressure, whereas the old ideal-gas floor would have reset it. Consider flooring Eint directly against `rho * eint_min` for the tabulated backend.

**2. `src/hydro/NSCBC_inflow.hpp:194` — subsonic-inflow boundaries now root-find through the table per ghost cell, with silent edge clamping (confirmed mechanism).** For tabulated problems, `ComputeEintFromTgas(rho, T_t)` here is a toms748 root-find (up to 32 table interpolations) through the managed registry per ghost cell per fillpatch, replacing one multiply. ShockCloud's wind boundary reaches it via `setInflowX1LowerLowOrder` (`testShockCloud.cpp:187`); if the prescribed wind temperature falls outside the table range at the extrapolated boundary density, the inflow Eint silently caps at the table edge.

**3. `src/problems/ShockCloud/testShockCloud.cpp:397` — the entropy derived variable was left on the direct table call while DiskGalaxy's identical diagnostic was migrated (confirmed).** ShockCloud uses `ResampledCooling::ComputeEntropyFromRhoEint(rho, Eint, tables)` (per-instance table handle) while `testDiskGalaxy.cpp:627` goes through `EOS<DiskGalaxy>` (process-global registry). The two data paths can diverge if tables are ever re-read, and the inconsistency invites copy-paste divergence.

All confirmed verdicts were verified directly against the code on this branch and the pre-PR files at merge-base `016201f`.

src/hydro/EOS.hpp
```diff
@@ -28,442 +30,613 @@
...
-		estate.mu = mean_molecular_weight_ / C::m_u;
-		eos(eos_input_rp, estate);
-		deint_dRho = estate.dedr;
-		deint_dP = 1.0 / estate.dpde;
-		dRho_dP = 1.0 / (estate.dpdr * C::k_B / boltzmann_constant_);
-		dP_dRho_s = estate.cs * estate.cs;
-		G = estate.G;
+		eos(eos_input_rt, chemstate);
+		return chemstate.dedT * chemstate.rho;
 	}
```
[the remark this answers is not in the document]
chongchonghe: addressed

src/QuokkaSimulation.hpp
chongchonghe: done. all  problems updated to use the new API.

src/conduction/ElectronConduction.hpp
chongchonghe: already fixed (table loading gate is is_tabulated, not cooling.enabled)

src/QuokkaSimulation.hpp
chongchonghe: already fixed (table loading gate is is_tabulated, not cooling.enabled)

src/cooling/EOSTabulatedRegistry.hpp
chongchonghe: fixed

src/hydro/EOS.hpp
chongchonghe: fixed: relative-only perturbation floor (eps * Eint, no absolute eps)

src/cooling/EOSTabulatedRegistry.hpp
chongchonghe: done

src/hydro/EOS.hpp
chongchonghe: done

src/cooling/ResampledCooling.hpp
chongchonghe: done

src/hydro/EOS.hpp
chongchonghe: done

src/problems/RandomBlast/testRandomBlast.cpp
chongchonghe: done

docs/superpowers/specs/2026-06-28-unified-eos-temperature-design.md
BenWibking, line 1: Did you mean to include this file in the PR?

scripts/bash/bootstrap.sh
BenWibking, line 1: Did you mean to include this file?

src/conduction/ElectronConduction.hpp
```diff
@@ -84,6 +84,10 @@ template <typename problem_t> class ElectronConduction
 		amrex::IntVect ng = amrex::IntVect(AMREX_D_DECL(state.nGrow(), state.nGrow(), state.nGrow()));
 		std::optional<decltype(tables.const_tables())> tables_dev;
 		if (params.eos_flag == EOSFlagforConduction::ResampledCooling) {
+			AMREX_ALWAYS_ASSERT_WITH_MESSAGE(quokka::EOS<problem_t>::is_tabulated,
+							 "conduction.eos_flag = ResampledCooling requires the EOSTabulated EOS backend. "
+							 "Add 'using EOSBackend = quokka::EOSTabulated<YourProblem>;' "
+							 "to your EOS_Traits specialization.");
 			tables_dev = tables.const_tables();
 		} else if (params.eos_flag != EOSFlagforConduction::ResampledCooling && params.eos_flag != EOSFlagforConduction::EOS) {
 			amrex::Abort("Invalid eos_flag value in ElectronConduction. Must be 0 (resampled cooling) or 1 (quokka::EOS).");
```
BenWibking: I think it might be better to remove `conduction.eos_flag` and instead keep `cooling.enabled`.

docs/markdown/parameters.md
```diff
@@ -96,12 +96,10 @@ These parameters are read in the `QuokkaSimulation<problem_t>::readParmParse()`
 
 These parameters are read in the `QuokkaSimulation<problem_t>::readParmParse()` function in `src/QuokkaSimulation.hpp`.
 
-| Parameter Name                       | Type          | Default                             | Description                                                                                                                                         |
-|--------------------------------------|---------------|-------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------|
-| cooling.enabled                      | Boolean (0/1) | `0` (Disabled)                      | If set to 1, turns on optically-thin radiative cooling as a Strang-split source term.                                                               |
```
BenWibking, line 102: I don't think we should get rid of this parameter. It is still useful to be able to disable cooling for testing purposes, even when using it to calculate the temperature.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
```diff
@@ -52,6 +52,7 @@ static_assert(AMREX_SPACEDIM == 3, "DiskGalaxy problem requires AMREX_SPACEDIM =
 template <> struct quokka::EOS_Traits<DiskGalaxy> {
 	static constexpr double gamma = 5. / 3.;
 	static constexpr double mean_molecular_weight = 0.6 * C::m_u;
```
BenWibking, line 54: I think we should disallow setting `mean_molecular_weight` when EOSTabulated is used.

src/QuokkaSimulation.hpp
```diff
@@ -162,7 +162,6 @@ template <typename problem_t> class QuokkaSimulation : public AMRSimulation<prob
 							      // override real star formation rate from the simulation if non-negative
 	quokka::PeHeatingTables<> peHeatingTables_;
 
-	int enableCooling_ = 0;
```
BenWibking, line 165: I think we should keep this for testing/debugging purposes.

src/QuokkaSimulation.hpp
```diff
@@ -1092,19 +1100,11 @@ auto QuokkaSimulation<problem_t>::addStrangSplitSourcesWithBuiltin(amrex::MultiF
 	// start by assuming cooling integrator is successful.
 	bool cool_success = true;
 	auto const applyCooling = [&]() {
-		if (enableCooling_ == 1) {
+		if constexpr (quokka::EOS<problem_t>::is_tabulated) {
```
BenWibking, line 1112: I think this should be:
```
if ((enableCooling_ == 1) && (quokka::EOS<problem_t>::is_tabulated)) {
```
It's fine to use the tabulated EOS but turn off the cooling source term for testing or debugging.

BenWibking (changes requested):
Two important comments below about the runtime parameters and the compile-time parameters.

docs/superpowers/specs/2026-06-28-unified-eos-temperature-design.md
chongchonghe: sure not. I removed it.

scripts/bash/bootstrap.sh
chongchonghe: this will not show as a modified file once merging in dev branch

docs/markdown/parameters.md
chongchonghe: Good point. Done.

src/QuokkaSimulation.hpp
chongchonghe: done.

src/problems/DiskGalaxy/testDiskGalaxy.cpp
chongchonghe: I agree, but I'll leave this for a future PR because currently we need this variable to compute number density. 

When I add the new cooling module, I will make it compute mean molecular weight from the Grackle table. Currently, we don't do this, so we do need this variable for number density calculation for the ionization module and for the gas-dust coupling module.

src/conduction/ElectronConduction.hpp
chongchonghe: I removed `ResampledCooling` from ElectronConduction.hpp completely, because we always use ideal gas law to compute sound speed in the Riemann solver. That's what we should use in `ElectronConduction` as well, even if tabulated cooling is used.

chongchonghe:
@BenWibking Ready

src/QuokkaSimulation.hpp
```diff
@@ -1105,19 +1109,14 @@ auto QuokkaSimulation<problem_t>::addStrangSplitSourcesWithBuiltin(amrex::MultiF
...
+			if (enableCooling_ != 1) {
+				return;
+			}
```
BenWibking, line 1115: I think this is a somewhat confusing way to write it. Is there a reason not to put the call to `computeCooling` inside `if (enableCooling == 1) {`?

docs/markdown/parameters.md
```diff
@@ -97,12 +97,10 @@ These parameters are read in the `QuokkaSimulation<problem_t>::readParmParse()`
...
```
BenWibking, line 102: Can you restore `cooling.enabled` to the docs?

BenWibking (changes requested):
Two minor comments.

docs/markdown/parameters.md
chongchonghe: done

src/QuokkaSimulation.hpp
chongchonghe: I think this is actually a good practice to reduce the number of nested control flows or the amount of indentation.

BenWibking: I don't necessarily agree, but we can agree to disagree.

BenWibking (approved):
LGTM

PR #2020, opened 2026-06-29
feat: migrate DataTable to self-describing HDF5 group format; consolidate ResampledCooling

chongchonghe:
### Description

Migrates `quokka::DataTable` to a new self-describing HDF5 group format and rewrites `ResampledCooling` to load all five cooling outputs into a single `DataTable<2, 5>` with log–log interpolation.

**HDF5 format change:** The old reader expected three separate top-level groups (`/metadata`, `/grids`, `/data`). The new format uses named groups (`tab1`, `tab2`, …) that are each self-contained: all metadata lives as attributes on the group (`Ndim`, `Nout`, `Nx`, `xlo`, `xhi`, `spacing`, `input_names`, `output_names`, `input_units`, `output_units`) and the single `data` dataset has shape `[Nout, Nx[0], Nx[1], …]` in row-major order. Domain-specific attributes (`include_pe`, `cloudy_H_mass_fraction`) live inside `tab1` alongside the standard attributes.

**DataTable changes:**
- `H5Reader(file_path, group_name = "tab1", output_spacings = {})` replaces the old multi-parameter reader. The third argument is an `std::array<SpacingType, Nout>` defaulting to all-linear.
- Per-output spacing: `DataTableGpuConst` now holds `output_spacings[Nout]` instead of a single `output_spacing`, and `interpolate`/`interpolate_single` apply the back-transform per output index. This lets different outputs use different log transforms during interpolation.
- Output spacing is **not** stored in HDF5 files — it is a property of the interpolation declared at the C++ call site, not a property of the data.
- `initializeStorage` saves physical bounds before any log transform into `xlo_physical_` / `xhi_physical_`, exposed via `coord_xlo()` / `coord_xhi()`.
- `setMetadata` moved from private to public; signature updated to accept `std::array<SpacingType, Nout>`.
- HDF5 dataspace extents validated before any buffer read (`H5Aget_space`/`H5Dget_space`) to catch shape mismatches early.
- `H5Aread` return values checked in `ResampledCooling`.

**ResampledCooling changes:**
- All five cooling outputs (`cooling_rate`, `temperature`, `sound_speed`, `pressure`, `entropy`) packed into one `DataTable<2, 5>` loaded with a single `H5Reader` call passing `{linear, fast_log, fast_log, fast_log, fast_log}`.
- Cooling rate (index 0) interpolated linearly (can be negative). Temperature, sound speed, pressure, and entropy (indices 1–4) interpolated in log–log space: `H5Reader` applies `FastMath::inverse_pow2` at load time, and `FastMath::pow2` recovers the physical value during lookup.
- HDF5 files always contain **raw physical values** — no pre-transform in Python scripts is needed.
- `resampledGpuConstTables` reduced to `all_tables + cloudy_H_mass_fraction + eint_min/eint_max`.
- Named output index constants (`COOLING_RATE_IDX`, `TEMPERATURE_IDX`, `SOUND_SPEED_IDX`, `PRESSURE_IDX`, `ENTROPY_IDX`) replace magic indices.

**Python scripts and HDF5 files:**
- `resample_grackle_cooling_tables.py` and `resample_cloudy_cooling_tables.py` write the `tab1` format with raw physical output values (no log transform).
- All five committed `.h5` files regenerated.
- Documentation pages added for `DataTable` and the cooling module.

**Verification:** `ResampledCoolingTest` passes with relative L1 error 0.0117 < 0.02 tolerance.

### Related issues
Relates to #1243, #1373, and #1128

### Checklist


src/cooling/ResampledCooling.cpp
```diff
@@ -12,88 +12,95 @@
...
-	// Check if file exists
 	if (!amrex::FileSystem::Exists(hdf5_file)) {
 		amrex::Abort("Resampled cooling table file does not exist!");
 	}
 
-	// Define coordinate names and fast_log setting
-	const std::vector<std::string> coord_names = {"rho", "eint"};
-	const int is_fast_log = 1;
+	// Read all 5 cooling outputs from the 'tab1' group in one call
+	resampledTables.all_tables = quokka::DataTable<2, 5>::H5Reader(hdf5_file, "tab1");
```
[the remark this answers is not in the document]
chongchonghe: the code will abort, so it's safe. We currently don't use that cooling table (@BenWibking is this right?) . We will upgrade the table later if we ever need to use them.

chongchonghe:
@BenWibking @markkrumholz Ready for review.

chongchonghe:
# PR #2020 Review — feat: migrate DataTable to self-describing HDF5 group format; consolidate ResampledCooling

**Branch:** `chong/cooling/use-DataTable2`  
**PR type:** Feature/Enhancement (SPEC → VERIFY → EDGE cycle)

---

## SPEC

From the PR description and diff, the stated contract:

1. **New `H5Reader` API**: `DataTable::H5Reader(file_path, group_name="tab1")` replaces the old multi-parameter form. Reads `Ndim`, `Nout`, `Nx`, `xlo`, `xhi`, `spacing`, name/unit arrays, and the `data` dataset from the named group.
2. **Single table for 5 outputs**: `DataTable<2, 5>` loads all five cooling quantities (`cooling_rate`, `temperature`, `sound_speed`, `pressure`, `entropy`) in one `H5Reader` call.
3. **Physical bounds accessors**: `coord_xlo()` / `coord_xhi()` return pre-log-transform bounds, used by `ComputeEgasFromTgas` for root-finding bounds.
4. **Callers pass physical coordinates**: Interpolation sites pass `{rho, eint}` directly; `DataTable` applies `fast_log` transform internally.
5. **`setMetadata` made public**: Allows external population of metadata after construction.
6. **Named index constants** replace magic indices: `COOLING_RATE_IDX=0`, `TEMPERATURE_IDX=1`, etc.
7. **HDF5 files regenerated**: Five `.h5` files updated to new `tab1` format; `CloudyData_UVB=HM2012_resampled_no_PE.h5` deleted and replaced by `CloudyData_UVB=HM2012_resampled_noPE.h5`.
8. **Correctness criterion**: `ResampledCoolingTest` relative L1 error < 0.02.

---

## VERIFY / EDGE Table

|---|---|---|
| VERIFY – golden path | `quokka buildrun -d 1d ResampledCoolingTest` on PR branch | Build succeeded; L1 error **0.01217 < 0.02** ✓ |
| VERIFY – HDF5 format | Python `h5py` inspect of new `CloudyData_UVB=HM2012_resampled.h5` | `tab1` group present; `Ndim=2`, `Nout=5`, `Nx=[30,200]`, `spacing=['fast_log','fast_log']`, data shape `(5,30,200)` — matches spec |
| VERIFY – `cloudy_H_mass_fraction` | Compare HDF5 attr (0.7157683773530885) vs old constant `1/(1+0.1*3.971)` | Bit-identical ✓ |
| EDGE – shielded table | `ResampledCoolingTest` with `CloudyData_UVB=HM2012_shielded_resampled.h5` | File loads and simulation completes without error; L1 differs from unshielded reference as expected (different physics) ✓ |
| EDGE – old filename removed | `grep "resampled_no_PE" inputs/ src/` | Zero hits — all four `toml` files now point to `noPE.h5` ✓ |

---

## Feature Assessment

The implementation is correct and matches the stated design:

- **`flatDataSize` includes `Nout`** (`DataTable.hpp:517`): returns `Nout * N_0 * N_1`, so the flat read buffer is correctly sized for a `[5, 30, 200]` dataset.
- **Physical bounds saved before log transform** (`DataTable.hpp:803-804` then `+817-818`): `xlo_physical_` / `xhi_physical_` are assigned from `x_mins` / `x_maxs` before `initializeStorage` overwrites `coord_min_` / `coord_max_` with their log-transformed equivalents. `coord_xlo()[1]` correctly provides physical `eint_min` for root finding in `ComputeEgasFromTgas`.
- **`interpolate_single` handles coordinate transform** (`DataTable.hpp:243-254`): Physical coordinates are transformed to fast-log space internally, so callers are correct to drop the `FastMath::fastlg()` calls.
- **Broadcast is complete**: The new `H5Reader` broadcasts `sizes`, `x_mins`, `x_maxs`, `spacing_types`, all name/unit string arrays, and `flat_data` to non-IO ranks. The extra domain-specific attributes (`include_pe`, `cloudy_H_mass_fraction`) are separately broadcast in `readResampledData`. No data is left IO-only.

---

## PR-Introduced Concerns

**1. Orphaned old-format HDF5 file breaks three input files (not CI-blocking, but will abort on manual runs)**

`extern/cooling/isrf_1000Go_grains_resampled.h5` was not converted to the new format. It still has the old layout (`/metadata`, `/grids`, `/data` groups) rather than a `tab1` group. Three input files reference it with `cooling_table_type = "resampled"`:

- `inputs/ShockCloud_32.toml`
- `inputs/ShockCloud_64.toml`
- `inputs/TabulatedCooling.toml`

After this PR, any attempt to run these inputs will hit:
```
AMREX_ALWAYS_ASSERT_WITH_MESSAGE: Failed to open HDF5 group: tab1
```

`ShockCloud` has `ADD_TEST OFF` so it is not in the CI test suite, and `TabulatedCooling` has no CMake target. These failures would only appear on manual runs, but they are surprising for anyone trying to use these inputs. **Recommendation:** either convert `isrf_1000Go_grains_resampled.h5` to the new format (preferred) or add a compatibility note in the input files.

**2. Double file open in `readResampledData`**

`H5Reader` opens and closes the HDF5 file internally. Immediately after, `readResampledData` (`ResampledCooling.cpp:476-504`) reopens the same file a second time to read `include_pe` and `cloudy_H_mass_fraction`. Both are correctness-neutral, but it is slightly wasteful and could confuse a reader. Consider passing these as output parameters from `H5Reader` (via a struct or an `extra_attrs` callback) in a follow-up.

**3. `include_pe` type mismatch (harmless, but fragile)**

The Python script writes `include_pe` as an `int64` attribute (`h5py` default for Python `int`). The C++ reader requests `H5T_NATIVE_INT` (32-bit). HDF5 performs automatic type conversion, so the read succeeds, but it would silently truncate if the attribute were ever stored as a non-integer type. Storing and reading as `H5T_NATIVE_INT` on both sides, or using `H5T_NATIVE_INT64`, would be cleaner.

**4. Stray `/// placeholder` comment**

`src/problems/ResampledCoolingTest/testResampledCoolingTest.cpp:11`:
```cpp
/// placeholder
```
This appears to be an accidentally committed placeholder docstring. Should be removed before merge.

---

None found that are independent of this PR.

---

chongchonghe:
> ## PR-Introduced Concerns
> **1. Orphaned old-format HDF5 file breaks three input files (not CI-blocking, but will abort on manual runs)**
> 
> `extern/cooling/isrf_1000Go_grains_resampled.h5` was not converted to the new format. It still has the old layout (`/metadata`, `/grids`, `/data` groups) rather than a `tab1` group. Three input files reference it with `cooling_table_type = "resampled"`:
> 
> * `inputs/ShockCloud_32.toml`
> * `inputs/ShockCloud_64.toml`
> * `inputs/TabulatedCooling.toml`
> 
> After this PR, any attempt to run these inputs will hit:
> 
> ```
> AMREX_ALWAYS_ASSERT_WITH_MESSAGE: Failed to open HDF5 group: tab1
> ```
> 
> `ShockCloud` has `ADD_TEST OFF` so it is not in the CI test suite, and `TabulatedCooling` has no CMake target. These failures would only appear on manual runs, but they are surprising for anyone trying to use these inputs. **Recommendation:** either convert `isrf_1000Go_grains_resampled.h5` to the new format (preferred) or add a compatibility note in the input files.

@BenWibking Do you use this table? Is it worth addressing? 

> **2. Double file open in `readResampledData`**
> 
> `H5Reader` opens and closes the HDF5 file internally. Immediately after, `readResampledData` (`ResampledCooling.cpp:476-504`) reopens the same file a second time to read `include_pe` and `cloudy_H_mass_fraction`. Both are correctness-neutral, but it is slightly wasteful and could confuse a reader. Consider passing these as output parameters from `H5Reader` (via a struct or an `extra_attrs` callback) in a follow-up.

This is intentional: not all tables have additional attributes. 

> **3. `include_pe` type mismatch (harmless, but fragile)**
> 
> The Python script writes `include_pe` as an `int64` attribute (`h5py` default for Python `int`). The C++ reader requests `H5T_NATIVE_INT` (32-bit). HDF5 performs automatic type conversion, so the read succeeds, but it would silently truncate if the attribute were ever stored as a non-integer type. Storing and reading as `H5T_NATIVE_INT` on both sides, or using `H5T_NATIVE_INT64`, would be cleaner.

Done. 

> **4. Stray `/// placeholder` comment**
> 
> `src/problems/ResampledCoolingTest/testResampledCoolingTest.cpp:11`:
> 
> ```c++
> /// placeholder
> ```
> 
> This appears to be an accidentally committed placeholder docstring. Should be removed before merge.

Will remove before merge.

BenWibking:
> > ## PR-Introduced Concerns
> > **1. Orphaned old-format HDF5 file breaks three input files (not CI-blocking, but will abort on manual runs)**
> > `extern/cooling/isrf_1000Go_grains_resampled.h5` was not converted to the new format. It still has the old layout (`/metadata`, `/grids`, `/data` groups) rather than a `tab1` group. Three input files reference it with `cooling_table_type = "resampled"`:
> > 
> > * `inputs/ShockCloud_32.toml`
> > * `inputs/ShockCloud_64.toml`
> > * `inputs/TabulatedCooling.toml`
> > 
> > After this PR, any attempt to run these inputs will hit:
> > ```
> > AMREX_ALWAYS_ASSERT_WITH_MESSAGE: Failed to open HDF5 group: tab1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `ShockCloud` has `ADD_TEST OFF` so it is not in the CI test suite, and `TabulatedCooling` has no CMake target. These failures would only appear on manual runs, but they are surprising for anyone trying to use these inputs. **Recommendation:** either convert `isrf_1000Go_grains_resampled.h5` to the new format (preferred) or add a compatibility note in the input files.
> 
> @BenWibking Do you use this table? Is it worth addressing?

Yes, I do use it. Please convert it.

src/cooling/ResampledCooling.cpp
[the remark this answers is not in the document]
BenWibking: I do use this table, so please convert it if you can.

chongchonghe:
Converted `extern/cooling/isrf_1000Go_grains_resampled.h5` from the old format to the new `tab1` group format using `scripts/python/convert_cooling_table_hdf5.py`. The script reads the old `/metadata`, `/grids`, `/data` layout and writes the self-describing group format, preserving metadata attributes from the original file as-is; `include_pe` defaulted to 0 since the attribute was absent from this file.

**Correctness verification:** Checked out `CloudyData_UVB=HM2012_resampled.h5` from the `development` branch (old format), applied the conversion script, and ran `ResampledCoolingTest` against the converted file. It passes (L1 error 0.012 < 0.02 tolerance), confirming the conversion script is strictly correct.

**Runtime test:** `ShockCloud` builds and runs 1 step without error, and `quokka::ResampledCooling::computeCooling()` appears in the profiler output, confirming the ISRF table is loaded and used. 

scripts/bash/convert_cooling_table_hdf5.py
```diff
@@ -0,0 +1,120 @@
...
+
+Usage:
+    python3 convert_cooling_table_hdf5.py file.h5 [file2.h5 ...]
+    python3 convert_cooling_table_hdf5.py file.h5 --output converted.h5
+    python3 convert_cooling_table_hdf5.py file.h5 --include_pe 1
+"""
+
+import argparse
+import os
+import sys
```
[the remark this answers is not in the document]
chongchonghe: done

chongchonghe:
@BenWibking ready

BenWibking:
@coderabbitai review

BenWibking:
Codex says:
```
  - [P2] Update remaining cooling utilities to the new HDF5 layout — /Users/benwibking/amrex_codes/quokka/extern/cooling/
    resample_grackle_cooling_tables.py:354-354
    When the regenerated/committed tables are used by existing Python helpers such as extern/cooling/
    integrate_cooling_zone.py::load_resampled_cooling_tables or test_cooling_onezone.py, those helpers still open /grids/
    fast_log_rho, /data/cooling_rates, and /metadata, but this writer now emits only /tab1/data and /tab1/grids/{rho,eint}. Those
    scripts now fail immediately on the updated HDF5 files, so please migrate the remaining Python readers or keep compatibility
    datasets with the format change.
```

This looks right to me.

chongchonghe:
@BenWibking All done.

BenWibking:
Ok, will take a look tomorrow.

extern/cooling/resample_cloudy_cooling_tables.py
```diff
@@ -351,52 +351,39 @@ def resample_cloudy_cooling_tables(cloudy_file, n_rho=100, n_eint=100,
...
+                          np.array(['rho', 'eint'], dtype='S'))
+        tab1.attrs.create('output_names',
+                          np.array(['cooling_rate', 'temperature', 'sound_speed',
+                                    'pressure', 'entropy'], dtype='S'))
+        tab1.attrs.create('input_units',
+                          np.array(['g/cm^3', 'erg/g'], dtype='S'))
+        tab1.attrs.create('output_units',
+                          np.array(['erg/cm^3/s/(g/cm^3)^2', 'K', 'cm/s',
+                                    'dyne/cm^2', 'erg*cm^2'], dtype='S'))
+        tab1.attrs['cloudy_H_mass_fraction'] = cloudy_H_mass_fraction
```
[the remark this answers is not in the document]
chongchonghe: we may want to remove cloudy table if we are not going to use it.

docs/superpowers/specs/2026-06-29-datatable-cooling-design.md
```diff
@@ -0,0 +1,204 @@
...
+    Nout            int           number of output quantities
+    input_names     str[Ndim]     e.g. ['rho', 'eint']
+    output_names    str[Nout]     e.g. ['cooling_rate', 'temperature', ...]
+    input_units     str[Ndim]     e.g. ['g/cm^3', 'erg/g']
+    output_units    str[Nout]     e.g. ['erg/cm^3/s/(g/cm^3)^2', 'K', ...]
+    xlo             float[Ndim]   lower bounds in physical units (not log)
+    xhi             float[Ndim]   upper bounds in physical units (not log)
+    spacing         str[Ndim]     per-dimension: 'linear', 'log', or 'fast_log'
+    [extra attrs]                 domain-specific (e.g. include_pe, cloudy_H_mass_fraction)
+  /data             dataset       shape [Nout, Nx[0], Nx[1], ...], row-major, physical values
```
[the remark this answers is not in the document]
chongchonghe: The data file always store the raw table value. the DataTable class decides what spacing to store the output at when reading the data table. Therefore, the data file itself should not store output_spacing_ .

extern/cooling/integrate_cooling_zone.py
```diff
@@ -14,33 +14,33 @@
...
+        'rho':  rho_grid,
+        'eint': eint_grid,
+        'fast_log_rho':  fast_log2(rho_grid),
+        'fast_log_eint': fast_log2(eint_grid),
+        'cooling_rates': all_data[0],
+        'temperatures':  all_data[1],
+        'sound_speeds':  all_data[2],
+        'pressures':     all_data[3],
+        'entropies':     all_data[4],
+        'metadata':      metadata,
```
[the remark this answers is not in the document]
chongchonghe: done.

BenWibking:
@chongchonghe Let me know when you're done making changes, and then I'll review it.

chongchonghe:
@BenWibking All tests passed. I just cleaned up design docs. Ready for review.

docs/markdown/cooling_module.md
```diff
@@ -0,0 +1,165 @@
...
+
+Available table files:
+
+| File | UV background | Photoelectric heating |
+|------|---------------|-----------------------|
+| `CloudyData_UVB=HM2012_resampled.h5` | HM2012 | included |
+| `CloudyData_UVB=HM2012_resampled_noPE.h5` | HM2012 | excluded |
+| `CloudyData_UVB=HM2012_shielded_resampled.h5` | HM2012 (shielded) | included |
+| `CloudyData_UVB=HM2012_shielded_resampled_noPE.h5` | HM2012 (shielded) | excluded |
+| `isrf_1000Go_grains_resampled.h5` | ISRF 1000 G₀ + grains | not applicable |
```
BenWibking: PE heating is included in this table.

docs/markdown/cooling_module.md
```diff
@@ -0,0 +1,165 @@
...
+
+To regenerate the Grackle-based tables, run `extern/cooling/resample_grackle_cooling_tables.sh`. To regenerate `isrf_1000Go_grains_resampled.h5`, run `extern/cooling/resample_cloudy_cooling_tables.py` directly.
+
+**Note:** The `output_spacing` metadata is intentionally absent from the HDF5 files. Output spacing is a property of the interpolation, not of the data; it is declared at the C++ call site in `ResampledCooling.cpp`. HDF5 files always contain raw physical values. When `H5Reader` loads a `fast_log` output, it applies `FastMath::inverse_pow2` (Newton iteration) to each value at load time, so the internal buffer stores the interpolation-space representation that `FastMath::pow2` inverts during lookup.
```
BenWibking: Ok, that may be true for the entries within the table, but the spacing of the table axes themselves has to be determined when creating the HDF5 files. Uniform linear spacing in rho and e produces a different HDF5 table than uniform spacing in fastlog_2.

So I think this is not actually a spacing, but a transform applied to the entries within the table before/after interpolation. I would call it something like `output_transform` instead of `output_spacing`.

I would also define the fastlog_2 function here or link to the Doxygen documentation, so users can construct tables with abscissas corresponding to uniform spacing in fastlog_2.

docs/markdown/cooling_module.md
```diff
@@ -0,0 +1,165 @@
...
+- **Attributes** on `tab1`:
+
+  | Attribute | Type | Description |
+  |-----------|------|-------------|
+  | `Ndim` | int32 | Number of input dimensions (2) |
+  | `Nout` | int32 | Number of output quantities (5) |
+  | `Nx` | int32[2] | Grid sizes `[n_rho, n_eint]` |
+  | `xlo` | float64[2] | Physical lower bounds `[rho_min, eint_min]` |
+  | `xhi` | float64[2] | Physical upper bounds `[rho_max, eint_max]` |
+  | `spacing` | bytes[2] | Coordinate spacing type (`"fast_log"` for both) |
```
BenWibking: I think the Type should be "string"?

This is the actual spacing parameter.

docs/markdown/cooling_module.md
```diff
@@ -0,0 +1,165 @@
...
+  | `input_names` | bytes[2] | `["rho", "eint"]` |
```
BenWibking: should this be "string[2]"?

docs/markdown/cooling_module.md
```diff
@@ -0,0 +1,165 @@
...
+  | `output_names` | bytes[5] | Names for the five outputs |
```
BenWibking: string[5]?

docs/markdown/cooling_module.md
```diff
@@ -0,0 +1,165 @@
...
+  | `input_units` | bytes[2] | Physical units of inputs |
```
BenWibking: string[2]?

docs/markdown/cooling_module.md
```diff
@@ -0,0 +1,165 @@
...
+  | `output_units` | bytes[5] | Physical units of outputs |
```
BenWibking: string[5]?

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+Ndim
+Nx[0], Nx[1], ...
+Nout
+input_name_0, input_name_1, ...
+output_name_0, output_name_1, ...
+input_unit_0, input_unit_1, ...
+output_unit_0, output_unit_1, ...
+xlo[0], xlo[1], ...
+xhi[0], xhi[1], ...
+spacing[0], spacing[1], ...   # linear / log / fast_log
```
BenWibking, line 46: This spacing is stored in the file. Is this the same as `SpacingType::linear`?

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+
+Each input dimension has an independent spacing type, set per-dimension when the table is constructed or read:
+
+| Spacing | Stored coordinate | Use case |
+|---------|-------------------|----------|
+| `SpacingType::linear` | physical value | uniformly spaced grids |
+| `SpacingType::log` | natural log | log-spaced grids (standard) |
+| `SpacingType::fast_log` | `FastMath::lg(x)` (bit-level log₂ approx) | hot paths on GPU |
+
+Callers always pass **physical** values to `interpolate`; the spacing transform is handled internally.
```
BenWibking, line 25: Is this actually a transform applied to the table entries themselves, not to the coordinates, e.g. rho and e?

See my earlier comment about the "spacing_type" in the cooling module documentation.

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+### On the host
+
+```cpp
+auto host_table = my_datatable.const_tables_host();  // pinned-memory view
+auto vals = host_table.interpolate(point);
+```
+
+## Out-of-bounds behavior
+
+With the default `OutOfBounds::clamp` policy, coordinates outside `[xlo, xhi]` are silently clamped to the nearest boundary before interpolation (nearest-neighbor extrapolation). With `OutOfBounds::fail`, the code aborts via `AMREX_ALWAYS_ASSERT_WITH_MESSAGE`.
```
BenWibking: Note that `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` does not trigger on GPU.

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+
+## MPI
+
+All data is read on the I/O processor and broadcast automatically to non-I/O ranks by `H5Reader` and `CSVReader`. No special MPI handling is needed at the call site.
+
+## Output spacing
```
BenWibking: I think there is a nomenclature confusion here. I think this should be called an output _transform_, not an output spacing. See my earlier comments.

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+|---------|-------------|-----------------|--------------------------|
+| `linear` | physical value | physical value | buffer value |
+| `fast_log` | physical value | `FastMath::inverse_pow2(physical)` | `FastMath::pow2(buffer)` |
+| `log` | physical value | `ln(physical)` | `exp(buffer)` |
+
+HDF5 files always store **raw physical values** regardless of spacing. When `H5Reader` loads a `fast_log` output, it applies `FastMath::inverse_pow2` (Newton iteration) element-wise at load time so that subsequent bilinear interpolation happens in log space. No transform is needed in the Python table-generation scripts. See [Cooling module](cooling_module.md) for a concrete example.
+
+### Why `fast_log` is as accurate as a true log–exp pair
+
+`fast_pow2` is a ~10% approximation to `2^x`, and a naive `fast_log2` is a ~10% approximation to `log₂(x)`. Their naive composition is **not** the identity: `fast_pow2(fast_log2(q)) ≠ q`. Storing `fast_log2(q)` in the table would leave a ~10% error at every grid point.
```
BenWibking: This is not correct. The round-trip inverse error only happens at localized points, and it has nothing to do with how good each function approximates 2^x or log_2(x).

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+
+`FastMath::inverse_pow2` is different: it uses Newton iteration to find `v` such that `fast_pow2(v) = q` **to machine precision**. It is the exact mathematical inverse of `fast_pow2`, not an approximation to log₂. This has two consequences:
+
+1. **Grid points are exact.** The buffer stores `inverse_pow2(q_i)`. At query time, `fast_pow2(inverse_pow2(q_i)) = q_i` to machine precision — no approximation error at all.
+
+2. **Between grid points, accuracy is determined by table resolution, not by the ~10% deviation of `fast_pow2` from `2^x`.** Linear interpolation in `inverse_pow2`-space followed by `fast_pow2` is geometrically equivalent to log-space interpolation, with second-order error `O(h²)` — the same as a true log–exp pair.
```
BenWibking: I'm not sure what "geometrically equivalent" means here. It is not equivalent. Fastlog is a linear interpolation between integer-valued floating-point exponents.

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+
+The underlying principle is that any smooth monotone bijection can be used for interpolation with full accuracy, as long as the exact forward and inverse transforms are applied consistently. Here `(inverse_pow2, fast_pow2)` form that exact bijection by construction. The 10% approximation error of `fast_pow2` relative to `2^x` is irrelevant because both directions are consistent with each other.
```
BenWibking: I agree with this paragraph. I think the earlier paragraphs are not mathematically correct.

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Key API
+
+| Method | Description |
+|--------|-------------|
+| `H5Reader(file, group, output_spacings)` | Static factory: read from HDF5 group. `output_spacings` defaults to all `linear`. |
```
BenWibking: output_spacing -> output_transform. See earlier comments.

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+| `CSVReader(file, output_spacing)` | Static factory: read from CSV file; same spacing applied to all outputs. |
```
BenWibking: output_spacing -> output_transform.

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+#include "util/DataTable.hpp"
+
+// Cooling rate (index 0) is linear; T/cs/P/S (indices 1-4) are fast_log.
+quokka::DataTable<2, 5> tbl = quokka::DataTable<2, 5>::H5Reader(
+    "extern/cooling/CloudyData_UVB=HM2012_resampled.h5", "tab1",
+    {quokka::SpacingType::linear,   // cooling rate — can be negative
+     quokka::SpacingType::fast_log, // temperature
+     quokka::SpacingType::fast_log, // sound speed
+     quokka::SpacingType::fast_log, // pressure
+     quokka::SpacingType::fast_log  // entropy
```
BenWibking: Maybe `quokka::TransformType` instead?

docs/markdown/data_table.md
```diff
@@ -0,0 +1,192 @@
...
+    });
+
+// Pass the GPU view to a kernel
+auto gpu_tbl = tbl.const_tables();
+amrex::ParallelFor(box, [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept {
+    std::array<amrex::Real, 2> pt = {rho(i,j,k), eint(i,j,k)};
+    auto vals = gpu_tbl.interpolate(pt);
+    // vals[0] = cooling rate (raw), vals[1] = T (K), vals[2] = cs (cm/s), ...
```
BenWibking, line 217: It would be good to also provide an `interpolate_single` example here.

src/cooling/ResampledCooling.cpp
```diff
@@ -12,88 +12,100 @@
...
+	const amrex::Real rho_max = resampledTables.all_tables.coord_xhi()[0];
+	const amrex::Real eint_min = resampledTables.all_tables.coord_xlo()[1];
+	const amrex::Real eint_max = resampledTables.all_tables.coord_xhi()[1];
 
-	amrex::Print() << std::format("\tDensity range: {} to {} g/cm^3 ({} steps).\n", resampledTables.rho_min, resampledTables.rho_max, n_rho);
-	amrex::Print() << std::format("\tSpecific energy range: {} to {} erg/g ({} steps).\n", resampledTables.eint_min, resampledTables.eint_max, n_eint);
-	amrex::Print() << std::format("\tPhotoelectric heating: {}.\n", is_pe_enabled ? "enabled" : "disabled");
+	amrex::Print() << std::format("\tDensity range: {} to {} g/cm^3 ({} steps).\n", rho_min, rho_max, n_rho);
+	amrex::Print() << std::format("\tSpecific energy range: {} to {} erg/g ({} steps).\n", eint_min, eint_max, n_eint);
+	amrex::Print() << std::format("\tPhotoelectric heating: {}.\n", resampledTables.include_pe ? "enabled" : "disabled");
```
BenWibking: Maybe this can be changed to `? "included in table" : "NOT included in table"`

src/cooling/ResampledCooling.hpp
```diff
@@ -25,69 +24,52 @@
 namespace quokka::ResampledCooling
 {
 
+// Output indices into the DataTable<2, 5> for the five cooling quantities
+constexpr int COOLING_RATE_IDX = 0;
+constexpr int TEMPERATURE_IDX = 1;
+constexpr int SOUND_SPEED_IDX = 2;
+constexpr int PRESSURE_IDX = 3;
+constexpr int ENTROPY_IDX = 4;
```
BenWibking, line 32: Can you use an enum instead?

src/problems/ResampledCoolingTest/testResampledCoolingTest.cpp
```diff
@@ -8,6 +8,7 @@
 /// \file testResampledCoolingTest.cpp
 /// \brief Defines a test problem for cooling integrator accuracy using the ResampledCooling module.
 ///
+/// placeholder
```
BenWibking: Can you delete this?

src/util/DataTable.hpp
```diff
@@ -66,8 +67,8 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> s
 
 	std::array<int, Ndim> sizes{};
 
-	// Output spacing for return values: linear, log, or fast_log
-	SpacingType output_spacing = SpacingType::linear;
+	// Per-output spacing for return values: linear, log, or fast_log
+	std::array<SpacingType, Nout> output_spacings{};
```
BenWibking: output_transforms?

src/util/DataTable.hpp
```diff
@@ -197,13 +198,11 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> s
 		// Part 2: Perform n-dimensional interpolation for all outputs
 		auto values = interpolate_from_indices(interp);
 
-		// Part 3: Convert from log space if output values are stored in log
-		if (output_spacing == SpacingType::fast_log) {
-			for (int i = 0; i < Nout; ++i) {
+		// Part 3: Convert from log space per output
+		for (int i = 0; i < Nout; ++i) {
+			if (output_spacings[i] == SpacingType::fast_log) {
```
BenWibking: output_transforms

src/util/DataTable.hpp
```diff
@@ -197,13 +198,11 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> s
...
 				values[i] = FastMath::pow2(values[i]);
-			}
-		} else if (output_spacing == SpacingType::log) {
-			for (int i = 0; i < Nout; ++i) {
+			} else if (output_spacings[i] == SpacingType::log) {
```
BenWibking: same

src/util/DataTable.hpp
```diff
@@ -267,10 +266,10 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> s
 		// Part 2: Perform n-dimensional interpolation for single output
 		amrex::Real value = interpolate_single_from_indices(interp, output_index);
 
-		// Part 3: Convert from log space if output values are stored in log
-		if (output_spacing == SpacingType::fast_log) {
+		// Part 3: Convert from log space for this output
+		if (output_spacings[output_index] == SpacingType::fast_log) {
```
BenWibking: same

src/util/DataTable.hpp
```diff
@@ -267,10 +266,10 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> s
...
 			value = FastMath::pow2(value);
-		} else if (output_spacing == SpacingType::log) {
+		} else if (output_spacings[output_index] == SpacingType::log) {
```
BenWibking: same

src/util/DataTable.hpp
```diff
@@ -452,8 +453,8 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> c
 	std::array<std::string, Ndim> input_units_{};
 	std::array<std::string, Nout> output_units_{};
 
-	// Output spacing type: "linear" or "fast_log"
-	SpacingType output_spacing_ = SpacingType::linear;
+	// Per-output spacing type: linear, log, or fast_log
+	std::array<SpacingType, Nout> output_spacings_{};
```
BenWibking: same

src/util/DataTable.hpp
```diff
@@ -538,17 +539,18 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> c
...
-			 const std::array<std::string, Ndim> &input_units, const std::array<std::string, Nout> &output_units, SpacingType output_spacing)
+			 const std::array<std::string, Ndim> &input_units, const std::array<std::string, Nout> &output_units,
+			 std::array<SpacingType, Nout> output_spacings)
 	{
 		input_names_ = input_names;
 		output_names_ = output_names;
 		input_units_ = input_units;
 		output_units_ = output_units;
-		output_spacing_ = output_spacing;
+		output_spacings_ = output_spacings;
```
BenWibking: same

src/util/DataTable.hpp
```diff
@@ -538,17 +539,18 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> c
 		return index;
 	}
 
+      public:
 	void setMetadata(const std::array<std::string, Ndim> &input_names, const std::array<std::string, Nout> &output_names,
-			 const std::array<std::string, Ndim> &input_units, const std::array<std::string, Nout> &output_units, SpacingType output_spacing)
+			 const std::array<std::string, Ndim> &input_units, const std::array<std::string, Nout> &output_units,
+			 std::array<SpacingType, Nout> output_spacings)
```
BenWibking: same

src/util/DataTable.hpp
```diff
@@ -727,13 +729,13 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> c
...
-		    dcoord_,	    // dcoord array
-		    sizes_,	    // sizes array
-		    output_spacing_ // output spacing (converted to enum)
+		    data_tables,     // array of data tables
+		    coord_min_,	     // coord_min array
+		    coord_max_,	     // coord_max array
+		    spacing_types_,  // spacing types array (converted to enum)
+		    dcoord_,	     // dcoord array
+		    sizes_,	     // sizes array
+		    output_spacings_ // per-output spacing
```
BenWibking: same

src/util/DataTable.hpp
```diff
@@ -756,7 +758,7 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> c
 		}
 
 		DataTableGpuConst<Ndim, Nout, oob_policy> tables{coord_tables,	 data_tables, coord_min_, coord_max_,
-								 spacing_types_, dcoord_,     sizes_,	  output_spacing_};
+								 spacing_types_, dcoord_,     sizes_,	  output_spacings_};
```
BenWibking: same

src/util/DataTable.hpp
```diff
@@ -1332,144 +1340,257 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> c
 
 		DataTable table;
 		table.initializeCommonFlat(x_mins, x_maxs, sizes, spacing_types_enum, flat_data);
-		table.setMetadata(input_names, output_names, input_units, output_units, output_spacing_bcast);
+		std::array<SpacingType, Nout> output_spacings_arr{};
+		output_spacings_arr.fill(output_spacing_bcast);
```
BenWibking: same

src/util/DataTable.hpp
```diff
@@ -1332,144 +1340,257 @@ template <int Ndim, int Nout = 1, OutOfBounds oob_policy = OutOfBounds::clamp> c
...
+				// Apply per-output transform (inverse_pow2 for fast_log, log for log).
+				// HDF5 stores raw physical values; the transform converts to interpolation space.
+				flat_data.resize(static_cast<std::size_t>(total));
+				amrex::Long const spatial_size = total / Nout;
+				for (amrex::Long k = 0; k < total; ++k) {
+					auto const val = static_cast<amrex::Real>(raw_data[static_cast<std::size_t>(k)]);
+					int const out_idx = static_cast<int>(k / spatial_size);
+					if (output_spacings[static_cast<std::size_t>(out_idx)] == SpacingType::fast_log) {
+						flat_data[static_cast<std::size_t>(k)] = FastMath::inverse_pow2(val);
+					} else if (output_spacings[static_cast<std::size_t>(out_idx)] == SpacingType::log) {
```
BenWibking: same

BenWibking (changes requested):
I think the overall design is good.

The main issue I see is that there is a nomenclature overload in the use of "output spacings". I think the C++ compile-time "output spacings" are actually output transforms, since they have nothing to do with the spacing of the table axes themselves, but instead transform/un-transform the table entries before and after interpolation. So I think in most places "spacing" should be replaced with "transform". I've added comments everywhere I noticed this.

I think there are a few mathematically incorrect or confusing statements about fastlog2. The part about it being a monotone smooth bijective transform is correct and the important thing. I think the level to which is approximates 2^x or log_2(x) is not important and is not relevant to why Newton's method is used to obtain a machine-precision-accurate transform composition.

There are also a few places that are labeled as "bytes" but I think should be "string", or should be expanded to "fixed-length byte string".

BenWibking:
Codex found a minor issue with the old Python scripts:
```
• The core C++ target builds, but the changed Python table loader breaks existing one-zone cooling scripts that still expect the
  legacy metadata keys. This is an actionable regression in committed utilities.

  Review comment:

  - [P2] Preserve cooling metadata keys for one-zone scripts — /Users/benwibking/amrex_codes/quokka/extern/cooling/
    integrate_cooling_zone.py:30-30
    With the new tab1 group, metadata now contains attrs like xlo/xhi instead of the legacy rho_min, rho_max, eint_min, and eint_max
    keys. The existing callers in extern/cooling/test_cooling_onezone.py and test_cloudy_cooling_onezone.py still read
    tables['metadata']['rho_min'] immediately after load_resampled_cooling_tables(...), so those scripts now raise KeyError as soon
    as they load a converted table. Please add compatibility aliases here or update both callers.
```

docs/markdown/data_table.md
chongchonghe: Yes, they are the spacing of the grids (input) and they need to be stored in the file.

docs/markdown/data_table.md
chongchonghe: yes, you have to pass `fast_log(n)` to the table, not  `n`. These are real spacing of the grids, not transformation of the data values.

docs/markdown/data_table.md
chongchonghe: added '... on CPUs'

docs/markdown/data_table.md
chongchonghe: Well, the round-trip inverse error is the sum/product of the error of the two functions. However, I think the 10% approximation is not an accurate statement. It is the maximum error, not an average error, so I removed that.

docs/markdown/data_table.md
chongchonghe: I have rephrased it. See if that makes sense to you.
What I mean here is that it has the same geometrical effect on transforming the data values, and it works equally well as a exp-log pair.

docs/markdown/data_table.md
chongchonghe: Since this is used in the input spacing, the grid spacing, and output transfer, I think I'll leave it as SpacingType.

docs/markdown/data_table.md
BenWibking: No, I don't think the round-trip error (how much it differs from the original value) is the combination of the error of the two functions with respect to how well they approximate 2^x or log_2(x).

docs/markdown/data_table.md
chongchonghe: done

src/cooling/ResampledCooling.cpp
chongchonghe: done

src/util/DataTable.hpp
chongchonghe: done

docs/markdown/data_table.md
BenWibking: I don't understand this comment. What is `n`?

docs/markdown/data_table.md
BenWibking: Can you use a generic term instead? The issue I see is that users may confuse this as something that *only* applies to axis spacing, rather than something that applies to axis spacing *and* the pre/post-interpolation transform.

What do you mean by "input spacing"?

docs/markdown/data_table.md
BenWibking: Did you push the new version? I don't see any changes to this section.

docs/markdown/cooling_module.md
chongchonghe: done.

chongchonghe:
@BenWibking All addressed.

docs/markdown/data_table.md
```diff
@@ -0,0 +1,245 @@
...
+|-----------|-------------|-----------------|--------------------------|
+| `linear` | physical value | physical value | buffer value |
+| `fast_log` | physical value | `FastMath::inverse_pow2(physical)` | `FastMath::pow2(buffer)` |
+| `log` | physical value | `ln(physical)` | `exp(buffer)` |
+
+HDF5 files always store **raw physical values** regardless of transform. When `H5Reader` loads a `fast_log` output, it applies `FastMath::inverse_pow2` (Newton iteration) element-wise at load time so that subsequent bilinear interpolation happens in log space. No transform is needed in the Python table-generation scripts. See [Cooling module](cooling_module.md) for a concrete example.
+
+### Why `fast_log` is as accurate as a true log–exp pair
+
+`fast_pow2` and naive `fast_log2` are each a few-percent approximation to `2^x` and `log₂(x)`, respectively. Their direct composition is **not** the identity: `fast_pow2(fast_log2(q)) ≠ q`. So if the table stored `fast_log2(q)`, each grid point would carry a noticeable reconstruction error.
```
BenWibking: The direct composition is the identity in infinite precision arithmetic, but due to rounding error, there is a catastrophic loss of precision just below each value of 2^n.

docs/markdown/data_table.md
BenWibking: @chongchonghe Can you address this?

docs/markdown/data_table.md
BenWibking: @chongchonghe Can you address this?

BenWibking (changes requested):
I think there is one additional clarification needed in the docs (see comment below).

There are also two comments from the last review that weren't fully addressed.

docs/markdown/data_table.md
chongchonghe: `n` is the Hydrogen number density in the resampled cooling table.

chongchonghe: What I meant is that this spacing type here (log and fast_log) applies to how the grids are spaced; they are not how the data values are transformed. So your earlier comment about space_type versus transform_type does not apply.

docs/markdown/data_table.md
chongchonghe: I have already updated the text to be intentionally vague here. I wrote "the round-trip error can be significant", which I remember you have shown in a figure previously. It doesn't matter exactly what those errors are; our trick reduces that to machine accuracy. That's the point.
Regarding how this error is propagated, I think we can prove mathematically that the error is bounded by the combination of the two errors. 
I'm happy to discuss this further, but it shouldn't block this PR.

docs/markdown/data_table.md
chongchonghe: strange. should be pushed now.

docs/markdown/data_table.md
chongchonghe: I'm not sure what you mean here, but I have edited the texts for better clarity.

docs/markdown/data_table.md
chongchonghe: > What do you mean by "input spacing"?

that's repeating " grid spacing", a typo. 

OK, anyway. Renaming it to TransformType is fine; I have not strong opinion on that. 

Done.

src/cooling/ResampledCooling.hpp
chongchonghe: It’s not a good idea because you have to do a static_cast to integer before passing it to the interpolate_single function. That just overcomplicates things, and I don’t think it is necessary.

It is also not a good idea to build this into the DataTable class because we don’t want to enforce enums for the output interaction. I can foresee a future application where we use tens of outputs in a table, and we don’t want to be forced to use enums.

docs/markdown/data_table.md
BenWibking: Ah, I see. That was not clear to me from reading this. I  would be fine with calling that `SpacingType`.

BenWibking: What made this confusing is that all of the examples (that you list below) in this file where this is used mean in it the opposite way - that is, they all refer to the transform type of the table entries.

BenWibking: I think for this section (and only this section), they should be labeled as `SpacingType`, even if it is just an alias to their corresponding `TransformType`.

This way, it is clear that when writing C++ code to use DataTable, you only ever need to pass `TransformType` to the `H5Reader` constructor. Only when creating the HDF5 file itself do users need to set the `SpacingType`. I think this is worth stating explicitly in the documentation.

BenWibking (changes requested):
Ok, I think there is one clarification in the docs that would be helpful regarding the difference between `SpacingType` and `TransformType`.

chongchonghe:
I think the section title `## Coordinate spacing` made it clear that we are talking about grid spacing. The term "transform" is also accurate here because we are indeed transforming the input into a fast log space before we do the linear interpolation. 

This transformation on the input is the same as the transformation in the output, so there's no reason to distinguish them.

BenWibking:
Codex produced two new potential issues:
```
  Full review comments:

  - [P2] Preserve the old luminosity transform key — /Users/benwibking/amrex_codes/quokka/src/simulation.hpp:1110-1110
    For decks that still set the existing particles.rad_table_output_spacing key, this query leaves rad_table_output_transform_ at
    its default fast_log. Linear luminosity tables are then loaded with the wrong output transform; using the old key in
    ParticleRadiation.toml leaves it unused and makes the energy check miss by about 3e-3. Please query the old name as a fallback or
    otherwise migrate the documented parameter in the same change.

  - [P2] Keep the renamed Cloudy source reachable — /Users/benwibking/amrex_codes/quokka/docs/markdown/cooling_module.md:45-45
    The isrf_1000Go_grains.h5 source table was renamed to isrf_1000Go_grains-deprecated.h5, but the Cloudy one-zone helper still
    opens ./isrf_1000Go_grains.h5, so a clean checkout can no longer run that comparison/regeneration workflow. Please update the
    helper/regeneration instructions to pass the renamed source file or keep the old basename available.
```

BenWibking:
> I think the section title `## Coordinate spacing` made it clear that we are talking about grid spacing. The term "transform" is also accurate here because we are indeed transforming the input into a fast log space before we do the linear interpolation.
> 
> This transformation on the input is the same as the transformation in the output, so there's no reason to distinguish them.

I disagree with this. I think there is a reason to distinguish them, and it is this: there are no keyword arguments in C++, so it is ambiguous whether the arguments provided to the `H5Reader` constructor apply to the coordinate axes or to the table values. This makes it clear that `TransformType` applies to the table values, and `SpacingType` applies to the coordinate axes, even if internally, one is just an alias for the other.

src/cooling/ResampledCooling.hpp
BenWibking: Ok, that's a good point. For future applications when a lot of components (like rate tables), we should write a macro to generate these indices for us, but that can be done in a future PR.

src/problems/ResampledCoolingTest/testResampledCoolingTest.cpp
BenWibking: @chongchonghe

chongchonghe:
> > I think the section title `## Coordinate spacing` made it clear that we are talking about grid spacing. The term "transform" is also accurate here because we are indeed transforming the input into a fast log space before we do the linear interpolation.
> > This transformation on the input is the same as the transformation in the output, so there's no reason to distinguish them.
> 
> I disagree with this. I think there is a reason to distinguish them, and it is this: there are no keyword arguments in C++, so it is ambiguous whether the arguments provided to the `H5Reader` constructor apply to the coordinate axes or to the table values. This makes it clear that `TransformType` applies to the table values, and `SpacingType` applies to the coordinate axes, even if internally, one is just an alias for the other.

The variable name `spacing_types` is used extensively in DataTable.hpp to indicate the input spacing, whereas `output_transforms` is used for output transform.  so it should be pretty clear. 

@BenWibking  I think we are getting into very picky details here. We don't want to waste time on little details like naming conventions. I think you also previously said something like this. So, can we proceed?

src/problems/ResampledCoolingTest/testResampledCoolingTest.cpp
chongchonghe: I am using this dummy change to trigger 'quick' and 'romc-quick' CI, so I intentionally leave it for the very last.

chongchonghe:
> Codex produced two new potential issues:
> 
> ```
>   Full review comments:
> 
>   - [P2] Preserve the old luminosity transform key — /Users/benwibking/amrex_codes/quokka/src/simulation.hpp:1110-1110
>     For decks that still set the existing particles.rad_table_output_spacing key, this query leaves rad_table_output_transform_ at
>     its default fast_log. Linear luminosity tables are then loaded with the wrong output transform; using the old key in
>     ParticleRadiation.toml leaves it unused and makes the energy check miss by about 3e-3. Please query the old name as a fallback or
>     otherwise migrate the documented parameter in the same change.
> 
>   - [P2] Keep the renamed Cloudy source reachable — /Users/benwibking/amrex_codes/quokka/docs/markdown/cooling_module.md:45-45
>     The isrf_1000Go_grains.h5 source table was renamed to isrf_1000Go_grains-deprecated.h5, but the Cloudy one-zone helper still
>     opens ./isrf_1000Go_grains.h5, so a clean checkout can no longer run that comparison/regeneration workflow. Please update the
>     helper/regeneration instructions to pass the renamed source file or keep the old basename available.
> ```

Addressed.

BenWibking (dismissed):

BenWibking:
> @BenWibking I think we are getting into very picky details here. We don't want to waste time on little details like naming conventions. I think you also previously said something like this. So, can we proceed?

I think what I said was it wasn't worth it to rename functions or variables (because that tends to introduce bugs). However, I think it is usually worth it to take some time get the naming right the first time, especially when designing interfaces or public APIs that will be used again many times in the future like `DataTable`.

I have approved this PR.

chongchonghe:
> > @BenWibking I think we are getting into very picky details here. We don't want to waste time on little details like naming conventions. I think you also previously said something like this. So, can we proceed?
> 
> I think what I said was it wasn't worth it to rename functions or variables (because that tends to introduce bugs). However, I think it is usually worth it to take some time get the naming right the first time, especially when designing interfaces or public APIs that will be used again many times in the future like `DataTable`.
> 
> I have approved this PR.

OK, fair point. The name alias thing you recommended can be added later if necessary without introducing compatibility issue.

chongchonghe:
@BenWibking removed a placeholder and merged in dev branch. Can you approve again?

BenWibking (approved):

chongchonghe:
@BenWibking I'm not sure what's blocking this from merging.

BenWibking (approved):

BenWibking:
That is amusing. I'm not sure either. I tried re-triggering the doc build actions. Let's see if that fixes it.

BenWibking:
It looks like that worked. Somehow the doc CI actions failed. Probably a Github reliability issue.

PR #2042, opened 2026-07-06
Add O(v/c) work term to the photoionization network

### LINKED ISSUE #1988: Add v/c work term in photoionization
**Describe the proposal**

A clear and concise description of any alternative solutions you've considered.

**Additional context**

Conversation from #1956  

@BenWibking Are $\mathcal{O}(v/c)$ terms included in the equations solved for photoionization?

@chongchonghe No, but you've raised a good and important point. 

The radiation pressure from FUV photons is often comparable to thermal pressure in ionized gas, so I think it's important to include it. I would choose a simple approach: do a single Forward Euler update to include the work term AFTER photochemistry solve. I'll do it in a separate PR.

Ben: I think that's fine, but I think it should be a _backward_ Euler update. That is required for it to be asymptotic-preserving. For case B it is never in the asymptotic diffusion limit, but in general, it could be.

ChongChong: No, we don't do backward Euler on Flux. It's unnecessary. For the RHD module, we do a Picard iteration on the work term. We can do something similar for FUV, but I'll test and see if it is necessary.

chongchonghe:
### Description

Adds the O(v/c) radiation-pressure **work term** to the photoionization operator-split
step. Previously the ionizing-photon flux was attenuated by the burn but the absorbed
photon momentum was never delivered to the gas, and the gas total energy was
reconstructed assuming unchanged kinetic energy. This PR deposits that momentum and
accounts for the resulting kinetic energy, and also adds a new test problem to validate this
following the design of the earlier WIP PR #2007 by @BenWibking (closes #1988).

**`src/radiation/photochemistry.hpp`:**
- Gated on a compile-time switch `do_vc_work = (beta_order >= 1) && is_hydro_enabled`
  — the same O(v/c) gate the regular RHD source terms use. Problems with
  `beta_order == 0` (DTypeFront, OneZonePhotoionization) are unaffected.
- Deposits the absorbed photon momentum `dP = -(F_after − F_before)/c²`, summed over
  chem bands. The momentum is tied to the physical photon flux `F = cE`; the reduced
  speed of light only sets the absorption rate in the chemistry solve.
- Recomputes the gas total energy from the (unchanged) internal energy and the updated
  momentum via `ComputeEgasFromEint`, so the absorbed momentum appears purely as
  kinetic energy.
- `dP` is computed unconditionally in the flux write-back loop and the application is
  gated by a runtime `if (do_vc_work)` on the constexpr flag, to avoid NVCC's
  "extended `__device__` lambda cannot first-capture variable in `constexpr-if`" error
  (verified with a local `3d-cuda` build).
- An optional strict-conservation path (`debit_work_term_from_radiation`, **off by
  default**) debits the work-term energy `(c/ĉ)·dEkin` from the chem-band photon field
  and rescales the flux to preserve `|F| ≤ cE`. It is off by default because conserving
  the ionizing photon number density matters more for correct front propagation than
  strict gas+radiation energy conservation.

Updated `DTypeFront` test to `beta_order = 1` to validate that the radiation pressure term
is not important for the dynamics at all for the parameters in the problem. 
**New test `src/problems/DTypeFrontVC/`** (ported from PR #2007): a `beta_order >= 1`,
purely x-directed radiation beam on uniform gas. It verifies directly, by conservation,
that the deposited gas momentum equals the absorbed photon momentum `(F_before −
F_after)/c²`, that no transverse momentum is created, that the internal energy is
unchanged, and that the total-energy increase equals the kinetic-energy gain.

### Effect of enabling the work term on DTypeFront

The `DTypeFront` problem (`beta_order == 0`) intentionally does **not** activate this
PR's work term — it tests pure thermal-pressure D-type front expansion against the
Spitzer analytic solution. During development we tested what happens when the momentum
coupling is turned on (`beta_order = 1`), first with a buggy formula and then with the
correct `/c²` formula:

| Build | Formula | r_analytical (cm) | r_eff (cm) | (r_eff − r_ana)/dx |
|-------|---------|-------------------|------------|---------------------|
| #14474 (GPU, 64³, full) | buggy `dp = −ΔF/(c·ĉ)` + photon debit | 9.76e18 | 1.98e19 | **40.3** cells |
| #14475 (GPU, 64³, full) | buggy `dp = −ΔF/(c·ĉ)`, no debit | 9.76e18 | 1.54e19 | **22.8** cells |
| local (CPU, 16³, 4 t_s) | correct `dp = −ΔF/c²`, momentum on | 4.127e18 | 3.833e18 | −0.30 cells |
| local (CPU, 16³, 4 t_s) | correct `dp = −ΔF/c²`, momentum off | 4.127e18 | 3.822e18 | −0.31 cells |
| tolerance | — | — | — | 2.60 cells |

The two GPU CI rows (#14474/#14475) used an early implementation that over-injected
momentum by `c/ĉ ≈ 1000×` (the stored chem-band flux is `F = cE`, not `ĉE`). Those
huge deviations are an artifact of that bug, not the physical radiation force.

With the **correct** `dP = −(F_after − F_before)/c²` formula (bottom two rows),
the momentum kick has a negligible effect on the D-type front radius: it shifts
r_eff by only **+0.011 cells** relative to the control run, both passing within
the 2.6-cell tolerance. The radiation pressure from ionizing photons is sub-dominant
relative to the thermal pressure of the ~10⁴ K ionized gas — a physically plausible
result — so the Spitzer solution remains a valid check for the thermal-only
physics tested by `DTypeFront`.

### Related issues
Closes #1988.

### Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a direct photoionization radiation–gas momentum coupling validation test for multidimensional simulations.
  - Added configuration support for testing O(v/c) photoionization momentum deposition and energy behavior.
  - Radiation updates now account for momentum transferred to gas during photoionization.

- **Bug Fixes**
  - Improved consistency between gas momentum, kinetic energy, and absorbed radiation during O(v/c) interactions.
  - Added optional radiation-energy accounting for radiation-pressure work.

### DIFF
--- a/src/radiation/photochemistry.hpp
+++ b/src/radiation/photochemistry.hpp
@@ -138,23 +138,81 @@ auto computePhotoChemistry(amrex::MultiFab &mf, std::array<amrex::MultiFab const
 			for (int nn = 0; nn < NumSpec; ++nn) {
 				state(i, j, k, RadSystem<problem_t>::scalar0_index + nn) = photochemstate.xn[nn] * spmasses[nn];
 			}
+			// Attenuate the photon field and accumulate the momentum transferred to the gas.
+			// Absorbed ionizing photons deposit their momentum in the gas: dp = -(F_new - F_old) / (c * chat),
+			// matching the RSLA-consistent radiation-force convention used by the M1 source terms.
+			const Real inv_c_chat = 1.0_rt / (C::c_light * photochemstate.c_hat);
+			Real dMomX = 0.0_rt;
+			Real dMomY = 0.0_rt;
+			Real dMomZ = 0.0_rt;
+			amrex::GpuArray<Real, NumChemBands> dMomMag{};
+			Real dMomMagSum = 0.0_rt;
 			for (int nn = 0; nn < NumChemBands; ++nn) {
-				state(i, j, k, firstChemIndex + Physics_NumVars::numRadVarsPerGroup * nn) =
-				    photochemstate.rn[0 + MicrophysicsNumRadVarsPerGroup * nn] * chemBandQuanta[nn];
-				state(i, j, k, firstChemFxIndex + Physics_NumVars::numRadVarsPerGroup * nn) =
-				    photochemstate.rn[1 + MicrophysicsNumRadVarsPerGroup * nn] *
-				    state(i, j, k, firstChemFxIndex + Physics_NumVars::numRadVarsPerGroup * nn);
-				state(i, j, k, firstChemFyIndex + Physics_NumVars::numRadVarsPerGroup * nn) =
-				    photochemstate.rn[1 + MicrophysicsNumRadVarsPerGroup * nn] *
-				    state(i, j, k, firstChemFyIndex + Physics_NumVars::numRadVarsPerGroup * nn);
-				state(i, j, k, firstChemFzIndex + Physics_NumVars::numRadVarsPerGroup * nn) =
-				    photochemstate.rn[1 + MicrophysicsNumRadVarsPerGroup * nn] *
-				    state(i, j, k, firstChemFzIndex + Physics_NumVars::numRadVarsPerGroup * nn);
+				const int eIdx = firstChemIndex + Physics_NumVars::numRadVarsPerGroup * nn;
+				const int fxIdx = firstChemFxIndex + Physics_NumVars::numRadVarsPerGroup * nn;
+				const int fyIdx = firstChemFyIndex + Physics_NumVars::numRadVarsPerGroup * nn;
+				const int fzIdx = firstChemFzIndex + Physics_NumVars::numRadVarsPerGroup * nn;
+
+				state(i, j, k, eIdx) = photochemstate.rn[0 + MicrophysicsNumRadVarsPerGroup * nn] * chemBandQuanta[nn];
+
+				// flux_factor = F_new / F_old (photon flux attenuation from the burn)
+				const Real flux_factor = photochemstate.rn[1 + MicrophysicsNumRadVarsPerGroup * nn];
+				const Real FxOld = state(i, j, k, fxIdx);
+				const Real FyOld = state(i, j, k, fyIdx);
+				const Real FzOld = state(i, j, k, fzIdx);
+
+				const Real dpx = (1.0_rt - flux_factor) * FxOld * inv_c_chat;
+				const Real dpy = (1.0_rt - flux_factor) * FyOld * inv_c_chat;
+				const Real dpz = (1.0_rt - flux_factor) * FzOld * inv_c_chat;
+				dMomX += dpx;
+				dMomY += dpy;
+				dMomZ += dpz;
+				dMomMag[nn] = std::sqrt(dpx * dpx + dpy * dpy + dpz * dpz);
+				dMomMagSum += dMomMag[nn];
+
+				state(i, j, k, fxIdx) = flux_factor * FxOld;
+				state(i, j, k, fyIdx) = flux_factor * FyOld;
+				state(i, j, k, fzIdx) = flux_factor * FzOld;
 			}
+
 			// Quokka uses rho*eint
 			const Real dEint = (photochemstate.e * photochemstate.rho) - Eint;
 			state(i, j, k, RadSystem<problem_t>::gasInternalEnergy_index) += dEint * energy_update_factor;
-			state(i, j, k, RadSystem<problem_t>::gasEnergy_index) += dEint * energy_update_factor;
+
+			// Update the gas momentum by conservation with the absorbed photons.
+			const Real xmom_new = xmom + dMomX;
+			const Real ymom_new = ymom + dMomY;
+			const Real zmom_new = zmom + dMomZ;
+			state(i, j, k, RadSystem<problem_t>::x1GasMomentum_index) = xmom_new;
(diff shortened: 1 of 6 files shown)


src/radiation/photochemistry.hpp
```diff
@@ -138,23 +154,97 @@ auto computePhotoChemistry(amrex::MultiFab &mf, std::array<amrex::MultiFab const
...
-				state(i, j, k, firstChemFzIndex + Physics_NumVars::numRadVarsPerGroup * nn) =
-				    photochemstate.rn[1 + MicrophysicsNumRadVarsPerGroup * nn] *
-				    state(i, j, k, firstChemFzIndex + Physics_NumVars::numRadVarsPerGroup * nn);
+				const Real FxOld = state(i, j, k, fxIdx);
+				const Real FyOld = state(i, j, k, fyIdx);
+				const Real FzOld = state(i, j, k, fzIdx);
+				// dp = (F_before - F_after) / c^2 = (1 - flux_factor) * F_before / c^2
+				const Real dpx = (1.0_rt - flux_factor) * FxOld * inv_c2;
+				const Real dpy = (1.0_rt - flux_factor) * FyOld * inv_c2;
+				const Real dpz = (1.0_rt - flux_factor) * FzOld * inv_c2;
```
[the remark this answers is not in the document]
chongchonghe: No. The rational is explained in the comments a few lines above the code.

chongchonghe:
@markkrumholz @BenWibking Ready for review.

BenWibking:
Can you add an input file that tests a regime where the momentum deposition is O(1) compared to the pressure force?

Mark's paper provides a formula for this as a function of background density and luminosity (Equation 8): https://ui.adsabs.harvard.edu/abs/2009ApJ...703.1352K/abstract

src/radiation/photochemistry.hpp
```diff
@@ -138,23 +151,97 @@ auto computePhotoChemistry(amrex::MultiFab &mf, std::array<amrex::MultiFab const
 			for (int nn = 0; nn < NumSpec; ++nn) {
 				state(i, j, k, RadSystem<problem_t>::scalar0_index + nn) = photochemstate.xn[nn] * spmasses[nn];
 			}
+			// Update the chem-band photon number density, attenuate the flux by the burn's rn[1] factor, and
+			// accumulate the momentum the gas absorbs from the O(v/c) work term: dP = -(F_after - F_before) / c^2,
+			// summed over chem bands. The absorbed photon momentum is tied to the physical photon flux F = c E; the
+			// reduced speed of light only changes the absorption rate used by the chemistry solve. dMom is computed
+			// here unconditionally (cheap) and only applied when do_vc_work is true -- computing it outside the
+			// `if constexpr (do_vc_work)` block avoids NVCC's "cannot first-capture in constexpr-if" restriction.
```
BenWibking, line 159: You can instead first-capture into a variable here:
`auto captured_var = do_vc_work;`

Then you can do:
`if constexpr (captured_var) {`

src/radiation/photochemistry.hpp
```diff
@@ -138,23 +151,97 @@ auto computePhotoChemistry(amrex::MultiFab &mf, std::array<amrex::MultiFab const
...
+				state(i, j, k, RadSystem<problem_t>::gasEnergy_index) = ::quokka::EOS<problem_t>::ComputeEgasFromEint(
+				    rho, xmom_new, ymom_new, zmom_new, state(i, j, k, RadSystem<problem_t>::gasInternalEnergy_index), Emag);
+
+				// Optionally enforce strict gas+radiation energy conservation: remove the work-term energy
+				// (c / chat) * dEkin from the chem-band photon field, split across bands in proportion to each band's
+				// momentum contribution, and rescale each band's flux by the same factor its energy density changes so
+				// the reduced flux |F| / (c E) is preserved (keeps the M1 closure F / c <= E from being violated by the
+				// debit). Disabled by default: conserving the ionizing photon number density is more important than
+				// strict energy conservation for correct ionization-front propagation (see debit_work_term_from_radiation).
+				if constexpr (debit_work_term_from_radiation) {
```
BenWibking, line 220: Following our discussion yesterday, I'm not sure if it makes sense to leave this as a option. I would prefer to just delete this code (but leave the comment explaining why).

src/radiation/photochemistry.hpp
chongchonghe: right, but it's only used once, so I won't bother define a new parameter. I removed the unnecessary comment about constexpr.

src/radiation/photochemistry.hpp
chongchonghe: I intentionally leave it there, just in case we ever prefer a energy-preserving scheme. I think the inline comments are clear enough to avoid confusion.

chongchonghe:
> Can you add an input file that tests a regime where the momentum deposition is O(1) compared to the pressure force?
> 
> Mark's paper provides a formula for this as a function of background density and luminosity (Equation 8): https://ui.adsabs.harvard.edu/abs/2009ApJ...703.1352K/abstract

This requires setting up a problem in the radiation-pressure-dominated regime given by the parameter in Eq. 8 of the paper, and compare it against an approximate solution Eqs. 11-13 with a large tolerance (~10-20%). 
@James471 Can you create a PR to do that?

chongchonghe:
@BenWibking I'll leave the radiation pressure-dominated regime test for @James471 in a separate PR. I would prefer to merge in this PR as is. It's ready for review now.

BenWibking (approved):
Looks good to me.

PR #2047, opened 2026-07-09
Lower one zone photoionization test tolerance

James471:
### Description
This PR lowers the tolerances for the one zone photoionization tests, otherwise they are very slow on GPUs.

### Checklist

## Summary by CodeRabbit

* **Improvements**
  * Updated photochemistry integration settings for the standard and energy-enabled one-zone configurations.
  * Relaxed absolute and relative numerical tolerances to support more practical simulation runs.
  * Retained the existing radiation failure threshold while adjusting species, radiation, and energy-related tolerance values.

### DIFF
--- a/inputs/OneZonePhotoionization.toml
+++ b/inputs/OneZonePhotoionization.toml
@@ -36,17 +36,17 @@ photoionization.temperature = 1e3
 
 # Validation test against Julia reference at 1e-6 L1 error tolerance.
 # Use very tight tolerances (effectively pure relative) to match the reference.
-integrator.atol_spec                     = 1e-14
-integrator.atol_enuc                     = 1e-14
-integrator.atol_rad_num                  = 1e-14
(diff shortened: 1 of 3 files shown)


chongchonghe:
@James471 I'm afraid this is not the right thing to do. In a production run, we would have to choose one solver, so it's going to be ROS2S. So, it is not helpful if RODAS5P runs fast and ROS2S is slow for this problem. 

I think the right thing to do is to change the tolerance such that it passes the test and runs fast. @James471 Can you do that?

chongchonghe:
@James471 To trigger a test problem with `azp run quick`, you need to modify the problem in `src/problems/PROBLEM` . So, you can do a dummy change to the cpp file and trigger `azp run quick` and `azp run rocm-quick`

chongchonghe:
@James471 seems like this is ready. LGTM. I'll mark it ready and approve. Please merge it if you think it's ready. If you need to further change it, I'll review again later.

chongchonghe (approved):

Issue #2073, opened 2026-07-11
Resistive diffusion reference uses mean heating


BenWibking:
> *This was generated by AI during triage.*

## Summary
The `MHDResistiveDiffusion` reference solution uses domain-averaged Ohmic heating at every cell, so its pointwise total-energy reference is not the analytical solution of the test problem.

## Severity
`High`

## Affected File
`src/problems/MHDResistiveDiffusion/testMHDResistiveDiffusion.cpp`

## Affected Function / Symbol
`QuokkaSimulation<MHDResistiveDiffusion>::computeReferenceSolution(...)`

## Audit Metadata
- Source log: `issues/new/resistive-diffusion-reference-uses-mean-heating.md`
- Finding tags: correctness/MHD/test

## Proposed Patch
Replace the uniform heating increment in `computeReferenceSolution()` with the local `cos^2(kx)` expression and construct the total-energy reference from the local internal and magnetic energies. Keep the auxiliary `gasInternalEnergy` reference consistent with whether that component is actually evolved when dual energy is disabled.

Strengthen the regression by checking the spatial amplitude of total energy or derived internal energy, not only a global relative norm. Choose parameters and a tolerance that make replacing `cos^2(kx)` by its mean fail reliably.

## Why This Is a Bug
The test evolves

```text
B_z(x,t) = B_0 exp(-eta k^2 t) sin(kx).
```

The current reference adds the same internal-energy increment to every cell:

```text
B_0^2 / 4 * (1 - exp(-2 eta k^2 t)).
```

That expression is the domain average. Locally, the current density is proportional to `cos(kx)`, so integrating `eta |J|^2` gives

```text
Delta e_int(x,t) = B_0^2 / 2 * (1 - exp(-2 eta k^2 t)) * cos^2(kx).
```

The present `E_tot` reference therefore has the wrong spatial structure and can hide errors in the resistive energy flux. The current 3D `MHDResistiveDiffusion` CTest passes despite this mismatch.

AstroKriel:
Resolved by #2085

PR #2085, opened 2026-07-11
Fix resistive energy flux averaging; replace MHDResistiveDiffusion test

AstroKriel:
### Description

Investigating #2073 led to #2086: a real bug in `AddResistiveEnergyFlux()` (wrong edge-averaging coefficient, `0.25` instead of `0.5`), and the discovery that `MHDResistiveDiffusion` cannot validate that function regardless of whether the coefficient is right, because its field configuration is never in mechanical equilibrium. See #2086 for the full investigation, evidence, and reasoning behind each change below.

This PR:
- Fixes the `0.25 -> 0.5` coefficient in `AddResistiveEnergyFlux` (`src/hydro/mhd_system.hpp`).
- Removes `MHDResistiveDiffusion`, and repoints `FastWaveConvergence`/`SlowWaveConvergence`'s resistivity-abort messages at `AlfvenWaveLinearConvergence`.
- Adds `MHDResistiveEnergyFluxKernel`: a static test that calls `AddResistiveEnergyFlux()` directly on a hand-built field, with no `QuokkaSimulation`/`evolve()` involved. This is the first unit/algorithm test in the repo (`docs/markdown/maintainer_guidelines.md` defines the category; none existed before this).

### Validation

`MHDResistiveEnergyFluxKernel`: with the fix, all three checked points match independently-computed expected values to `rel_err=0`. Reverting to the original `0.25` coefficient makes all three fail with exactly `rel_err=0.5`, confirming the test discriminates the bug rather than passing regardless.

Since `AddResistiveEnergyFlux` returns immediately when resistivity is off, this change is a no-op for ideal MHD. Confirmed via the full MHD smoke-test set: `SlowWaveConvergence` passes (2nd-order convergence, rate 1.99), `BrioWuShockTube` and `OrszagTang` both exit cleanly with no NaN/crash.

### Related issues

- #2073: fixed by removing the test whose own reference bug made it unable to catch this class of error in the first place.
- #2086: fixed directly; see that issue for the full investigation.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added validation for resistive energy-flux calculations using synthetic magnetic-field configurations.
  * Registered the resistive energy-flux validation problem for three-dimensional builds.

* **Bug Fixes**
  * Corrected resistive energy-flux averaging.
  * Improved the resistive diffusion analytic reference and weak-field validation.
  * Updated resistivity validation guidance for fast- and slow-wave tests.

* **Chores**
  * Simplified the resistive energy-flux test configuration and reduced its placeholder grid size.
  * Removed the obsolete resistive diffusion test build and test registration.


AstroKriel:
@markkrumholz, just a heads up that `quick`/`rocm-quick` fails because it's trying to build `MHDResistiveDiffusion`, which this PR removes; `release [GPU, 3D]` check has passed on ROCm and is still running on nvcc.

markkrumholz (approved):
Approving and merging, noting the reason for the failed quick pipeline tests.

PR #2087, opened 2026-07-12
Fix particle container layout after AMR regridding

YaoguangPei:
### Description
This PR fixes stale particle-grid metadata after AMR regridding. It adds an interface for synchronizing each particle container with the updated `Geometry`, `BoxArray`, and `DistributionMap`, then performs particle redistribution using the new grid layout. This prevents particle iterators from referencing outdated grid indices or MPI ownership after restarting and regridding.
### Related issues
None

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved particle handling after adaptive mesh regridding.
  * Particle data now stays synchronized with updated grid layouts across affected simulation levels.
  * Improved particle redistribution during 3D subcycled time steps.

### DIFF
--- a/src/particles/PhysicsParticles.hpp
+++ b/src/particles/PhysicsParticles.hpp
@@ -183,6 +183,9 @@ class PhysicsParticleDescriptorBase
 	// Redistribute particles at level lev and above with ngrow ghost cells
 	virtual void redistribute(int lev, int ngrow) const = 0;
 
+	// Synchronize the particle container with an AMR level after regridding.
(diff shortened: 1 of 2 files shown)


BenWibking:
I don't think this is the correct fix. `AmrParticleContainer(AmrCore*)` automatically tracks the `AmrCore` grid hierarchy, so manually redoing this should not be necessary, and may also cause some edge cases to break.

However, I think something might be broken with the code I added to restart with refinement on every level when regridding is allowed. I will try to create a PR to fix that.

@YaoguangPei Are you able to create a minimal reproducer for the restart issue you're seeing?

BenWibking:
@YaoguangPei Can you try this one-line fix instead:
https://github.com/quokka-astro/quokka/pull/2089/changes#r3566726708

PR #2089, opened 2026-07-12
Reattach particles to simulation class after restart refinement

BenWibking:
#### Problem

Universal restart refinement temporarily assigns geometry and grid layouts from the checkpoint file to physics particle containers. This detaches an `AmrParticleContainer` from the actual `AmrCore` `ParGDB` used for the simulation. Due to this, after a later AMR regrid, the particle container's `BoxArray` and `DistributionMapping` are no longer in sync with the actual AMR grids used by the AMRSimulation/AMRCore class.

This issue only affected users if they restarted a simulation with universal refinement, and then ran then simulation with AMR regridding enabled. This would then cause a crash due to invalid memory accesses.

#### Solution

- Reattach the particle containers to `AmrCore::GetParGDB()` before redistributing particles at the end of the restart procedure
- Extended `BinaryOrbitCIC` to verify this resolves the original bug as part of the existing CheckpointRestartRefine action.
- Confirmed the new test fails on `development` at the first post-restart regrid before applying the fix.
- Full `inputs/checkpoint_restart_refine_test.sh` with four MPI ranks: existing hydro comparison passed, `fcompare` passed, and the new particle layout check passed with 16 particles.

### Related issues
N/A

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for restarting particle containers while refining the AMR hierarchy.
  * Added optional verification to confirm particle layouts remain consistent with the refined hierarchy.
  * Extended automated coverage for particle checkpoint and restart-refinement workflows.

* **Bug Fixes**
  * Improved particle container setup during refined restarts to preserve compatibility with the current AMR layout.

* **Tests**
  * Added validation for successful particle checkpoint creation and refined restart behavior.
  * Reduced unnecessary runtime output and profiling during AMR test runs.

### DIFF
--- a/src/simulation.hpp
+++ b/src/simulation.hpp
@@ -5218,6 +5218,7 @@ void AMRSimulation<problem_t>::restartParticleContainerWithRefinement(std::uniqu
 			particles->SetParticleBoxArray(lev, current_ba[lev]);
 			particles->SetParticleDistributionMap(lev, current_dm[lev]);
 		}
+		particles->Define(this->GetParGDB());
 
 		// Redistribute particles to refined grid
 		particles->Redistribute();
--- a/.github/workflows/checkpoint-restart-refine.yml
+++ b/.github/workflows/checkpoint-restart-refine.yml
@@ -76,7 +76,7 @@ jobs:
       working-directory: ${{runner.workspace}}/build
       shell: bash
       # Execute the build.  You can specify a specific target with "--target <NAME>"
-      run: cmake --build . --config $BUILD_TYPE --parallel 4 --target HydroBlast3D
+      run: cmake --build . --config $BUILD_TYPE --parallel 4 --target HydroBlast3D BinaryOrbitCIC
 
     - name: Checkpoint/Restart with Refinement Test
       working-directory: ${{github.workspace}}/tests
--- a/inputs/checkpoint_restart_refine_test.sh
+++ b/inputs/checkpoint_restart_refine_test.sh
@@ -255,6 +255,9 @@ fi
 
 mpirun --use-hwthread-cpus -n $NPROC $BUILD_DIR/src/problems/BinaryOrbitCIC/BinaryOrbitCIC ../inputs/BinaryOrbitCICAMR_checkpoint_restart.toml \
     restartfile=$particle_checkpoint 'amr.n_cell=64 64 64' amr.blocking_factor=16 amr.max_grid_size=16 regrid_interval=1 \
-    max_timesteps=2 checkpoint_interval=-1 plotfile_interval=-1 suppress_output=1 tiny_profiler.enabled=0 problem.verify_particle_layout=1
+    max_timesteps=2 checkpoint_interval=-1 plotfile_interval=-1 suppress_output=1 tiny_profiler.enabled=0 problem.verify_particle_layout=1 || {
+    echo "❌ PARTICLE LAYOUT CHECK FAILED"
+    exit 1
+}
 
 echo "✅ PARTICLE LAYOUT CHECK PASSED: Restart-refined particle container tracks the regridded AMR hierarchy."
--- a/inputs/checkpoint_restart_refine_test.sh
+++ b/inputs/checkpoint_restart_refine_test.sh
@@ -94,7 +94,8 @@ $SED_INPLACE 's/do_subcycle = 0/do_subcycle = 1/' fine_amr.toml
 
 echo "=== Step 1: Create coarse AMR checkpoint (32^3 base + 1 AMR level) ==="
 # Run coarse AMR simulation to create checkpoints
-mpirun --use-hwthread-cpus -n $NPROC $BUILD_DIR/src/problems/HydroBlast3D/HydroBlast3D coarse_amr.toml stop_time=$CHECKPOINT_TIME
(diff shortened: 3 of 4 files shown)


src/simulation.hpp
```diff
@@ -5218,6 +5218,7 @@ void AMRSimulation<problem_t>::restartParticleContainerWithRefinement(std::uniqu
 			particles->SetParticleBoxArray(lev, current_ba[lev]);
 			particles->SetParticleDistributionMap(lev, current_dm[lev]);
 		}
+		particles->Define(this->GetParGDB());
```
BenWibking, line 5221: This is the one-line fix.

chongchonghe:
# PR #2089 Review: Reattach particles to simulation class after restart refinement

**Branch:** `agent/fix-restart-particle-layout`  
**Reviewer:** Claude Code  
**Date:** 2026-07-13  
**Classification:** Bug fix → RED → GREEN → EDGE cycle

---

## RED / GREEN / EDGE Table

|---|---|---|
| RED (development) | `BinaryOrbitCIC BinaryOrbitAMR.toml max_timesteps=1 checkpoint_interval=1` → then restart with `amr.n_cell=64 64 64 regrid_interval=1 max_timesteps=2` | Simulation completed without crash on this platform (memory access non-corrupting), but code inspection at `simulation.hpp:5215–5220` confirms `particles->Define(this->GetParGDB())` is absent — the `ParGDB` pointer is left detached after refinement, exactly as described by the PR author |
| GREEN (PR branch) | Identical restart scenario with `problem.verify_particle_layout=1` added (new assertion from this PR) | All timesteps completed; `AMREX_ALWAYS_ASSERT_WITH_MESSAGE(layout_matches, ...)` did NOT fire — particle container correctly tracks AMR hierarchy post-fix |
| EDGE – standard run | `BinaryOrbitCIC BinaryOrbitCIC.toml` (no restart, no verify) | `max deviation = 0.051 cell widths`, `Test passed` × 2 — no regression |
| EDGE – pre-existing restart path | `BinaryOrbitCICAMR_checkpoint_restart.toml` (existing TOML, no verify flag) | `max deviation = 0.056 cell widths`, `Test passed` × 2 — existing restart+refine path unaffected |

---

## Root Cause Analysis

`AMRSimulation::restartParticleContainerWithRefinement` (`src/simulation.hpp:5178–5227`) handles the universal-refinement restart path. After reading particles from the coarse checkpoint, it restores the refined geometry using `SetParticleGeometry`, `SetParticleBoxArray`, and `SetParticleDistributionMap` (lines 5216–5219). These calls update the particle container's internal arrays but do **not** update the `ParGDB` (Particle Grid Database) — the AMReX object that links an `AmrParticleContainer` to its parent `AmrCore`.

The `ParGDB` is what AMReX consults when regridding: all registered particle containers are expected to share the same `ParGDB` as the simulation's `AmrCore`. Because `Define(GetParGDB())` was missing, the particle container after restart refinement held a stale `ParGDB` reference — one set up from the checkpoint's grid layout rather than the live simulation. On the next regrid, `AmrCore::regrid()` updates the simulation's `ParGDB` and all correctly-attached containers, but the detached particle container is left with a stale pointer. Subsequent particle operations (redistribution, deposition) then access the wrong grid metadata, causing invalid memory accesses (platform-dependent; a crash on some systems, silent data corruption on others).

**Fix** (`simulation.hpp:5221`, one line): `particles->Define(this->GetParGDB())` immediately after restoring the geometry, before `Redistribute()`. This re-registers the particle container with the simulation's `ParGDB` so it tracks all future regrids correctly.

The accompanying test extension in `testBinaryOrbitCIC.cpp` adds `verify_particle_layout` mode: after each timestep, it asserts that `CICParticles->GetParGDB() == GetParGDB()` and that the `BoxArray` and `DistributionMapping` on every level match the simulation's. This turns the previously-silent mismatch into a deterministic, immediate assertion failure.

---

## PR-Introduced Concerns

**Minor: `TestFailed` in non-particle checkpoint step is pre-existing, not introduced**

When running `BinaryOrbitAMR.toml` with `max_timesteps=1`, one of the two final test checks prints `Test failed`. This pre-dates this PR (present on `development` too) and is caused by zero-mass particles at that early stage — it is not a regression from this PR.

**Minor: `verify_particle_layout` is a global static, not thread-safe**

`verify_particle_layout` at `testBinaryOrbitCIC.cpp:34` is a file-scope `static bool` set via `pp.query`. This is fine for the current serial-init pattern, but is the same pattern as the pre-existing `do_split_particles` / `split_factor` statics. Not introduced by this PR, not a blocker.

**Minor: shell test does not assert the exit code of the `mpirun` restart command**

In `inputs/checkpoint_restart_refine_test.sh:259`, the restart `mpirun` command is followed immediately by the success echo with no `|| exit 1` guard. If `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` fires (non-zero exit), the shell script would still print `✅ PARTICLE LAYOUT CHECK PASSED`. A stricter guard would be:

```bash
mpirun ... problem.verify_particle_layout=1 || { echo "❌ PARTICLE LAYOUT CHECK FAILED"; exit 1; }
echo "✅ PARTICLE LAYOUT CHECK PASSED: ..."
```

This is a minor robustness gap; the existing CI workflow captures exit codes at the job level so the CI would still fail, but the success message being printed on failure is confusing.

---

None identified that are relevant to block this PR.

---

BenWibking:
@chongchonghe I've fixed the error message. The other issues appear to be unrelated.

chongchonghe (approved):

BenWibking:
It just failed to install the dependencies when Github ran the container. It's re-running now.

PR #2126, opened 2026-07-24
Enable mixed thermal + chemical multigroup bands, harden the matter-radiation Newton solve, and add DTypeFront1D

chongchonghe:
### Description

Enables mixed thermal + chemical multigroup radiation, fixes three Newton-solver convergence bugs, and adds `DTypeFront1D`, a 1D dust-reprocessing test that exercises them.

**Mixed thermal + chemical bands.** Chemistry bands may now occupy the last `NChemBands` groups of a multigroup run rather than all-or-nothing. They are decoupled from the thermal gas–radiation energy exchange: their source is removed from the Newton solve and injected directly into `Erad` afterwards, so it survives at zero opacity and does not pollute the gas energy balance. They emit no blackbody, and emission above the first chemistry-band boundary is dropped rather than renormalized. The solver now also injects the source into transparent groups (`tau == 0`), which it previously left unchanged.

**Newton convergence fixes.** Three separate defects in the multigroup energy-exchange solve, each of which could abort a run.

1. *Wrong group Planck temperature derivative.* `ComputeThermalRadiationTempDerivativeMultiGroup` returned `4 a T^3 f_g`, differentiating the `a T^4` prefactor while holding the group energy fractions `f_g` fixed. That is accurate for the group carrying the spectral peak but wrong for a group on the Wien tail, where `f_g` rises roughly as `T^20`. The resulting Jacobian entry was 5x too small in a reported failing cell, making the Newton step 3.5x too long, so the iteration overshot and diverged. Replaced by the exact derivative, which integrates by parts into the same normalized Planck integral already used for the energy fractions.

2. *Catastrophic cancellation in the group energy.* With `R_g` as the unknown, each iteration recovers `Erad_g = kappaPoverE_g * (4 pi B_g / c - R_g / tau_g)`, and for `tau_g << 1` those two terms agree to within a factor `tau_g`. Optically thin groups now take `Erad_g` as the unknown instead and recover `R_g`, which does not cancel in that regime; thick groups keep `R_g`, since near radiative equilibrium it is `R_g` that comes out of a cancelling difference. The map is affine, so converged answers are unchanged outside round-off. The existing round-off floor on the convergence test is retained for the groups still based on `R_g`.

3. *Unbracketed one-variable solve.* `BackwardEulerOneVariable` ran a bare Newton iteration and returned a `-1` sentinel on failure, surfacing as "dust temperature is negative". It now brackets the root by multiplicative marching and runs Newton guarded by bisection.

Measured on a 216-case sweep over group boundaries, opacities (seven orders of magnitude in both directions) and injected flux: converged cases go from 115 to 212, and matter-radiation non-convergence aborts from 38 to 0. The four that remain are all in the dust-temperature solve at the most extreme flux.

Defect 1 is the cause of the non-convergence @james471 reported (`Newton-Raphson iteration for matter-radiation coupling failed to converge!` with `photoionize.kappa1 = 1.0e-2`, `photoionize.kappa2 = 1.0e4`). That case reproduces on `development` and passes on this branch.

**`enable_dE_constrain` now defaults to `false`** — a global default affecting every radiation problem. Required here: with it enabled `DTypeFront1D` still fails to converge.

**New problem `DTypeFront1D`.** A 1D planar H II region test — the plane-parallel analogue of the Strömgren sphere — that doubles as a dust-reprocessing test. A 128-cell slab of cold neutral hydrogen carries IR, optical and ionizing groups. A central ionizing source drives an ionization front outward into the neutral gas, photoheating what it ionizes, until it stalls near the planar Strömgren column `N_S = F_ion/(alpha_B n_0)`; the measured ionized column, 3.3e20 cm^-2 per side, sits in the 3.3–3.8e20 range that `alpha_B` spans between 8000 K and 10^4 K. A co-located non-ionizing optical source is absorbed by dust, which settles at radiative equilibrium (~130 K) and re-emits into the IR. Gray dust opacities and both photon fluxes are runtime parameters, and the dust temperature is solved separately with the gas–dust collisional coupling switched off, so it is set purely by radiative equilibrium.

The source sits at the centre and each half of it is injected beamed away from the centre (`radFluxSource = -c*E` left of centre, `+c*E` right), which is what keeps the budgets clean. Both boundaries are reflecting, and a reflecting wall is a momentum source whose bookkeeping value the reduced speed of light amplifies by `c/chat`; with the source central and the run stopped before either front arrives, no wall is ever reached. Beaming is the other half: an isotropic source injects no net momentum, so the momentum budget would measure how fast the M1 closure can beam an isotropic source rather than what the solver did with the momentum (0.78 and drifting with resolution, versus 1.004 and flat when beamed).

### Validation

`DTypeFront1D` checks, with measured values:

- IR + optical conserve the energy injected into the optical band (1.0000000). Photoionization drains only the chemistry band, and with the gas thermally decoupled the dust just shuffles energy between the two thermal groups.
- Linear momentum, two statements. The outward momentum (each cell signed by `sgn(x - x_source)`) held by the gas plus the two sourced bands is 1.0038 of the injected `2 Phi E_photon t / c` (tol 1%), and it is flat under refinement — 1.0038 / 1.0041 at 128 / 1024 cells. The gas carries 0.37 of it, so this is the check that exercises the radiation force; unlike the energy budget it carries no `chat/c` factor. Separately the signed total over the gas and all three bands vanishes (1.8e-16 of the injected scale, tol 1e-10), which catches any asymmetry between the two wings. The dust-re-emitted IR band is excluded from the budget — nothing injects momentum into it, its outward flux is generated by transport down the pressure gradient, and including it would add 9.2% since the RSLA reservoir `F/(c*chat)` is inflated by `c/chat`.
- Optical light front against `chat * t` (4.2%, tol 10%; one cell of front-position quantization at this resolution, falling to 0.7% by 1024 cells).
- Dust reprocessing: 0.367 of the injected light moved into the IR, and the surviving optical matches the analytic `exp(-tau)` integral (1.0032, tol 10%).
- Ionizing photon budget: 4.0e21 injected, 2.7e21 absorbed, ionized column 6.7e20 cm^-2, with the surviving fraction 0.316 bounded in [0.02, 1).

The checks were verified to discriminate by deliberately breaking what they cover: mis-scaling the chemistry-band source by `chat/c` collapses the surviving ionizing fraction to 8e-5 (the photon-budget inequality alone does *not* catch this — fewer photons produce a proportionally smaller ionized column — which is why the surviving fraction is bounded below). Deleting the gas momentum kick drops the outward-momentum ratio to 0.644 and the gas share from 0.37 to 0.021.

Widening the source slab was tried and does not help: `photoionize.source_cells` = 1, 2, 4, 8 gives a momentum ratio of 1.0038, 1.0083, 1.0105, 1.0118 beamed and 0.78, 0.74, 0.63, 0.47 isotropic, and a wide slab also pushes the measured light front outward. One cell per side is shipped. Both knobs (`photoionize.beamed`, `photoionize.source_cells`) are kept so the comparison is reproducible.

All 30 `Rad*` and `DType*` 1D regression tests pass. `RadMarshak`, `RadhydroPulseMGconst`, `RadSuOlson` and `RadDustMG` error norms are unchanged.

**#2171 is merged.** The directed radiation source (`AddRadSource` with a reduced-flux buffer) was split out into #2171, which has since landed in `development`; this branch is merged up to it, so the feature no longer appears in this PR's diff. `DTypeFront1D` consumes it to beam each half of its injection outward. What remains here from that area is the prose documentation of the hook in `docs/markdown/radiation_integrator.md`, plus the dust-solver fix below.

Docs: `docs/markdown/tests/DTypeFront1D.md` describes the problem, its analytic solution and every answer check; `docs/markdown/radiation_integrator.md` documents the `SetRadSource` hook and the scaling of both buffers.

### Known limitation: the O(v/c) work term

With `beta_order = 1` the work term is added to the gas internal energy inside the coupling solve and moved to kinetic energy afterwards by subtracting `dEkin_work`. The two are meant to cancel, but the work term is lagged from the previous outer iteration while `dEkin_work` comes from the freshly updated momentum, so they do not cancel exactly and the subtraction can drive `Egas` negative. A cap (`work_term_min_eint_fraction`) keeps a fixed fraction of the internal energy alive. **The cap does not conserve energy**, and it is retained here rather than fixed.

It is not a timestep symptom. Refining `DTypeFront1D` so `dt` halves at each step (coarse steps 16 / 32 / 64, radiation subcycles per step flat at 8.31), the capped *fraction* of radiation cell-updates is invariant:

| cells | capped cell-updates | total radiation cell-updates | fraction |
|---|---|---|---|
| 128 | 4,570 | 17,024 | 26.8% |
| 256 | 18,520 | 68,096 | 27.2% |
| 512 | 72,462 | 273,408 | 26.5% |

It is nonetheless benign: the cap can only bind where `dEkin_work` already exceeds the internal energy available, i.e. in cold cells the radiation has evacuated, so `Egas` and hence the discarded energy are minuscule regardless of how many cells are involved. `DTypeFront1D` closes its thermal-band energy budget to `1.0000000` with ~27% of its radiation cell-updates capping. The limitation is documented at both cap sites; no runtime counter is emitted, since that would mean an atomic on one address in a large fraction of every radiation kernel. The two candidate fixes — dropping the work term from the coupling step, or reformulating the solve in total gas energy — change answers for every `beta_order = 1` problem and are tracked in #2173 for a separate PR.

Verified that the cap is still load-bearing rather than vestigial: removing it changes the shipped answers, and a Debug build then aborts on `Assertion 'Eint > 0.' failed`. The same unclamped subtraction existed in the single-group solver and now carries the same guard.

### Related issues
N/A. Follow-up tracked in #2173 (O(v/c) work-term energy conservation).

### Checklist


BenWibking:
One neat trick I learned about from an old paper is that you can put the Newton solve for the radiation/matter energy much closer to the root if you do 1 nonlinear Picard iteration first. I'm not sure if that will help the issues you saw here, but I wanted to note it for the future.

chongchonghe:
ignore the ROCm error. That's a system failure. rocm-quick passed.

chongchonghe:
> One neat trick I learned about from an old paper is that you can put the Newton solve for the radiation/matter energy much closer to the root if you do 1 nonlinear Picard iteration first. I'm not sure if that will help the issues you saw here, but I wanted to note it for the future.

Good to know, although it won't help for the bugs fixed in this PR. 

The bugs are two fold:

1. roundoff errors, where the residual is below machine epsilon, so we have to set the correct absolute tolerance. 
2. an inaccurate Jacobian: I have been ignoring the partial B_g / partial T term (I really shouldn't), and that causes overshooting to negative `Egas` in extreme causes where multigroup opacities spans more than 4 orders of magnitude.

chongchonghe:
# PR 2126 review — mixed thermal + chemical bands, beamed radiation sources, `DTypeFront1D`

**Cycle:** feature PR → SPEC → VERIFY → EDGE.

**Environment:** ROCm/HIP box (no CUDA toolkit, no GPU device). CPU checks with preset `1d`; GPU checks are compile-only with preset `1d-hip`, which is what catches device-lambda and device-code restrictions. Base `origin/development` @ `9b1a0c97`; PR head `49157af7`.

## SPEC — the contract extracted from the PR and the diff

| # | Claim | How it is exercised |
|---|---|---|
| S1 | Chemistry bands may occupy the last `NChemBands` groups of a multigroup run; they emit no blackbody, are removed from the Newton solve, and their source is injected directly into `Erad` afterwards | `DTypeFront1D` groups 0–1 thermal, group 2 chemical; checks 1 and 5 |
| S2 | `SetRadSource` gains a `radFluxSource` buffer (`3*g + n`); `radFluxSource = c * radEnergySource` injects radiation with reduced flux 1 | check 3 |
| S3 | The flux source is scaled exactly like the energy source: `chat/c` for a thermal group, unscaled for a chemistry band | checks 1 and 5, and the `c/chat = 1000` ratio between the two shipped fluxes |
| S4 | Exact group Planck temperature derivative replaces `4 a T^3 f_g` | Jacobian only; verified analytically plus regression norms |
| S5 | Optically thin groups (`tau < 1`) switch their Newton unknown from `R_g` to `Erad_g`; converged answers unchanged outside round-off | regression norms of existing `Rad*` tests |
| S6 | `BackwardEulerOneVariable` brackets the root and guards Newton with bisection; tolerance falls back to `|f(x0)|` when the caller's scale is zero | `DTypeFront1D` runs with `dust_gas_interaction_coeff = 0`, which makes that scale identically zero |
| S7 | The solver injects the source into transparent groups (`tau == 0`), which it previously left unchanged | **see finding 1 — only true for the non-dust solver** |
| S8 | `enable_dE_constrain` now defaults to `false` — a global default affecting every radiation problem | regression sweep |
| S9 | New problem `DTypeFront1D` with five answer checks | direct run |

## Evidence

|---|---|---|
| VERIFY — golden path | `quokka config/build -d 1d DTypeFront1D`, then run with the shipped `inputs/DTypeFront1D.toml` | All five checks pass, reproducing the PR's quoted numbers: energy ratio `1.0000000`; front `0.33%` from `chat*t`; `1 - f = 4.9e-15`; reprocessed `0.552`, optical/analytic `1.039`; surviving ionizing fraction `0.198`, `3.2e21` absorbed ≥ `3.0e20` ionized column |
| VERIFY — S4/S5 algebra | Hand derivation against `radiation_system.hpp:629-670`, `:1594-1607`, `source_terms_multi_group.hpp:106-147`, `radiation_dust_system.hpp:88-122` | Both correct. Details below |
| VERIFY — GPU | `quokka config/build -d 1d-hip DTypeFront1D` | Compiles and links clean (`gfx90a`). The only two warnings come from `QuokkaSimulation.hpp:1268`, which this PR does not touch |
| EDGE — extreme opacity | `DTypeFront1D ... photoionize.kappa1=1.0e-2 photoionize.kappa2=1.0e4` (the case the PR reports as previously aborting) | Runs to completion, **no** `Newton-Raphson iteration for matter-radiation coupling failed to converge!`, and thermal-band energy conservation still exact (ratio `1`). The front/reprocessing checks fail, which is expected and not a defect: at `tau_front = 20` the beam is absorbed within the first cells, so those two checks are only meaningful at the shipped parameters |
| EDGE — transparent sourced group | `DTypeFront1D ... photoionize.kappa2=0.0` and `... kappa1=0.0 kappa2=0.0` | **Finding 1.** Reduced flux at the injection cell `f = 1.46e12`, and only `0.112` of the injected energy present |
| EDGE — regression sweep | `ctest -R "^(Rad\|DType)"` on both branches, first without `-ffp-exception-behavior=strict` | 24/29 fail on the PR branch — **and 23/28 fail identically on `origin/development`**. The backtrace puts the trap in `RadSystem::ComputeFluxes` on the base branch, i.e. hyperbolic transport code this PR does not touch. Environmental, not a PR regression. See the pre-existing note below |
| EDGE — regression sweep, corrected | Both branches rebuilt with `-DCMAKE_CXX_FLAGS=-ffp-exception-behavior=strict` (the flag `.ci/azure-pipelines-quick-amdgpu.yml` passes), then the same 6 tests plus `DTypeFront1D` | All pass on both branches. Error norms compared below |

### Error norms, base vs PR, identical build flags

| test | base (`9b1a0c97`) | PR (`49157af7`) | relative change |
|---|---|---|---|
| `RadMarshak` | 1.819322185e-2 | 1.819322185e-2 | bit-identical |
| `RadMatterCoupling` | 2.379512315e-7 | 2.379512315e-7 | bit-identical |
| `RadSuOlson` | 2.673767479e-2 | 2.673767479e-2 | bit-identical |
| `RadhydroPulseMGconst` | 4.194961044e-3 | 4.194961044e-3 | bit-identical |
| `RadDustMG` | 7.237950272e-4 | 7.237950248e-4 | 3e-9 |
| `RadhydroShockMultigroup` | 6.373328322e-3 | 6.373328028e-3 | 5e-8 |

This is exactly the signature claims S4/S5 predict, and it is the strongest single piece of evidence in this review. The single-group problems, which the rebasing and the new group Planck derivative cannot reach, come out bit-for-bit identical — which also shows the `enable_dE_constrain` flip (S8) changed nothing for them. The two multigroup problems move only in the 9th and 8th significant figure, consistent with a change that is an affine re-parametrization in exact arithmetic and differs only in round-off. `DTypeFront1D` also passes under the FPE traps, so the new problem raises no floating-point exception.

## Analysis

### S4 — the exact group Planck derivative is correct

Group emission is `e_g = a T^4 [P(x_{g+1}) - P(x_g)]` with `x = h nu / (k T)`, so `x_g` itself depends on `T` with `dx_g/dT = -x_g/T`. Differentiating,

$$\frac{\mathrm{d}e_g}{\mathrm{d}T} = a T^3 \left[ D(x_{g+1}) - D(x_g) \right], \qquad D(x) \equiv 4 P(x) - x P'(x) .$$

With `P` the normalized Planck integral, `P'(x) = (15/\pi^4) x^3/(e^x-1)`, hence `x P'(x) = (15/\pi^4) x^4/(e^x-1)`. That is exactly what `radiation_system.hpp:661` computes, `4. * integrate_planck_from_0_to_x(x) - (x^4/(exp(x)-1)) / gInf`, with `gInf = \pi^4/15` (`planck_integral.hpp:23`). Limits check out: `D(x) \to (15/\pi^4) x^3/3 > 0` as `x \to 0`, and `D(\infty) = 4` recovers `d(aT^4)/dT`. `D` is the integral of a positive kernel, so it is increasing and every group entry stays positive — which is what the Jacobian sign needs.

### S5 — `RebaseThinGroupsOntoErad` is consistent, including the cross term

With `F_g = E_g - E_{0,g} - R_g - \mathrm{Src}_g` and `F_0 = \Delta E_{gas} + (c/\hat c)\sum_g R_g - \dots`, the unrebased entries are `Jgg = -\kappa_{PoverE}/\tau - 1`, `Jg0 = \kappa_{PoverE} B'/c_v`, `J0g = c/\hat c`, `J00 = 1`. Substituting `R_g = (4\pi B_g/c - E_g/\kappa_{PoverE})\tau + w` and `scale = dR_g/dE_g = -\tau/\kappa_{PoverE}`:

- `dF_g/dE_g = Jgg \cdot scale = 1 + \tau/\kappa_{PoverE}` ✓ (matches direct differentiation)
- `dF_g/dx|_E = -\tau B'/c_v = Jg0 \cdot scale` ✓
- `dF_0/dE_g = J0g \cdot scale` ✓
- `dF_0/dx|_E = J00 + (c/\hat c)\sum_g \tau_g B'_g/c_v`, and the code's `J00 -= J0g[g]*scale*Jg0[g]` gives exactly that ✓ — and it correctly uses the *unscaled* `J0g`/`Jg0`, because the `J00` line precedes the three scaling lines in the loop body.

The same derivation holds for `ComputeJacobianForGasAndDustDecoupled` (there `x = T_d`, `J00 = 0`, `J0g = 1`, `Jg0 = \kappa_{PoverE} \cdot d(4\pi B/c)/dT_d`), because the identity the routine relies on, `dR_g/dx|_E = -scale \cdot Jg0`, follows from the shared structure of `F_g` in both Jacobians. Restricting the rebase to `dust_model == 2` is the right call: `ComputeJacobianForGasAndDust` has already eliminated the `T_d(\sum R_g)` coupling in the `R_g` unknowns, so its columns are not a plain change of variable.

Note that with `use_D_as_base = false` the rebase is applied on **every** multigroup radiation problem, not just the new one. That is the change with the widest blast radius in this PR, alongside S8.

### S6 — `BackwardEulerOneVariable`

The fallback `scale = max(compare, |f(x0)|)` is needed and correct: `ComputeDustTemperatureBateKeto` passes `Lambda_compare = N_d \sqrt{T_{gas}} T_{gas}`, which is identically zero when `dust_gas_interaction_coeff = 0` — precisely the configuration `DTypeFront1D` runs in — so the old `|f| < rel_tol * compare` test was unsatisfiable. The bracketing direction is right: `rhs` is decreasing in `T_d`, so `j0 < 0`, and `dir = (j0*f0 > 0) ? -1 : 1` reproduces the sign of the Newton step `-f/j`. The bisection guard is written as the negation of the in-bracket test so a NaN step also falls back to bisection — a good detail.

## PR-Introduced Concerns

### 1. The `tau == 0` source injection is missing from both dust solvers, leaving `|F| >> cE` in the state — **verified by running it**

The transparent-group injection

```cpp
for (int g = 0; g < nGroups_; ++g) {
    if (!(tau[g] > 0.0)) {
        EradVec_guess[g] = Erad0Vec[g] + Src[g];
    }
}
```

exists only in `SolveGasRadiationEnergyExchange` (`src/radiation/source_terms_multi_group.hpp:455-459`). Neither `SolveGasDustRadiationEnergyExchange` (`src/radiation/radiation_dust_system.hpp:606`) nor `SolveGasDustRadiationEnergyExchangeWithPE` (`:988`) has an equivalent. Meanwhile `UpdateFlux` adds `Src_flux` **unconditionally** for every group (`src/radiation/source_terms_multi_group.hpp:2069-2071` of the diff, i.e. `Frad_t0[n] = consPrev(...) + Src_flux[n][g]`).

So in any dust-enabled multigroup problem, a sourced group with zero opacity now receives the **flux** source but not the **energy** source. Before this PR both were dropped, which was at least self-consistent; the PR makes them inconsistent.

This is not hypothetical for the problem the PR adds. `kappa1` and `kappa2` in `testDTypeFront1D.cpp:101-102` both default to `0.0` and are only lifted by the shipped TOML, and `DTypeFront1D` uses `enable_dust_gas_thermal_coupling_model = true` with `dust_model == 2`, so it runs through `SolveGasDustRadiationEnergyExchange`. Running the shipped input with the optical opacity turned off:

```
$ DTypeFront1D ../inputs/DTypeFront1D.toml photoionize.kappa2=0.0
Thermal bands (IR + optical): 2.719606295e+10 (injected 2.423816461e+11)
Test FAILED: thermal-band energy is 0.1122034749 of injected (expected 1 within 1%).
Reduced flux at the injection cell: 1 - f = -1.45673215e+12 (f expected in [0.9, 1.000001])
Test FAILED: injected radiation exceeds the free-streaming limit (f = 1.45673215e+12); the flux source is too large.
```

`f = 1.5e12` is a conserved state with `|F|` twelve orders of magnitude above the free-streaming limit — exactly the failure mode check 3's upper bound was written to detect, and the check does detect it. The 11% of energy that does appear is generated downstream by transport acting on a flux with no matching energy, so the run is also not energy-conserving.

Suggested fix: mirror the same loop at the end of both dust solvers. Alternatively, gate `Src_flux` on the same condition the energy source uses, so the two can never diverge.

### 2. Chemical-band emission is zeroed but its temperature derivative is not (single-group path)

`src/radiation/source_terms_single_group.hpp:192` now reads

```cpp
fourPiBoverC = (nGroupsThermal_ == 0) ? 0.0 : ComputeThermalRadiationSingleGroup(T_d);
```

but `:292` still takes `d_fourpiboverc_d_t = ComputeThermalRadiationTempDerivativeSingleGroup(T_d)`, which returns `4 a T^3` unconditionally. For a single chemistry band (`nGroupsThermal_ == 0`) the residual therefore says emission is zero while the Jacobian says it grows as `4 a T^3`. The same asymmetry exists in the `nGroups_ == 1` branch of `ComputeThermalRadiationTempDerivativeMultiGroup` (`radiation_system.hpp:636-638`), though that branch is unreachable since `nGroups_ <= 1` dispatches to the single-group solver.

This is inert today: every existing chemistry-band problem — `DTypeFront`, `DTypeFrontVC`, `OneZonePhotoionization`, `StromgrenSphere`, `StromgrenSphereRSLA` — sets `nGroups = 1`, `NUM_CHEM_BANDS = 1` **and** `ComputePlanckOpacity = ComputeEnergyMeanOpacity = 0`, so `tau == 0` and the negligible-optical-depth branch runs instead. It becomes a wrong-Jacobian bug the moment anyone gives a chemistry band a nonzero opacity. Worth guarding the derivative the same way, so the two can't drift apart.

Related and worth calling out in the PR description: for those same five problems the emission term changes from `a T^4` to `0`. That is a genuine improvement — the single-group `ComputeThermalRadiationSingleGroup` returns the *whole* frequency-integrated `a T^4`, which has no business being fed into a Lyman-continuum band — but it is a physics change to five existing problems, currently masked only by their zero opacities. It deserves a line in the description.

### 3. `enable_dE_constrain = false` is a global default flip driven by one new problem

`radiation_system.hpp:45`. This safeguard is consumed at four sites — `source_terms_single_group.hpp:342`, `source_terms_multi_group.hpp:420`, `radiation_dust_system.hpp:589` and `:971` — so the flip reaches every radiation problem in the code base, single-group and multigroup alike, not only the dust ones. The PR does disclose it, and the 1D regression sweep is the evidence for it, but the 3D-only radiation problems (`RadhydroShell`, `StromgrenSphere*`, `DTypeFront`, `DTypeFrontVC`) are not covered by that sweep and should be exercised in CI before merge.

### 4. `radFluxSource` is allocated and zeroed for every radiation problem

`QuokkaSimulation.hpp:3172` allocates an `amrex::MultiFab` with `3 * nGroups` components and 3 ghost cells, and `setVal(0.0)` runs on it twice per radiation subcycle, whether or not the problem overrides `SetRadSource`. For the multigroup problems with many groups this is a real memory and bandwidth cost paid by problems that never use the feature. Not a correctness issue, but worth a thought — e.g. allocating lazily, or a trait that says whether the problem sets a flux source.

### 5. Smaller points

- `BackwardEulerOneVariable` can now spend up to 200 extra `rhs()` evaluations bracketing before it gives up, and for the multigroup case each `rhs()` evaluates the full Planck spectrum and the opacity model for every group. In the common case the bracket is found in a few steps, but the failure path got considerably more expensive, and it is per-cell and per-outer-iteration.
- The monotonicity argument the bracketing relies on ("emission and the collisional term both grow with the dust temperature") holds for the shipped opacity models, but `kappa_P(T_d)` is user-supplied and a steeply decreasing opacity law could in principle make `kappa_P * B` non-monotone. Fine as an assumption; worth stating as one in the comment.
- `fb` in `BackwardEulerOneVariable` is maintained but never read — hence the `amrex::ignore_unused(fb)` at the end. The bracket update only needs `fa`. Dropping `fb` would remove both the dead store and the suppression.
- The new round-off convergence path, `Fg_abs_sum < 10 * Fg_roundoff`, sums the allowance across all groups. A group with a very large `4 pi B_g / c` therefore buys headroom that can mask a genuine residual in a low-energy group in the same cell. The pre-existing relative test `cscale * Fg_abs_sum / Etot0 < resid_tol` has the same aggregate character, so this is consistent with the existing design rather than new, but it is a place where a mixed thermal + chemistry configuration — a chemistry band many orders of magnitude below the thermal blackbody — could be affected once chemistry bands carry nonzero opacity.
- The comment on the transparent-group injection says "Transparent groups do not couple to the gas (their Rvec contribution is zero)". With `beta_order_ == 1` and the work term enabled, a `tau0 == 0` group gets `Rvec[g] = work_local[g]`, not zero (`source_terms_multi_group.hpp:298`). Harmless for the problems in tree, but the comment overstates it.
- The chemistry band's radiation-group upper edge (`radBoundaries[3] = 1e19` Hz) and its photochemistry band edge (`CHEM_BANDS "3.29e15, 1.50e16"`) disagree. The header comment explains why this is harmless here, and it is — the group is transparent — but it means `E_photon` used for the source normalization is the mean over `[3.29e15, 1.5e16]` while transport treats the group as reaching `1e19` Hz.
- Cosmetic: `1 - f` is quoted as `6e-15` in the PR description, `4e-15` in the code comment, `3e-15` in `docs/markdown/tests/DTypeFront1D.md`, and I measured `4.9e-15`. It is a round-off-level quantity so any of these is fine, but one number in all three places would read better.

- `force_rad_floor_in_iteration = false` (`radiation_system.hpp:40`) means nothing keeps `EradVec_guess` non-negative during the Newton iteration, while `AMREX_ASSERT(min(EradVec_guess) >= 0.0)` fires at the end in a debug build. The rebasing makes `Erad_g` the directly-advanced variable for thin groups, which changes *how* it could go negative but neither introduces nor fixes the underlying gap. Pre-existing; does not block this PR; would need its own PR.
- `SetRadSource` in `testDTypeFront1D.cpp:304-319` **assigns** to `radEnergy` rather than adding, while the base-class comment says users should *add*. In 1D this is safe (`particleRegister_.depositRadiation` is `#if AMREX_SPACEDIM == 3`), and the pre-existing `DTypeFront`/`StromgrenSphere` overrides do the same, so this is an existing convention mismatch rather than something this PR introduces. Pre-existing; does not block this PR.
- Built with amdclang **without** `-ffp-exception-behavior=strict`, `origin/development` @ `9b1a0c97` fails 23 of 28 1D `Rad*` tests, all with an FPE trap inside `RadSystem::ComputeFluxes` at the first step. Adding the flag makes them pass. The AMD GPU CI pipeline passes that flag but `.ci/azure-pipelines-quick.yml` and the CPU presets do not, so a plain local `quokka config -d 1d` build cannot run the radiation test suite. Entirely pre-existing — reproduced on the base branch, in code this PR does not touch — but it cost real time to diagnose here and will cost it again for the next person. Would need its own PR; a note in the docs or a preset change would do.

## Verdict

The physics and the numerics hold up. I checked the two subtle derivations by hand and both are right: the exact group Planck derivative reduces correctly to `a T^3 [D(x_{g+1}) - D(x_g)]` with the stated closed form for `D`, and all four Jacobian updates in `RebaseThinGroupsOntoErad` — including the `J00` cross term and the ordering that makes it use the unscaled `J0g`/`Jg0` — are consistent with the residual definitions in both the gas and the decoupled-dust Jacobians. The base-vs-PR norm comparison confirms the claim that converged answers are unchanged: four tests bit-identical, two multigroup ones moved by 3e-9 and 5e-8. `DTypeFront1D` reproduces every number in the PR description, compiles for HIP, and the extreme-opacity case that motivated the Newton work runs to completion with exact energy conservation.

The one thing I would fix before merge is **finding 1**: the `tau == 0` energy-source injection went into `SolveGasRadiationEnergyExchange` only, while `UpdateFlux` applies the flux source unconditionally, so in the dust solvers a transparent sourced group gets flux without energy. `DTypeFront1D` is itself a dust-solver problem whose opacities default to zero in the source, and running the shipped input with `photoionize.kappa2=0.0` leaves the injection cell at reduced flux `1.5e12` and loses 89% of the injected energy. It is a few lines to mirror the loop into both dust solvers. Findings 2 and 3 are worth a look but need not block.

Two coverage gaps worth noting for CI rather than for the diff: the `enable_dE_constrain` flip and the chemistry-band emission change both reach 3D-only problems (`RadhydroShell`, `StromgrenSphere`, `StromgrenSphereRSLA`, `DTypeFront`, `DTypeFrontVC`) that the 1D sweep above cannot exercise.

chongchonghe:
**Verdict:** 2 blocking — the round-1 finding is unchanged and still reproduces, and the new work-term cap loses energy silently.
**Checked:** `1d` + `1d-hip` at `0b1c5dca`; 9/9 pass incl. the new `RadStreamingFluxSource`. Error norms bit-identical to `49157af7` on all six regression tests, and within 3e-9 / 5e-8 of `development`. Verified fixed: the `coeff_n == 0` branch guard and the cooling-Jacobian `1/c_v` — both real bugs.

**Blocking**

1. **`tau == 0` energy source dropped in both dust solvers, while `UpdateFlux` applies the flux source unconditionally** — `radiation_dust_system.hpp:611` and `:1000` have no equivalent of the loop at `source_terms_multi_group.hpp:455`. `DTypeFront1D photoionize.kappa2=0.0` (both opacities default to `0.0`, `testDTypeFront1D.cpp:114-115`) retains 11.2% of injected energy with `max |F|/(cE) = 1.48e12`. Round 1's reduced-flux bound caught this and was deleted; `RadStreamingFluxSource` replaces it on the gas-only path, which was already correct.
2. **Work-term cap is non-conserving with no diagnostic** — `source_terms_multi_group.hpp:603`. Its own comment says "not energy conserving", but there is no counter or warning, so a capped run still reports success. Hits every `beta_order = 1` multigroup problem.

**Non-blocking**

- **Chem-band emission zeroed but its derivative is not** — `source_terms_single_group.hpp:188` vs `:288`, still `4 a T^3`; inert only because every chem-band problem has zero opacity.
- **Cap missing from the single-group path** — `work_term_min_eint_fraction` is referenced only at `source_terms_multi_group.hpp:603`; the same negative-`Eint` route is still open there.
- **Stale comment** — `testDTypeFront1D.cpp:140` says `beta_order = 0`; the code sets `1`.

| Phase | Ran | Observed |
|---|---|---|
| Regression | `ctest` — `DTypeFront1D`, `RadStreaming`, `RadStreamingFluxSource`, `RadMarshak`, `RadSuOlson`, `RadDustMG`, `RadhydroPulseMGconst`, `RadMatterCoupling`, `RadhydroShockMultigroup` | 9/9 pass |
| Golden path | `DTypeFront1D`, shipped input | energy ratio 1, momentum ratio 1.003844, signed momentum 1.8e-17 of injected, reprocessed 0.367, optical/analytic 1.0032, surviving ionizing 0.316 |
| Norms | same 6 tests, base vs `49157af7` vs `0b1c5dca` | `RadMarshak` 1.819322185e-2 / 1.819322185e-2 / 1.819322185e-2; `RadDustMG` 7.237950272e-4 / 7.237950248e-4 / 7.237950248e-4; `RadhydroShockMultigroup` 6.373328322e-3 / 6.373328028e-3 / 6.373328028e-3; `RadSuOlson`, `RadMatterCoupling`, `RadhydroPulseMGconst` identical across all three |
| EDGE | `DTypeFront1D photoionize.kappa2=0.0` | `thermal-band energy is 0.1119759074 of injected`, `outward linear momentum is 0.1202021945 of injected`; instrumented (then reverted) `max |F|/(cE)`: group 1 = 1.484423574e+12 vs 0.999999997 shipped |
| GPU | `1d-hip` compile | clean; only pre-existing `QuokkaSimulation.hpp:1268` warning |

Note: the 1D radiation suite needs `-DCMAKE_CXX_FLAGS=-ffp-exception-behavior=strict` on this box — without it 23/28 tests trap an FPE in `ComputeFluxes` on `development` too. Pre-existing.

</details>

chongchonghe:
**Verdict:** Both round-2 blockers fixed and verified. 1 blocking left — the shipped test emits the new non-conservation warning on every run.
**Checked:** `1d` at `4d9ac2cc`, 9/9 pass. Reproducer `photoionize.kappa2=0.0` now gives energy ratio **1** (was 0.1119759074) and `max |F|/(cE)` = **1** (was 1.484423574e+12). Norms bit-identical across all four heads. Fix itself verified: `tau` is initialized where the new loop reads it (`radiation_dust_system.hpp:301,759`, set on the `n==0` pass), chem bands can't be double-injected (`Src[g]` zeroed upstream), and `iteration_failure_counter` was correctly resized 3→4 at its only allocation.

**Blocking**

1. **Shipped `DTypeFront1D` trips its own new warning** — `capped in 4576 cell-updates ... Energy is not conserved`, against ~17k radiation cell-updates (2048 zone-updates × 8.31 subcycles). Check 1 sums `Erad` only and the cap moves gas internal→kinetic, so no check can see the loss. Fires in none of the other 8 tests. Apply one of the remedies the warning names, or say in the header why it's acceptable.

**Non-blocking**

- **The fixed path has no CI coverage** — `DTypeFront1D/CMakeLists.txt:101` registers only the shipped TOML, where the new loop is a no-op (`tau > 0` for groups 0-1; group 2's `Src` is zeroed upstream), so Check 5 passes trivially at `max 1`. Deleting the injection loop again would keep the suite green. A `kappa2=0` variant needs Checks 2/4 and the momentum budget gated on `kappa2 > 0` — I ran it: Check 4 aborts by design and momentum lands at 1.0130 vs 1% tol, while Check 1 correctly gives 1.
- **Capped count isn't a cell count** — `source_terms_multi_group.hpp:874` increments per `UpdateFlux` call, which runs twice per subcycle plus once per work-term iteration; the message says "cell-updates".
- **Chem-band emission zeroed, derivative not** — `source_terms_single_group.hpp:188` vs `:288`, still `4 a T^3`. Carried from round 2, re-checked, unchanged.

| Phase | Ran | Observed |
|---|---|---|
| Regression | `ctest` — `DTypeFront1D`, `RadStreaming`, `RadStreamingFluxSource`, `RadMarshak`, `RadSuOlson`, `RadDustMG`, `RadhydroPulseMGconst`, `RadMatterCoupling`, `RadhydroShockMultigroup` | 9/9 pass |
| Golden path | `DTypeFront1D`, shipped input | energy 1, momentum 1.003844, reprocessed 0.3667, optical/analytic 1.0032, ionizing 0.3157, `max |F|/(cE)` = 1; work-cap warning, 4576 events |
| Fix verification | `DTypeFront1D photoionize.kappa2=0.0` | energy **1** (was 0.1119759074); `max |F|/(cE)` group 1 = **1** (was 1.484423574e+12) |
| Cap scope | `ctest -V`, grep the warning across the other 8 tests | fires in none |
| Norms | 6 tests, `development` / `49157af7` / `0b1c5dca` / `4d9ac2cc` | bit-identical across all four: `RadMarshak` 1.819322185e-2, `RadSuOlson` 2.673767479e-2, `RadMatterCoupling` 2.379512315e-7, `RadhydroPulseMGconst` 4.194961044e-3, `RadDustMG` 7.237950248e-4, `RadhydroShockMultigroup` 6.373328028e-3 (last two vs base 7.237950272e-4 / 6.373328322e-3) |

</details>

chongchonghe:
Thanks — round-2 findings addressed. Both blocking items are in as of 14ae3ba2.

**Finding 1 (`tau == 0` energy source missing from the dust solvers) — fixed.** The injection loop is mirrored into both dust solvers, so a transparent sourced group is no longer left at `Erad0` while `UpdateFlux` applies its flux source. Reproduced your edge case before and after:

| `photoionize.kappa2=0` | before | after |
|---|---|---|
| thermal-band energy | 0.1120 | 1.0000000 |
| `max \|F\|/(cE)`, sourced group | 1.48e12 | 1 |

The shipped configuration is unchanged and reproduces your instrumented numbers exactly (0.514 / 0.999999997 / 1).

**And the check you said was lost is back, permanently.** `DTypeFront1D` now carries a reduced-flux invariant over every cell and group, `|F| <= c E`. It is a property the M1 closure cannot violate rather than a tuned tolerance, and running with `photoionize.kappa2 = 0` exercises exactly the dust-solver path that `RadStreamingFluxSource` cannot reach.

**N1 (work-term cap) — measured rather than assumed, and the outcome is not what the old comment claimed.**

First, the cap is still load-bearing: removing it changes the shipped answers, and a Debug build then aborts on `Assertion 'Eint > 0.' failed`. Neither the Finding 1 fix nor the `coeff_n > 0` branch-selection fix obsoleted it.

Second, the claim that reaching it means the radiation timestep is too long is wrong. Refining so `dt` halves at each step (coarse steps 16 / 32 / 64, subcycles per step flat at 8.31), the capped *fraction* is invariant:

| cells | capped cell-updates | total radiation cell-updates | fraction |
|---|---|---|---|
| 128 | 4,570 | 17,024 | 26.8% |
| 256 | 18,520 | 68,096 | 27.2% |
| 512 | 72,462 | 273,408 | 26.5% |

So it is a structural limitation of the lagged O(v/c) work term, not user error, and the remedy the comment suggested does not work.

Third, on your specific ask for a counter: we tried it and then removed it. It reconciles with the energy budget closing to seven digits because the cap can only bind where `dEkin_work` already exceeds the internal energy available — cold cells the radiation has evacuated — so the discarded energy is minuscule however many cells are involved. Reporting it would mean an atomic on a single address in ~27% of every radiation kernel, which is not worth paying for in a production GPU run. Instead the limitation is documented at both cap sites and in the PR description with the numbers above, and the two candidate fixes (drop the work term from the coupling step, or reformulate the solve in total gas energy) are tracked in #2173 for a separate PR, since either changes answers for every `beta_order = 1` problem.

**N2 — not multigroup-specific, so fixed.** The single-group solver had the identical unclamped `Egas_guess -= dEkin_work` and now carries the same guard.

**N3 — fixed.** The stale `beta_order = 0` comment now describes `beta_order = 1` and why the momentum check needs it.

**Not addressed:** findings 2, 3, 5a–5d and N4, as non-blocking. One caveat worth recording: at `kappa2 = 0` the momentum budget reads 1.013 against a 1% tolerance, so the thin path is not yet a clean test variant — the reduced-flux invariant is what guards it for now.

**Verification.** 30/30 of the `Rad*` / `DType*` 1D suite pass. Error norms match your table; the one apparent difference (`RadMatterCoupling`, 8th digit) is our build, not this change — stashing the changes and rebuilding at `0b1c5dca` gives the same `2.379512088e-07` here, so it is the `-ffp-exception-behavior=strict` flag on your box.

chongchonghe:
### Update: round-2 review addressed, and merged up to `development`

Merged `development` at `2ee86233`, which now contains #2171. The flux-source plumbing and the six unrelated problem files it touched have dropped out of this PR's diff as a result — 23 files down to 14. One textual conflict in `source_terms_multi_group.hpp`, where our fused `src_scale` loop uses the `nGroupsThermal_` alias and additionally carries the chem-band decoupling block; kept ours, since `nGroupsThermal_ == nGroups_ - NChemBands` makes the two sides equivalent on the line they share.

**Finding 1 — fixed.** The `tau == 0` energy-source injection is mirrored into both dust solvers, so a transparent sourced group is no longer left at `Erad0` while `UpdateFlux` applies its flux source:

| `photoionize.kappa2=0` | before | after |
|---|---|---|
| thermal-band energy | 0.1120 | 1.0000000 |
| `max \|F\|/(cE)`, sourced group | 1.48e12 | 1 |

The shipped configuration is unchanged and reproduces your instrumented numbers exactly (0.514 / 0.999999997 / 1).

**The check you flagged as lost is back.** `DTypeFront1D` carries a reduced-flux invariant over every cell and group, `|F| <= c E` — a property the M1 closure cannot violate rather than a tuned tolerance. Running with `photoionize.kappa2 = 0` exercises the dust-solver path that `RadStreamingFluxSource` cannot reach.

**N1 — measured, and the old comment was wrong.** The cap is still load-bearing: removing it changes the shipped answers and a Debug build then aborts on `Assertion 'Eint > 0.' failed`. But it is not a timestep symptom. Halving `dt` at each refinement (coarse steps 16 / 32 / 64, subcycles per step flat at 8.31) leaves the capped *fraction* invariant:

| cells | capped cell-updates | total radiation cell-updates | fraction |
|---|---|---|---|
| 128 | 4,570 | 17,024 | 26.8% |
| 256 | 18,520 | 68,096 | 27.2% |
| 512 | 72,462 | 273,408 | 26.5% |

On your specific suggestion of a counter: we implemented one, then removed it. It would mean an atomic on a single address in ~27% of every radiation kernel, which is not worth paying for in a production GPU run — the existing failure counters are cheap only because they fire almost never. The reconciliation with a seven-digit energy budget is that the cap can only bind where `dEkin_work` already exceeds the internal energy available, i.e. cold cells the radiation has evacuated, so the discarded energy is minuscule however many cells are involved. That is now documented at both cap sites and in the PR description with the numbers above, and the two candidate fixes are tracked in #2173 for a separate PR, since either changes answers for every `beta_order = 1` problem.

**N2 — fixed.** Not multigroup-specific after all: the single-group solver had the identical unclamped `Egas_guess -= dEkin_work` and now carries the same guard.

**N3 — fixed.** The stale `beta_order = 0` comment now describes `beta_order = 1`.

**Not addressed:** findings 2, 3, 5a–5d and N4, as non-blocking. One caveat: at `kappa2 = 0` the momentum budget reads 1.013 against a 1% tolerance, so the thin path is not yet a clean test variant — the reduced-flux invariant is what guards it.

**Verification.** 30/30 of the `Rad*` / `DType*` 1D suite pass post-merge, and `DTypeFront1D`'s momentum ratio is bit-identical across the merge at 1.003844149. Error norms match your table; the one apparent difference (`RadMatterCoupling`, 8th digit) is our build rather than this change — stashing and rebuilding at `0b1c5dca` gives the same `2.379512088e-07` here, so it is the `-ffp-exception-behavior=strict` flag on your box.

markkrumholz (dismissed):

chongchonghe:
@markkrumholz Fixed a clang-tidy error and make planer DType test run in 3d as well. Can you approve again?

markkrumholz (approved):

PR #2138, opened 2026-07-31
Add RyuJones2aShockTube: MHD shock-tube correctness test with an exact solution

AstroKriel:
### Description
This PR adds `RyuJones2aShockTube`, the MHD Riemann problem from Figure 2a of Ryu & Jones (1995). Unlike `BrioWuShockTube`, this shock tube has an exact solution, so it probes a complementary shock regime for MHD correctness.

Computing that exact solution means solving the MHD Riemann problem, which I do with a standalone tool, [MHDRiemannSolver](https://github.com/AstroKriel/MHDRiemannSolver). Since correctness here is checked against that external solver rather than an in-repo analytic formula, this problem is registered with `ADD_TEST OFF` (buildable and runnable, like `ShockCloud`/`DiskGalaxy`, but not part of the automated ctest suite).

Figure below compares Quokka's output against the exact solution (this is from the MHD paper where we compare different EMF scheme combinations).

ryu-jones-2a-shock-tube-scheme-comparison

### Related issues
N/A.

### Checklist
- [x] I have added a link to any related issues (if applicable; see above) — n/a, see Related issues.
- [ ] I have added tests for any new physics that this PR adds to the code — n/a, correctness verified externally against MHDRiemannSolver, not via ctest (see Description).

## Summary by CodeRabbit

- **New Features**
  - Added the Ryu–Jones 2a magnetohydrodynamic shock-tube problem for three-dimensional simulations.
  - Included configurable initial and boundary conditions, adaptive mesh refinement, timestep controls, and MHD evolution settings.
  - Added a ready-to-use configuration for running the shock-tube problem.

- **Documentation**
  - Expanded and refreshed the test-problem catalog with additional hydro, MHD, radiation, dust, particle, gravity, and passive-scalar scenarios.
  - Corrected capability details and clarified convergence-specific problem names.


chongchonghe:
# PR #2138 Review — `RyuJones2aShockTube`

**Type:** feature (new problem) → SPEC → VERIFY → EDGE cycle
**Base:** `origin/development` @ `76ec71c1` · **PR head:** `5454d819`
**Platform:** Linux x86_64, CPU preset `3d` (Release), GPU preset `3d-hip` (ROCm 6.4.4, gfx90a — compile only, no device on this box)

**Verdict: approve.** The problem reproduces the RJ2a solution to 0.05 % on every intermediate state at 512 cells (0.35 % at 128), at two resolutions and with both the shipped and the default EMF schemes. All findings below are minor/cosmetic; none block merge.

---

## SPEC — contract extracted from the PR description and diff

| # | Claim | Source |
|---|---|---|
| S1 | New 3D-only problem `RyuJones2aShockTube`, buildable, not in the ctest suite (`ADD_TEST OFF`) | `src/problems/RyuJones2aShockTube/CMakeLists.txt` |
| S2 | Initial state = Ryu & Jones (1995), ApJ 442, 228, Fig. 2a: γ=5/3; L = (ρ,p,vx,vy,vz) = (1.08, 0.95, 1.2, 0.01, 0.5), R = (1, 1, 0, 0, 0); Bx = 2/√(4π) constant, By_L = 3.6/√(4π), By_R = 4/√(4π), Bz = 2/√(4π) constant | `testRyuJones2aShockTube.cpp:46–61` |
| S3 | Runs to `stop_time = 0.2` on `[0,1]³`, 512×16×16, `ext_dir` in x + periodic in y,z, discontinuity at x = 0.5 | `inputs/RyuJones2aShockTube.toml` |
| S4 | Dirichlet BCs hold the L/R states fixed on both x faces (cell-centred and face-centred) | `testRyuJones2aShockTube.cpp:147–200` |
| S5 | Output matches the exact MHD Riemann solution (validated externally against `MHDRiemannSolver`, not in-repo) | PR description |
| S6 | `docs/.../table_of_all_problems.md` regenerated with `parse_problems.py` | PR commit `7ffee4b6` |

---

## Evidence

|---|---|---|
| VERIFY — build (CPU) | `quokka config -d 3d --delete …` + `quokka build -d 3d RyuJones2aShockTube -j 16` | `RyuJones2aShockTube SUCCESS` (S1) |
| VERIFY — build (GPU) | `quokka config -d 3d-hip --delete …` + `quokka build -d 3d-hip RyuJones2aShockTube -j 12` | `RyuJones2aShockTube SUCCESS` for `gfx90a`. Two warnings, both from `src/QuokkaSimulation.hpp:1214` (`-Wgpu-maybe-wrong-side`), i.e. pre-existing core code, none from the new file |
| VERIFY — golden path (S2/S3/S4) | `RyuJones2aShockTube inputs/RyuJones2aShockTube.toml` → 766 steps to t = 0.2, 756 s serial | Ran clean. Slice at the outermost cells reproduces the RJ2a L/R states to ≲1e-14: L = ρ 1.08, P 0.95, v (1.2, 0.01, 0.5), By 1.0155412503859613; R = ρ 1.0, P 1.0, v 0, By 1.1283791670955126. Bx = 0.56418958354775628 in **every** cell (spread 0.0e+00) — matches 2/√(4π) to the last bit |
| VERIFY — wave structure (S5) | 1-D slice of `plt0000766` via `amrex_fextract`, discontinuity detector on relative jumps in (ρ,P,v,B) | All **7** RJ2a waves resolved in the right order: fast shock (x≈0.32) · rotational discontinuity + slow shock (x≈0.54–0.56) · **contact** (x≈0.607) · slow shock + rotational discontinuity (x≈0.70) · fast shock (x≈0.96). The contact shows the correct signature — Δρ = −0.166 with ΔP = −0.0017, Δvx = +0.0013, ΔBy = +0.0016 |
| VERIFY — quantitative (S5) | Plateau means of ρ between the waves, P recovered from `gasEnergy − Ekin − Emag` | See table below — agreement with the commonly quoted RJ2a intermediate densities is ≤ 0.05 % everywhere at 512 cells |
| VERIFY — docs (S6) | `python3 docs/markdown/tests/parse_problems.py`, then `diff` against the committed file | Byte-identical. The 186-line table churn is genuine script output, not hand editing |
| VERIFY — CI input validation | `python3 scripts/python/check_inputs_toml.py` | `validated 173 TOML input files`, exit 0 |
| EDGE — resolution | Same input at 128×16×16 | Same plateaus (see table); no negative ρ or P |
| EDGE — non-default EMF scheme | 128×16×16 with `mhd.emf_*_scheme` **removed** (falls back to `FelkerStone2017`/`LondrilloDelZanna2004`) | Ran clean; same plateaus to ≤ 0.2 % of the Balsara2025 run. The problem is not tied to the `Balsara2025` schemes the input file selects |
| EDGE — off-face discontinuity | 125×16×16 (`blocking_factor_x = 1`), 1 step, inspected `plt0000000` | Deliberately puts x = 0.5 inside a cell, so the cell-centred and face-centred initialisers disagree for that one cell. Total energy stays exactly consistent (`Etot − eint − Ekin − Emag = 0`) because `postInitialization()` rebuilds it — see analysis. Not a bug |
| EDGE — adjacent regression | `quokka buildrun -d 3d BrioWuShockTube -j 16` | `BrioWuShockTube SUCCESS`, 874 steps, 832.7 s |

### Intermediate densities

| Region | RJ95 reference | 512 cells, Balsara2025 | 128 cells, Balsara2025 | 128 cells, default EMF |
|---|---|---|---|---|
| L (upstream) | 1.0800 | 1.0800 | 1.0800 | 1.0800 |
| behind left fast shock | 1.4903 | 1.4904 | 1.4901 | 1.4900 |
| behind left slow shock | 1.6343 | 1.6339 | 1.6292 | 1.6291 |
| behind contact | 1.4735 | 1.4729 | 1.4732 | 1.4707 |
| behind right slow shock | 1.3090 | 1.3089 | 1.3090 | 1.3091 |
| R (upstream) | 1.0000 | 1.0000 | 1.0000 | 1.0000 |

The reference column is the set of intermediate densities commonly quoted for RJ2a in the MHD-code literature; it is an independent cross-check, not a substitute for the author's `MHDRiemannSolver` comparison. Across the contact the numerical solution keeps P (1.9319 vs 1.9291) and B (1.4128/0.4377 vs 1.4138/0.4381) continuous while ρ jumps 1.6341 → 1.4717, which is exactly the contact-discontinuity condition.

---

## Analysis

**Does the implementation match the stated design? Yes.** The file is a faithful clone of `testBrioWuShockTube.cpp` with the RJ2a state vector substituted, plus the non-zero `vy`/`vz` and the third field component that BrioWu does not have. Everything RJ2a-specific is correct:

- `Bx = Bz = std::numbers::inv_sqrtpi` is exactly 2/√(4π) = 1/√π (`testRyuJones2aShockTube.cpp:58,61`).
- `By_L = 1.0155412503859613` and `By_R = 1.1283791670955125` are the correctly rounded doubles for 3.6/√(4π) and 4/√(4π) — I checked both against `3.6/math.sqrt(4*pi)` and `4/math.sqrt(4*pi)`; the second literal round-trips to the same double as `…126`, so there is no precision loss.
- `mhd.emf_averaging_scheme`/`emf_compute_scheme = "Balsara2025"` are valid enumerators (`src/hydro/mhd_system.hpp:28`, `AMREX_ENUM(EMFAvgScheme, LondrilloDelZanna2004, Balsara2025)`), and the EDGE run shows the problem also works without them.
- The domain is large enough for the run time: the leading fast waves reach x ≈ 0.32 and x ≈ 0.96 at t = 0.2, so nothing interacts with the `ext_dir` boundaries and the fixed Dirichlet states never over-constrain the solution.
- `ADD_TEST OFF` follows established precedent (`ShockCloud`, `DiskGalaxy`, `RandomBlast`, `NscbcVortex`, `DustyOrszagTang`, …) and the problem is still glob-picked by `src/problems/CMakeLists.txt`, so CI compiles it on every build — that is where the passing CUDA/HIP jobs on this PR come from.

**One thing worth knowing for anyone reading this file:** the `Emag` term added to `energy_index` at `testRyuJones2aShockTube.cpp:112` is dead. `AMRSimulation::setInitialConditions()` → `QuokkaSimulation::postInitialization()` calls `updateInitialMagneticEnergyFromFaceField()` (`src/QuokkaSimulation.hpp:1870`, called from `postInitialization()` at `:1936`), which overwrites `energy_index` with `eint + Ekin + Emag(face-averaged B)`. I confirmed this empirically with the 125-cell run: in the one cell where the cell-centred branch picks the right state but the face-centred branch picks `By_L`, the stored total energy comes out consistent with `By_L`, i.e. with the face field, not with the value the problem file wrote. So the pressure in this problem is set solely by `internalEnergy_index`, and the cell-centred/face-centred coordinate mismatch discussed under "concerns" cannot corrupt it.

---

## PR-Introduced Concerns

All minor. Ordered by how much they matter.

1. **`plotfile_prefix = "snapshots/plt"` is not cleaned by repo tooling** (`inputs/RyuJones2aShockTube.toml:30`). This is the only one of the 173 input files that writes plotfiles into a subdirectory — the other 12 files that set a prefix all write into the CWD. `quokka clean` runs `rm -rf plt* slice* chk* part_* *.csv *.pdf *.png Backtrace.*` inside `tests/`, so `tests/snapshots/` survives and accumulates across runs. (`.gitignore` does cover it via the blanket `/tests/` rule, so this is housekeeping, not repo hygiene.) Either use a flat prefix like `"rj2a_plt"` or add `snapshots` to `clean_tests()` in `scripts/bash/quokka`.

2. **Runtime is not the reason to keep `ADD_TEST OFF`.** The 512-cell run took **756 s** serial here, against **833 s** for `BrioWuShockTube`, which *is* registered as a ctest and likewise has no in-code error check — it only asserts that the run completes. Registering RJ2a the same way would cost less than the MHD test already in the suite and would catch NaNs/crashes/regressions in the MHD solver on this shock regime. Worth reconsidering, or at least adding a reduced-resolution variant. As it stands the `quick`/`rocm-quick` pipelines can only *compile* this problem, so the "triggered GPU tests for changed problems" checklist item gives build coverage only.

3. **Dead includes** (`testRyuJones2aShockTube.cpp:11,14,16,17,19,20,27,29`): `matplotlibcpp.h`, `math/interpolate.hpp`, `<format>`, `<fstream>`, `<string>`, `<unordered_map>`, `util/ArrayUtil.hpp`, `util/fextract.hpp`. `problem_main()` only calls `setInitialConditions()` and `evolve()`, so none of these are used. Inherited verbatim from `testBrioWuShockTube.cpp`, but since this is a brand-new file it is a good moment to drop them.

4. **`RadSystem<>` indices in the boundary condition, `HydroSystem<>` indices in the initial condition** — same file, two conventions (`:164–169`, `:176–181` vs `:108–113`). They alias to the same values so the behaviour is correct, but radiation is disabled for this problem and `HydroSystem<>` would read consistently with the rest of the file. Inherited from BrioWu.

5. **Redundant zero-fill** at `:173–175`: `high_bdr_cells` is already value-initialised by `{}` on the preceding line, and `low_bdr_cells` at `:163` correctly relies on that. Inherited from BrioWu.

6. **Cell-centred vs face-centred x convention** in `setInitialConditionsOnGridFaceVars` (`:126`): `x1_L = prob_lo[0] + i * dx[0]` is the correct coordinate for the x-face component, but for the y- and z-face components the x location is the cell centre, `prob_lo[0] + (i + 0.5) * dx[0]`. With the shipped input (512 cells on `[0,1]`) x = 0.5 lands exactly on a cell face and both conventions give identical answers; my 125-cell EDGE run shows they diverge by one cell when the discontinuity falls inside a cell. Harmless here and harmless for the total energy (see analysis), but it means the problem is only safe to re-run at resolutions where 0.5 stays on a face. Inherited from BrioWu — worth fixing in both files together rather than in this PR.

---

- **`gasInternalEnergy` in the plotfile is not the thermodynamic internal energy at shocks.** With `hydro.use_dual_energy = 0` the separately advected `internalEnergy_index` drifts from `gasEnergy − Ekin − Emag` by up to **2.3 %** (median 0.68 %), peaking at the contact at x ≈ 0.614. Anyone plotting pressure for this shock tube must derive it from `gasEnergy`, not from `gasInternalEnergy`, or the intermediate-state pressures come out wrong by a couple of percent. This is general Quokka dual-energy behaviour, not something this PR introduces; it is called out only because it directly affects how one analyses this new problem. Pre-existing, does not block this PR, needs a separate PR (or a documentation note) if it is worth addressing at all.

- **`-Wgpu-maybe-wrong-side` in the HIP build** at `src/QuokkaSimulation.hpp:1214` — "capture host side class data member by this pointer in device or host device lambda function". Present on `development` too, unrelated to this PR. Pre-existing, does not block this PR, needs a separate PR.

---

chongchonghe:
LGTM. 

The "PR-Introduced Concerns 1" is the only one worth addressing, but should not block this PR. The test is turned off anyway.

chongchonghe (approved):

chongchonghe (approved):

PR #2143, opened 2026-08-01
Add MHDSmallScaleDynamo: turbulence-driven small-scale dynamo test problem

AstroKriel:
### Description
This PR adds `MHDSmallScaleDynamo`, an ideal-MHD test problem for the turbulent small-scale dynamo (SSD): pure solenoidal turbulent driving amplifies a weak seed magnetic field through kinematic exponential growth until it saturates.

The setup uses isothermal turbulence driven by the `TurbGen` library (Federrath), with a single-mode ABC (Arnold-Beltrami-Childress) seed field so the initial magnetic energy is concentrated at one wavenumber. No explicit resistivity or viscosity is applied; dissipation is purely numerical.

Since there's no exact analytic solution for turbulence-driven SSD growth/saturation to compare against, this problem is not part of the automated ctest suite.

I validated this test on a 256^3 domain driven over `k in [1,3]`, and find the correct SSD evolution and signatures. A clean kinematic exponential growth phase is resolved followed by the nonlinear backreaction phase, and then finally a sustained saturated state, with the magnetic power spectrum peaking on the smallest scales.

ssd

### Related issues
N/A.

### Checklist
- [x] I have added a link to any related issues (if applicable; see above): n/a, see Related issues.
- [ ] I have added tests for any new physics that this PR adds to the code: n/a, no exact analytic solution exists for turbulence-driven SSD growth/saturation; validated via a 256^3 GPU run showing physically consistent kinematic growth and saturation (see Description).

## Summary by CodeRabbit

* **New Features**
  * Added a three-dimensional small-scale dynamo magnetohydrodynamics simulation.
  * Added configurable periodic geometry, adaptive resolution, timestep limits, resistivity, viscosity, output intervals, and turbulent driving.
  * Added ABC seed magnetic-field initialization with configurable wavenumber and field fraction.
  * Added isothermal fluid behavior and deterministic turbulence seeding.
  * Added validation and warnings for seed-field and driving parameters.
* **Tests**
  * Registered the simulation for three-dimensional builds.


AstroKriel:
Note that `quick` fails with `No ctest tests found for changed problems`, which is not actually due to an issue with this PR: `MHDSmallScaleDynamo` intentionally registers no ctest (`ADD_TEST OFF`), and `azure-pipelines-quick.yml`'s test-name resolution doesn't yet handle that case gracefully.

chongchonghe:
# PR 2143 review — `MHDSmallScaleDynamo`

Reviewed as a **feature PR** (SPEC → VERIFY → EDGE). The PR adds three new files and touches no existing source, so the regression surface is structurally nil; the review therefore concentrates on whether the new problem delivers the contract its own input file documents.

Platform: CPU build `3d` (GCC, Release) and GPU build `3d-hip` (ROCm, `gfx90a`). No CUDA toolkit and no GPU device on this box, so the HIP preset was **compiled** but not run; GPU runtime behaviour remains CI's job.

## SPEC — the contract extracted from the PR description, the diff, and `inputs/MHDSmallScaleDynamo.toml`

| Input | Documented meaning | Verified? |
|---|---|---|
| `setup.seed_b_wavenumber` | integer mode $n$, so $k_0 = 2\pi n / L$; must be a positive integer | yes |
| `setup.seed_b_fraction` | "$E_\mathrm{mag}/E_\mathrm{kin}$ at $t=0$; $B_0 = \sqrt{\mathrm{fraction}} \times$ `target_vdisp`" | **no — off by a factor of 3** |
| `turbulence.*` | solenoidal parabolic driving, $k \in [1,3]$, $t_\mathrm{decay} = L/(k_\mathrm{driv} v)$ | yes |
| `mhd.resistivity`, `hydro.shear_viscosity`, `hydro.bulk_viscosity` | all zero, so ideal MHD with purely numerical dissipation | yes |

| Output / effect | Verified? |
|---|---|
| uniform $\rho = 1$, $v = 0$, isothermal $c_s = 1$ | yes |
| face-centred ABC seed field built as a discrete curl, hence $\nabla \cdot B = 0$ to machine precision | yes |
| excluded from the automated ctest suite (`ADD_TEST OFF`) | yes |
| built only for `AMReX_SPACEDIM == 3` | yes |
| kinematic exponential growth → saturation of the small-scale dynamo | **not testable locally** — see Analysis |

## Evidence

|---|---|---|
| VERIFY — CPU build | `quokka config/build -d 3d MHDSmallScaleDynamo` | `MHDSmallScaleDynamo SUCCESS` |
| VERIFY — GPU build | `quokka config/build -d 3d-hip MHDSmallScaleDynamo` | `MHDSmallScaleDynamo SUCCESS` for `gfx90a`. The only warning is `QuokkaSimulation.hpp:1267` (`-Wgpu-maybe-wrong-side`), which is pre-existing core code this PR does not touch. No warning originates in `testMHDSmallScaleDynamo.cpp`, so the two new `AMREX_GPU_DEVICE` lambdas are clean — they capture only value types (`GpuArray` for `dx`/`prob_lo`, the `quokka::direction` enum, and host-computed scalars), never a host pointer. |
| VERIFY — golden path | shipped TOML reduced to $32^3$ / 60 steps | ran cleanly to $t = 0.462$, exit 0, no NaN, no abort. TurbGen echoed exactly the configuration the input file claims: parabola spectrum, $k_\mathrm{min}=1$, $k_\mathrm{max}=3$, characteristic $k=2$, solenoidal weight $1.0$, autocorrelation time $1.0 = L/(k_\mathrm{driv} v)$. |
| VERIFY — claim "$\nabla \cdot B = 0$" | independent numpy re-implementation of the exact face formulas from `testMHDSmallScaleDynamo.cpp`, at $N = 32$ and $N = 256$ | $\max(\Delta x \, \lvert \nabla \cdot B \rvert / \lvert B \rvert) = 2.1 \times 10^{-16}$, against Quokka's own projection tolerance of $2 \times 10^{-14}$. The discrete curl telescopes exactly because the same edge-centred $A$ values are shared between neighbouring faces, so the seed field is divergence-free by construction rather than by projection. |
| VERIFY — claim "$E_\mathrm{mag}/E_\mathrm{kin} = $ `seed_b_fraction`" | Quokka's own `Initial gas+radiation energy` print, cross-checked against numpy | Code prints $3.702055615\times10^{-11}$; numpy gives $3.702056\times10^{-11}$ — an exact match, so the re-implementation is faithful. But with `seed_b_fraction` $=10^{-10}$ and `target_vdisp` $=0.5$ the ratio is $E_\mathrm{mag}/E_\mathrm{kin} = 2.96\times10^{-10}$ at $N=32$ and $2.9994\times10^{-10}$ at $N=256$, i.e. **3× the documented value**. |
| VERIFY — `ampl_factor` retuning | read `extern/turbulence/TurbGen.h:340-352` and `:419` | Correct, and worth recording because the naive Kolmogorov argument suggests otherwise. TurbGen raises the user's `ampl_factor` to the 1.5th power, so the injection rate scales as $f^3$ and the steady-state $v_\mathrm{rms}$ as $f^1$. TurbGen's own auto-adjust does exactly `ampl_factor *= (v_target/v_current)^1.5` on the already-cubed internal value, which is linear in the user-facing factor. The author's retuning $0.5 \times (0.5/0.302) = 0.828$ therefore follows the library's own prescription. |
| EDGE — `seed_b_wavenumber = 0` | run | Assert fires exactly as documented: `Assertion 'seed_b_wavenumber > 0' failed ... setup.seed_b_wavenumber must be a positive integer`, SIGABRT. |
| EDGE — `seed_b_wavenumber = 4` | run | Runs clean. Initial magnetic energy $3.561\times10^{-11}$ against $3.702\times10^{-11}$ for $n=2$ — the seed amplitude is $k$-independent as designed ($a \propto 1/k$ and $B = kA$), the residual 4% being the $O((k\Delta x)^2)$ truncation of the discrete curl at $32^3$. |
| EDGE — `setup.seed_b_fraction` omitted | run | **Silent degenerate run.** Initial magnetic energy is exactly $0$, `relative conservation error = nan`, and the simulation runs to completion with exit 0 and no warning at all. |
| EDGE — ctest exclusion | `ctest -N` in `build/3d` | 135 tests, `MHDSmallScaleDynamo` absent. `ADD_TEST OFF` works. |
| EDGE — dimensionality guard | `quokka config -d 2d`, then `ninja -t targets all` | No executable target in the 2d configuration; the link rule `CXX_EXECUTABLE_LINKER__MHDSmallScaleDynamo_Release` exists only in `build/3d`. CMake still creates an empty per-directory build folder in 2d, which is its normal behaviour for every guarded problem and is harmless. |
| EDGE — adjacent regression | `ctest -R "^(MHDBlast\|Turbulence)$"` in `build/3d` | `Turbulence` **Passed** in 196 s, with velocity dispersion $0.1961$ against a target of $0.2$, i.e. a relative error of $0.0197$ against a tolerance of $0.075$. This is the most relevant neighbour because it exercises the same TurbGen driving path that this problem depends on. `MHDBlast` hit ctest's built-in 1500 s timeout on this single-core CPU box, which is a wall-clock limitation of the machine rather than a signal about the PR — the diff is purely additive (3 new files, 230 insertions, 0 deletions, `git diff --name-status` reports `A` for all three), so no existing test can regress. |

## Analysis

The implementation matches the PR's stated design. The vector potential is the standard ABC form with all three amplitudes equal, $A = a(\sin kz + \cos ky,\ \sin kx + \cos kz,\ \sin ky + \cos kx)$, which is force-free with $B = \nabla \times A = kA$. Taking the curl discretely on the staggered grid, rather than sampling $B$ analytically at face centres, is the right choice: it makes the initial field divergence-free to round-off, so the MAC projection in `QuokkaSimulation::projectFaceCenteredMagneticField` has nothing to correct. The face-variable indexing convention (`Physics_Indices<...>::mhdFirstIndex` branched on `grid_elem.dir_`) matches `MHDBlast` and the other MHD problems exactly.

The cell-centred initial conditions, including setting `energy_index` and `internalEnergy_index` to zero under an isothermal $\gamma = 1$ EOS, are identical to the existing `Turbulence` problem, so they follow established practice in this repo.

Declaring `ResistivityModel::constant` and `ViscosityModel::constant` in the traits is required, not gratuitous: `QuokkaSimulation.hpp:663` asserts if `mhd.resistivity` appears in the input file under any other model. With the coefficients at zero the PR's "ideal MHD" claim genuinely holds, because `QuokkaSimulation.hpp:285` and `:296` gate the diffusion solvers on the coefficient being non-zero, and the parabolic timestep limits at `:915` and `:934` are gated the same way. So the zero-valued keys cost nothing.

The one claim I could not test is the physics headline: kinematic exponential growth followed by saturation. That needs a magnetic Reynolds number above the dynamo's critical value, which numerical-dissipation-limited schemes typically reach only around $256^3$ and above, and it needs many turnover times. Neither is affordable on a CPU here, and a $32^3$ run would show decay for reasons that have nothing to do with this PR. The author's own $256^3$ GPU validation, including the magnetic power spectrum peaking at the grid scale, is the appropriate evidence, and the plot in the PR description is consistent with a standard SSD. I take no issue with `ADD_TEST OFF` given there is no analytic solution to compare against.

## PR-Introduced Concerns

**1. `setup.seed_b_fraction` is documented as the initial magnetic-to-kinetic energy ratio, but delivers three times that.** `inputs/MHDSmallScaleDynamo.toml:47` says `E_mag/E_kin at t=0; B0 = sqrt(fraction) * target_vdisp`. For the ABC field with equal amplitudes, $\langle \lvert B \rvert^2 \rangle = 3 k^2 a^2$, not $k^2 a^2$, because each of the three components contributes unit mean square. With `seed_vecpot_amplitude` $= \sqrt{f}\,v/k$ at `testMHDSmallScaleDynamo.cpp:83` this gives $\langle \lvert B \rvert^2 \rangle = 3 f v^2$, and since Quokka defines the magnetic energy as $\tfrac{1}{2}B^2$ (`hydro_system.hpp:691`), the delivered ratio is $E_\mathrm{mag}/E_\mathrm{kin} = 3f$. I measured $2.9994 f$ at $N = 256$, converging to exactly $3$. A $3\times$ error on a $10^{-10}$ seed has no bearing on the science, and the growth rate is unaffected, but the comment states a contract the code does not honour, and anyone using this to set a specific initial ratio will be wrong by $3\times$. Either divide the amplitude by $\sqrt{3}$, or reword to $B_\mathrm{rms} = \sqrt{3f} \times$ `target_vdisp`.

**2. Omitting `setup.seed_b_fraction` or `turbulence.target_vdisp` silently produces a dynamo run with no magnetic field.** At `testMHDSmallScaleDynamo.cpp:65-67` the file-scope defaults are `seed_b_fraction = 0.0` and `target_vdisp = 0.0`, and both are read with `ParmParse::query`, which leaves the default in place when the key is absent. The comment on line 64 says "defaults match input file", but the input file ships `1.0e-10` and `0.5`, so they do not. Because $B = 0$ is an exact fixed point of the induction equation, the run then completes normally as a pure hydrodynamic turbulence simulation, with zero magnetic energy forever and no diagnostic that anything is wrong — I confirmed exit 0 with `Initial gas+radiation energy = 0`. Since the file already asserts `seed_b_wavenumber > 0`, the consistent fix is an `AMREX_ALWAYS_ASSERT_WITH_MESSAGE(seed_b_fraction > 0.0, ...)` alongside it, and either the same for `target_vdisp` or defaults that genuinely match the TOML. Correcting the comment on line 64 either way would help.

**3. The seed-field box length is taken from the $x$-extent alone, decoupled from `turbulence.length`.** `testMHDSmallScaleDynamo.cpp:81` computes `box_length = prob_hi[0] - prob_lo[0]`, while the driving takes its box size from the independent `turbulence.length` key. In the shipped cubic unit box these agree, but a user who edits `geometry.prob_hi` without also editing `turbulence.length` — or who tries a non-cubic box — gets a seed field and a driving field referring to different box sizes, with no warning. Low severity given the intended use, but cheap to guard with an assert that the two agree.

**Energy "conservation error" is meaningless for isothermal driven runs.** The end-of-run summary reports `relative conservation error = 3.08e8` for `gasEnergy`. Under $\gamma = 1$ the energy equation is not solved, yet `turbulentDriving::applyDriving` still accumulates `dE` into `energy_index` (`TurbulentDriving.hpp:97`), so the diagnostic compares a field that is never evolved against one that is monotonically incremented. I confirmed this is pre-existing rather than something the new problem introduces: the existing `Turbulence` test prints the same `Initial gasEnergy = 0` followed by a non-zero absolute conservation error, and it still passes. It does not block this PR and would need a separate PR to address.

## Verdict

The setup is correct where it matters most: the seed field is divergence-free to round-off, the driving configuration matches what the input file documents, the GPU build is clean, the dimensionality guard and ctest exclusion both work, and the ideal-MHD claim holds. Concern 1 is a one-line documentation or normalisation fix, and concern 2 is a two-line assert that would turn a silent no-op into an immediate error. Neither affects the validation shown in the PR description. I would be happy to see this merged once concerns 1 and 2 are addressed.

markkrumholz (changes requested):
The changes look fine. Please address the sqrt(3) issue the AI review found (no need to change code, just explicitly state that the initialized strength is per-component in the comments / documentation), and then I will approve.

AstroKriel:
@markkrumholz thanks! I've added a few gates to check TOML-params, just to be safe, and also fixed the `sqrt(3)` normalisation.

markkrumholz (approved):

BenWibking:
Just a thought: it might be worth adding this problem to the nightly test suite, possibly replacing MHDBlast.

PR #2144, opened 2026-08-01
Extend MHD tests: expose setup.num_periods instead of stop_time

AstroKriel:
### Description
This PR replaces the hardcoded `stop_time` in `AlfvenWaveLinear`, `AlfvenWaveCircular`, `FastWave`, `SlowWave`, `EntropyWave`, and `FieldLoop` with a `setup.num_periods` parameter, computed from each test's own wave period (or crossing time, for `FieldLoop`) at runtime. Each TOML previously hardcoded a raw `stop_time` that was really some number of periods worked out by hand (`SlowWave.toml` even carried a comment saying so). `MHDBalsaraVortex` already computed `stop_time` this way, just under a different name (`setup.num_orbits`); renamed to `setup.num_periods` here for consistency with the rest.

For the first five wave tests and `MHDBalsaraVortex`, the new default reproduces the prior `stop_time` exactly: no behaviour change, pure convenience exposure/naming. `FieldLoop` is the one exception: its old `stop_time = 3.0` corresponded to a non-round ~1.30 domain crossings, so rather than preserve that number, the default is normalised to `num_periods = 1` (i.e. the loop now runs for exactly one full crossing). This is a deliberate, small behaviour change for `FieldLoop` only.

### Related issues
N/A.

### Checklist
- [x] I have added a link to any related issues (if applicable; see above): n/a, see Related issues.
- [ ] I have added tests for any new physics that this PR adds to the code: n/a, no new physics; behaviour is preserved exactly for 6 of 7 problems (see Description for the one exception, `FieldLoop`).

## Summary by CodeRabbit

* **New Features**
  * Simulation duration can now be configured by specifying the desired number of physical periods.
  * Wave and field-loop simulations calculate stop times automatically from their characteristic periods or domain size.
  * Balsara vortex runs support fractional period counts for more flexible durations.

* **Bug Fixes**
  * Invalid or nonpositive period values are rejected.
  * Conflicting fixed stop-time settings now produce clear configuration errors.
  * Convergence tests retain their intended fixed-duration behavior.

### DIFF
--- a/inputs/AlfvenWaveCircular.toml
+++ b/inputs/AlfvenWaveCircular.toml
@@ -27,7 +27,6 @@ plotfile_interval = -1
 plotfile_prefix = "alfven_wave_circular_plt"
 checkpoint_interval = -1
 cfl = 0.3
-stop_time = 5.0
 max_timesteps = 8000
 
 hydro.rk_integrator_order = 2
@@ -37,6 +36,7 @@ hydro.use_dual_energy = 0
 setup.run_convergence = false
 setup.run_sim = true
 setup.error_tol = 0.003
+setup.num_periods = 5
 
 mhd.emf_compute_scheme = "Balsara2025"
 mhd.emf_averaging_scheme = "Balsara2025"
--- a/inputs/AlfvenWaveLinear.toml
+++ b/inputs/AlfvenWaveLinear.toml
@@ -27,7 +27,6 @@ plotfile_interval = 1000
 plotfile_prefix = "alfven_wave_linear_plt"
 checkpoint_interval = -1
 cfl = 0.3
-stop_time = 2.0
 max_timesteps = 10000
 
 hydro.rk_integrator_order = 2
@@ -37,6 +36,7 @@ hydro.use_dual_energy = 0
 setup.run_convergence = false
 setup.run_sim = true
 setup.error_tol = 0.005
+setup.num_periods = 2
 setup.num_modes_x = 1
 setup.num_modes_y = 0
 setup.num_modes_z = 0
--- a/inputs/EntropyWave.toml
+++ b/inputs/EntropyWave.toml
@@ -24,7 +24,6 @@ do_tracers = 1
 plotfile_interval = -1
 
 cfl = 0.3
-stop_time = 1.0
 max_timesteps = 10000
 
 hydro.rk_integrator_order = 2
@@ -36,6 +35,7 @@ mhd.emf_reconstruction_order = 3
 setup.run_convergence = false
 setup.run_sim = true
 setup.error_tol = 0.002
+setup.num_periods = 1
 setup.num_modes_x = 1
 setup.num_modes_y = 0
 setup.num_modes_z = 0
--- a/inputs/FastWave.toml
+++ b/inputs/FastWave.toml
@@ -29,7 +29,6 @@ do_subcycle = 0
 plotfile_interval = -1
 checkpoint_interval = -1
 cfl = 0.3
-stop_time = 0.7071068
 max_timesteps = 2000
 
 hydro.rk_integrator_order = 2
@@ -39,6 +38,7 @@ hydro.use_dual_energy = 0
 setup.run_convergence = false
 setup.run_sim = true
 setup.error_tol = 0.002
+setup.num_periods = 1
 setup.num_modes_x = 1
 setup.num_modes_y = 0
 setup.num_modes_z = 0
--- a/inputs/FieldLoop.toml
+++ b/inputs/FieldLoop.toml
@@ -27,7 +27,6 @@ do_subcycle = 0
 plotfile_interval = -1
 checkpoint_interval = -1
 cfl = 0.2
-stop_time = 3.0
 max_timesteps = 3e4
 
 hydro.rk_integrator_order = 2
@@ -41,6 +40,7 @@ mhd.emf_averaging_scheme = "LondrilloDelZanna2004"
 # *****************************************************************
 # Field loop setup
 # *****************************************************************
+setup.num_periods = 1
 setup.loop_radius = 0.3
 setup.loop_center_x = 0.0
 setup.loop_center_y = 0.0
--- a/inputs/MHDBalsaraVortex.toml
+++ b/inputs/MHDBalsaraVortex.toml
@@ -36,4 +36,4 @@ hydro.use_dual_energy = 0
 setup.vortex_Mach = 0.1
 setup.vortex_b_magn = 0.1
 setup.advection = 1
-setup.num_orbits = 3
+setup.num_periods = 3
--- a/inputs/SlowWave.toml
+++ b/inputs/SlowWave.toml
@@ -29,8 +29,6 @@ do_subcycle = 0
 plotfile_interval = -1
 checkpoint_interval = -1
 cfl = 0.3
-# stop_time = 1 slow-wave period at angle=45 deg, c_s = vA = 1, |k| = 2 pi.
-stop_time = 1.8477590650225735
 max_timesteps = 20000
 
 hydro.rk_integrator_order = 2
@@ -44,6 +42,7 @@ mhd.emf_averaging_scheme = "LondrilloDelZanna2004"
 setup.run_convergence = false
 setup.run_sim = true
 setup.error_tol = 0.002
+setup.num_periods = 1
 setup.num_modes_x = 1
 setup.num_modes_y = 0
 setup.num_modes_z = 0
--- a/src/problems/AlfvenWaveCircularConvergence/testAlfvenWaveCircularConvergence.cpp
+++ b/src/problems/AlfvenWaveCircularConvergence/testAlfvenWaveCircularConvergence.cpp
@@ -331,8 +331,8 @@ auto problem_main() -> int
 			amrex::ParmParse const pp("setup");
 			pp.query("num_periods", num_periods);
 		}
-		if (num_periods <= 0.0) {
-			amrex::Abort("setup.num_periods must be > 0.");
+		if (!std::isfinite(num_periods) || num_periods <= 0.0) {
+			amrex::Abort("setup.num_periods must be finite and > 0.");
 		}
 		{
 			double unused_stop_time = 0.0;
--- a/src/problems/AlfvenWaveCircularConvergence/testAlfvenWaveCircularConvergence.cpp
+++ b/src/problems/AlfvenWaveCircularConvergence/testAlfvenWaveCircularConvergence.cpp
@@ -218,10 +218,17 @@ void QuokkaSimulation<AlfvenWaveCircular>::computeReferenceSolution_fc(amrex::Mu
 	}
 }
 
-auto runWaveTest(int nx, int ny, int nz) -> double
+// one wave period (num_modes and alfven_speed are fixed constants for this problem); shared by
+// runWaveTest() (always exactly one period) and problem_main()'s run_sim path (num_periods * this)
+auto computeWavePeriod() -> double
 {
 	const double wavelength = 1.0 / num_modes;
-	const double max_time = wavelength / alfven_speed;
+	return wavelength / alfven_speed;
+}
+
+auto runWaveTest(int nx, int ny, int nz) -> double
+{
+	const double max_time = computeWavePeriod();
 	const int max_timesteps = std::max(20000, nx * 100);
 
 	// Set grid dimensions using AMReX parameter system
@@ -283,6 +290,13 @@ auto problem_main() -> int
 		pp.query("run_convergence", run_convergence);
 		pp.query("run_sim", run_sim);
 		pp.query("error_tol", error_tol);
+
+		double unused_num_periods = 0.0;
+		if (run_convergence && !run_sim && (pp.query("num_periods", unused_num_periods) != 0)) {
+			amrex::Abort("setup.num_periods has no effect when setup.run_convergence=true and setup.run_sim=false; the "
+				     "convergence sweep always uses a fixed one-period run length. Remove setup.num_periods or set "
+				     "setup.run_sim=true.");
+		}
 	}
 
 	// AlfvenWaveCircularConvergence does not model resistivity; abort early rather than silently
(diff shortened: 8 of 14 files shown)


chongchonghe:
## PR #2144 Review — Extend MHD tests: expose `setup.num_periods` instead of `stop_time`

**Type:** Feature / enhancement → SPEC → VERIFY → EDGE cycle
**Head:** `283d18e7` (`extend/mhd-num-periods`) · **Base:** `origin/development`
**Build:** CPU `3d` preset, Release. All 7 affected targets compile and link clean; `quokka-pre-commit.sh` passes (clang-format, YAML, merge conflicts).
**GPU build:** not run — the diff touches only host-side `problem_main()` code. No `AMREX_GPU_DEVICE` lambdas, no `amrex::ParallelFor`, no core template changes. (This box has ROCm but no CUDA and no GPU device.)

**Verdict:** the core claim holds — I verified the stop-time equivalence both analytically and by running every affected test. Five concerns below are worth addressing before merge; none of them invalidate the approach.

---

### SPEC — extracted contract

| # | Claim | Source |
|---|---|---|
| 1 | `setup.num_periods` replaces hardcoded `stop_time` in 6 wave TOMLs | PR description |
| 2 | For `AlfvenWaveCircular`, `AlfvenWaveLinear`, `EntropyWave`, `FastWave`, `SlowWave`, `MHDBalsaraVortex` the new default reproduces the prior `stop_time` **exactly** | PR description |
| 3 | `MHDBalsaraVortex`: pure rename `num_orbits` → `num_periods`, no behaviour change | PR description |
| 4 | `FieldLoop`: deliberate change — old `stop_time = 3.0` (≈1.30 crossings) → `num_periods = 1` (exactly one crossing) | PR description |
| 5 | `num_periods` is a runtime knob: changing it changes the run length | implied by diff |

---

### Evidence

|---|---|---|
| VERIFY — claim 2, analytic | Recomputed each new formula from the constants and TOML inputs | `AlfvenWaveCircular` 5·(1/1)/1 = **5.0** (old 5.0 ✓); `AlfvenWaveLinear` 2·1/1 = **2.0** (old 2.0 ✓); `EntropyWave` 1·1/1 = **1.0** (old 1.0 ✓); `SlowWave` 1/c_s(45°) = **1.8477590650225735** (old 1.8477590650225735 ✓, bit-identical); `FastWave` 1/c_f(90°) = **0.70710678118654746** vs old literal `0.7071068` — differs by 2.7e-8 relative (see note below) |
| VERIFY — claim 2, runtime | Built PR branch, ran all 5 wave TOMLs | Final coarse-step times 4.9984, 1.9987, 0.99994, 0.7060, 1.8462 — each the last step *before* the exact stop time. Error norms 1.33e-09, 3.77e-10, 3.39e-10, 9.43e-10, 2.96e-10, all ≫3 orders below their tolerances. All PASS. |
| VERIFY — claim 3 | `MHDBalsaraVortex ../inputs/MHDBalsaraVortex.toml` | Ended at t ≈ 328.67 = 3·√(10²+10²)/(0.1·√(5/3)); error norm 1.369e-03. Matches the pre-PR `num_orbits = 3` formula exactly — rename only. `grep -r num_orbits` over the whole tree returns nothing, so the rename is complete. |
| VERIFY — claim 4 | `FieldLoop ../inputs/FieldLoop.toml` | `Simulation ended at t = 2.310176197 (100%)`, exit 0, 2000 coarse steps. That is exactly `hypot(2.0, 1.15625)`. Old `stop_time = 3.0` → 3.0/2.310176 = **1.2986** crossings, matching the PR's "~1.30". Claim confirmed; FieldLoop now runs ~23% shorter. |
| VERIFY — claim 5 | `EntropyWave.toml setup.num_periods=0.5` / `=3` | Stopped at 0.49997 / 2.99981 — scales linearly as intended. |
| EDGE — boundary | `EntropyWave.toml setup.num_periods=0` and `=-1` | Both **exit 0 and report PASS** with error norm 6.37e-18. Zero timesteps taken. See Concern 1. |
| EDGE — boundary | `MHDBalsaraVortex.toml setup.num_periods=0.5` | `terminate called after throwing an instance of 'std::runtime_error'` → `MPI_Abort(6)`. See Concern 4. |
| EDGE — mode interaction | `EntropyWaveConvergence.toml setup.num_periods=5` vs default | Identical: "Overall convergence rate: 2.09", PASSED, in both. `num_periods` is silently ignored. See Concern 2. |
| EDGE — precedence | `EntropyWave.toml stop_time=0.1` | Stopped at 0.09999 — an explicit `stop_time` still wins. Explained by `AMRSimulation::rereadRuntimeParameters()` (`src/simulation.hpp:1165-1170`), called at the top of `evolve()`, which re-runs `readParameters()` and its `pp.queryWithParser("stop_time", stopTime_)` (`src/simulation.hpp:937`). |
| EDGE — adjacent regression | `EntropyWaveConvergence.toml` (a TOML the PR does not touch) | Convergence rate 2.09, "✓ Richardson convergence test PASSED (target error 2.000e-09 reached)". No regression. The 5 `*Convergence.toml` files never set `stop_time` — the convergence driver computes `max_time` itself — so removing `stop_time` from the *run_sim* TOMLs cannot affect them. |

**Note on the FastWave 2.7e-8 discrepancy:** this is not a defect. The old TOML literal `0.7071068` was a 7-significant-figure truncation of 1/√2; the new code computes the value in full double precision. The new value is *more* correct, not different. Strictly, claim 2 is "exact" for 5 of 6 problems and "exact to 7 s.f." for `FastWave`. The measured error norm is unchanged in magnitude (9.4e-10 vs a 2e-3 tolerance).

---

### Analysis

The change is well-motivated. Every one of these TOMLs carried a magic `stop_time` that was really *N periods* worked out by hand — `SlowWave.toml` even shipped a comment saying so, and that comment is exactly the kind of thing that goes stale when the physics constants change. Deriving the stop time from the problem's own dispersion relation at runtime means the number can no longer drift out of sync with `angle_between_k_b0`, `num_modes_*`, or the background state. That is a genuine robustness win, not just cosmetics.

The implementations are placed correctly. In all five wave tests the new block sits after `k_magn` / `angle_between_k_b0_rad` are populated (e.g. `testSlowWaveConvergence.cpp:604-615` sits below the setup parse at `:557-576`, where `angle_between_k_b0_rad` is set at `:561` and `k_magn` at `:576`), so no use-before-init. Wave speeds are `constexpr` (`sound_speed = 1.0`, `alfven_speed = b0_magn/√ρ`), so the formulas are evaluated against the same constants the initial conditions use.

The magnetosonic formulas themselves are correct: `c_{f,s}² = ½[(a²+v_A²) ± √((a²+v_A²)² − 4a²v_A²cos²θ)]` with `+` for fast (`testFastWaveConvergence.cpp:604-605`) and `−` for slow (`testSlowWaveConvergence.cpp:613-614`); the Alfvén branch uses `c_A = v_A|cosθ|` (`testAlfvenWaveLinearConvergence.cpp:577`), all standard.

On `stop_time` precedence: because `rereadRuntimeParameters()` re-reads `stop_time` at the top of `evolve()`, an explicitly-supplied `stop_time` silently overrides the computed value. This is pre-existing harness behaviour that `MHDBalsaraVortex` already relied on, and "explicit user input wins" is a defensible rule — but the PR now makes six more problems depend on it implicitly, and nothing tells a user who sets both that their `num_periods` was discarded. Worth a one-line comment at minimum.

---

### PR-Introduced Concerns

**1. `num_periods` has no input validation; non-positive values make the test pass vacuously.**
`setup.num_periods=0` and `setup.num_periods=-1` both produce `stopTime_ ≤ 0`, run zero timesteps, and exit 0 with "error norm 6.368370e-18 (tol = 2.000000e-03)" — a green PASS for a simulation that never advanced. A negative value is worse than a no-op: it turns a correctness test into an unconditional success. These same files already validate their other setup parameters (`testFieldLoop.cpp:234` aborts on `loop_radius <= 0`, `testMHDBalsaraVortex.cpp:228-229` aborts on `vortex_radius <= 0`), so guarding is the established convention here and the new parameter is the odd one out. Suggest an `amrex::Abort("setup.num_periods must be > 0.")` in each of the seven sites.

**2. `setup.num_periods` is silently ignored in `run_convergence` mode.**
The new code lives only in the `if (run_sim)` branch. The convergence driver `runWaveTest()` computes its own single-period `max_time` and never queries `num_periods` (`testSlowWaveConvergence.cpp:461` + `:519`, `testFastWaveConvergence.cpp:452` + `:510`, `testAlfvenWaveLinearConvergence.cpp:436` + `:496`, `testEntropyWaveConvergence.cpp:332` + `:390`, `testAlfvenWaveCircularConvergence.cpp:223-224` + `:266`). I confirmed this empirically: `EntropyWaveConvergence.toml setup.num_periods=5` gives byte-identical output to the default run. Fixing the run length for a Richardson study is probably the *intended* design, but as written a user who sets `num_periods` in a convergence config gets no error and no effect. Either honour it in `runWaveTest()` or warn/abort when it is set alongside `run_convergence = true`.

**3. `FieldLoop`: `domain_diagonal` is only the true crossing time for the *default* advection angle.**
`testFieldLoop.cpp:251-256` sets `stopTime_ = num_periods * hypot(Lx, Ly)` and the accompanying comment asserts "one period is one full crossing of the domain diagonal". That holds only because the default `advection_angle_deg` at `:245` is itself derived from the domain aspect ratio — but `:246` makes it user-settable (exposed by #2135, already on `development`). With `|v| = 1` and any other angle, `t = hypot(Lx, Ly)` no longer returns the loop to its start:

| `advection_angle_deg` | displacement after `t = 2.310176`, in units of (Lx, Ly) |
|---|---|
| 59.967 (default) | (1.0000, 1.0000) ✓ |
| 45 | (0.8168, 1.4128) ✗ |
| 30 | (0.5775, 1.7303) ✗ |

`FieldLoop` has no pass/fail criterion today (`problem_main` ends `sim.evolve(); return 0;`), so this cannot fail CI — but the parameter's documented meaning becomes wrong the moment anyone touches the angle, and that is precisely the situation the PR is trying to eliminate elsewhere. Deriving the time from the actual velocity (`Lx / |v_x|`, which equals `hypot(Lx, Ly)` at the default angle) would at least be self-consistent; otherwise the comment should state the assumption explicitly.

**4. `MHDBalsaraVortex` keeps `num_periods` as `int` while the other six use `double`.**
`testMHDBalsaraVortex.cpp:221` declares `int num_periods = 1` (carried over from `int num_orbits`), so `setup.num_periods=0.5` aborts with `std::runtime_error` / `MPI_Abort(6)`, whereas the same input works fine on all five wave tests. Since the PR's stated rationale for the rename is "consistency with the rest", the type should follow the name — `double num_periods = 1.0`, dropping the two `static_cast<double>` at `:256` and `:260`.

**5. Duplicated wave-speed and period formulas.**
The fast/slow magnetosonic expressions now appear twice in the same file — once in `runWaveTest()` and again in the new `problem_main()` block (`testFastWaveConvergence.cpp:426` vs `:604-605`; `testSlowWaveConvergence.cpp:461` region vs `:613-614`), as does `wavelength = 2π/k_magn` across all five. The two copies can drift independently, which reintroduces the class of staleness this PR set out to remove. A small file-local helper (`wavePeriod()`) used by both call sites would let `runWaveTest()` and the `run_sim` path share one definition — and would also resolve Concern 2 naturally. Related cosmetic wrinkle: the pre-existing copies use the non-ASCII identifier `cosθ` while the new code uses `cos_theta`, so the same file now spells the same quantity two ways.

---

### Pre-existing Issues (Out of Scope)

- **`FieldLoop` has no correctness check.** `problem_main` unconditionally `return 0` after `evolve()`, so the test only verifies that the run does not crash. It computes a `divB` derived variable but never asserts on it. Pre-existing; does not block this PR; would need a separate PR. Worth noting only because Concern 3's impact is limited by it.
- **Non-ASCII identifiers (`cosθ`, `θ`, `sinθ`) in `testFastWaveConvergence.cpp` and `testSlowWaveConvergence.cpp`.** Pre-existing; does not block this PR. The PR author already chose ASCII for new code (commit `470c5b73`), so a follow-up sweep would just finish the job.

---

chongchonghe:
## Follow-up review — `93a10bef`, `76c32604`

All five concerns from my earlier review are addressed. Re-verified on a fresh `3d` CPU build of `76c32604`: all 7 targets compile, `quokka-pre-commit.sh` passes.

| # | Concern | Fix | Verified |
|---|---|---|---|
| 1 | no validation on `num_periods` | `!isfinite \|\| <= 0` abort at all 7 sites | `=0` / `=-1` now abort with "setup.num_periods must be finite and > 0." |
| 2 | silently ignored in convergence mode | abort when `run_convergence && !run_sim` and the key is set | `EntropyWaveConvergence.toml setup.num_periods=5` now aborts |
| 3 | `FieldLoop` diagonal valid only at the default angle | comment now states the assumption | the `vx/vy = Lx/Ly` claim checks out |
| 4 | vortex `num_periods` was `int` | now `double`, casts dropped | `setup.num_periods=0.5` runs instead of `MPI_Abort(6)` |
| 5 | duplicated period formulas | `computeWavePeriod()` shared by `runWaveTest()` and `problem_main()` | error norms bit-identical (below) |

**No regressions.** Every number matches my pre-fix run to the last digit: `AlfvenWaveCircular` t=5 / 1.331002516e-09 · `AlfvenWaveLinear` t=2 / 3.767014623e-10 · `EntropyWave` t=1 / 3.388766465e-10 · `FastWave` t=0.7071067812 / 9.434014397e-10 · `SlowWave` t=1.847759065 / 2.956209724e-10 · `MHDBalsaraVortex` t=328.6335345 / 1.368562133e-03 · `FieldLoop` t=2.310176197. Convergence sweeps: `EntropyWaveConvergence` 2.09, `SlowWaveConvergence` 2.16, both PASSED.

I also checked the new `stop_time` guard cannot misfire: none of the TOMLs these 7 executables consume sets `stop_time`, and `pp_root`'s empty prefix matches the one `simulation.hpp:893` reads it under.

### Two minor leftovers — both non-blocking, your call

**1. The convergence guard misses the `run_convergence && run_sim` case.** It only fires when `run_sim` is false, so with both true it stays silent and the two legs disagree. Running `EntropyWave.toml setup.run_convergence=true setup.run_sim=true setup.num_periods=2` gives `Simulation ended at t = 2` (2 periods) alongside a convergence sweep still using 1 period, with no warning. Dropping `&& !run_sim` would close it.

**2. `stop_time` went from a working override to a hard abort.** Failing loudly is defensible and matches the resistivity abort just above it, but it does remove a convenient escape hatch — shortening one of these runs for a quick smoke test now crashes instead. `amrex::Warning` plus honouring the value would give the same "no silent surprise" guarantee while keeping it usable.

markkrumholz (approved):

PR #2145, opened 2026-08-01
Enable resistivity support in the CurrentSheet test problem

AstroKriel:
### Description

`CurrentSheet`'s `Physics_Traits` never sets `resistivity_model`, so it defaults to `none`. Since #2050, setting `mhd.resistivity` on a problem in that state hard-aborts at startup with a message telling you to set `resistivity_model = ResistivityModel::constant` in its `Physics_Traits` specialization; this PR is doing so and making it standard.

This is needed to exercise a resistive `CurrentSheet` setup used to validate the Joule-heating implementation for an upcoming paper. Since the trait is a compile-time switch with no cost or behavioural effect unless `mhd.resistivity` is explicitly set, it's enabled generally here rather than gated to that one use case, matching how other MHD test problems (e.g. `AlfvenWaveLinear`) already expose it.

`mhd.resistivity` defaults to `0.0`, so this is a no-op for every existing `CurrentSheet` config.

### Related issues

N/A

### Checklist

- [ ] I have added tests for any new physics that this PR adds to the code: n/a, no new physics; existing default behaviour is preserved bit-for-bit.

### DIFF
--- a/src/problems/CurrentSheet/testCurrentSheet.cpp
+++ b/src/problems/CurrentSheet/testCurrentSheet.cpp
@@ -32,6 +32,7 @@ template <> struct quokka::EOS_Traits<CurrentSheet> {
 template <> struct Physics_Traits<CurrentSheet> : DefaultPhysicsTraits {
 	static constexpr bool is_hydro_enabled = true;
 	static constexpr bool is_mhd_enabled = true;
+	static constexpr ResistivityModel resistivity_model = ResistivityModel::constant; // eta defaults to 0; no-op unless set
 };
 
 // constants


chongchonghe:
# PR #2145 Review — Enable resistivity support in the CurrentSheet test problem

**Verdict: Approve.** The change is correct, minimal, matches the existing precedent byte-for-byte, and I verified the "no-op by default" claim is bit-for-bit true. Two minor notes below, neither blocking.

**Classification:** Feature / enhancement (SPEC → VERIFY → EDGE). Because the PR also unblocks an existing hard-abort, I additionally ran the RED step on the base branch.

**Setup:** base `origin/development` @ `9a46029d`, PR head @ `c8221760`. Source diff between the two is exactly 1 line in 1 file (`git diff origin/development --stat -- 'src/*'`). Preset `3d` (CurrentSheet is 3D-only), CPU; plus a `3d-hip` GPU compile. Reduced input: `inputs/CurrentSheet.toml` with `stop_time = 0.05` (20 steps) and `plotfile_interval = 20`.

---

## Evidence

|---|---|---|
| RED (base) | base build + `mhd.resistivity = 1.0e-3` | **Aborts, exit 6** — `Assertion 'resistivity_key_present' failed` at `QuokkaSimulation.hpp:639`, i.e. exactly the #2050 message the PR describes. Bug/blocker reproduced. |
| VERIFY 1 — no-op claim | PR build, default input (no `mhd.resistivity`), vs base plotfile | **All field data byte-identical.** `diff -r -x metadata.yaml` over `plt0000020` reports zero differences; only `metadata.yaml` differs, in the recorded `git_hash_quokka`. Per-step `Coarse STEP … at t = …` sequence also identical across all 20 steps. |
| VERIFY 2 — abort gone | PR build + `mhd.resistivity = 1.0e-3` + `hydro.use_dual_energy = 0` | Runs to completion, **exit 0**, 20 steps, no NaNs in output. |
| VERIFY 3 — resistivity actually applied | PR build, eta=0 vs eta=1e-3 (both `use_dual_energy=0`) | Plotfiles **differ** → the trait is genuinely wired through the EMF/energy-flux path, not silently accepted. |
| VERIFY 4 — resistive CFL path | PR build + `mhd.resistivity = 0.1` | dt = **2.44140625e-4**, exactly the predicted `cfl·dx_min/(2·eta/dx_min)` = `0.2·(1/64)/12.8`. Step count 20 → 205. Confirms `QuokkaSimulation.hpp:883-895` is correctly active. |
| EDGE 1 — negative eta | PR build + `mhd.resistivity = -1.0e-3` | Correctly rejected: `Assertion 'mhdResistivity_ >= 0.0' failed` at `QuokkaSimulation.hpp:626`. This assertion is only reachable for CurrentSheet *because* of this PR (previously it aborted earlier at :639). |
| EDGE 2 — dual-energy guard | PR build + `mhd.resistivity = 1.0e-3`, default `use_dual_energy` | Correctly aborts at `QuokkaSimulation.hpp:289` (Ohmic heating not in the aux energy equation). Guard from #2050 works on this problem. See Note 2. |
| EDGE 3 — adjacent regression | `AlfvenWaveLinearConvergence` (3d, CPU) | **PASSED** — overall convergence rate 2.07, target error 2.000e-10 reached. |
| GPU build | `quokka config/build -d 3d-hip CurrentSheet` (ROCm 6.4, gfx90a) | **SUCCESS.** Required here because the trait change newly instantiates the resistive device lambdas (`ApplyResistiveCorrection`, `AddResistiveEnergyFlux`) for this problem — the exact code carrying the `amrex::ignore_unused` first-capture workaround. No device-lambda capture errors. |

*(This box has a ROCm toolchain but no GPU device, so the HIP result is a compile-only check; runtime GPU behaviour is CI's job.)*

---

## Analysis

`Physics_Traits<CurrentSheet>` never set `resistivity_model`, so it inherited `ResistivityModel::none` from `DefaultPhysicsTraits` (`physics_info.hpp:37`). Since #2050, `QuokkaSimulation.hpp:628-639` hard-aborts when `mhd.resistivity` appears in the input while the trait is not `constant`. Setting the trait to `constant` moves the problem into the `hpp.query("resistivity", mhdResistivity_)` branch at `:625`, making the runtime knob live.

The added line is **byte-identical** to the existing precedent at `testAlfvenWaveLinearConvergence.cpp:46`, comment included — good consistency. 121 chars, within the 160 limit.

The no-op claim holds because every consumer is guarded: the dt constraint at `QuokkaSimulation.hpp:883` is wrapped in `if (mhdResistivity_ != 0.0)`, and both flux paths multiply by `resistivity`, so with eta=0 they add exactly `0.0`. I confirmed this empirically at the byte level rather than by inspection alone.

## PR-Introduced Concerns

**1. The "no cost" wording in the PR description is not accurate (minor, non-blocking).**

The description says the trait has "no cost or behavioural effect unless `mhd.resistivity` is explicitly set." The *behavioural* half is verified true (bit-for-bit). The *cost* half is not: `AddResistiveEnergyFlux` early-returns only for `ResistivityModel::none` (`mhd_system.hpp:1168`) — there is no runtime `eta == 0` guard. So with `constant` + eta=0 it still launches its full face-loop kernel every stage. In my 20-step run:

- base (`none`): `MHDSystem::AddResistiveEnergyFlux()` **absent from the profile entirely**
- PR (`constant`, eta=0): **60 calls, 0.1462 s = 2.30 %** of profiled time (total wall 6.342 s → 6.367 s, +0.4 %)

The EMF-side correction is effectively free (`ComputeEMF_FelkerStone2017` 0.3462 s → 0.3232 s, within noise). So the real overhead is one redundant kernel sweep computing zeros.

This is inherent to the `constant` model and is equally true of `AlfvenWaveLinearConvergence` today — it is *not* a defect introduced by this diff. Suggested action: just soften the wording to "negligible cost". A genuine fix (an `if (resistivity == 0.0) { return; }` guard at the top of `AddResistiveEnergyFlux`) touches shared code and would change behaviour for other problems, so it belongs in a separate PR, not here.

**2. `inputs/CurrentSheet.toml` still cannot run resistively as shipped (minor, worth a one-line follow-up).**

`hydro.use_dual_energy` defaults to 1, and resistivity requires it off (`QuokkaSimulation.hpp:286-289`). So after this PR, adding `mhd.resistivity = 1e-3` to the shipped input *still* hard-aborts — just at `:289` instead of `:639`. I had to add `hydro.use_dual_energy = 0` myself to get a resistive run.

Every other MHD input already sets it (`AlfvenWaveLinear.toml:35`, `AlfvenWaveCircular.toml:35`, `AlfvenWaveLinearConvergence.toml:31`, `BrioWuShockTube.toml:37`, …); `CurrentSheet.toml` is the outlier. Given the PR's stated motivation is to exercise a resistive CurrentSheet for the Joule-heating paper, adding that one line to `inputs/CurrentSheet.toml` would make the problem usable out of the box. Note this is *not* a no-op for the default test — turning off dual energy changes the non-resistive run too — so it is a judgement call for the author, not an obvious win. Fine to leave as-is if the resistive setup lives in a separate input file.

- **HIP warnings unrelated to this PR.** The 3d-hip build emits 2 `-Wgpu-maybe-wrong-side` warnings at `QuokkaSimulation.hpp:1214` and `:1963`, both from `[this]`-capturing density-floor lambdas. Unrelated to resistivity; pre-existing; does not block this PR.
- **CurrentSheet has no pass/fail criterion.** `problem_main()` always returns 0, so the test is a smoke test only — nothing validates the reconnection solution. Pre-existing, orthogonal to this PR, would need its own PR.

## Not Checked

`clang-format` is not installed on this box, so I did not run a formatter check. The added line is byte-identical to the existing `AlfvenWaveLinearConvergence` line, so style consistency follows by construction.

chongchonghe (approved):

PR #2146, opened 2026-08-02
Fix HydroWaveConvergence: broadcast run_sim status across MPI ranks

AstroKriel:
### Description
I noticed `run_sim`'s pass/fail check computes its error norm independently on every rank and never broadcasts it, unlike `quokka::richardson::run()` (`src/util/richardson.cpp:73`), which already broadcasts its own error from the IO processor for exactly this reason: `fextract()` only gathers the full comparison profile onto rank 0, so non-root ranks can disagree with rank 0 about pass/fail, and `mpirun` reports the whole run as failed if any rank exits nonzero, even when rank 0's own result is within tolerance. I bring `run_sim`'s two error-norm computations (`runWaveTest()` and the duplicate inline block in `problem_main()`) in line with `richardson.cpp`'s existing broadcast pattern.

### Related issues

_(none)_

### Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved consistency of HydroWave convergence error-norm calculations across parallel MPI runs.
  - Results now remain synchronized across processing ranks, supporting more reliable convergence evaluation.

### DIFF
--- a/src/problems/HydroWaveConvergence/testHydroWaveConvergence.cpp
+++ b/src/problems/HydroWaveConvergence/testHydroWaveConvergence.cpp
@@ -150,7 +150,9 @@ auto runWaveTest(int nx, int ny, int nz) -> double
 		// ε = || Δ U || = [&sum_k (Δ Uk)2]^{1/2}
 		err_sq += dU_k * dU_k;
 	}
-	const amrex::Real epsilon = std::sqrt(err_sq);
+	amrex::Real epsilon = std::sqrt(err_sq);
+	// fextract only gathers the full comparison to the IO processor; broadcast so every rank agrees
+	amrex::ParallelDescriptor::Bcast(&epsilon, 1, amrex::ParallelDescriptor::IOProcessorNumber());
 
 	return epsilon;
 }
@@ -202,7 +204,9 @@ auto problem_main() -> int
 			}
 			err_sq += dU_k * dU_k;
 		}
-		const amrex::Real epsilon = std::sqrt(err_sq);
+		amrex::Real epsilon = std::sqrt(err_sq);
+		// fextract only gathers the full comparison to the IO processor; broadcast so every rank agrees
+		amrex::ParallelDescriptor::Bcast(&epsilon, 1, amrex::ParallelDescriptor::IOProcessorNumber());
 
 		amrex::Print() << std::format("\nrun_sim error norm = {:.6e}  (tol = {:.6e})\n", static_cast<double>(epsilon), error_tol);
 		if (epsilon > error_tol) {


chongchonghe:
# Review: PR #2146 — Fix HydroWaveConvergence: broadcast `run_sim` status across MPI ranks

**Classification:** bug fix → RED → GREEN → EDGE cycle.
**Verdict:** ✅ Approve. The bug reproduces on `development`, the PR fixes it, real failures are still detected, and the convergence path is numerically unchanged.

**Environment:** 1d preset, Release, GCC, MPICH 4.1.2 (`AMReX_MPI=ON`), CPU only. Base = `origin/development` @ `9a46029d`; PR = `aefb030f` (branch `fix/hydrowave-status-broadcast`). No GPU device code in the diff, so no GPU build was required.

---

## Evidence

|---|---|---|
| RED — mechanism (base) | `mpirun -np 4 HydroWaveConvergence ../inputs/HydroWave.toml` with a temporary per-rank `AllPrint` of `epsilon` | Ranks disagree: rank 0 `5.690129478e-09` (global), rank 1 `6.623520e-09`, rank 2 `4.782217e-09`, rank 3 `6.621414e-09` — up to **±16 %** off rank 0 |
| RED — failure (base) | `mpirun -np 4 … ../inputs/HydroWave.toml setup.error_tol=6.0e-9` | Rank 0 prints `run_sim error norm = 5.690129e-09 (tol = 6.000000e-09)` → **passes**, yet ranks 1 and 3 exit nonzero → **`mpirun` EXIT=1** |
| GREEN — PR branch | identical command, `setup.error_tol=6.0e-9`, 4 ranks | **`mpirun` EXIT=0**; per-rank diagnostic shows all four ranks report the identical `5.690129478e-09` |
| GREEN — default tol | `mpirun -np 4 … ../inputs/HydroWave.toml` and serial run | `5.690129e-09 (tol = 1.0e-08)`, EXIT=0 in both; serial value unchanged from base |
| EDGE 1 — real failures still caught | `mpirun -np 4 … setup.error_tol=1.0e-12` (PR) | `mpirun` **EXIT=1** — the broadcast does not mask genuine failures |
| EDGE 2 — adjacent test | `mpirun -np 4 … ../inputs/HydroWaveFc.toml` (PR) | `run_sim error norm = 6.439856e-09 (tol = 1.0e-08)`, EXIT=0 |
| EDGE 3 — convergence path, base vs PR | `mpirun -np 4 HydroWaveConvergence` (no input → `run_convergence`) on **both** branches | Byte-identical error sequence (`5.029443e-09 → 1.251883e-11` over nx 128→2048), overall rate **2.16** (expected 2.0, tol 0.3), `Richardson convergence test PASSED`, EXIT=0 on both — no numeric change, no deadlock from the second broadcast |

The RED failure is not contrived: it just picks a tolerance inside the band where rank 0 and the other ranks genuinely disagree (`5.69e-09 < tol < 6.62e-09`). The shipped `HydroWave.toml` tolerance (`1.0e-8`) sits only a factor ~1.5 above the *largest* rank-local norm, so the same disagreement is one tolerance tweak — or one platform/compiler shift — away from failing CI for a run that rank 0 considers correct.

---

## Analysis

**Root cause.** `fextract()` (`src/util/fextract.cpp:176–224`) `Gatherv`s the slice onto the IO processor and then swaps in the assembled global arrays **only** under `if (ParallelDescriptor::IOProcessor())` (`src/util/fextract.cpp:222`). Non-root ranks keep their rank-local slice. `testHydroWaveConvergence.cpp` then computes

```
dU_k = Σ_i |ΔU_k,i| / nx_final
```

where both `nx_final` and the sum are rank-local on non-root ranks — i.e. a *local mean* over 32 points rather than the *global mean* over 128. Each rank then makes its own pass/fail decision (`testHydroWaveConvergence.cpp:212`), and MPI launchers report the job as failed if any rank exits nonzero. Measured divergence above confirms this exactly.

**The fix.** Broadcasting `epsilon` from `IOProcessorNumber()` before the comparison (`:155`, `:209`) is the semantically correct repair, not a reduction: a non-root rank's partial slice has no valid global denominator, so its norm cannot be meaningfully combined — rank 0's value is the only correct one. This mirrors the pattern already used at `src/util/richardson.hpp:73`.

**Collective safety.** Both `Bcast` sites are reached by every rank: `run_sim` comes from `ParmParse`, so the `if (run_sim)` guard is identical on all ranks, and `runWaveTest()` is invoked by all ranks from inside `richardson::run`'s loop. Confirmed empirically — no hang across all multi-rank runs above.

---

## PR-Introduced Concerns

1. **The `runWaveTest()` broadcast (`:155`) is redundant.** `richardson::run` already broadcasts the callable's return value at `src/util/richardson.hpp:73`, and `runWaveTest()` is only called from there (`testHydroWaveConvergence.cpp:236`). I verified this is a pure no-op: base and PR produce identical convergence sequences and rates. It is harmless — a second collective that every rank reaches — and defensible as defence-in-depth if `runWaveTest()` is ever called outside `richardson::run`. **Non-blocking**; drop it or keep it, but the PR description implies both sites were broken when only the `problem_main()` one was.

2. **`epsilon` loses `const`** at both sites, which `CLAUDE.md` discourages. Unavoidable given `Bcast` takes a pointer, and it matches the existing style at `richardson.hpp:72–73`. **No action needed.**

---

1. **The same un-broadcast `fextract` pass/fail pattern exists in ~35 other problems.** A concrete, verified example: `src/problems/RadStreaming/testRadStreaming.cpp:207–217` computes `rel_err_norm = err_norm / sol_norm` from rank-local vectors and initialises `status = 1`. On a rank whose local slice is empty, both sums are zero → `rel_err_norm` is `NaN` → `NaN < rel_err_tol` is false → that rank exits 1 even when rank 0 passed. Other files matching the pattern include the `RadMarshak*`, `Radhydro*`, `RadMatterCoupling*`, `Dust*`, `NscbcChannel`, `ParticleSink`, `ResampledCoolingTest` and `SN` tests. Pre-existing, does not block this PR, needs a separate PR (ideally a shared helper rather than 35 copies of the same `Bcast`).

2. **The `run_sim` block duplicates `runWaveTest()`'s error-norm computation almost verbatim** (`:138–153` vs `:194–207`), and this PR now duplicates the broadcast too. Factoring the norm into one helper would stop the two copies drifting. Pre-existing duplication, does not block this PR, separate PR.

3. **Non-root ranks still execute the whole (meaningless) norm loop** before the broadcast, indexing `val_exact.at(n)[i]` for `i < nx_final` where `nx_final` comes from the *final* `fextract` and `val_exact` from the *initial* one. Safe today because both calls share the same `BoxArray`/`DistributionMap` so local sizes match, but guarding the computation with `if (ParallelDescriptor::IOProcessor())` would be strictly more robust and cheaper. Pre-existing, adjacent to the diff, does not block this PR.

---

chongchonghe (approved):

PR #2147, opened 2026-08-02
eps_ff_eff is now a ramp

aditivijayan:
### Description
This PR adds a ramp to `eps_ff` to alleviate creation of high density central region in the M82 simulations. Instead of a constant `eps_ff` for all densities as density increases and `lambda_J` decreases the efficiency of star formation goes as exp(dx/lambdaJ).  

### Related issues

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved stochastic stellar particle formation by adapting formation probabilities to local cell size and Jeans length.
  * Cells larger than the Jeans length now receive an exponentially increased effective star-formation efficiency, capped at 100%.
  * This provides more responsive and physically consistent stellar particle creation across varying simulation resolutions while preserving existing behavior in other conditions.

### DIFF
--- a/src/particles/particle_creation.hpp
+++ b/src/particles/particle_creation.hpp
@@ -452,8 +452,8 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 			// Create increasing ramp for eps_ff_
 			const amrex::Real Jdx_over_LambdaJ = J * dx[0] / LambdaJ;
 			amrex::Real eps_ff_eff = eps_ff_;
-			if (Jdx_over_LambdaJ > 1.0) {
-				eps_ff_eff = std::min(eps_ff_ * std::exp(Jdx_over_LambdaJ), 1.0);
+			if(Jdx_over_LambdaJ > 1.0) {
+				eps_ff_eff = std::min(eps_ff_ * std::exp(Jdx_over_LambdaJ - 1.), 1.0);
 			}
 			const amrex::Real nominal_prob_star_formation = (eps_ff_eff / eps_star) * (dt / t_ff);
 			// force P_sf to 1 if we are very far below the Jeans length (as determined by J_truncate)
--- a/src/particles/particle_creation.hpp
+++ b/src/particles/particle_creation.hpp
@@ -453,7 +453,7 @@ template <> struct ParticleCreationTraits<ParticleType::StochasticStellarPop> {
 			const amrex::Real Jdx_over_LambdaJ = J * dx[0] / LambdaJ;
 			amrex::Real eps_ff_eff = eps_ff_;
 			if(Jdx_over_LambdaJ > 1.0) {
-				eps_ff_eff *= std::min(eps_ff_ * std::exp(Jdx_over_LambdaJ), 1.0);
+				eps_ff_eff = std::min(eps_ff_ * std::exp(Jdx_over_LambdaJ), 1.0);
 			}
 			const amrex::Real nominal_prob_star_formation = (eps_ff_eff / eps_star) * (dt / t_ff);
 			// force P_sf to 1 if we are very far below the Jeans length (as determined by J_truncate)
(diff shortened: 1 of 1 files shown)


markkrumholz (changes requested):
I agree with the AI comment to be careful to avoid overflows in computing the exponential. Also, the way you have set this up, eps_ff = 0.01 when dx / lambda_J = 1, which is not consistent with the existing prescription where eps_ff becomes 0.01 at the threshold, which is when J_truncate * dx = lambda_J. I therefore suggest changing the exponential factor from dx / lambda_J to J_truncate * dx / lambda_J, so that you keep eps_ff at the threshold fixed, and only increase it at densities above the threshold.

aditivijayan:
> I agree with the AI comment to be careful to avoid overflows in computing the exponential. Also, the way you have set this up, eps_ff = 0.01 when dx / lambda_J = 1, which is not consistent with the existing prescription where eps_ff becomes 0.01 at the threshold, which is when J_truncate * dx = lambda_J. I therefore suggest changing the exponential factor from dx / lambda_J to J_truncate * dx / lambda_J, so that you keep eps_ff at the threshold fixed, and only increase it at densities above the threshold.

Actually in this implementation, eps_ff = 0.01 when dx / lambda_J = 1/J, whcih means that J*dx = lambda_J. 
When density is such that lambda_J < J_truncate * dx, eps_ff is already equal to infinity.

As I understand it (and please correct if I am mistaken)--

In the repo version-
1. if lambda_J < J_trunc * dx, eps_ff = infinity, meaning probability of star formation = unity.
2. if J_trunc * dx < lambda_J < J * dx, eps_ff = 0.01 and probabilty of star formation = eps_ff/eps_star * (dt/t_ff)

What we want is-
1. if lambda_J < J_trunc * dx, eps_ff = infinity, meaning probability of star formation = unity.
2. J_trunc * dx < lambda_J < J * dx, eps_ff = 0.01 * exp(J*dx/lambda_j), and probability of SF= eps_ff/eps_star * (dt/t_ff).

markkrumholz:
OK, I think I misunderstood the meaning of J_trunc -- I thought this was where star formation turned on, not where it went to infinity. I agree with your statement of what we want, i.e., 

> 1. if lambda_J < J_trunc * dx, eps_ff = infinity, meaning probability of star formation = unity.
> 2. J_trunc * dx < lambda_J < J * dx, eps_ff = 0.01 * exp(J*dx/lambda_j), and probability of SF= eps_ff/eps_star * (dt/t_ff).

So I think the only problem in your code as it stands is that your expression for eps_ff is eps_ff = 0.01 * exp(dx/lambda_j), i.e., missing the factor of J in the equation you wrote down.

Once that is implemented, and you've addressed the AI comment, let me know and I can re-review.

aditivijayan:
> OK, I think I misunderstood the meaning of J_trunc -- I thought this was where star formation turned on, not where it went to infinity. I agree with your statement of what we want, i.e.,
> 
> > 1. if lambda_J < J_trunc * dx, eps_ff = infinity, meaning probability of star formation = unity.
> > 2. J_trunc * dx < lambda_J < J * dx, eps_ff = 0.01 * exp(J*dx/lambda_j), and probability of SF= eps_ff/eps_star * (dt/t_ff).
> 
> So I think the only problem in your code as it stands is that your expression for eps_ff is eps_ff = 0.01 * exp(dx/lambda_j), i.e., missing the factor of J in the equation you wrote down.
> 
> Once that is implemented, and you've addressed the AI comment, let me know and I can re-review.

It is done.

markkrumholz (changes requested):
Sorry, realized it should be 

`std::min(eps_ff_ * std::exp(Jdx_over_LambdaJ - 1), 1.0)`

so that at the threshold we get the target eps_ff_ exactly at threshold,`Jdx_over_LambdaJ = 1`.

aditivijayan:
> Sorry, realized it should be
> 
> `std::min(eps_ff_ * std::exp(Jdx_over_LambdaJ - 1), 1.0)`
> 
> so that at the threshold we get the target eps_ff_ exactly at threshold,`Jdx_over_LambdaJ = 1`.

It is done.

If it looks good, I can update the docs.

markkrumholz (approved):

chongchonghe (changes requested):
I am gonna set a block on this PR because we don't want to apply this to the development branch unless we have tested it thoroughly and demonstrated it produces better results in general. My impression is Aditi created this PR just to build the container and test the code on Frontier.

PR #2155, opened 2026-08-10
Removed unnecessary charge conservation.

XRByte:
### Description
Integrators like VODE are inherently charge conserving. Removed explicit charge conservation from `chemistry.hpp`. If future integrators require it, it can be included in `actual_network_data.cpp` in microphysics/quokka-networks for that particular network and is easily generated using JAFF.

### Related issues
None

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated chemistry integration to avoid automatically applying hard-coded electron number-density corrections.
  * Removed automatic mass-fraction renormalization during chemistry calculations.


BenWibking:
We do need to test whether this holds empirically for the PrimordialChem network.

The reason this is necessary is because if there is a bug in the Jacobian (even a small one that doesn't affect the ability to obtain the correct solution to integration tolerance), linear invariant preservation is lost.

XRByte:
> We do need to test whether this holds empirically for the PrimordialChem network.
> 
> The reason this is necessary is because if there is a bug in the Jacobian (even a small one that doesn't affect the ability to obtain the correct solution to integration tolerance), linear invariant preservation is lost.

There is already a `balance_charge` function in  `actual_network_data.cpp`. We can call it if it's not called anywhere.

XRByte:
In order to make a clean solution to this, I need to mark the functions in `actual_network_data.cpp` ([here](https://github.com/AMReX-Astro/Microphysics/blob/59c48535461d1310efcf3fbf4970e4c19e1eac64/networks/primordial_chem/actual_network_data.cpp)) with  `AMREX_GPU_HOST_DEVICE`. This can be done to all networks. Is there a reason this hasn't been done in Microphysics?

markkrumholz:
@BenWibking: @XRByte and I just discussed how to proceed on this. Here's my take based on this conversation:

1. The current implementation of charge conservation needs to go because it is hardwired and won't work for general chemistry networks of the type we're now testing; the code as written breaks tests that need to be done for the JAFF paper, which is being written now. So some form of this PR needs to go through.
2. For the longer term there is a question of how to keep the possibility of a manual enforcement of charge conservation, while getting rid of the hard-coded version that works for primordial chem but breaks everything else. My suggestion for this is that we can use the capability that is built into microphysics, but that we need to do a PR to microphysics to add GPU annotations to this function so that we can call it from GPU as part of our chemistry update. @XRByte will generate a PR to microphysics to make this change. Hopefully it will go through quickly, since it's a < 1 line change.
3. While we're waiting on microphysics, I suggest that we create an issue in Quokka noting that we are currently not enforcing charge conservation manually, and pointing to the microphysics PR to explain why. We will link to that issue in this PR. We can then add back the charge conservation once the microphysics PR is merged. Alternatively, if someone wants to run sooner, we can always make another fork of microphysics with our change incorporated and point Quokka to that.
4. Long-long-term, this is another version of the reason we're planning to move away from microphysics: because we don't control that repo, we can't guarantee that PRs get merged in a timely fashion, and when we encounter a blocking problem as is the case here, we are left with unappealing choices.

Thoughts on this resolution?

BenWibking:
> @BenWibking: @XRByte and I just discussed how to proceed on this. Here's my take based on this conversation:
> 
> 1. The current implementation of charge conservation needs to go because it is hardwired and won't work for general chemistry networks of the type we're now testing; the code as written breaks tests that need to be done for the JAFF paper, which is being written now. So some form of this PR needs to go through.
> 2. For the longer term there is a question of how to keep the possibility of a manual enforcement of charge conservation, while getting rid of the hard-coded version that works for primordial chem but breaks everything else. My suggestion for this is that we can use the capability that is built into microphysics, but that we need to do a PR to microphysics to add GPU annotations to this function so that we can call it from GPU as part of our chemistry update. @XRByte will generate a PR to microphysics to make this change. Hopefully it will go through quickly, since it's a < 1 line change.
> 3. While we're waiting on microphysics, I suggest that we create an issue in Quokka noting that we are currently not enforcing charge conservation manually, and pointing to the microphysics PR to explain why. We will link to that issue in this PR. We can then add back the charge conservation once the microphysics PR is merged. Alternatively, if someone wants to run sooner, we can always make another fork of microphysics with our change incorporated and point Quokka to that.
> 4. Long-long-term, this is another version of the reason we're planning to move away from microphysics: because we don't control that repo, we can't guarantee that PRs get merged in a timely fashion, and when we encounter a blocking problem as is the case here, we are left with unappealing choices.
> 
> Thoughts on this resolution?

Sure, that's all fine with me.

All I wanted to suggest is that someone just test that the Jacobian is exactly consistent with the rhs by doing a simple finite difference calculation for the PopIII network before merging this PR. Claude can do it. That just ensures there's no suprises when running that network in Quokka after this is merged. If there's a problem, then we can just create an issue so anyone who tries to use PopIII before the longer-term solution is merged is aware of the issue.

XRByte:
I have compared the numerical jacobian with the analytical jacobian for the primordial chem network and it's consistent. Since the PR to microphtysics will require a GPU decorator for only the primordial chem network which is already consistent and I am not adding a new network, I think it won't be fruitful to make the PR. I think this is ready for review.

chongchonghe:
@XRByte From our discussion on Wednesday, it seems to me that it's trivial to add a GPU decorator to this `balance_charge` function, and you don't have to touch Microphysics. You can simply add the decorator to the function in the hpp file in `quokka/src/networks` . Just want to make check that, considering this, this PR is what you actually want to do. 

Also, I am maintaining a fork of Microphysics at https://github.com/chongchonghe/Microphysics-fork , and quokka is currently pointing at the HEAD of that fork, and will continue to be. So, if you want to make a change to Microphysics, just create a PR to my fork and I can approve immediately.

XRByte:
Done.

markkrumholz (approved):

chongchonghe (approved):

PR #2158, opened 2026-08-10
Fix gas internal energy when MHD is enabled: subtract magnetic energy at all reconstruction sites

chongchonghe:
### Description

Under MHD, `energy_index` holds $E_\mathrm{int} + E_\mathrm{kin} + E_\mathrm{mag}$ while `internalEnergy_index` holds $E_\mathrm{int}$ alone. This PR sweeps every call of `EOS::ComputeSoundSpeed` / `ComputePressure` / `ComputeTgasFromEint` / `ComputeEintFromEgas` / `ComputeEgasFromEint`, every hand-written `Etot - Ekin`, and every mutation of `energy_index` in `src/`, and fixes the places that forgot $E_\mathrm{mag}$.

> **Note on #2157.** That PR fixes the same class of bug in `src/particles/particle_creation.hpp` and is in the merge queue but not yet in `development`. My sweep found those two sites independently; I have removed them from this PR so there is no duplicate fix and no conflict. This PR does not touch `particle_creation.hpp` at all.

Everything in `src/hydro/`, `src/radiation/`, `src/cooling/`, `src/conduction/`, `src/dust/`, `src/io/` was already correct and is unchanged. `src/chemistry/Chemistry.hpp` passes `0.0` but is guarded by a `static_assert(!is_mhd_enabled)` directly above it, so it is fine as-is. The remaining bugs were in `QuokkaSimulation.hpp` and two files in `src/particles/`.

**What was broken and is now fixed:**

- `QuokkaSimulation::printCellProperties` used `Eint = Etot - Ekin`, so the pressure and sound speed printed in the CFL diagnostic were too large under MHD. It now reads the face-centered field from `state_new_fc_` and goes through `quokka::EOS::ComputeEintFromEgas`. Diagnostic output only; no effect on the solution.
- `particle_deposition.hpp` `addCompositeBufferToState`: three separate errors. The positivity limiter compared $E_\mathrm{int}+E_\mathrm{kin}$ against a total energy that includes $E_\mathrm{mag}$; `e_kinetic_max` was inflated by $E_\mathrm{mag}$, so SN momentum deposition was under-limited; and the value written to `internalEnergy_index` was $E_\mathrm{int}+E_\mathrm{mag}$. All three now go through `ComputeEgasFromEint` / `ComputeEintFromEgas` with the cell's magnetic energy.
- `particle_deposition.hpp` `addThermalOnlyBufferToState`: the same error in the final internal energy.
- `particle_accretion.hpp` `UpdateHydroState` scaled the entire `energy_index` by the accretion down-factor $f$, scaling $E_\mathrm{mag}$ along with the gas — the same bug #2157 fixes at particle creation, but on the accretion path, which #2157 does not touch. Accreting gas does not change $B$ (and cannot, without violating $\nabla \cdot B = 0$), so the old code left the reconstructed $E_\mathrm{int}$ short by $(1-f)\,E_\mathrm{mag}$: spurious cooling proportional to the accreted mass fraction. `UpdateHydroState` gains a `state_fc` parameter, supplied by its single caller `applyAccretion`.

I used $E_\mathrm{tot}' = f\,(E_\mathrm{tot} - E_\mathrm{mag}) + E_\mathrm{mag}$ here rather than #2157's "subtract the removed internal + kinetic energy" form. Both are correct, but the form used there reads `internalEnergy_index` and so changes the result whenever the auxiliary and conserved internal energies have drifted apart — including for pure-hydro problems. Subtracting $E_\mathrm{mag}$ keeps non-MHD results bit-for-bit identical, which matters more on the accretion path since it runs every step on every cell.

**Found but deliberately not fixed:** `QuokkaSimulation::PreInterpState` / `PostInterpState` convert $E_\mathrm{tot}$ to $(E_\mathrm{tot} - E_\mathrm{kin})/\rho$ before AMR interpolation, which under MHD is $(E_\mathrm{int} + E_\mathrm{mag})/\rho$. The round trip is exact within a cell, but coarse-fine ghost cells get a coarse-interpolated $E_\mathrm{mag}$ that does not match the separately (divergence-preserving) interpolated face-centered $B$. Fixing this requires $B$ on the *coarse temporary* MultiFab built inside `FillPatchWithData`, whose BoxArray and DistributionMapping differ from `state_new_fc_`; that means changing the interp-hook interface in `simulation.hpp`, which is well beyond this PR. Worth a follow-up issue.

**Verified by** building and running with the `3d` preset: `SN` (MHD on, $B_0 = 10^{-7}$ G, `SN_scheme = "SN_thermal_or_thermal_momentum"` — directly exercises the fixed deposition path on a magnetised background), plus `ParticleSinkFormation`, `ParticleAccretion`, `ParticleCreation`, `ParticleSF` and `ParticleDeposition` as regression coverage. All pass. Non-MHD problems are bit-for-bit unaffected because `ComputeMagneticEnergy` returns exactly `0.0` when MHD is disabled. Because `UpdateHydroState` now captures face-centered data inside a device lambda, I also ran a `3d-cuda` compile check on `SN` and `ParticleSinkFormation` (compile and link only; this machine has `nvcc` but no GPU device).

### Related issues
N/A — overlaps with #2157, which covers the `particle_creation.hpp` sites; those are excluded here.

### Checklist

### DIFF
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -954,6 +954,24 @@ template <typename problem_t> void QuokkaSimulation<problem_t>::printCellPropert
 	// print density, velocity magnitude, temperature, adiabatic sound speed
 	amrex::Vector<amrex::Real> cell_values = amrex::get_cell_data(state_new_cc_[lev], index);
 
+	// when MHD is enabled, the cell-centered total energy also contains the magnetic energy, which must be
+	// subtracted to obtain the gas internal energy. The magnetic field is face-centered, so it is read from
+	// state_new_fc_, which shares the DistributionMapping of state_new_cc_ (so the same MPI rank owns both).
+	amrex::Real Emag = 0.;
+	if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
+		for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+			amrex::IntVect index_hi = index;
+			index_hi[idim] += 1;
+			const amrex::Vector<amrex::Real> b_lo = amrex::get_cell_data(state_new_fc_[lev][idim], index);
+			const amrex::Vector<amrex::Real> b_hi = amrex::get_cell_data(state_new_fc_[lev][idim], index_hi);
+			if (!b_lo.empty() && !b_hi.empty()) {
+				const amrex::Real b_cc =
+				    0.5 * (b_lo[Physics_Indices<problem_t>::mhdFirstIndex] + b_hi[Physics_Indices<problem_t>::mhdFirstIndex]);
+				Emag += 0.5 * b_cc * b_cc;
+			}
+		}
+	}
+
 	// cell_values is *only* filled on the MPI rank that holds the box with this cell
 	// (NOTE: for Cray MPICH, standard output is NOT ordered with respect to different ranks.)
 	if (!cell_values.empty()) {
@@ -967,8 +985,7 @@ template <typename problem_t> void QuokkaSimulation<problem_t>::printCellPropert
 		const amrex::Real vx3 = px3 / rho;
 		const amrex::Real vsq = (vx1 * vx1) + (vx2 * vx2) + (vx3 * vx3);
 		const amrex::Real vel_mag = std::sqrt(vsq);
-		const amrex::Real Ekin = 0.5 * rho * vsq;
-		const amrex::Real Eint = Etot - Ekin;
+		const amrex::Real Eint = ::quokka::EOS<problem_t>::ComputeEintFromEgas(rho, px1, px2, px3, Etot, Emag);
 		const amrex::Real P = ::quokka::EOS<problem_t>::ComputePressure(rho, Eint);
 		const amrex::Real cs = ::quokka::EOS<problem_t>::ComputeSoundSpeed(rho, P);
 
--- a/src/particles/particle_accretion.hpp
+++ b/src/particles/particle_accretion.hpp
@@ -506,11 +506,22 @@ void UpdateParticleMassAndMomentum(ContainerType *container, amrex::MultiFab &st
 	}
 }
 
-template <typename problem_t> void UpdateHydroState(amrex::MultiFab &state, amrex::MultiFab &accretion_rate)
+template <typename problem_t>
+void UpdateHydroState(amrex::MultiFab &state, amrex::MultiFab &accretion_rate, std::array<amrex::MultiFab, AMREX_SPACEDIM> const *state_fc)
 {
 	const BL_PROFILE("SinkAccretionUtils::UpdateHydroState()");
 	const auto &local_accretion_rate_arr = accretion_rate.arrays();
 	const auto &state_arr = state.arrays();
+
+	// the magnetic field is unchanged by accretion, so the magnetic energy must be excluded when the gas
+	// energies are scaled down below (it is identically zero unless MHD is enabled)
+	std::array<amrex::MultiArray4<const amrex::Real>, AMREX_SPACEDIM> state_fc_arrs{};
+	const bool has_state_fc = (state_fc != nullptr);
+	if (has_state_fc) {
+		for (int dir = 0; dir < AMREX_SPACEDIM; ++dir) {
+			state_fc_arrs[dir] = (*state_fc)[dir].const_arrays();
+		}
+	}
 
 	amrex::ParallelFor(state, [=] AMREX_GPU_DEVICE(int bx, int i, int j, int k) noexcept {
 		const double accretion_rate_cell = local_accretion_rate_arr[bx](i, j, k);
@@ -518,12 +529,24 @@ template <typename problem_t> void UpdateHydroState(amrex::MultiFab &state, amre
 		AMREX_ASSERT(accretion_rate_cell > -1.0);
 		const double accretion_down_factor = 1.0 + accretion_rate_cell;
 		AMREX_ASSERT(accretion_down_factor > 0.0);
+
+		std::array<amrex::Array4<const amrex::Real>, AMREX_SPACEDIM> fab_fc{};
+		if (has_state_fc) {
+			for (int dir = 0; dir < AMREX_SPACEDIM; ++dir) {
+				fab_fc[dir] = state_fc_arrs[dir][bx];
+			}
+		}
+		const auto *const fab_fc_ptr = has_state_fc ? &fab_fc : nullptr;
+		const double Emag = HydroSystem<problem_t>::ComputeMagneticEnergy(i, j, k, fab_fc_ptr);
+
 		state_arr[bx](i, j, k, HydroSystem<problem_t>::density_index) *= accretion_down_factor;
 		state_arr[bx](i, j, k, HydroSystem<problem_t>::x1Momentum_index) *= accretion_down_factor;
 		state_arr[bx](i, j, k, HydroSystem<problem_t>::x2Momentum_index) *= accretion_down_factor;
 		state_arr[bx](i, j, k, HydroSystem<problem_t>::x3Momentum_index) *= accretion_down_factor;
 		state_arr[bx](i, j, k, HydroSystem<problem_t>::internalEnergy_index) *= accretion_down_factor;
-		state_arr[bx](i, j, k, HydroSystem<problem_t>::energy_index) *= accretion_down_factor;
+		// scale only the gas part of the total energy, leaving the magnetic energy untouched
+		state_arr[bx](i, j, k, HydroSystem<problem_t>::energy_index) =
+		    (state_arr[bx](i, j, k, HydroSystem<problem_t>::energy_index) - Emag) * accretion_down_factor + Emag;
 
 		// update passive scalars
 		for (int n = 0; n < Physics_Traits<problem_t>::numPassiveScalars; ++n) {
@@ -580,7 +603,7 @@ void applyAccretion(ContainerType *container, amrex::MultiFab &state, amrex::Mul
 	UpdateParticleMassAndMomentum<ContainerType, problem_t>(container, state, scale_down, state_fc, lev, mass_index, time, dt, mdot_index, ang_mom_index);
 
 	// Step 4: Update the hydro state. We do this at last because the original state is needed for updating particles in step 3.
-	UpdateHydroState<problem_t>(state, state_accretion_rate);
+	UpdateHydroState<problem_t>(state, state_accretion_rate, state_fc);
 }
 
 } // namespace SinkAccretionUtils
--- a/src/particles/particle_creation.hpp
+++ b/src/particles/particle_creation.hpp
@@ -314,11 +314,14 @@ AMREX_GPU_DEVICE inline void initializeSinkLikeParticles(PType *particles, int n
 
 	// update cell density to be the threshold density
 	const amrex::Real scale_factor = rho_J / cell_density;
+	// the magnetic field is unchanged by particle creation, so only the gas part of the total energy is scaled
+	// (Emag is identically zero unless MHD is enabled)
+	const amrex::Real Emag = HydroSystem<problem_t>::ComputeMagneticEnergy(i, j, k, fab_fc);
 	state_arr(i, j, k, HydroSystem<problem_t>::density_index) = rho_J;
 	state_arr(i, j, k, HydroSystem<problem_t>::x1Momentum_index) *= scale_factor;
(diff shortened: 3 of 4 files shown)


chongchonghe:
# PR #2158 Review — Fix gas internal energy when MHD is enabled

**Verdict:** the fix is correct, and I confirmed both bugs by watching them fail on `development` first. The `printCellProperties` error reproduces exactly, to 7 significant figures. The accretion-path error reproduces as spurious cooling at the sink cell. Non-MHD results are bit-for-bit unchanged, as claimed. The GPU (HIP) compile passes. One `ParticleSF` test failure I found is **pre-existing** and fails identically on `development`.

Classified as a **bug fix**, so I ran the RED $\to$ GREEN $\to$ EDGE cycle.

- Base: `development` @ `87e82941` (this already contains #2157).
- PR head: `5d83f46e`. It contains #2157 through its merge of `development`, so there is no duplicate fix and no conflict — the PR description is accurate on this point. The PR is behind `development` by exactly one commit, `71e07e07`, a CodeQL action version bump, which is CI-only and harmless.
- Preset `3d`, `-DQUOKKA_PYTHON=OFF`. GPU check used the `3d-hip` preset (this machine has ROCm/`hipcc`, not CUDA).

## Why `ParticleSinkFormation` is the right test for this

This problem is strongly magnetized, which makes the bug large instead of marginal. With $\rho_0 = m_p$, $\mu = m_p$, $T_0 = 10$ K, $\gamma = 5/3$ and $B_0 = 10^{-7}$:

$$E_\mathrm{int} = 2.070973 \times 10^{-15}, \qquad E_\mathrm{mag} = \tfrac{1}{2}B_0^2 = 5 \times 10^{-15}, \qquad E_\mathrm{mag}/E_\mathrm{int} = 2.414 \quad (\beta = 0.276).$$

The magnetic energy is $2.4$ times the internal energy, so forgetting to subtract it is a factor-of-$3.4$ error in $E_\mathrm{int}$, not a rounding effect. This is why I used it in preference to `SN`, where $B_0 = 10^{-7}$ against supernova energies gives $E_\mathrm{mag}/E_\mathrm{SN} \sim 10^{-8}$ and the same bug is numerically invisible.

## Evidence

|---|---|---|
| **RED** (`development`) | `ParticleSinkFormation`, step 0 CFL diagnostic | `cs = 6.853611e+04` cm/s. Analytic value is `3.709088e+04`. **Wrong by $1.848\times$.** |
| **RED** (`development`) | Same run, plotfile at step 20, energy consistency at the sink | $(E_\mathrm{tot}-E_\mathrm{kin}-E_\mathrm{mag})$ sits $1.4128\times10^{-4}$ **below** the auxiliary $E_\mathrm{int}$ at cell $(16,16,16)$, the sink cell. Spurious cooling. |
| **GREEN** (PR) | Identical `ParticleSinkFormation` run | `cs = 3.709088e+04` — **matches the analytic value to all 7 digits.** |
| **GREEN** (PR) | Same plotfile check | Worst deviation is $-2.4355\times10^{-16}$, i.e. round-off. Zero cells deviate by more than $10^{-6}$ (base had 1). |
| **GREEN** (PR) | Global `gasEnergy` conservation | Loss shrinks from `-4.666576787e+42` to `-4.642048308e+42`: $2.45\times10^{40}$ erg of magnetic energy is no longer destroyed. |
| **EDGE** | `ParticleDeposition` (non-MHD), base vs PR | All **371 physics output lines identical**; only the performance timing differs. Bit-for-bit claim verified. |
| **EDGE** | `ParticleAccretion` (MHD), base vs PR | `rel_diff = 0.0002510797953` on both, identical. No regression. |
| **EDGE** | `SN` (MHD, exercises the fixed deposition path) | Passes on both. Conservation errors move by $\sim10^{-9}$ relative; only cooling-substep statistics differ. |
| **EDGE** | GPU compile, `3d-hip` preset | `ParticleSinkFormation`, `ParticleAccretion`, `SN` all compile and link cleanly. |
| **EDGE** | `ctest` on 7 particle/MHD tests | 6 pass. `ParticleSF` fails — **also fails identically on `development`** (see below). |
| **EDGE** | `quokka-pre-commit.sh` | clang-format, YAML, merge-conflict checks all pass. |

### The RED number is exactly predicted, not approximately

On `development`, `printCellProperties` computed $E_\mathrm{int} = E_\mathrm{tot} - E_\mathrm{kin}$, which under MHD is really $E_\mathrm{int} + E_\mathrm{mag}$. The predicted wrong sound speed is therefore

$$c_s^\mathrm{wrong} = \sqrt{\gamma(\gamma-1)(E_\mathrm{int}+E_\mathrm{mag})/\rho} = 6.853611\times10^{4}\ \mathrm{cm\,s^{-1}},$$

against the printed `6.853611e+04`. The ratio to the true value is $\sqrt{1+E_\mathrm{mag}/E_\mathrm{int}} = 1.84779$, and the observed ratio is $1.84779$. Agreement to five decimal places leaves no doubt about the mechanism.

### How I read the plotfiles

I wrote a small independent AMReX plotfile reader rather than trusting a derived field. It is validated against the code's own output: summing `gasEnergy` over the domain and multiplying by the cell volume gives `3.5077057432e+45`, against the code-printed `Initial gasEnergy = 3.507705743e+45` — agreement to 10 significant figures. `gasInternalEnergy` matches equally well, and $E_\mathrm{mag}$ recovered from the plotfile B-fields in `SN` is `5.0000e-15`, exactly $\tfrac{1}{2}B_0^2$.

## Analysis

The root cause is that `energy_index` holds $E_\mathrm{int}+E_\mathrm{kin}+E_\mathrm{mag}$ under MHD while `internalEnergy_index` holds $E_\mathrm{int}$ alone, and three code paths reconstructed the gas internal energy without subtracting $E_\mathrm{mag}$.

- `src/QuokkaSimulation.hpp:988` — was `Eint = Etot - Ekin`, now goes through `ComputeEintFromEgas` with $E_\mathrm{mag}$ read from `state_new_fc_`. Diagnostic output only; it cannot change the solution. I verified the face lookup is safe: the face-centered `BoxArray` shares the `DistributionMapping` with the cell-centered one, and for direction `idim` the face box spans $[\mathrm{lo}, \mathrm{hi}+1]$, so both `index` and `index+e_idim` are always found on the rank that owns the cell. `amrex::get_cell_data` compares with `Box::contains`, which ignores index type, so this works as intended.
- `src/particles/particle_accretion.hpp:549` — `UpdateHydroState` scaled the whole `energy_index` by the accretion factor $f$, dragging $E_\mathrm{mag}$ down with it. Since accretion cannot change $B$ without violating $\nabla\cdot B = 0$, the correct update is $E_\mathrm{tot}' = f(E_\mathrm{tot}-E_\mathrm{mag}) + E_\mathrm{mag}$, which is what the PR now does. This is the change that actually affects the solution, and it is the one my RED/GREEN pair measures.
- `src/particles/particle_deposition.hpp` — `addCompositeBufferToState` had three coupled errors and `addThermalOnlyBufferToState` one. After the fix the comparison `e_int_plus_kinetic <= e_tot_new` is consistent, because both sides now include $E_\mathrm{mag}$; on `development` the left side excluded it while the right side included it. `e_kinetic_max = e_tot_new - e_int_new_tmp - Emag` is the correct budget for the momentum limiter, and the quadratic solve for $\lambda$ uses it consistently.

The author's choice of the $E_\mathrm{tot}' = f(E_\mathrm{tot}-E_\mathrm{mag})+E_\mathrm{mag}$ form over #2157's "subtract the removed energies" form is well reasoned and I agree with it. Because `ComputeMagneticEnergy` returns exactly `0.0` when MHD is off, $(E-0)\cdot f + 0$ is bitwise identical to $E\cdot f$ for positive $E$, including under FMA contraction. My `ParticleDeposition` comparison confirms this empirically.

The new device-lambda code in `UpdateHydroState` follows the idiom already used by `ComputeScaleDown` in the same file: hoist `const_arrays()` outside the lambda, index by the box id inside, and take the address of a lambda-local `std::array` of `Array4`. This is the established pattern, and the HIP compile confirms it is GPU-safe.

**Independent sweep.** I checked the PR's claim that the sweep is complete by grepping `src/` for every remaining hand-written $E_\mathrm{tot} - E_\mathrm{kin}$ and every `energy_index` mutation. Every other site in `src/hydro/`, `src/dust/`, `src/radiation/` and `src/hydro/EOS.hpp` already subtracts `magnetic_energy`. The only remaining site is `PreInterpState` (below). I also scanned all MHD-enabled problem files in `src/problems/` and found no instance of the same pattern. The sweep is complete.

## PR-Introduced Concerns

1. **No test fails without this fix.** This is the main gap. `ParticleAccretion` is isothermal ($\gamma = 1$), so `energy_index` is dynamically inert there and the test gives bit-identical `Mdot` with and without the fix — it can never catch a regression of the accretion path. `ParticleSinkFormation` has no assertion on energy. `SN` is affected only at the $10^{-9}$ level. So if someone reverts `particle_accretion.hpp:549` tomorrow, the whole suite still goes green. Consider adding a cheap assertion that $E_\mathrm{tot} - E_\mathrm{kin} - E_\mathrm{mag}$ agrees with `internalEnergy_index` in a magnetized sink problem — my RED/GREEN check above is exactly such a test and separates base from PR by four orders of magnitude ($1.4\times10^{-4}$ against $2.4\times10^{-16}$).

2. **Stale variable name.** In `addCompositeBufferToState`, `e_int_plus_kinetic` now also contains the magnetic energy, so the name no longer describes the value. The added comment says so, but renaming it to something like `e_tot_at_fixed_sie` would prevent the next reader from repeating this very bug. Cosmetic.

3. **`printCellProperties` cost on GPU.** The new code adds $2\times$`AMREX_SPACEDIM` = 6 `amrex::get_cell_data` calls, each of which allocates a `Gpu::DeviceVector`, launches a kernel and does a device-to-host copy, on every rank and every coarse step when `amr.v = 1`. This is a diagnostic path so the impact is small, but on GPU each call is a synchronizing copy. Hoisting the block inside the existing `if (!cell_values.empty())` guard would confine the cost to the one rank that actually prints.

4. **Silent fallback.** `if (!b_lo.empty() && !b_hi.empty())` skips that direction's contribution when a face is not found, which would silently under-count $E_\mathrm{mag}$ and print a plausible but wrong sound speed. My analysis above says it cannot trigger, so this is defensive code that can only hide a fault rather than surface it. Very minor.

1. **`ParticleSF` aborts with a floating-point exception.** It fails on the PR branch and fails identically on `development` @ `87e82941`, so it is not caused by this PR and does not block it. The backtrace points into `quokka::ResampledCooling::computeCooling<ParticleSFProblem>`, reached from `addStrangSplitSourcesWithBuiltin`, aborting during the first hydro step. The cooling tables it references are present on disk. `ParticleSF` is a non-MHD problem, and I verified this PR leaves non-MHD results bit-for-bit unchanged, so the two are independent. Needs a separate issue.

2. **The derived `temperature` field written to plotfiles appears to be uninitialized.** In `ParticleSinkFormation` the initial condition sets `rho_e = CV * T0 * rho`, so temperature must be exactly $10$ K in every cell at $t=0$. The `temperature` component instead ranges over $[-4.11\times10^{4}, 2.31\times10^{4}]$ in the step-0 plotfile and $[-1.32\times10^{9}, 2.07\times10^{-15}]$ at step 1, with the maximum equal to the ambient $E_\mathrm{int}$ value rather than any temperature. The six conserved components in those same files read back correctly to 10 significant figures against the code's own conservation sums, so this is not a reader artifact. This is unrelated to the PR and needs its own investigation before being treated as confirmed.

3. **`PreInterpState` / `PostInterpState` and MHD.** The author flagged this in the PR description and deliberately left it alone; I confirmed the description is accurate. `src/QuokkaSimulation.hpp:2101` converts $E_\mathrm{tot}$ to $(E_\mathrm{tot}-E_\mathrm{kin})/\rho$, which under MHD is $(E_\mathrm{int}+E_\mathrm{mag})/\rho$. The round trip is exact within a cell, so it is harmless on a single level, but coarse-fine ghost cells receive a coarse-interpolated $E_\mathrm{mag}$ that does not match the separately divergence-preserving interpolated face-centered $B$. Fixing it needs $B$ on the coarse temporary MultiFab inside `FillPatchWithData`, whose BoxArray and DistributionMapping differ from `state_new_fc_`, so it would mean changing the interp-hook interface in `simulation.hpp`. Agreed that this belongs in a follow-up issue, not here.

## Recommendation

Approve on correctness once the draft status is lifted and GPU CI has been triggered. The one thing genuinely worth adding before merge is a regression test that fails without the accretion fix, since nothing in the current suite does.

markkrumholz (approved):

PR #2162, opened 2026-08-13
Allow heating_rate_external to depend on position

chongchonghe:
### Description

`heating_rate_external` could only vary in time: the parser was compiled with `registerVariables({"time", "dt"})` and evaluated once per level per Strang step on the host, so the resulting scalar was added uniformly to every cell. This PR makes it a function of position as well.

The parser is now compiled as `compile<5>()` over `{x, y, z, time, dt}` and evaluated per cell, at cell centres, inside the cooling kernel. `y` and `z` are zero when `AMREX_SPACEDIM` is smaller than 2 or 3, matching how `density_floor_expr` handles reduced dimensionality. Existing expressions that use only `time` and `dt` (e.g. `inputs/TallBoxSf.toml`) keep working unchanged.

Implementation:

- `src/simulation.hpp`: register `x`, `y`, `z` as parser variables and skip them when resolving ParmParse constants.
- `src/cooling/ResampledCooling.hpp`: new `ExternalHeatingRate` struct that carries the compiled `ParserExecutor<5>` plus the level's `prob_lo`/`dx`/`time`/`dt` and turns `(i, j, k)` into a heating rate. It is device-safe and returns zero when default-constructed (no parser set), so the no-expression path is unchanged. `computeCooling` takes it as an optional argument and adds it to the spatially uniform photoelectric heating rate per cell.
- `src/QuokkaSimulation.hpp`: `computeExternalHeatingRate(time, dt)` is replaced by `makeExternalHeatingRate(lev, time, dt)`, which builds that struct from the level geometry.

### Verification

New test `ResampledCoolingTestSpatialHeating` (`inputs/ResampledCoolingTestSpatialHeating.toml`) reuses the `ResampledCoolingTest` executable with `heating_rate_external = "if(x > 5.0e26, 1.0e-25, 0.0)"`, so only the upper half of the domain in $x$ is heated. The box is made large enough that the sound-crossing time is about 200 times the stop time, and `max_dt` keeps the timestep at the same cadence as the uniform test; without this the heated gas simply expands across the whole box and the contrast is erased. The test asserts $T(x_{\rm last}) / T(x_{\rm first}) > 10$; the measured ratio is 1473 in both 1D and 3D, and the unheated half reaches exactly the same final temperature as the uniform test ($T_{\rm final}/T_{\rm initial} = 9.253 \times 10^{-4}$).

Ran locally:

- `ResampledCoolingTest` (1D and 3D) — passes, L1 error against the Grackle reference is 0.0117 (tolerance 0.02), unchanged from `development`.
- `ResampledCoolingTestSpatialHeating` (1D and 3D) — passes.
- `ResampledCoolingTestV2` (3D, untouched by this PR) — passes.
- `HydroShocktube` (1D, untouched) — passes.
- CUDA compile check (`1d-cuda`, `ResampledCoolingTest`) — the parser executor is now captured in a `AMREX_GPU_DEVICE` lambda, so this confirms the device path compiles.
- 3D CPU build of `TallBoxSf`, the existing user of `heating_rate_external`.

### Related issues
N/A

### Checklist

## Summary by CodeRabbit

* **New Features**
  * External heating expressions now support spatial variation using `x`, `y`, and `z`, in addition to time-based variables.
  * Heating is evaluated at cell centers, enabling different heating rates across the simulation domain.
  * Added a spatial-heating cooling test configuration and validation.

* **Documentation**
  * Updated parameter and cooling documentation to describe spatially varying external heating and dimensionality-dependent behavior.


BenWibking:
This makes sense to me.

What is the use case for `dt` in the heating expression?

chongchonghe:
> This makes sense to me.
> 
> What is the use case for `dt` in the heating expression?

No particular use so far. I just registered it anyway, since time and dt usually come together. 

One usage case I can think of is to insert a fix total amount of energy in one time step, by setting the heating rate to `Q / dt`.

chongchonghe:
# Review of PR #2162 — Allow `heating_rate_external` to depend on position

**Classification:** feature / enhancement → SPEC → VERIFY → EDGE cycle.

**Platform:** CPU build (GCC, 48 cores) plus a ROCm/HIP device compile. This box has `hipcc` and `/opt/rocm` but no CUDA toolkit and no GPU device, so the GPU check is a compile-only check (`1d-hip`); runtime GPU behaviour remains CI's job.

**Verdict:** the feature works as described, and every claim in the PR description that I could check reproduced. Backward compatibility is exact — bit-identical results on `development` and on the PR branch for the existing time-only expressions. No PR-introduced blocking concerns; two minor, non-blocking notes are listed at the end.

## SPEC — the contract extracted from the description and diff

| # | Claim | Where it comes from |
|---|---|---|
| 1 | `heating_rate_external` is compiled over `{x, y, z, time, dt}` and evaluated per cell at cell centres | `src/simulation.hpp:1061-1062`, `src/cooling/ResampledCooling.hpp:85-108` |
| 2 | `y` and `z` are zero when `AMREX_SPACEDIM` is smaller than 2 or 3 | `src/cooling/ResampledCooling.hpp:96-105` |
| 3 | Expressions that use only `time` and `dt` keep working unchanged | `inputs/TallBoxSf.toml:57`, `inputs/TallBoxSf_regression.toml:58` |
| 4 | Unit constants `yr`, `kyr`, `Myr`, `Gyr` still resolve from ParmParse; `x`, `y`, `z`, `time`, `dt` are skipped | `src/simulation.hpp:1047-1052` |
| 5 | A default-constructed `ExternalHeatingRate` returns zero, so the no-expression path is unchanged | `src/cooling/ResampledCooling.hpp:90-92` |
| 6 | The new test `ResampledCoolingTestSpatialHeating` asserts $T(x_{\rm last})/T(x_{\rm first}) > 10$, measured 1473 in 1D and 3D | `inputs/ResampledCoolingTestSpatialHeating.toml`, `src/problems/ResampledCoolingTest/testResampledCoolingTest.cpp:229-247` |
| 7 | The parser executor is captured in an `AMREX_GPU_DEVICE` lambda, so the device path must compile | `src/cooling/ResampledCooling.hpp:170-175` |

## Evidence

|---|---|---|
| VERIFY — golden path, 1D | `quokka run -d 1d --filter ResampledCooling` on the PR branch | Both tests pass. `Spatial heating check: T(x_last)/T(x_first) = 1473.091496 (required > 10)`; unheated half `T_final/T_initial = 0.000925291305`, against `0.0009253102749` for the uniform `ResampledCoolingTest` — the two halves behave as the description says. Claim 6 confirmed. |
| VERIFY — golden path, 3D | `quokka run -d 3d --filter ResampledCooling` | 3/3 pass (`ResampledCoolingTest`, `ResampledCoolingTestSpatialHeating`, `ResampledCoolingTestV2`). Ratio `1473.03355`. |
| VERIFY — 2D | `quokka run -d 2d --filter ResampledCooling` | 2/2 pass. The new test is registered unconditionally in `CMakeLists.txt`, so it runs in 2D CI as well; it works there. |
| VERIFY — claim 1, feature is genuinely new | Base branch (`origin/development`, `71e07e07`) run with `heating_rate_external="if(x>5.0e26,1.0e-25,0.0)"` | Aborts: `Unknown variable x in Parser expression`. On the PR branch the same expression runs and produces the spatial contrast. This is the clean "before" evidence that the capability did not exist. |
| VERIFY — claim 2, `y` and `z` vanish in 1D | 1D, `heating_rate_external` set to `1.0e-25`, then to `1.0e-25*(y+z+1.0)`, then to `1.0e-25*(0.0*x+1.0)` | All three give `T_final/T_initial = 1.363038753` and a spatial ratio of exactly `1`. In 1D the expressions are equivalent, as claimed. |
| VERIFY — `y` and `z` are really wired in 3D | 3D, `if(y>5.0e26,...)` vs `if(y<5.0e26,...)`, and the same pair in `z` | The extracted ray sits at $y, z < 5\times10^{26}$, and the results flip exactly as they should: `0.0009253164269` (ray not heated) vs `1.363038753` (ray heated), identically for `y` and for `z`. The PR's own test only exercises `x`, so this closes the gap. |
| EDGE — backward compatibility, real existing user | The exact `TallBoxSf` expression `max(min(1.0, 2.0 - time / (32 * Myr)), 0.0024) * 2e-26` run on `ResampledCoolingTest.toml`, on base and on the PR branch | Bit-identical on both: `T_final/T_initial = 0.0009260054408`, relative L1 error `0.02316209324`. Claims 3 and 4 confirmed — the `Myr` constant still resolves and per-cell evaluation reproduces the old host-side single evaluation exactly. |
| EDGE — backward compatibility, uniform constant | `heating_rate_external="1.0e-25"` on `ResampledCoolingTest.toml`, base vs PR | Bit-identical: `1.363038775` on both. |
| EDGE — no-expression path unchanged | `ResampledCoolingTest.toml` (no heating), base vs PR | Bit-identical: `T_final/T_initial = 0.0009253102749`, relative L1 error `0.01167824563` against tolerance `0.02`. Claim 5 confirmed, and this matches the number quoted in the PR description. |
| EDGE — error handling | `heating_rate_external="1.0e-25*bogus_symbol"` | Aborts with `Unknown variable bogus_symbol in Parser expression ...`. The wider skip list did not weaken symbol validation. |
| EDGE — regression, adjacent test | `ResampledCoolingTestV2` (3D, untouched by the PR) | Passes. |
| GPU — device compile | `quokka config -d 1d-hip` then `quokka build -d 1d-hip ResampledCoolingTest` | `ResampledCoolingTest SUCCESS`. Claim 7 confirmed on the ROCm path. The only two `-Wgpu-maybe-wrong-side` warnings come from pre-existing `[this]` captures at `src/QuokkaSimulation.hpp:1268` and `src/QuokkaSimulation.hpp:2017`, both in the density-floor code that this PR does not touch. Nothing warns on the new capture. |
| Build — existing user compiles | `quokka build -d 3d TallBoxSf` | `TallBoxSf SUCCESS`. |

One anomaly during testing turned out to be my own artefact rather than a code problem: passing an expression that begins with a parenthesis as a command-line ParmParse override, such as `heating_rate_external="(1.0e-25*(time/Myr)*0.0)+1.0e-25"`, gets mis-tokenised and silently evaluates to roughly zero. The same expression placed in a TOML file evaluates correctly (`1.363038753`). This is ParmParse command-line tokenisation, is not specific to this parameter, and is unrelated to the PR. I mention it only so the number is not mistaken for a real finding.

## Analysis

The design matches what the description states. The old code evaluated the parser once per level per Strang step on the host and passed a scalar into `computeCooling`; the new code moves the evaluation inside the cooling kernel. The mechanism is `quokka::ResampledCooling::ExternalHeatingRate` (`src/cooling/ResampledCooling.hpp:85-108`), a small aggregate holding `amrex::ParserExecutor<5>` plus the level's `prob_lo`, `dx`, `time` and `dt`, with a device-callable `operator()(i, j, k)` that builds the cell centre and calls the parser.

Three details make this safe, and I checked each one:

- The zero path is guarded by `if (!parser)`. `amrex::ParserExecutor::operator bool` is declared `AMREX_GPU_HOST_DEVICE` at `extern/amrex/Src/Base/Parser/AMReX_Parser.H:101-104` and returns `m_device_executor != nullptr` on device, so a default-constructed instance correctly returns zero on both sides. The bit-identical no-heating numbers above confirm this in practice.
- `computeCooling` takes `ExternalHeatingRate const &` and the kernel captures it with `[=]`, which copies the referenced aggregate into the lambda. The aggregate holds only value types and the executor's own pointers, and the owning `amrex::Parser` (`heatingRateExternalParser_`) lives in the simulation object for the whole run, so there is no host-pointer capture and no lifetime problem. The HIP compile agrees.
- `makeExternalHeatingRate(lev, ...)` is called from `addStrangSplitSourcesWithBuiltin` at `src/QuokkaSimulation.hpp:1171`, where `lev` is the function's own level parameter, and it uses `geom[lev].ProbLoArray()` and `geom[lev].CellSizeArray()`. That is the correct per-level geometry, uses the array forms required by the GPU-safety rules in `CLAUDE.md`, and is consistent with how `applyTurbulence` a few lines below obtains its cell sizes. Cell centres on refined levels will therefore be right, although the test suite does not exercise AMR for this feature.

`grep` over `src/` shows a single call site for `computeCooling` and no leftover references to the removed `computeExternalHeatingRate`, so the replacement is complete.

The new test is well posed. The box is large enough that the sound-crossing time is far longer than `stop_time`, so each half evolves at nearly constant density and the contrast is not smeared out; the measured ratio of 1473 sits three orders of magnitude above the threshold of 10, which leaves no tolerance sensitivity to worry about. The `!(T_ratio_x > ...)` form correctly fails on NaN, and the comment says so. Calling `fextract` before the `IOProcessor` guard is right, since `fextract` is collective.

The documentation changes in `docs/markdown/parameters.md` and `docs/markdown/cooling_module.md` accurately describe the new behaviour, including the reduced-dimensionality rule and the distinction between the spatially uniform SFH photoelectric term and the per-cell parser term.

## PR-Introduced Concerns

Neither of these blocks the PR.

1. **The new test's temperature diagnostic ignores kinetic energy** (`src/problems/ResampledCoolingTest/testResampledCoolingTest.cpp:235-238`). It computes internal energy as `Etot - active_magnetic_energy_initial`, which reuses the assumption stated in `computeAfterTimestep` that there is no kinetic energy. That assumption holds for the uniform test but is no longer strictly true for the spatial-heating test, which deliberately creates a pressure discontinuity at $x = 5\times10^{26}$ and therefore some gas motion. The effect is negligible here because the box is huge and the run is short, and the assertion has a factor of ~150 of headroom, so the reported ratio of 1473 is not in doubt. Still, `HydroSystem<...>::internalEnergy_index` is already among the components returned by `fextract`, so subtracting the kinetic term (or reading that component) would make the printed temperature exact rather than approximate. Worth a one-line change if you touch the file again.

2. **The `ExternalHeatingRate const &external_heating = {}` default argument is unused** (`src/cooling/ResampledCooling.hpp:140`). There is exactly one caller and it always passes the argument explicitly. The default is harmless, but it is unexercised API surface: if a future caller forgets the argument, external heating silently disappears rather than failing to compile. Making the parameter required would remove that possibility.

**Performance note, not a concern:** the parser is now evaluated once per cell per Strang step even when the expression has no spatial dependence, which is the situation for both existing users (`TallBoxSf.toml`, `TallBoxSf_regression.toml`). I did not measure this — the tests available here are far too small to resolve it — but a parser bytecode evaluation should be negligible next to the per-cell cooling ODE integration that immediately follows it. Flagging it only so the trade-off is on record.

1. `src/QuokkaSimulation.hpp:1268` and `src/QuokkaSimulation.hpp:2017` capture `this` in `AMREX_GPU_HOST_DEVICE` lambdas in the density-floor path, which produces `-Wgpu-maybe-wrong-side` warnings on the HIP build. These are the only device warnings in the `1d-hip` build of `ResampledCoolingTest`. They are in the `else` branches that call `densityFloor(...)` when no parser expression is set, are untouched by this PR, and do not block it. They would need a separate PR — the fix is to hoist the needed members into local values before the lambda, as the parser branches already do.

markkrumholz (approved):

PR #2167, opened 2026-08-16
Add Strömgren-volume photoionization feedback module

chongchonghe:
### Description

Adds a Strömgren-volume photoionization feedback module: a cheap subgrid alternative to the existing M1 + photochemistry path, for runs where H II regions are unresolved. No photons are transported. For each star particle the module finds the volume whose recombinations exactly consume the particle's ionizing photon budget $Q$, marks that gas ionized, and holds it at a fixed $T_{\rm HII}$; the resulting overpressure drives the expansion.

The method follows the Strömgren-volume technique of Kessel-Deynet & Burkert (2000) and Dale et al. (2007b), in the form used for stellar feedback by Hopkins et al. (2018, FIRE-2, appendix E), and assumes the on-the-spot approximation.

**Implementation note.** FIRE-2 sorts nearby cells by distance and walks outward consuming the budget, which on an AMReX grid would mean gathering a cell list across boxes, ranks and levels. That is avoidable: because cells are consumed in order of increasing distance, the consumed set is always a distance-prefix, which is exactly a ball. The walk is therefore equivalent to finding the single radius $R_{\rm St}$ at which the enclosed recombination rate equals $Q$ — and this equivalence is exact for an arbitrary, non-uniform density field, not just a uniform one. The module exploits this to replace the cross-rank sort with a radial-bin histogram plus one small MPI reduction per source. Overlapping sources are processed in descending $Q$ order against a shared ionized mask, so photons are never spent twice.

The ionizing photon rate is assigned once at birth from the stellar mass, stored in a new per-particle component on `ToyStellarModel` (`nExtraReal` 0 → 1) and not refreshed as the particle accretes. It is a least-squares fit to all twelve luminosity-class-V rows of Martins, Schaerer & Hillier (2005) Table 1 — their recalibration of the Sternberg, Hoffmann & Pauldrach (2003) grid — regressing the tabulated $\log_{10} Q_0$ on $x \equiv \log_{10}(M_{\rm spec}/M_\odot)$:

$$\log_{10} Q_0\ [\mathrm{s}^{-1}] = 34.3149 + 15.8931\,x - 4.0825\,x^2,$$

spanning $16.46$–$58.34\,M_\odot$ (O9.5V to O3V) with rms residual $0.036$ dex and maximum residual $0.064$ dex.

A quadratic is used rather than a cubic deliberately: a cubic lowers the rms only from $0.036$ to $0.035$ dex, but its leading coefficient is negative, so it peaks at $71\,M_\odot$ and declines thereafter — it would assign a $150\,M_\odot$ star less ionizing flux than a $37\,M_\odot$ one. Outside the tabulated range the fit is not extrapolated: below $16.46\,M_\odot$ $Q$ is zero (extrapolating gives $10^{43.4}$ photons/s at $5\,M_\odot$ against a true $\sim10^{38}$, and low-mass stars dominate a sampled IMF by number), and above $58.34\,M_\odot$ the mass is clamped so $Q$ saturates at the O3V value rather than following the fit into its turnover.

The module runs from `particleMeshInteraction` on the finest level, matching the existing supernova-deposition convention. It is off by default (`stromgren.enabled = 0`), so behaviour is unchanged for every existing problem.

**Known limitations**, all documented in `docs/markdown/photoionization.md`: the ionized region is always spherical (no champagne flows or shadowing — inherent to the approximation, and the reason the angular/HEALPix variant was considered and deferred); finest-level only; a budget unexhausted within `R_max_cells` is discarded, as in FIRE-1; the EOS's fixed $\mu$ means users should set $T_{\rm HII} \approx 2\times10^4$ K to get the right overpressure; no radiation pressure, and MHD is rejected at runtime.

**Review fixes (`eea6649e`).** A review pass found four real issues, all fixed: the `x_ion` output slot was guarded in neither direction (an index aliasing a mass scalar would have silently corrupted the composition via `EnforceLimits`, and an out-of-range one silently wrote nothing) — both now abort with the reason; the heating added the same increment to both energy variables instead of setting them from one internal energy as `depositSN` does; `x_ion` was written raw into a slot the rest of the code treats as a conserved density (`rho*s`), and now stores `rho * x_ion`; and the birth sentinel's comment wrongly claimed AMReX zero-initializes particle reals, so `ParticleStarEvolution` could read indeterminate memory. Two CI variants were added for the previously hand-checked overlap and subgrid paths.

**Verified by** four tests plus a regression run:

- `StromgrenVolumeFeedback` — uniform medium, one $30\,M_\odot$ star ($Q = 7.65\times10^{48}\ \mathrm{s}^{-1}$ from the calibration). The equivalent radius of the ionized region matches the analytic $R_{\rm St} = (3Q/4\pi\alpha_B n_H^2)^{1/3}$ to **0.0064 cells** (tolerance: one cell), exercising the real birth-assignment path rather than the override. The same test validates the calibration against five rows of Martins et al. Table 1 (all within $0.052$ dex), and checks both range edges: a $10\,M_\odot$ star gets no ionizing photons, and a $150\,M_\odot$ star gets exactly the clamped $58.34\,M_\odot$ rate.
- `StromgrenVolumeGradient` — same binary, linear density gradient and an off-centre source. Here the analytic radius does not apply (the measured radius is 1.14 cells away from it, so this is a genuinely different configuration), and the test instead confirms that the recombination rate inside the ionized region equals $Q$ to better than $10^{-6}$ — i.e. exact photon conservation on the non-uniform path.
- `StromgrenVolumeSubgrid` — $Q$ small enough that $R_{\rm St} \approx 0.005$ cells, the unresolved regime the module exists for. A one-cell tolerance is vacuous there, so this variant asserts relative accuracy; measured error is $2\times10^{-7}$.
- `StromgrenVolumeOverlap` — two co-located sources. The combined region matches $2^{1/3} R_{\rm St}$ to 0.007 cells and recombinations total exactly $2Q$, confirming photons are not spent twice.
- `ParticleStarEvolution` (exercises the Star particle layout that changed) passes its physics checks unchanged — mean $\dot{M} = 6.3118\times10^{14}$ g/s against the analytic Bondi rate, identical to before this PR. See the note below on its FPE-trap failure.
- CUDA compile check at `3d-cuda` passes. It initially caught a real bug — an extended `__device__` lambda first-capturing in an `if constexpr` context — which is fixed in the second commit.

Note: `ParticleStarEvolution` fails under `ctest` on aarch64 CPU builds, where the FPE traps are active. This is pre-existing and unrelated to this PR — the trap fires inside `HydroSystem::ComputeFluxes` at step 1, before any particle code runs, and the identical failure reproduces on pristine `origin/development`. It passes on x86/HIP and when run without the traps.

Design notes and the literature review are in `docs/superpowers/specs/2026-08-16-stromgren-volume-photoionization-design.md`.

### Related issues

N/A

### Checklist


chongchonghe:
**Verdict:** 1 blocking guard fix; the algorithm itself verified correct under harder tests than the PR ships.
**Checked:** `3d-hip` (ROCm 6.4, gfx90a) build + GPU runs. `StromgrenVolumeFeedback` `|R_eff - R_St| = 0.006709` cells, `StromgrenVolumeGradient` recombination `1e+49` vs `Q = 1e+49`, `ParticleStarEvolution` passed. Both PR numbers reproduce to the digit.

**Blocking**
1. **`x_ion_scalar_index` is guarded in neither direction** — `particle_photoionization.hpp:342`. The check is only `< nscalars_`, but mass scalars occupy the *first* slots of that same block (`radiation_system.hpp:884`; `numPassiveScalars = numMassScalars + N`, `physics_info.hpp:39`), so `x_ion_scalar_index=0` on a problem with mass scalars overwrites a species partial density every step and `EnforceLimits` renormalizes it into the composition (`hydro_system.hpp:1077`) — silently, and in exactly the galaxy/GMC use case this module targets. Out-of-range fails silently too: with `numPassiveScalars=1`, `x_ion_scalar_index=5` runs to completion with `x_ion` all zeros.

**Non-blocking**
- **Birth sentinel reads indeterminate memory** — `stellar_models.hpp:95`. The comment claims AMReX zero-initializes particle reals; `InitFromAsciiFile` does not (`AMReX_Particle.H:358` has no initializer), as this PR's own `photoionization.md` states. `ParticleStarEvolution` never zeroes the new `Q` slot (`testParticleStarEvolution.cpp:95`), so a positive garbage value would pin `Q` for the run.
- **Dual-energy update inconsistent** — `particle_photoionization.hpp:359,368`. `dEint` is derived from `E_tot - Ekin` but added to *both* energy variables; `depositSN` instead resyncs `e_int = e_tot - Ekin` (`particle_deposition.hpp:776`).
- **`x_ion` written raw into a slot that is `rho*s` elsewhere** — `particle_photoionization.hpp:373` vs `particle_deposition.hpp:786`.
- **Overlap and subgrid paths have no CI coverage** — both are headline claims; I verified them by hand (below).

| Phase | Ran | Observed |
|---|---|---|
| Build | `3d-hip`, `StromgrenVolumeFeedback` + `ParticleStarEvolution` | Both SUCCESS. No warning from the new module (the `-Wgpu-maybe-wrong-side` ones are pre-existing, `QuokkaSimulation.hpp:1268/2017`). |
| Verify | `ctest StromgrenVolumeFeedback` | Passed 9.6 s. `Q(M_star)=1e+49`, `R_St=10.87185` cells, `R_eff=10.87856`, diff `0.006709` cells. |
| Verify | `ctest StromgrenVolumeGradient` | Passed 2.3 s. diff `1.143189` cells; recombination `1e+49` vs `Q=1e+49`. |
| Verify | `ctest ParticleStarEvolution` | Passed 3.7 s. `dM/dt = 6.3118e+14` vs Bondi `6.2202e+14`, tol 0.02. |
| Edge | `stromgren.enabled=0` | `R_eff = 0`, `FAIL: no gas was ionized`. The pass above is not vacuous. |
| Edge | `R_max_cells=4` (`R_St≈10.9`) | Warns `1 source(s) did not exhaust their ionizing photon budget`; `R_eff=4.058` cells. |
| Edge | `Q_ion=1e40`, `stop_time=1e6` (gas frozen) | `R_eff=2.096707e+15` vs analytic `2.096704e+15` cm — rel. err `1.4e-6` at `R_St=0.011` cells. Subgrid path exact. |
| Edge | `Q_ion=1e40`, default `stop_time` | `R_eff` up by 1.1664x, but recombination still exactly `1e+40`. Real hydro expansion lowering central density, not a bug. |
| Edge | 2 co-located sources | `R_eff=13.6976740` vs predicted `2^(1/3)*R_St=13.6976739` cells; recombination exactly `2e+49`. No double-spending. |
| Edge | `ParticleStarEvolution` + `stromgren.enabled=1` | Aborts at `particle_photoionization.hpp:221` with the MHD message. Rejection works. |

</details>

chongchonghe:
All four findings validated against the code, none were false positives. All are now fixed in `eea6649e`.

**Blocking: `x_ion_scalar_index` guard.** Confirmed. `ComputeMassScalars` reads `scalar0_index + n` for `n < nmscalars_`, so mass scalars really do occupy the *first* slots of the block that `numPassiveScalars` counts, and `EnforceLimits` renormalizes whatever lands there back into the composition. The module now rejects an index below `numMassScalars` or at/above `numPassiveScalars`, aborting with the reason instead of silently corrupting or silently doing nothing:

```
Assertion `par.x_ion_scalar_index < nscalars' failed ... Msg: stromgren.x_ion_scalar_index
is out of range; it must be < Physics_Traits::numPassiveScalars.
```

**`x_ion` written raw into a `rho*s` slot.** Confirmed — `EnforceLimits` rescales scalars by `rho/rho_new` and `depositSN`'s own comment says "scalars are conserved densities". Now stores `rho * x_ion`; the test divides by `rho` to recover the fraction, and the docs say so.

**Dual-energy.** Confirmed. Now sets both variables from one internal energy (`e_tot = Ekin + Eint_new`, `e_int = Eint_new`), matching `depositSN`, rather than adding the same delta to each — which preserved any drift instead of landing on `T_HII`.

**Birth sentinel.** Confirmed, and the misleading comment was mine — corrected to state that `InitFromAsciiFile` leaves the remaining reals *indeterminate*. `ParticleStarEvolution` now zeroes from `StarParticleMdotIdx` onward, which also covers `mdot` — read by `evolve` before the accretion module first writes it.

**CI coverage for the overlap and subgrid paths.** Both are now CI tests rather than hand checks, and they reproduce your numbers:

| Test | Result |
|---|---|
| `StromgrenVolumeSubgrid` (`Q=1e40`, `R_St = 0.0054` cells) | relative radius error `1.96e-07`; recombination `1e+40` vs `Q=1e+40`. A one-cell tolerance is vacuous here, so this variant asserts *relative* accuracy. |
| `StromgrenVolumeOverlap` (2 co-located sources) | `R_eff = 13.70493` vs predicted `2^(1/3) R_St = 13.69767` cells; recombination `2e+49` vs `Q_total = 2e+49`. |

I also noticed the variant `add_test` entries were missing `${QuokkaTestParams}`, so they had been running *without* the FPE traps the main test gets. Fixed; all four pass with traps enabled.

**One correction to the record.** `ParticleStarEvolution` fails under `ctest` on this machine (aarch64 CPU) — but it is **not** caused by this PR. The trap fires in `HydroSystem::ComputeFluxes` during step 1, before any particle code runs, and I confirmed it by building pristine `origin/development` in a scratch worktree and reproducing the identical failure there. It passes when run without the FPE traps, which is why it looked green earlier and why your HIP build passed it. Pre-existing and platform-specific; not something this PR should fix, but worth flagging separately.

PR #2181, opened 2026-08-27
Support turning off hydro and radiation while leaving self-gravity on

chongchonghe:
### Description

Quokka could not run a problem with hydro and radiation both disabled: `computeMaxSignalLocal` aborted with *"At least one of hydro or radiation must be enabled! Cannot compute a time step."* Several other code paths also indexed hydro components of the cell-centred state, which holds only a single placeholder component when there is no hyperbolic system (`Physics_Indices::nvarTotal_cc == 1`).

This PR makes self-gravity usable on its own and adds a test for it.

Core changes:
- The timestep no longer requires a hyperbolic system. `computeMaxSignalLocal` returns a zero signal speed instead of aborting, and `computeTimestepAtLevel` leaves the hydro timestep unconstrained in that case so the particle CFL sets `dt`. If no enabled physics module constrains the timestep at all and neither `constant_dt` nor `max_dt` is set, `computeTimestepAtLevel` aborts with a clear message; the check lives there because `computeTimestep` later clips `dt` to `1.1 * dt_[lev]`, which would turn an unconstrained timestep into a large-but-finite one that no later check could recognise. The `timestep limited by ...` attribution is suppressed when nothing constrains `dt`, so an unconstrained step is never misreported as hydro-limited.
- Guarded the paths that index into hydro state components when there is no hydro state: `fillPoissonRhsAtLevel`, `applyPoissonGravityAtLevel`, `ApplyHydroStateFixup`, `CheckHydroStates` and `computeAfterEvolve`.
- `advanceSingleTimestepAtLevel` no longer copies six components out of a one-component state.
- `defineComponentNames` names the placeholder component, so plotfiles and conservation sums stay consistent.
- Added a `static_assert` requiring hydro or radiation when passive scalars or dust are enabled. Those components live inside the hydro/radiation block, so without either module `nvarTotal_cc` silently collapses to `nvarTotal_cc_adv` and drops them. This was reachable only in theory before; this PR makes the configuration real.

New test `BinaryOrbitCICGravityOnly`: the `BinaryOrbitCIC` binary orbit with hydro, MHD and radiation switched off, using the same particles, the same grid and the same 0.18 cell-width tolerance.

### Verification

Run at the same fixed timestep, `BinaryOrbitCIC` and `BinaryOrbitCICGravityOnly` report identical particle separations (6.179871901e12 cm and 6.164338886e12 cm at cycles 10 and 20) and an identical max deviation of 0.05482311283 cell widths, confirming that the gas in the original test is dynamically irrelevant. The new problem also runs clean under a Debug build with AMReX bounds checking.

The guard against an unconstrained timestep was checked directly with a zero-velocity particle IC: it aborts with `No enabled physics module constrains the timestep!`, and runs normally once `max_dt` or `constant_dt` is supplied.

Regression tests, all passing locally on CPU: the full `BinaryOrbitCIC*` suite (including the split, AMR and refactor-restart variants), `GravRadParticle3D` (radiation + gravity + particles with hydro disabled), `ParticleDeposition*`, `HydroShocktube`, `RadMarshak` (radiation-only) and `RadhydroShock`.

### Related issues
N/A

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for self-gravity simulations with particles when hydrodynamics and radiation are disabled.
  * Added a binary-orbit gravity-only simulation configuration and validation test.
  * Added safer timestep handling for simulations without hyperbolic physics.

* **Bug Fixes**
  * Invalid physics combinations, including passive scalars, dust, or electron conduction without hydro or radiation, now fail early.
  * Gravity-only simulations no longer perform irrelevant hydro operations or encounter zero-speed timestep errors.

* **Documentation**
  * Clarified compatibility and state-component behavior for gravity-only simulations.

### DIFF
--- a/docs/markdown/components.md
+++ b/docs/markdown/components.md
@@ -76,6 +76,7 @@ Notes:
 - `Cooling + particles` is exercised by `ParticleSF`, `TallBoxSf`, `RandomBlast`, and `DiskGalaxy` through their input files.
 - `Hydro + dust` is exercised by the dust dynamics problems `DustAdvection*`, `DustDamping*`, `DustSoundwave`, and `DustyShock`.
 - `MHD + dust` is exercised by charged-dust and dust-MHD problems including `DustDampedGyromotion`, `DustDampingMHDZeroB`, `DustHallPedersenDrift`, `DustLorentzShock`, `DustMagnetizedRDI`, `DustyAlfvenWave`, and `DustyOrszagTang`.
+- `Self-gravity + particles` with **no** hyperbolic module is supported: a problem may set `is_hydro_enabled = false` and `is_radiation_enabled = false` while keeping `is_self_gravity_enabled = true`, in which case the timestep is set by the particle CFL condition alone. This is tested by `BinaryOrbitCICGravityOnly`, which reproduces the `BinaryOrbitCIC` orbit without any gas.
 
 
 ## Cell-centred state vector layout
@@ -102,7 +103,7 @@ Defined in `src/physics_info.hpp`. Computes starting indices and total component
 
 **`nvarTotal_cc` calculation:**
 
-- If neither hydro nor radiation is enabled: `nvarTotal_cc = nvarTotal_cc_adv` (= 1, for pure advection).
+- If neither hydro nor radiation is enabled: `nvarTotal_cc = nvarTotal_cc_adv` (= 1). In `AdvectionSimulation` this single component is the advected density; in `QuokkaSimulation` (e.g. a gravity-only problem with particles) nothing evolves it, and it is named `placeholder` in plotfiles and conservation sums.
 - Otherwise: `nvarTotal_cc = numHydroVars + numPassiveScalars + numDustVarsPerGroup * nDustGroups * is_dust_enabled + numRadVarsPerGroup * nGroups * is_radiation_enabled`.
 
 ## Cell-centred component map
--- a/inputs/BinaryOrbitCICGravityOnly.toml
+++ b/inputs/BinaryOrbitCICGravityOnly.toml
@@ -1,0 +1,33 @@
+# *****************************************************************
+# Problem size and geometry
+# *****************************************************************
+geometry.prob_lo = [-2.5e13, -2.5e13, -2.5e13]
+geometry.prob_hi = [2.5e13, 2.5e13, 2.5e13]
+quokka.bc = ["periodic", "periodic", "reflecting"]
+
+# *****************************************************************
+# VERBOSITY
+# *****************************************************************
+amr.v = 1  # verbosity in Amr
+
+# *****************************************************************
+# Resolution and refinement
+# *****************************************************************
+amr.n_cell = [32, 32, 32]
+amr.max_level = 0  # number of levels = max_level + 1
+amr.blocking_factor = 32  # grid size must be divisible by this
+amr.max_grid_size = 32  # at least 128 for GPUs
+amr.n_error_buf = 3  # minimum 3 cell buffer around tagged cells
+amr.grid_eff = 1.0
+
+do_reflux = 1
+do_subcycle = 0
+
+checkpoint_interval = -1
+plotfile_interval = -1
+
+tiny_profiler.enabled = 0
+
+cfl = 0.3
+max_timesteps = 20
+stop_time = 5.0e7  # s
--- a/src/QuokkaSimulation.hpp
+++ b/src/QuokkaSimulation.hpp
@@ -487,6 +487,10 @@ template <typename problem_t> void QuokkaSimulation<problem_t>::defineComponentN
 	if constexpr (Physics_Traits<problem_t>::is_hydro_enabled || Physics_Traits<problem_t>::is_radiation_enabled) {
 		std::vector<std::string> hydroNames = {"gasDensity", "x-GasMomentum", "y-GasMomentum", "z-GasMomentum", "gasEnergy", "gasInternalEnergy"};
 		componentNames_cc_.insert(componentNames_cc_.end(), hydroNames.begin(), hydroNames.end());
+	} else {
+		// Physics_Indices::nvarTotal_cc still allocates one cell-centred component when there is no
+		// hyperbolic state; name it so that plotfiles and conservation sums stay consistent
+		componentNames_cc_.emplace_back("placeholder");
 	}
 	// add passive scalar variables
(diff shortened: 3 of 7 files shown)


chongchonghe:
**Verdict:** 1 blocking — the new "no physics constrains the timestep" guard can never fire, so the failure it exists to catch is silent instead.
**Checked:** `3d` / `3d-hip` / `3d-debug` builds of `BinaryOrbitCICGravityOnly`, `BinaryOrbitCIC`, `GravRadParticle3D`, all SUCCESS. New test passes at max deviation `0.06658809445` vs tol `0.18`, CPU and GPU bit-identical. At matched `constant_dt=1.0e4` the gravity-only and hydro runs are bit-identical (`6.223533407e12`, `6.175065741e12`, deviation `0.04795792545`) — the core claim holds.

**Blocking**
1. **New assert is unreachable** — `simulation.hpp:1395`. `dt_tmp[level] = min(dt_tmp[level], 1.1 * dt_[level])` with `dt_.resize(nlevs_max, 1.e100)` (`simulation.hpp:772`) clips the unconstrained `DBL_MAX` to `1.1e100` before the check, so `dt_0 < max()` is always true; with particles at rest the run prints `Simulation ended at t = 50000000 (100%)` / `No cell updates performed!` and exits 0 instead of aborting (`stop_time=1e200` gives `dt = 1.1000000000e+100`). Test the pre-clip `dt_tmp`, or have `computeTimestepAtLevel` report that nothing set a limit.

**Non-blocking**
- **`timestep limited by HYDRO` printed with hydro disabled** — `simulation.hpp:1324`. `std::min_element` returns the first minimum and `hydro_dt` leads the vector, so all-`DBL_MAX` misattributes to hydro — exactly the case above.
- **The two tests don't run the same orbit** — `testBinaryOrbitCICGravityOnly.cpp:12`. Neither input sets `constant_dt`, so this one runs at particle CFL `dt≈7.56e4` (≈0.79 orbit in 20 steps) vs `BinaryOrbitCIC`'s hydro CFL `3.6e4→1.66e4` (≈0.16 orbit); same 0.18 tolerance is a different bar. Setting `constant_dt = 1.0e4` in both would pin the equivalence.
- **Passive scalars + no hyperbolic module** — `physics_info.hpp:68`. `nvarTotal_cc` stays 1 while `defineComponentNames` now emits `placeholder` + N scalar names; pre-existing, but this PR makes the configuration supported.

| Phase | Ran | Observed |
|---|---|---|
| Build | `3d`, `3d-hip`, `3d-debug`, 3 targets | all SUCCESS; no device-lambda errors under `hipcc` |
| Timestep | `amr.v=1 max_timesteps=3` | `estimated particle timestep: 7.560830e+04`, `timestep limited by PARTICLES`, no abort |
| New test | `quokka run -d 3d` / `-d 3d-hip` | `0.06658809445` both, tol `0.18`, `Test passed` |
| Equivalence | both problems, `constant_dt=1.0e4 max_timesteps=20` | identical separations, deviation `0.04795792545` |
| Regression | `BinaryOrbitCIC` CPU+GPU, `GravRadParticle3D` | `0.05118959015` (both), `Relative error within tolerance.` |
| Plotfile | `plotfile_interval=10` | header: 1 component, `placeholder` (shipped input has `plotfile_interval = -1`) |
| AMR | `amr.max_level=1 amr.blocking_factor=16` | 2 levels, deviation `0.008815540938`, passed |
| Restart | `checkpoint_interval=10` then `restartfile=chk0000010` | `0.06658809445`, matches uninterrupted run |
| Debug | `3d-debug` with bounds checking | `0.06658809445`, clean |
| Particles at rest | zero-velocity IC | no abort; `No cell updates performed!`, exit 0 |

</details>

chongchonghe:
Thanks — the blocking finding is correct and is now fixed in 7458101.

**Blocking 1 (fixed).** You were right that `dt_tmp[level] = min(dt_tmp[level], 1.1 * dt_[level])` clips `DBL_MAX` to `1.1e100` before the check, so the guard was dead code. Moved it into `computeTimestepAtLevel`, where the pre-clip per-level minimum is still visible, and allowed `constant_dt` / `max_dt` as legitimate substitutes for a physics constraint. Verified with your zero-velocity IC:

- at rest, no override → aborts with `No enabled physics module constrains the timestep! Set constant_dt or max_dt in the inputs file.`
- at rest, `max_dt=1.0e4` → runs (`t = 200000`), as does `constant_dt=1.0e4`
- normal IC → `estimated particle timestep: 7.560830e+04`, `timestep limited by PARTICLES`, unchanged

**Non-blocking 1 (fixed).** The `timestep limited by HYDRO` misattribution is gone: the attribution print is now gated on the same `timestep_is_constrained` flag, so the all-`DBL_MAX` case either aborts or (with an override set) prints nothing rather than blaming hydro.

**Non-blocking 3 — passive scalars with no hyperbolic module (fixed).** Added the `static_assert` to `Physics_Indices`, so the configuration this PR makes reachable is now a compile error rather than a silent drop of the scalar/dust components:

```cpp
static_assert(Physics_Traits<problem_t>::is_hydro_enabled || Physics_Traits<problem_t>::is_radiation_enabled ||
                  (Physics_Traits<problem_t>::numPassiveScalars == 0 && !Physics_Traits<problem_t>::is_dust_enabled),
              "Passive scalars and dust require hydro or radiation to be enabled.");
```

Checked both directions: adding `numPassiveScalars = 1` to `BinaryOrbitCICGravityOnly` fails with *"Passive scalars and dust require hydro or radiation to be enabled."*, and every in-tree problem still compiles — the only problems with both hydro and radiation off are `Advection` and `AdvectionSemiellipse`, and neither declares scalars or dust.

**Non-blocking 2 — matched `constant_dt`: deliberately not changed.** This is a decision made in awareness of the trade-off, not an oversight. `constant_dt > 0` short-circuits `computeTimestep` (`dt_0 = constantDt_`), so pinning the shipped test would stop it exercising the particle-CFL path that this PR fixes — the default run is the only in-tree coverage of that path. The equivalence itself has been verified twice by hand, by both of us, at matched `constant_dt`. Note also that no arrangement of these tests mechanically *compares* the two binaries; each only checks its own tolerance.

Regression after the fixes: `BinaryOrbitCICGravityOnly`, `BinaryOrbitCIC`, `GravRadParticle3D`, `ParticleDeposition` all pass, and a full build of every problem in the tree is clean: `1d` 52/52 SUCCESS and `3d` 107/107 SUCCESS, 0 FAIL in both.

markkrumholz (approved):

PR #2189, opened 2026-08-30
Fix convergence tolerance for weak nonzero gas-dust coupling

### LINKED ISSUE #2188: PR #2126 causes a single-group radiation+dust convergence regression for weak but non-zero gas-dust coupling
**Describe the bug**

PR #2126 introduced a convergence regression in the single-group radiation+dust solver for weak but non-zero gas-dust coupling.

Before PR #2126, the same checkpoint completes step 3005 successfully with a maximum of approximately 4 Newton iterations. Starting from the merge commit of PR #2126, the dust-temperature solve requires approximately 46–47 iterations and eventually aborts with:

‘Newton-Raphson iteration for dust temperature failed to converge or dust temperature is negative!’

The regression is caused by the residual scaling introduced in ‘BackwardEulerOneVariable’.

Before PR #2126, the convergence test used:

‘abs(f(Td)) < 1e-8 × compare’

where:

‘compare ≈ D × Tgas’

and:

‘D = coeff_n × sqrt(Tgas)’

PR #2126 changed the residual scale to:

‘scale = max(compare, abs(f0))’

‘abs(f(Td)) < 1e-8 × scale’

In weakly coupled cells, ‘D’ is approximately ‘1e-24’. Therefore, ‘compare’ is very small, while ‘abs(f0)’ can be much larger. Using ‘max(compare, abs(f0))’ substantially relaxes the convergence tolerance relative to the physical gas-dust coupling scale.

The single-group outer coupled solver reconstructs the dust temperature using:

‘Td = Tgas - (R - work) / D’

Therefore, the propagated temperature error is:

‘delta_Td = -delta_R / D’

Because ‘D’ is very small, a small residual energy error from the inner dust-temperature solve is amplified by ‘1 / D’. The outer Newton iteration can then cross ‘Td = 0’ and enter the non-physical negative-temperature branch.

**To Reproduce**

1. Compile a single-group radiation-hydrodynamics problem with radiation+dust thermal coupling enabled, such as the modified ‘Turbulence’ problem.

2. Use weak but non-zero gas-dust coupling with the following relevant configuration:

‘radiation.dust_gas_interaction_coeff = 2.5e-34’

‘c_hat_over_c = 1.0e-4’

Use a cooling table without dust cooling so that the dust thermal evolution is handled by the radiation+dust solver.

3. Restart from the checkpoint immediately before step 3005 and run with 64 MPI processes.

4. Compile and run commit ‘8967c156’, the direct parent of PR #2126.

The result is:

‘step 3005 completed successfully’

‘mean Newton iterations ≈ 1.193’

‘max Newton iterations = 4’

5. Compile and run commit ‘5ca52427’, the merge commit of PR #2126, using the same checkpoint and input parameters.

The result is:

‘max Newton iterations ≈ 46–47’

‘Newton-Raphson iteration for dust temperature failed to converge or dust temperature is negative!’

‘MPI_ABORT’

6. Compile and run the latest tested official commit ‘0dd16b9a’ using the same checkpoint and input parameters.

The result is:

‘mean Newton iterations ≈ 1.258’

‘max Newton iterations = 46’

‘Newton-Raphson iteration for dust temperature failed to converge or dust temperature is negative!’

7. On commit ‘0dd16b9a’, change only:

‘scale = max(compare, abs(f0))’

back to:

‘scale = compare’

Leave the bracketed Newton/bisection implementation and all other code unchanged.

The result is:

‘step 3005 completed successfully’

‘max Newton iterations = 4’

This single-line comparison identifies the residual scaling change as the direct trigger of the regression.

**Additional context**

The regression boundary is:

‘6ff08310’, self-gravity disabled:

‘step 3005 completed successfully’

‘mean Newton iterations ≈ 1.193’

‘max Newton iterations = 4’

‘8967c156’, direct parent of PR #2126, self-gravity disabled:

‘step 3005 completed successfully’

‘mean Newton iterations ≈ 1.193’

‘max Newton iterations = 4’

‘5ca52427’, merge commit of PR #2126:

‘step 3005 failed’

‘max Newton iterations ≈ 46–47’

‘dust temperature became negative’

‘0dd16b9a’, latest tested official version, self-gravity disabled:

‘step 3005 failed’

‘mean Newton iterations ≈ 1.258’

‘max Newton iterations = 46’

‘0dd16b9a’, latest tested official version, self-gravity enabled:

‘step 3005 failed with the same timestep, radiation substep count, Newton statistics, and error location’

The self-gravity-enabled version only performs the additional MLMG Poisson solve. Therefore, the failure is not caused by an interaction between self-gravity and radiation+dust.

Two additional isolated changes were tested:

1. Removing the early-exit condition:

‘abs(dx) < 1e-6 × abs(Td)’

Result:

‘still failed’

‘max Newton iterations = 46’

2. Changing ‘enable_dE_constrain’ from ‘false’ back to ‘true’

Result:

‘still failed’

‘max Newton iterations = 46’

Therefore, neither condition independently causes the regression.

Changing the outer coupled variable to:

‘y = log(Td)’

‘Td = exp(y)’

also fixes the failing checkpoint:

‘step 3005 completed successfully’

‘the simulation entered step 3006’

‘mean Newton iterations ≈ 1.156’

‘max Newton iterations = 5’

A possible fix is to distinguish the exactly decoupled case ‘D = 0’ from the weakly but non-zero case ‘0 < D << 1’.

For ‘D > 0’, the convergence tolerance should remain tied to the physical coupling scale:

‘scale = compare’

For ‘D = 0’, the solver could enter the decoupled branch directly or use ‘abs(f0)’ only as a fallback scale.

Using ‘scale = max(compare, abs(f0))’ for all cells treats weakly coupled cells like decoupled cells and makes the residual tolerance too loose relative to ‘D × Tgas’.

A regression test should be added for the single-group radiation+dust solver with weak but non-zero gas-dust coupling.

YaoguangPei:
### Description

This PR fixes the weak but non-zero gas-dust coupling convergence regression in `BackwardEulerOneVariable`. It preserves the caller-provided physical residual scale whenever `compare > 0`, uses `abs(f0)` only for the exactly decoupled case, and adds a regression test to `RadDust` that distinguishes the corrected behavior from the overly relaxed tolerance introduced by PR #2126.

### Related issues

Fixes #2188

### Checklist

- [ ] *(For quokka-astro org members)* I have triggered GPU tests for changed problems with `/azp run quick` and `/azp run rocm-quick`, or `/azp run` if this PR changes base modules affecting multiple problems. Not applicable because I am not a quokka-astro organization member.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved radiation solver convergence for cases with weak but non-zero physical scales.
  * Ensured convergence checks consistently use the configured residual scale.
  * Improved reliability when solving decoupled or near-zero-scale conditions.

* **Tests**
  * Added GPU coverage validating solver convergence and residual accuracy under weak-scale conditions.
  * Test failures now correctly report a failing process status.

### DIFF
--- a/src/radiation/radiation_system.hpp
+++ b/src/radiation/radiation_system.hpp
@@ -1637,18 +1637,10 @@ AMREX_GPU_DEVICE auto RadSystem<problem_t>::BackwardEulerOneVariable(RHSFunction
 	// The caller supplies the physical scale against which its residual must be measured. Preserve every
 	// positive scale, however small: weak but non-zero gas-dust coupling relies on a correspondingly tight
 	// inner tolerance because the outer solve reconstructs T_d with a factor inversely proportional to that
-	// coupling. Scaling instead by a larger initial residual can therefore amplify the accepted inner error
-	// enough to drive the outer temperature negative.
-	//
-	// Exactly decoupled dust has a zero physical scale. Only in that case, fall back to the initial residual
-	// so the criterion requests a relative residual reduction rather than an impossible strict zero.
+	// coupling. Clamp a non-positive scale to zero, which accepts only an exact-zero initial residual.
 	const double f0 = rhs(x0);
-	AMREX_ASSERT(compare >= 0.0);
-	if (f0 == 0.0) {
-		return x0;
-	}
-	const double scale = (compare > 0.0) ? compare : std::abs(f0);
-	if (std::abs(f0) < rel_tol * scale) {
+	const double scale = std::max(0.0, compare);
+	if (std::abs(f0) <= rel_tol * scale) {
 		return x0;
 	}
 
--- a/src/radiation/radiation_system.hpp
+++ b/src/radiation/radiation_system.hpp
@@ -1634,13 +1634,20 @@ AMREX_GPU_DEVICE auto RadSystem<problem_t>::BackwardEulerOneVariable(RHSFunction
 	const double rel_change_tol = 1.0e-6;
 	const int max_iter_td = 100;
 
-	// Tolerance scale. The caller passes the physical scale its residual should be measured against, but for
-	// the dust temperature that scale is the gas-dust collisional term, which vanishes identically when the
-	// coupling coefficient is set to zero. The tolerance would then be zero and could never be met. Fall
-	// back to the size of the initial residual so the criterion degrades to a relative reduction instead of
-	// something unsatisfiable.
+	// The caller supplies the physical scale against which its residual must be measured. Preserve every
+	// positive scale, however small: weak but non-zero gas-dust coupling relies on a correspondingly tight
+	// inner tolerance because the outer solve reconstructs T_d with a factor inversely proportional to that
+	// coupling. Scaling instead by a larger initial residual can therefore amplify the accepted inner error
+	// enough to drive the outer temperature negative.
+	//
+	// Exactly decoupled dust has a zero physical scale. Only in that case, fall back to the initial residual
+	// so the criterion requests a relative residual reduction rather than an impossible strict zero.
 	const double f0 = rhs(x0);
-	const double scale = std::max(compare, std::abs(f0));
+	AMREX_ASSERT(compare >= 0.0);
+	if (f0 == 0.0) {
+		return x0;
+	}
+	const double scale = (compare > 0.0) ? compare : std::abs(f0);
 	if (std::abs(f0) < rel_tol * scale) {
 		return x0;
 	}
--- a/src/radiation/radiation_system.hpp
+++ b/src/radiation/radiation_system.hpp
@@ -1637,9 +1637,10 @@ AMREX_GPU_DEVICE auto RadSystem<problem_t>::BackwardEulerOneVariable(RHSFunction
 	// The caller supplies the physical scale against which its residual must be measured. Preserve every
 	// positive scale, however small: weak but non-zero gas-dust coupling relies on a correspondingly tight
 	// inner tolerance because the outer solve reconstructs T_d with a factor inversely proportional to that
-	// coupling. Clamp a non-positive scale to zero, which accepts only an exact-zero initial residual.
+	// coupling. When no positive physical scale exists, fall back to the initial residual so the solver
+	// requests a relative residual reduction instead of an impossible strict zero.
 	const double f0 = rhs(x0);
-	const double scale = std::max(0.0, compare);
+	const double scale = (compare > 0.0) ? compare : std::abs(f0);
 	if (std::abs(f0) <= rel_tol * scale) {
 		return x0;
 	}
--- a/src/problems/RadDust/testRadDust.cpp
+++ b/src/problems/RadDust/testRadDust.cpp
@@ -5,6 +5,7 @@
 #ifdef HAVE_PYTHON
 #include "util/matplotlibcpp.h"
 #endif
+#include "AMReX_GpuAsyncArray.H"
 #include "AMReX_Print.H"
 #include "QuokkaSimulation.hpp"
 #include "fundamental_constants.H"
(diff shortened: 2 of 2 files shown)


src/radiation/radiation_system.hpp
```diff
@@ -1634,13 +1634,20 @@ AMREX_GPU_DEVICE auto RadSystem<problem_t>::BackwardEulerOneVariable(RHSFunction
...
+	//
+	// Exactly decoupled dust has a zero physical scale. Only in that case, fall back to the initial residual
+	// so the criterion requests a relative residual reduction rather than an impossible strict zero.
 	const double f0 = rhs(x0);
-	const double scale = std::max(compare, std::abs(f0));
+	AMREX_ASSERT(compare >= 0.0);
+	if (f0 == 0.0) {
+		return x0;
+	}
+	const double scale = (compare > 0.0) ? compare : std::abs(f0);
```
chongchonghe: should this be `const double scale = std::max(compare, std::abs(f0));` , the original code?

chongchonghe (changes requested):
see my comment

src/radiation/radiation_system.hpp
YaoguangPei: This change is intentional. For weak but non-zero dust coupling, compare is the physical residual scale and can be much smaller than abs(f0). Using std::max(compare, std::abs(f0)) would loosen the tolerance and reproduce the premature-convergence bug covered by the new regression test. We only fall back to abs(f0) when compare == 0, while f0 == 0 is handled by the early return.

src/radiation/radiation_system.hpp
```diff
@@ -1634,13 +1634,20 @@ AMREX_GPU_DEVICE auto RadSystem<problem_t>::BackwardEulerOneVariable(RHSFunction
...
+	// positive scale, however small: weak but non-zero gas-dust coupling relies on a correspondingly tight
+	// inner tolerance because the outer solve reconstructs T_d with a factor inversely proportional to that
+	// coupling. Scaling instead by a larger initial residual can therefore amplify the accepted inner error
+	// enough to drive the outer temperature negative.
+	//
+	// Exactly decoupled dust has a zero physical scale. Only in that case, fall back to the initial residual
+	// so the criterion requests a relative residual reduction rather than an impossible strict zero.
 	const double f0 = rhs(x0);
-	const double scale = std::max(compare, std::abs(f0));
+	AMREX_ASSERT(compare >= 0.0);
```
chongchonghe: scale = std::max(0.0, compare)
if (std::abs(f0) <= rel_tol * scale) {
		return x0;
}

src/radiation/radiation_system.hpp
chongchonghe: OK. It's the problem of my original code, not yours. This code snippet doesn't make sense:

```cpp
	const double scale = std::max(compare, std::abs(f0));
	if (std::abs(f0) < rel_tol * scale) {
		return x0;
	}
```

because `std::abs(f0) < 1e-8 * std::abs(f0)` will never be met.

YaoguangPei: Sorry for the late reply and thanks for the confirm. I have updated this PR, and waiting for the merge if there have nothing wrong

chongchonghe:
**Verdict:** Ship it — no blocking issues. The new test fails on base and passes here, and all 9 adjacent radiation-dust tests keep bit-identical error norms.
**Checked:** presets `1d` and `1d-cuda` (sm_80, compile-only — no device here). RED on `development`: `solution = 0.01171875, residual = 8.86e-09` → RadDust FAIL. GREEN on PR: `solution = 0.01, residual = 1.66e-24` → SUCCESS. `pre-commit` clean.

**Non-blocking**
1. **Residual clause of the assertion has ±1 ulp of headroom** — `testRadDust.cpp:118`. `residual < 1e-23` holds at 0 and ±1 ulp from `1e-2` (`1.65e-24`, `6.6e-24`, `8.3e-24`) but fails at ±2 ulp (`1.16e-23`, `1.32e-23`); it also moves with codegen (`6.23e-25` exact, `1.65e-24` as `x*x*x*x`, `0` as `(x*x)*(x*x)`). It adds no discrimination — base returns `0.01171875`, which `|solution - 1e-2| < 1e-12` already rejects by 1.7e9 — so drop it or widen to `1e-20`.
2. **No GPU runtime has seen this test** — every check on this PR is a compile job. Worth `/azp run quick` and `/azp run rocm-quick` before merge.
3. **Synthetic quartic, not the dust path** — `testRadDust.cpp:99`. The `Td = Tgas - R/(N_d·sqrt(Tgas))` amplification from #2188 stays uncovered; no input exercises a tiny non-zero `dust_gas_interaction_coeff`.
4. **`status = 1` only on the IO rank** — `testRadDust.cpp:192`. The test runs on all ranks; deterministic, so harmless today.

| Phase | Ran | Observed |
|---|---|---|
| RED | `development` + PR's `testRadDust.cpp` only, solver at base | `solution = 0.01171875`, `residual = 8.859e-09` → **FAIL**. Loop breaks on its first midpoint because `scale = max(1e-16, |f0|) = 1`. |
| GREEN | + PR's `radiation_system.hpp` | `solution = 0.01`, `residual = 1.664e-24` → **SUCCESS**; `Relative L1 error norm = 0.0003586751597` (identical to base) |
| GPU | `1d-cuda`, `-DCMAKE_CUDA_ARCHITECTURES=80` | 160/160 objects, link clean |
| EDGE | `ctest -R` over DTypeFront1D, RadDust, RadDustMG, RadLineCooling, RadLineCoolingMG, RadLineCoolingMG_coupled, RadMarshakDust, RadMarshakDustPE | 8/8 Passed; `RadiationMarshakDustPE-decoupled` Passed separately |
| EDGE | base vs PR, 9 binaries, error norms + Newton counts | All norms bit-identical. Only change: `RadMarshakDust` mean iterations `20.5035 → 21.005`, max `44 → 44` |

`dust_gas_interaction_coeff` covered: `0.0` (DTypeFront1D, the `compare == 0` fallback #2126 added), `1e-20`, `1e-2`, `1e6`, `1e20`.

</details>

PR #2191, opened 2026-08-31
Spitzer thermal Conduction

aditivijayan:
### Description
This PR expands the current thermal conduction module to include Spitzer thermal conduction wherein the conductivity depends on T^2.5. 

The major modifications are-

1. `src/simulation.hpp `— adds `conductionType_` ("constant" or "spitzer") and splits the conduction timestep estimate into two paths:

    - `constant`: the original single-formula estimate (using the domain-min density), now applied AMREX_SPACEDIM-independently (dropped the old / AMREX_SPACEDIM factor).
    - `spitzer`: a new per-cell amrex::ParReduce that evaluates the actual local Spitzer conductivity (K0·T^2.5) at every cell to find the true minimum timestep and its location — needed because the nonlinear T-dependence means the timestep-limiting cell isn't necessarily where density is minimum.

2. ` src/conduction/ElectronConduction.hpp` — replaced a plain 2nd-order central-difference conduction flux with a full PPM/PLM face-reconstruction pipeline (ReconstructPrimVar) that mirrors the hydro solver's own reconstruction (same reconstruction_order, plm_limiter, and now the same ng_reconstruct ghost width). Also added spitzer_scaling to switch κ(T) between K0 (constant) and K0·T^2.5 (Spitzer) at the face-conductivity evaluation. 

3. `src/QuokkaSimulation.hpp` — wires conductionType_ through from input-file parsing (conduction.conduction_type, validated against {constant, spitzer}), and threads a properly-computed ng_reconstruct (matching MinimumHydroRiemannGhost(...) + 1, the same formula the hydro flux path uses) into ElectronConductionParams — this is the ghost-cell-width fix from earlier in this session, for the NaN crash at higher resolution.

4. `src/problems/ThermalConduction/testThermalConduction.cpp` — the test itself now validates both conduction types from one binary. Spitzer conduction is validated by estimating slopes from a full resolution (32/64/128) convergence test and compared against the [Pattle self-similar solution](https://ui.adsabs.harvard.edu/abs/1959QJMAM..12..407P/abstract). The slope is compared with 1.0 since the Pattle solution contains a sharp slope and the error is dominated by the error at the edges. 

## Evolution for nx=128

image

## Solution from 3D runs.

image

## Comparing with the Analytical solution (dashed line)

image

## Relative Error 
image

## Convergence Test

image

### Related issues

### Checklist
- [x ] I have added tests for any new physics that this PR adds to the code.

## Summary by CodeRabbit

* **New Features**
  * Added selectable constant and Spitzer models for electron thermal conduction.
  * Added temperature-dependent conductivity, timestep calculations, and temperature-floor protection for Spitzer conduction.
  * Improved interface reconstruction with configurable order, slope limiting, and ghost-cell handling.
  * Updated thermal conduction inputs to use the Spitzer model and support output suppression control.

* **Bug Fixes**
  * Improved conduction flux calculations for more consistent results.

* **Tests**
  * Expanded tests with Spitzer convergence studies and constant-conduction AMR validation.


aditivijayan:
@markkrumholz - I have started this PR to receive comments on the code changes. Happy to do any other tests that you recommend.

aditivijayan:
The test is failing because of ParticleSink problem.

BenWibking:
This suggests it is dividing by zero somewhere. Maybe you can re-run this test in a debugger?
```
Angular momentum Galilean invariance check:
Particle 0 (Phase2) angular momentum: Lx=-1.4740e+28 Ly=-1.7777e+43 Lz=-nan
Abort(8) on node 0 (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 8) - process 0
```

aditivijayan:
> This suggests it is dividing by zero somewhere. Maybe you can re-run this test in a debugger?
> 
> ```
> Angular momentum Galilean invariance check:
> Particle 0 (Phase2) angular momentum: Lx=-1.4740e+28 Ly=-1.7777e+43 Lz=-nan
> Erroneous arithmetic operation
> Abort(8) on node 0 (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 8) - process 0
> ```

Sure can!

BenWibking:
Actually, I suspect it is trying to take `pow(T, 2.5)` where T is somehow negative (maybe in ghost cells?).

aditivijayan:
> Actually, I suspect it is trying to take `pow(T, 2.5)` where T is somehow negative (maybe in ghost cells?).

No problem except `ThermalConduction` should take T^2.5.

aditivijayan:
This PR is ready for review.

markkrumholz (changes requested):
There are a bunch of seemingly-unrelated changes under extern/agora_data and extern/cooling. Can you remove those from the PR?

BenWibking (changes requested):
Codex identified these issues, which I think are real:

```
  - [P1] Derive the Spitzer timestep from physical internal energy — src/simulation.hpp:1304-1305

    In high-Mach or MHD cells, the auxiliary internalEnergy_index can differ from the internal energy derived from total energy, while
    ComputeExplicit uses ComputeInternalEnergy. If the auxiliary value is colder, the reduction underestimates (T^{2.5}), chooses an
    excessively large timestep, and can destabilize the explicit conduction update. Use the same total-energy-derived thermodynamic state as
    the flux calculation.

  - [P2] Populate composition before EOS calls — src/conduction/ElectronConduction.hpp:119-120

    When numMassScalars > 0, this optional is now always empty even though the previous implementation populated it with
    RadSystem::ComputeMassScalars. Composition-dependent EOS backends such as EOSMicrophysics then initialize species abundances to invalid
    sentinel values, producing incorrect or NaN temperatures and conduction fluxes. Retain the cell composition and provide a consistent
    face composition for the saturation calculation.

  - [P2] Do not add a background to the Pattle reference — src/problems/ThermalConduction/testThermalConduction.cpp:105-109

    The Pattle profile is a self-similar solution for zero-background nonlinear diffusivity (\kappa(T)=\kappa_0T^{2.5}). Adding Tfloor
    changes the coefficient to depend on Tfloor + Tpulse, causing diffusion outside the nominal compact front and changing its evolution.
    The resulting profile is therefore not an exact reference for the equation being solved, so the measured error need not converge to
    zero.

  - [P2] Require a negative convergence slope — src/problems/ThermalConduction/testThermalConduction.cpp:327-329

    A convergent error curve has a negative log(error) versus log(Nx) slope, but taking the absolute value also accepts a solution whose
    error grows as Nx^1 or faster. Require the slope itself to be sufficiently negative so a divergent Spitzer implementation cannot pass
    this regression test.
```

aditivijayan:
> There are a bunch of seemingly-unrelated changes under extern/agora_data and extern/cooling. Can you remove those from the PR?

Done.

aditivijayan:
> Codex identified these issues, which I think are real:
> 
> ```
>   - [P1] Derive the Spitzer timestep from physical internal energy — src/simulation.hpp:1304-1305
> 
>     In high-Mach or MHD cells, the auxiliary internalEnergy_index can differ from the internal energy derived from total energy, while
>     ComputeExplicit uses ComputeInternalEnergy. If the auxiliary value is colder, the reduction underestimates (T^{2.5}), chooses an
>     excessively large timestep, and can destabilize the explicit conduction update. Use the same total-energy-derived thermodynamic state as
>     the flux calculation.
> 
>   - [P2] Populate composition before EOS calls — src/conduction/ElectronConduction.hpp:119-120
> 
>     When numMassScalars > 0, this optional is now always empty even though the previous implementation populated it with
>     RadSystem::ComputeMassScalars. Composition-dependent EOS backends such as EOSMicrophysics then initialize species abundances to invalid
>     sentinel values, producing incorrect or NaN temperatures and conduction fluxes. Retain the cell composition and provide a consistent
>     face composition for the saturation calculation.
> 
>   - [P2] Do not add a background to the Pattle reference — src/problems/ThermalConduction/testThermalConduction.cpp:105-109
> 
>     The Pattle profile is a self-similar solution for zero-background nonlinear diffusivity (\kappa(T)=\kappa_0T^{2.5}). Adding Tfloor
>     changes the coefficient to depend on Tfloor + Tpulse, causing diffusion outside the nominal compact front and changing its evolution.
>     The resulting profile is therefore not an exact reference for the equation being solved, so the measured error need not converge to
>     zero.
> 
>   - [P2] Require a negative convergence slope — src/problems/ThermalConduction/testThermalConduction.cpp:327-329
> 
>     A convergent error curve has a negative log(error) versus log(Nx) slope, but taking the absolute value also accepts a solution whose
>     error grows as Nx^1 or faster. Require the slope itself to be sufficiently negative so a divergent Spitzer implementation cannot pass
>     this regression test.
> ```

These comments have been addressed.

markkrumholz (dismissed):

src/conduction/ElectronConduction.hpp
```diff
@@ -75,85 +97,104 @@ template <typename problem_t> class ElectronConduction
...
+			amrex::GpuArray<amrex::Real, nmscalars_> massArray_face{};
+			for (int n = 0; n < nmscalars_; ++n) {
+				massArray_face[n] = 0.5 * (massScalars_L[n] + massScalars_R[n]);
+			}
+			quokka::optional<amrex::GpuArray<amrex::Real, nmscalars_>> massScalars = massArray_face;
+			const amrex::Real Eint_face = ::quokka::EOS<problem_t>::ComputeEintFromTgas(rho_face, T_face, massScalars);
+			const amrex::Real Pgas_face = ::quokka::EOS<problem_t>::ComputePressure(rho_face, Eint_face, massScalars);
+			const amrex::Real cs_face = ::quokka::EOS<problem_t>::ComputeSoundSpeed(rho_face, Pgas_face, massScalars);
+
+			kappa_face = spitzer_scaling ? (kappa0 * std::pow(T_face, 2.5)) : kappa0;
```
BenWibking, line 146: I think it would be greatly preferable if the conductivity can be defined in a problem-specific function analogous to the way we treat opacities.

src/problems/ThermalConduction/testThermalConduction.cpp
```diff
@@ -212,28 +264,86 @@ auto problem_main() -> int
...
+		}
+		double const mean_x = sum_x / n;
+		double const mean_y = sum_y / n;
+		double const slope = (sum_xy - n * mean_x * mean_y) / (sum_xx - n * mean_x * mean_x);
+		double const intercept = mean_y - slope * mean_x;
+		amrex::Print() << std::format("\nBest-fit line: log(error) = {:.4f} * log(Nx) + {:.4f}\n", slope, intercept);
+
+		// error ~ Nx^slope for Pattle IC
+		amrex::Print() << std::format("Spitzer conduction convergence: slope = {:.4f} (-1 expected, converging faster is fine)\n", slope);
+		passed = slope <= -0.9;
```
BenWibking, line 327: Do you understand why it converges at order close to 1, rather than close to 2? I don't know the answer to this.

BenWibking (changes requested):
I have two outstanding concerns: one is physics, and the other is code architecture-related.

src/problems/ThermalConduction/testThermalConduction.cpp
markkrumholz: I think I do: this particular analytic solution (and it's the only one I know of for non-linear conduction) has the property that the temperature goes to 0 exactly at finite radius. The error is dominated by the error in the position of the expanding front of the region where T > 0, and of course you can't find the position of that front to better than one cell. It's analogous to a shock problem in hydrodynamics: the exact answer (assuming you're not resolving the viscous shock width) is a discontinuity, and you can't get that to better than a cell. I'd bet that if one defined an error only on the region inside where the solution is finite, it would converge at second order.

aditivijayan: I have actually done the experiments by just looking at "inside" points and restricting the errors within r1/2. Here's what I found-

1. For lower res- say, 32 or 64 the "inside" is very small and consequently it is not well resolved. You can see it in the error plots- the horns from the edge of the solution are quite wide compared to 128. Excluding the horns meant getting error from a handful of points.

2. Once we reach 128 the inside region is well resolved and any additional resolution does not substantially improve the accuracy of the solution.

markkrumholz: Wait, so the solution doesn't get any better at all past 128, i.e., convergence rate = 0? If that's what you mean, that is a bit worrying. In that case I would prefer to see a demonstration that we can get convergence for a problem without discontinuities. The way I'd suggest testing that is to use a Gaussian or sinusoidal initial condition so there are no discontinuities, run at very high resolution (say 4096 or 8192 cells), and use that as the "truth" against which you measure convergence.

aditivijayan: Well, the Pattle solution (at least in the paper you shared) is valid only for an initial delta function. Are there other references for a Gaussian or sinusoidal ICs?

BenWibking: I would also like to see the error convergence for the inside solution. I agree that should be second order. You can always initialize the problem with the Pattle/Barenblatt solution at an earlier time, and then evolve it to a later time. Then it should be possible to make it well resolved on coarser grids for any length of simulated time.

BenWibking: As you've set it up, I think the fix is just to make the domain smaller so the front hits the domain edge just after t=1.0.

BenWibking: This article (section 4.1) shows that Mark's hypothesis is correct, that is, you indeed can get second-order convergence if you measure the error away from the front: https://doi.org/10.1007/s10915-024-02642-x

aditivijayan: Okay, I can run the test and plot the results.

> I would also like to see the error convergence for the inside solution. I agree that should be second order. You can always initialize the problem with the Pattle/Barenblatt solution at an earlier time, and then evolve it to a later time. Then it should be possible to make it well resolved on coarser grids for any length of simulated time.

That's what I am doing right now. The test starts at 1=0.5tcond and evolves it to t=tcond.

markkrumholz: @aditivijayan and I discussed this in person, and she is going to try the experiment of running a continuous problem and seeing if the solution converges to a very high-res reference solution at second order or not. That will tell us if the convergence at less than second order if related to the Pattle solution in particular, or if it's general to the method. We can then go from there.

aditivijayan: Well, okay I have done this experiment and the results are telling. The convergence rate is closer to 2 when compared with an ultra high res solution (Nx=8192). 

image

The pairwise convergence is interesting-

  64 -> 128 : p = 2.092
   128 -> 256 : p = 0.565
   256 -> 512 : p = 2.778
   512 -> 1024: p = 0.690
  1024 -> 2048: p = 2.094
  
The rate is close to or better than  2 for some pairs and nearly flat others. I suspect perhaps where exactly the front lies makes substantial difference.

The convergence rate if we compared with the Pattle solution- 

image



Pairwise convergence rate p (error ~ Nx^-p):
    64 -> 128 : p = 1.744
   128 -> 256 : p = 0.977
   256 -> 512 : p = 0.984
   512 -> 1024: p = -0.349
  1024 -> 2048: p = 0.115
  
Note that the convergence rate is worse than what we were earlier because for higher resolutions (nx=256, 512, etc) the error remains flat. 
 
Also note this test is not identical to the previous test since I ran for a tad bit longer.

markkrumholz: OK, well the -1.6 is encouraging, though not quite the -2 one might have hoped for -- but do you have any idea why the convergence rate of the Pattle solution basically goes to zero at high resolution? That would seem to imply that the front position must be off by multiple cells at the highest resolutions. Is it?

aditivijayan: Yes, it is off by multiple cells.
The stars are the analytical solution and circles are numerical. 

image

Could it be related to the discreteness of the grid- similar to what we discussed for the Gaussian case for which the numerical solution matched the error function rather than the Gaussian?

markkrumholz: You could certainly eliminate that as a source of error by analytically integrating the exact solution over each cell in your initialization. But I wonder if this is more related to the floor temperature you need to adopt. You might try changing the floor and seeing if that makes a difference.

BenWibking: GPT-6 suggests the floor is the cause of the convergence issue:
```
    The Spitzer reference retains Efloor outside the expanding front, but this constant corresponds to roughly 4 million K for the configured
    EOS, not a representability-scale floor. Consequently the reference's integrated energy is the constant Pattle energy plus Efloor *
    (domain_length - 2*r1), which decreases as the front expands, unlike the conservative numerical evolution. This introduces a resolution-
    independent reference error and eventually spoils the convergence check. Use a genuinely negligible positive floor consistently with the
    simulation's temperature floor, or a reference that solves the finite-background problem.
```

BenWibking:
There are still two unaddressed issues from the previous AI review:
```
  - [P1] Pass cell composition to the Spitzer timestep EOS call — /Users/benwibking/amrex_codes/quokka/src/simulation.hpp:1305-1305
    With the chemistry EOS enabled, omitting mass scalars leaves every species abundance at the -1.0 sentinel initialized in
    EOSChem::ComputeTgasFromEint. The timestep therefore uses a temperature unrelated to the actual composition, whereas the conduction flux
    supplies the composition correctly. Since conductivity scales as T^2.5, this can substantially overestimate the stable timestep. Extract and
    pass the cell's mass scalars here.

  - [P2] Preserve physical internal energy when constructing temperature — /Users/benwibking/amrex_codes/quokka/src/conduction/
    ElectronConduction.hpp:106-109
    When hydro.use_dual_energy = 0, ApplyHydroStateFixup does not synchronize the auxiliary internal-energy component with total energy. This is
    also the required configuration for physical viscosity and resistivity, whose dissipative heating is not added to the auxiliary equation.
    Replacing ComputeInternalEnergy with this component therefore makes conduction use stale temperatures and miss that heating. Derive
    temperature from total minus kinetic and magnetic energy, and apply the same correction to the new Spitzer timestep calculation.
```

src/problems/ThermalConduction/testThermalConduction.cpp
BenWibking: Sorry, I just saw Mark's comment.

markkrumholz: Just closing the loop: @aditivijayan checked and the floor was indeed the problem. I think we will be able to declare victory on the convergence issue soon.

aditivijayan: With the floor now a third of the previous floor the convergence rate has definitely improved. 


image

The location of the front is off by one cell- 

image

BenWibking: Is there a reason not to use a *much* lower floor? I ran it with a floor of 3 K, and it seemed to work fine. The error power law (measured away from the front) also went to -2.

aditivijayan: Okay, I have done the experiment as suggeste by @markkrumholz. Instead of using a delta function as the IC, I used a gaussian profile and then evolved it using the Spitzer conduction. The profile looks like this-

image


I then compared these profiles with a high res run (at nx=4096). The errors look like this-


image


As you can see the errors are concentrated at the edge of the profile and the convergence rate is close to unity-

image


The convergence rate improves substantially if I estimate the error within 0.1 pc, very much in the interior. 

image

I am currently running nx=8192 just to check if that would improve the convergence rate.

aditivijayan: > Is there a reason not to use a _much_ lower floor? I ran it with a floor of 3 K, and it seemed to work fine. The error power law (measured away from the front) also went to -2.

Well, the conductivity depends on the mean molecular weight. When I had started adding conduction I realised that unless the mean molecular weight of the floor is same as that of the peak (for the Gaussian problem and constant conductivity) the match between the analytical solution and the numerical solution was off by a factor of mu. So I chose values of the floor so that the mu dependence gets cancelled out. 

@BenWibking - is there a way to fix mu?

BenWibking: I don't understand why it's off by a factor of mu. The conduction operator should diffuse the temperature itself, so as long as you are comparing the temperature at the end, I don't see how it matters, unless I'm missing something?

BenWibking: More precisely, if you have the correct factors of the specific heat $c_v$ on the right hand side, and you are comparing the temperature against the reference solution, then I don't it matters what c_v is, as long as it's computed self-consistently everywhere.

markkrumholz: I'm super-confused: in the profile you posted, it looks like there is no evolution whatsoever in the profile outside ~0.35 pc. The initial and final profiles are completely identical? Why?

aditivijayan: > I'm super-confused: in the profile you posted, it looks like there is no evolution whatsoever in the profile outside ~0.35 pc. The initial and final profiles are completely identical? Why?

Because there is a floor.

markkrumholz: @BenWibking, just to explain the background, since you weren't part of the in-person conversation: I suggested to @aditivijayan that to test whether the lower convergence order we were getting in testing again the Pattle solution was a result of the discontinuity in the solution, she should try a simulation starting from a smooth initial condition with no discontinuities. We don't have an analytic solution in that case, of course, but we could still check the convergence order by running at very high resolution and using that as our reference solution against which to check convergence. If we got second-order in that test, it would be strong evidence that the method itself converges at second order when applied to a smooth system.

markkrumholz: > > I'm super-confused: in the profile you posted, it looks like there is no evolution whatsoever in the profile outside ~0.35 pc. The initial and final profiles are completely identical? Why?
> 
> Because there is a floor.

OK, but why do we need a floor in this problem? The temperature is finite everywhere, not zero, so why not just turn off the floor? The floor seems like it's creating a discontinuity, which is exactly what we were trying to avoid.

aditivijayan: > > > I'm super-confused: in the profile you posted, it looks like there is no evolution whatsoever in the profile outside ~0.35 pc. The initial and final profiles are completely identical? Why?
> > 
> > 
> > Because there is a floor.
> 
> OK, but why do we need a floor in this problem? The temperature is finite everywhere, not zero, so why not just turn off the floor? The floor seems like it's creating a discontinuity, which is exactly what we were trying to avoid.

Can do.

markkrumholz: I should add: if the issue is that you're trying to avoid low temperatures where you need to worry about varying mu, just make your initial condition a Gaussian plus a constant, say 1e5 + a Gaussian to higher temperature, but then don't use any floors in the actual evolution. That will ensure that the temperature everywhere in the domain is high enough that we can ignore changes in mu.

aditivijayan: > I should add: if the issue is that you're trying to avoid low temperatures where you need to worry about varying mu, just make your initial condition a Gaussian plus a constant, say 1e5 + a Gaussian to higher temperature, but then don't use any floors in the actual evolution. That will ensure that the temperature everywhere in the domain is high enough that we can ignore changes in mu.

So for the problem at hand - evolving a Gaussian with Spitzer conductivity- we really care about the mean molecular weight right? Since the "true" solution is whatever the numerical solution is at nx=8192 with whatever IC I set it up with.

markkrumholz: I'm guessing you mean to say that we don't really care, and I think that's right. If the MMW is changing, who cares, since it's changing the same way in all the solutions we're comparing.

aditivijayan: Yes, that's what I meant- we don't really care what the mean molecular weight is for this specific problem- gaussian IC + spitzer.

aditivijayan: > More precisely, if you have the correct factors of the specific heat c v on the right hand side, and you are comparing the temperature against the reference solution, then I don't it matters what c_v is, as long as it's computed self-consistently everywhere.

@BenWibking - You are right in the sense that conductivity - $k_0 \  T^{5/2}$ - has no dependence on $\mu$. But the analytical solution for the test problems- i.e. the Pattle solution and also the evolution of a gaussian profile for a constant conductivity - assume that the diffusivity ($$D = \dfrac{k_0 T^{5/2}}{\rho c_v}$$) is spatially constant. Since $c_v$ depends on $\mu$ if the floor's $\mu$ is different from that of the peak, the diffusivity differs in the domain even though $k_0$ doesn't change. As a result the heat diffuses at different rates at different points and does not match the analytical solution, which assumes a single uniform D everywhere.

BenWibking: Right, sorry, I missed that. However, the diffusivity is unphysical for the floor anyway, so if c_v is wrong for the floor, why does it matter?

BenWibking: What I am thinking is: having a non-constant c_v breaks the self-similarity of the solution, yes, but so does having a floor in the first place. So if the only place where c_v is not constant is at the floor (and the rest of the solution is at much higher temperatures where c_v indeed is approximately constant), then I think it might not matter. I think this should be the case for this solution, since the Pattle solution has an extremely sharp front. In this situation, I think a smaller floor will always be better numerically.

aditivijayan: > What I am thinking is: having a non-constant c_v breaks the self-similarity of the solution, yes, but so does having a floor in the first place. So if the only place where c_v is not constant is at the floor (and the rest of the solution is at much higher temperatures where c_v indeed is approximately constant), then I think it might not matter. I think this should be the case for this solution, since the Pattle solution has an extremely sharp front. In this situation, I think a smaller floor will always be better numerically.

Right I understand now. Yes for a Pattle solution having a smaller floor is better- since I was modifying the code from the Gaussian solution for constant conductivity I kept the same floor. The floor was just inherited from previous set up. Once we have solved the convergence issue I can modify the code so that the Pattle solution has a small floor.

aditivijayan: ### Evolving an initial Gaussian profile through Spitzer conduction

The "discontinuity" in the profile is expected since not enough time has passed for the front to pass through. The conduction time for cooler gas is much longer.

image

### Comparing with nx=8192

image

### The convergence rate for error in the entire domain


image


### Convergence rate for inner region is better ($|x|<0.2$ pc)


image

BenWibking: I agree this looks much better. What floor did you use?

BenWibking: I think for the test pass criteria for the test problem it would make sense to check that the whole solution convergence around -1 and that the inner region converges around -2.

I am still slightly puzzled that the the inner region is -1.6 rather than -2.

aditivijayan: > I agree this looks much better. What floor did you use?

No floor!

aditivijayan: > I think for the test pass criteria for the test problem it would make sense to check that the whole solution convergence around -1 and that the inner region converges around -2.
> 
> I am still slightly puzzled that the the inner region is -1.6 rather than -2.

The Pattle solution does not converge to the analytical solution, even in the inner region, to -2. 

I am confused as well. In the [paper you shared earlier](https://link.springer.com/article/10.1007/s10915-024-02642-x), they are able to achieve second order convergence only for the log density method in which they are solving for log of density using a semi-implicit scheme. For their mixed method the order of convergence is close to unity.

aditivijayan:
> There are still two unaddressed issues from the previous AI review:
> 
> ```
>   - [P1] Pass cell composition to the Spitzer timestep EOS call — /Users/benwibking/amrex_codes/quokka/src/simulation.hpp:1305-1305
>     With the chemistry EOS enabled, omitting mass scalars leaves every species abundance at the -1.0 sentinel initialized in
>     EOSChem::ComputeTgasFromEint. The timestep therefore uses a temperature unrelated to the actual composition, whereas the conduction flux
>     supplies the composition correctly. Since conductivity scales as T^2.5, this can substantially overestimate the stable timestep. Extract and
>     pass the cell's mass scalars here.
> 
>   - [P2] Preserve physical internal energy when constructing temperature — /Users/benwibking/amrex_codes/quokka/src/conduction/
>     ElectronConduction.hpp:106-109
>     When hydro.use_dual_energy = 0, ApplyHydroStateFixup does not synchronize the auxiliary internal-energy component with total energy. This is
>     also the required configuration for physical viscosity and resistivity, whose dissipative heating is not added to the auxiliary equation.
>     Replacing ComputeInternalEnergy with this component therefore makes conduction use stale temperatures and miss that heating. Derive
>     temperature from total minus kinetic and magnetic energy, and apply the same correction to the new Spitzer timestep calculation.
> ```

Fixed these.

aditivijayan:
Cancelling workflows because the branch has the fake Gaussian problem.

markkrumholz:
I think this is all now working except for the mysterious question of why the convergence rate is 1.6 instead of 2. In the interest of checking if this is somehow the result of the non-linearity that develops in the current Gaussian test, I'd suggest slightly modifying the Gaussian initial conditions check by adding a constant temperature everywhere, to reduce the dynamic range on T from the current very large value to something a bit more modest -- say a factor of ~10 between the peak of the Gaussian and the minimum. Then run the test long enough that it is more than 1 diffusion time even at the minimum temperature. This should give us something a bit smoother, and we can see if that converges at second order.

aditivijayan:
> I think this is all now working except for the mysterious question of why the convergence rate is 1.6 instead of 2. In the interest of checking if this is somehow the result of the non-linearity that develops in the current Gaussian test, I'd suggest slightly modifying the Gaussian initial conditions check by adding a constant temperature everywhere, to reduce the dynamic range on T from the current very large value to something a bit more modest -- say a factor of ~10 between the peak of the Gaussian and the minimum. Then run the test long enough that it is more than 1 diffusion time even at the minimum temperature. This should give us something a bit smoother, and we can see if that converges at second order.


Omg finally! 

### Curve looks like this!

image


### Errors looks like this!

image

### And finally the convergence!

image


Currently running 4096 but I guess that's a bit pointless now...? -- Updated with 4096.

aditivijayan:
> > I think this is all now working except for the mysterious question of why the convergence rate is 1.6 instead of 2. In the interest of checking if this is somehow the result of the non-linearity that develops in the current Gaussian test, I'd suggest slightly modifying the Gaussian initial conditions check by adding a constant temperature everywhere, to reduce the dynamic range on T from the current very large value to something a bit more modest -- say a factor of ~10 between the peak of the Gaussian and the minimum. Then run the test long enough that it is more than 1 diffusion time even at the minimum temperature. This should give us something a bit smoother, and we can see if that converges at second order.
> 
> Omg finally!
> 
> ### Curve looks like this!
> image
> ### Errors looks like this!
> image
> ### And finally the convergence!
> image
> Currently running 4096 but I guess that's a bit pointless now...?

@markkrumholz - if this looks okay I will revert to the non-Gaussian original problem setup and use unity for slope comparison in the test.

markkrumholz:
I'd actually suggest leaving this version as the version of record, so that we demonstrate the -2 behavior. We'll still leave in the Pattle test which uses an analytic solution and which will show lower convergence rates, but that way we have a test that demonstrates that, for a smooth initial condition, we can get the expected convergence rate.

Once you've done that, let me know and I can re-approve.

markkrumholz:
Just because I realize I may have been a bit unclear, I think we should have two test problems for this physics: the Pattle test, for which we'll accept a convergence rate of -1 on the grounds of the discontinuity and the fact that you can't resolve the position the discontinuity to better than a cell, and this test, where we demand convergence at second order but test that convergence against a high-res reference solution rather than an analytic one. The comments can explain why we have both tests.

aditivijayan:
> Just because I realize I may have been a bit unclear, I think we should have two test problems for this physics: the Pattle test, for which we'll accept a convergence rate of -1 on the grounds of the discontinuity and the fact that you can't resolve the position the discontinuity to better than a cell, and this test, where we demand convergence at second order but test that convergence against a high-res reference solution rather than an analytic one. The comments can explain why we have both tests.

So, we would require both to pass? 

The GPU tests are 3D btw.

markkrumholz:
Yes, I think we should require both, since we want to test both things: second order convergence for smooth problems but without the ability to prove that we're converging to the physically correct solution, and linear convergence for a non-smooth problem but where we know the exact solution and can therefore verify that we're converging to the right answer.

src/conduction/ElectronConduction.hpp
BenWibking: @aditivijayan @markkrumholz This is still an open issue I would like to resolve (or at least make a plan to resolve) before merging this PR.

aditivijayan:
@BenWibking @markkrumholz - as discussed in the meeting yesterday I have added three new folders for testing constant conduction with AMR, spitzer with Gaussian IC and spitzer with Pattle solution. You may take a look. Thanks!

aditivijayan:
@markkrumholz - this PR is ready for review.

src/conduction/ElectronConduction.hpp
markkrumholz: @aditivijayan, looks like you haven't addressed this comment from @BenWibking. I though we had agreed to handle this via a constexpr physics trait, similar to the way viscosity is handled.

markkrumholz:
I thought we agreed on the Quokka call that you were going to change the switch between constant, Spitzer, and problem-specific conductivity by implementing a PhysicsTrait option, analogous to the way resistivity can now be set to zero, constant, or problem-specific.

aditivijayan:
> I thought we agreed on the Quokka call that you were going to change the switch between constant, Spitzer, and problem-specific conductivity by implementing a PhysicsTrait option, analogous to the way resistivity can now be set to zero, constant, or problem-specific.

My bad. I had interpreted it to be like future modification. I'm happy to do it. Can this PR be accepted as is and I could start another PR for that implementation?

markkrumholz:
> > I thought we agreed on the Quokka call that you were going to change the switch between constant, Spitzer, and problem-specific conductivity by implementing a PhysicsTrait option, analogous to the way resistivity can now be set to zero, constant, or problem-specific.
> 
> My bad. I had interpreted it to be like future modification. I'm happy to do it. Can this PR be accepted as is and I could start another PR for that implementation?

I'll defer to @BenWibking on which route he'd prefer. His change request is blocking, so he'll have to decide. I'm neutral between the two approaches.

src/conduction/ElectronConduction.hpp
aditivijayan: Here's another plan- for this PR let the structure be as is. I am currently testing the anisotropic conduction. I can add conductivity as a physics trait in that branch.

BenWibking: Ok, that works for me.

BenWibking:
Codex identified two remaining issues:
```
  - [P1] Preserve total-energy-derived internal energy during conduction — /Users/benwibking/amrex_codes/quokka/src/conduction/
    ElectronConduction.hpp:251-251
    With hydro.use_dual_energy = 0, the auxiliary internal-energy field is not synchronized with total energy by
    ApplyHydroStateFixup. Using that field here and subsequently rebuilding total energy from it discards differences caused by
    shock heating, viscosity, or resistivity—even when the conduction flux divergence is zero. Retain the total-energy-derived
    internal energy for this update, or update both energy fields by the same conservative flux increment without replacing
    total energy.

  - [P1] Bound the Spitzer timestep using neighboring face conductivities — /Users/benwibking/amrex_codes/quokka/src/
    simulation.hpp:1345-1351
    For a cold, low-density cell adjacent to hot, dense gas, the conductivity evaluated at the reconstructed face can greatly
    exceed the cold cell's own conductivity. Reducing cell-centered kappa(T)/(rho*c_v) therefore misses the large face
    conductivity divided by the receiving cell's small heat capacity. With weak or disabled saturation, the resulting explicit
    half-step can overshoot the neighboring temperature and become unstable even at the default conduction CFL. Estimate the
    bound using the face conductivities actually used by ComputeExplicit and each cell's heat capacity.
```

The first issue is the only one I'm concerned about now.

I think computing the timestep using the face-centered conductivities can be deferred to a future PR if it turns out to matter in practice.

BenWibking:
I'm fine with deferring the PhysicsTraits selection of the conductivity to a future PR.

aditivijayan:
> Codex identified two remaining issues:
> 
> ```
>   - [P1] Preserve total-energy-derived internal energy during conduction — /Users/benwibking/amrex_codes/quokka/src/conduction/
>     ElectronConduction.hpp:251-251
>     With hydro.use_dual_energy = 0, the auxiliary internal-energy field is not synchronized with total energy by
>     ApplyHydroStateFixup. Using that field here and subsequently rebuilding total energy from it discards differences caused by
>     shock heating, viscosity, or resistivity—even when the conduction flux divergence is zero. Retain the total-energy-derived
>     internal energy for this update, or update both energy fields by the same conservative flux increment without replacing
>     total energy.
> 
>   - [P1] Bound the Spitzer timestep using neighboring face conductivities — /Users/benwibking/amrex_codes/quokka/src/
>     simulation.hpp:1345-1351
>     For a cold, low-density cell adjacent to hot, dense gas, the conductivity evaluated at the reconstructed face can greatly
>     exceed the cold cell's own conductivity. Reducing cell-centered kappa(T)/(rho*c_v) therefore misses the large face
>     conductivity divided by the receiving cell's small heat capacity. With weak or disabled saturation, the resulting explicit
>     half-step can overshoot the neighboring temperature and become unstable even at the default conduction CFL. Estimate the
>     bound using the face conductivities actually used by ComputeExplicit and each cell's heat capacity.
> ```
> 
> The first issue is the only one I'm concerned about now.
> 
> I think computing the timestep using the face-centered conductivities can be deferred to a future PR if it turns out to matter in practice.

I have fixed the first one and will put in a fix for the second one in the next PR.

BenWibking:
Codex found one more issue. I can only get it to appear in practice if I compile with `-O0`, but this is important so it remains debuggable:
```
  - [P2] Avoid reading unallocated face fields in temperature diagnostics — /Users/benwibking/amrex_codes/quokka/src/problems/
    ThermalConductionSpitzerGaussian/testThermalConductionSpitzerGaussian.cpp:222-223
    When temperature is requested as a derived output, this accesses unallocated face-centered MultiFabs: this problem disables MHD, so AMRSimulation
    never defines those fields. Calling const_array(iter) dereferences a nonexistent FAB and can crash before the diagnostic kernel runs. Pass nullptr to
    ComputeInternalEnergy for these hydro-only problems, as the deleted test did. The same defect occurs in the new ConstantAMR and Pattle diagnostic
    implementations.
```

BenWibking:
Ok, due to the stochastic nature of AI code review, there are two more issues it reports that look real to me (issue 1 was spurious):

```
  2. The conduction timestep uses the wrong heat capacity for non-ideal EOS.
     Both conductivity branches assume a constant ideal-gas specific heat, while temperatures come from the selected EOS. This can underestimate
     diffusivity and permit an unsafe timestep. Use the EOS’s volumetric ComputeEintTempDerivative(rho, T, massScalars).

  3. Face density and composition are reconstructed inconsistently.
     The flux calculation combines reconstructed density with arithmetic averages of cell-centered partial densities. These can describe inconsistent
     composition and produce incorrect EOS results and saturation fluxes. Construct face composition consistently with face density.
```

Explanation of 2:
```
  The current timestep calculation instead assumes:

  c_v = k_B / (mean_molecular_weight * (gamma - 1));
  D = kappa / (rho * c_v);

  That is appropriate for a constant-composition ideal gas with those parameters. But the conduction flux obtains temperature from the selected EOS, whose
  heat capacity can differ because of composition or a tabulated energy–temperature relation.

  For example, if the assumed heat capacity is twice the actual EOS heat capacity, the code estimates half the actual diffusivity and allows a timestep
  twice as large. That can violate the explicit stability bound. An underestimate of heat capacity instead makes the timestep unnecessarily restrictive.

  The fix is to use the selected EOS consistently:

  const auto heat_capacity =
      quokka::EOS<problem_t>::ComputeEintTempDerivative(rho, T, massScalars);
  const auto diffusion_coefficient = kappa / heat_capacity;

  The derivative already returns volumetric heat capacity, so there is no additional factor of rho. Both conductivity branches need this treatment:
  constant conductivity does not imply constant heat capacity.
```

Explanation of 3:
```
• Issue 3 arises because the face calculation combines quantities constructed in two different ways:

  - Density: reconstruct each cell’s density to the interface, then average the left and right reconstructed values.
  - Species partial densities: average the original cell-centered values, without reconstruction.

  In src/conduction/ElectronConduction.hpp:156, this amounts to:

  rho_face = 0.5 * (rho_L_reconstructed + rho_R_reconstructed);
  massArray_face[n] = 0.5 * (massScalars_L[n] + massScalars_R[n]);

  Suppose neighboring cells have densities 2 and 4, with identical composition: 75% species A and 25% species B. Their partial densities are:

   Quantity             Left cell    Right cell    Cell-centered average
  ━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━
   Total density                2             4                        3
  ───────────────────  ───────────  ────────────  ───────────────────────
   Species A density          1.5             3                     2.25
  ───────────────────  ───────────  ────────────  ───────────────────────
   Species B density          0.5             1                     0.75

  The averaged partial densities sum to 3. But reconstruction uses surrounding cells and can produce a different face density—say 2.5. The EOS then
  receives:

  rho = 2.5
  rho_A = 2.25
  rho_B = 0.75

  These imply mass fractions 0.9 and 0.3, summing to 1.2. Thus, even spatially uniform composition becomes inconsistent at the face.

  A composition-dependent EOS can consequently calculate incorrect thermodynamic quantities. Conduction uses those quantities to calculate the sound speed
  and saturation flux, so the inconsistency can change the heat flux. An ideal EOS that ignores mass scalars hides the problem.

  The straightforward fix is to construct valid face mass fractions, then multiply them by the reconstructed rho_face to obtain consistent face partial
  densities.
```

aditivijayan:
> Ok, due to the stochastic nature of AI code review, there are two more issues it reports that look real to me (issue 1 was spurious):
> 
> ```
>   2. The conduction timestep uses the wrong heat capacity for non-ideal EOS.
>      Both conductivity branches assume a constant ideal-gas specific heat, while temperatures come from the selected EOS. This can underestimate
>      diffusivity and permit an unsafe timestep. Use the EOS’s volumetric ComputeEintTempDerivative(rho, T, massScalars).
> 
>   3. Face density and composition are reconstructed inconsistently.
>      The flux calculation combines reconstructed density with arithmetic averages of cell-centered partial densities. These can describe inconsistent
>      composition and produce incorrect EOS results and saturation fluxes. Construct face composition consistently with face density.
> ```
> 
> Explanation of 2:
> 
> ```
>   The current timestep calculation instead assumes:
> 
>   c_v = k_B / (mean_molecular_weight * (gamma - 1));
>   D = kappa / (rho * c_v);
> 
>   That is appropriate for a constant-composition ideal gas with those parameters. But the conduction flux obtains temperature from the selected EOS, whose
>   heat capacity can differ because of composition or a tabulated energy–temperature relation.
> 
>   For example, if the assumed heat capacity is twice the actual EOS heat capacity, the code estimates half the actual diffusivity and allows a timestep
>   twice as large. That can violate the explicit stability bound. An underestimate of heat capacity instead makes the timestep unnecessarily restrictive.
> 
>   The fix is to use the selected EOS consistently:
> 
>   const auto heat_capacity =
>       quokka::EOS<problem_t>::ComputeEintTempDerivative(rho, T, massScalars);
>   const auto diffusion_coefficient = kappa / heat_capacity;
> 
>   The derivative already returns volumetric heat capacity, so there is no additional factor of rho. Both conductivity branches need this treatment:
>   constant conductivity does not imply constant heat capacity.
> ```
> 
> Explanation of 3:
> 
> ```
> • Issue 3 arises because the face calculation combines quantities constructed in two different ways:
> 
>   - Density: reconstruct each cell’s density to the interface, then average the left and right reconstructed values.
>   - Species partial densities: average the original cell-centered values, without reconstruction.
> 
>   In src/conduction/ElectronConduction.hpp:156, this amounts to:
> 
>   rho_face = 0.5 * (rho_L_reconstructed + rho_R_reconstructed);
>   massArray_face[n] = 0.5 * (massScalars_L[n] + massScalars_R[n]);
> 
>   Suppose neighboring cells have densities 2 and 4, with identical composition: 75% species A and 25% species B. Their partial densities are:
> 
>    Quantity             Left cell    Right cell    Cell-centered average
>   ━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━
>    Total density                2             4                        3
>   ───────────────────  ───────────  ────────────  ───────────────────────
>    Species A density          1.5             3                     2.25
>   ───────────────────  ───────────  ────────────  ───────────────────────
>    Species B density          0.5             1                     0.75
> 
>   The averaged partial densities sum to 3. But reconstruction uses surrounding cells and can produce a different face density—say 2.5. The EOS then
>   receives:
> 
>   rho = 2.5
>   rho_A = 2.25
>   rho_B = 0.75
> 
>   These imply mass fractions 0.9 and 0.3, summing to 1.2. Thus, even spatially uniform composition becomes inconsistent at the face.
> 
>   A composition-dependent EOS can consequently calculate incorrect thermodynamic quantities. Conduction uses those quantities to calculate the sound speed
>   and saturation flux, so the inconsistency can change the heat flux. An ideal EOS that ignores mass scalars hides the problem.
> 
>   The straightforward fix is to construct valid face mass fractions, then multiply them by the reconstructed rho_face to obtain consistent face partial
>   densities.
> ```

Ok, I have addressed both these issues in the most recent pushes.

aditivijayan:
@BenWibking @markkrumholz - are there other changes you would like to see in this PR?

BenWibking:
> @BenWibking @markkrumholz - are there other changes you would like to see in this PR?

Codex says both of these are still issues:
  - Species normalization: src/conduction/ElectronConduction.hpp:160 divides cell-centered partial densities by reconstructed face densities. A focused harness using the actual reconstruction and normalization formula changed uniform fractions (0.4, 0.6) into (0.54545, 0.81818), summing to 1.36364. All densities remained positive. This confirms incorrect composition passed to the EOS; it affects composition-dependent EOS configurations, while the fixed-composition ideal EOS ignores these scalars.

  - Temperature diagnostics: I freshly compiled both ConstantAMR and Pattle drivers with -O0 -g. Both segfaulted during plotfile output after one timestep with derived_vars=temperature; both controls without that option completed successfully. Both backtraces identify ComputeDerivedVar → const_array → fabPtr, confirming the unallocated-field access.

Is this a wrong diagnosis, or do these still need to be fixed?

aditivijayan:
> > @BenWibking @markkrumholz - are there other changes you would like to see in this PR?
> 
> Codex says both of these are still issues:
> 
> * Species normalization: src/conduction/ElectronConduction.hpp:160 divides cell-centered partial densities by reconstructed face densities. A focused harness using the actual reconstruction and normalization formula changed uniform fractions (0.4, 0.6) into (0.54545, 0.81818), summing to 1.36364. All densities remained positive. This confirms incorrect composition passed to the EOS; it affects composition-dependent EOS configurations, while the fixed-composition ideal EOS ignores these scalars.
> * Temperature diagnostics: I freshly compiled both ConstantAMR and Pattle drivers with -O0 -g. Both segfaulted during plotfile output after one timestep with derived_vars=temperature; both controls without that option completed successfully. Both backtraces identify ComputeDerivedVar → const_array → fabPtr, confirming the unallocated-field access.
> 
> Is this a wrong diagnosis, or do these still need to be fixed?

These have been fixed. I am trying to look for other Codex issues. Where can I find them?

BenWibking:
> These have been fixed. I am trying to look for other Codex issues. Where can I find them?

Codex does not appear to update its cloud-based code reviews when new changes are pushed. I have been running it locally.

You can run Codex locally in a terminal with this branch checked out, and then type `/review` and then select "Review against base branch (PR-style)" and select "development".

src/problems/ThermalConductionConstantAMR/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+quokka_add_problem(JOB_NAME ThermalConductionConstantAMR PRIORITY 100)
```
BenWibking: Was this just for testing?

src/problems/ThermalConductionPattle/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+quokka_add_problem(JOB_NAME ThermalConductionPattle PRIORITY 100)
```
BenWibking: Testing only?

src/problems/ThermalConductionSpitzerGaussian/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+quokka_add_problem(JOB_NAME ThermalConductionSpitzerGaussian PRIORITY 100)
```
BenWibking: Testing only?

src/problems/ThermalConductionConstant/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+quokka_add_problem(JOB_NAME ThermalConductionConstant PRIORITY 100)
```
BenWibking: Testing only?

BenWibking (commented):
It looks like you set the priority of the tests so they would run first. Was this just for testing this PR, or did you want to set these as the priority for the merged version?

BenWibking:
I re-ran Codex locally, and it says these are still issues:
```
  - [P2] Normalize the reconstructed face composition — /Users/benwibking/amrex_codes/quokka/src/conduction/
    ElectronConduction.hpp:163-168
    For spatially varying composition with three or more mass scalars, independently limited mass fractions no longer necessarily
    sum to one. A focused default-PPM check with consecutive cell compositions (0.1,0.6,0.3), (0.3,0.3,0.4), (0.4,0.4,0.2), and the
    last composition repeated produces face fractions summing to 1.033333, despite every donor being normalized. Multiplication by
    rho_face preserves this inconsistency, so composition-dependent EOS calls compute incorrect thermodynamics and saturated fluxes.
    Enforce a valid, normalized face composition before constructing the partial densities.

  - [P2] Avoid accessing undefined face fields in the temperature diagnostic — /Users/benwibking/amrex_codes/quokka/src/problems/
    ThermalConductionConstant/testThermalConductionConstant.cpp:171-175
    When derived_vars=temperature is requested for ThermalConductionConstant, this unconditionally accesses face-centered MultiFabs
    even though the problem disables MHD and those fields are never allocated. const_array(iter) dereferences a nonexistent FAB
    before the non-MHD energy calculation can ignore it, causing a crash in unoptimized/debug builds. Pass nullptr to
    ComputeInternalEnergy for this hydro-only problem, or guard the face-field accesses as in the SpitzerGaussian diagnostic.
```

I saw that you changed the mass scalar reconstruction, but I think we do need to normalize them to sum to unity, otherwise it could produce an unphysical (or `NaN`) temperature.

src/problems/ThermalConductionConstantAMR/CMakeLists.txt
aditivijayan: Yes.

src/problems/ThermalConductionPattle/CMakeLists.txt
aditivijayan: Yep.

src/problems/ThermalConductionSpitzerGaussian/CMakeLists.txt
aditivijayan: Indeed.

src/problems/ThermalConductionConstant/CMakeLists.txt
aditivijayan: Should I change it?

aditivijayan:
> It looks like you set the priority of the tests so they would run first. Was this just for testing this PR, or did you want to set these as the priority for the merged version?

The priority can be changed once this PR is finalised and ready to be merged.

aditivijayan:
@BenWibking this is ready. I tried Codex review and it didn't show any red flags.

src/problems/ThermalConductionConstant/CMakeLists.txt
BenWibking: Yes, I think it can be changed now. Everything else looks good to me, and I am ready to approve it.

BenWibking (changes requested):

BenWibking (approved):
Looks good to me! Approved.

aditivijayan:
> Looks good to me! Approved.

Woohoo! Didn't know PR approvals could bring joy to me. :D

PR #2202, opened 2026-09-03
iterate state-dependent dust coefficients at GIRK stages

### LINKED ISSUE #2206: Dust drag ignores transverse momentum in 1D and 2D
> *This was generated by AI during triage.*

## Summary

In reduced-dimensional simulations, dust drag leaves transverse relative velocities undamped and omits their contribution to drag heating. Priority: P1 (incorrect numerical evolution).

## Explanation

At commit `0adb99a404c0a65852e8502639f39c98a7c471e9`, `src/dust/DustSources.hpp:361-408` sizes velocity arrays and computes kinetic energy and relative speed using `AMREX_SPACEDIM`. The stage solves, heating calculation, and final momentum writes likewise loop only over mesh dimensions (`:459-559`, `:604-609`).

The conserved state has three momentum components independently of mesh dimension: `src/physics_numVars.hpp:10` assigns four variables to each dust species, and `src/dust/dust_system.hpp:72-75` transports both transverse components. A 1D mesh describes spatial dependence, not a restriction that all motion must be along x.

Consequently, a uniform gas/dust pair differing only in y velocity experiences no drag in 1D. In 2D, the same failure occurs for z velocity. If gas has transverse velocity, the sound-speed calculation also treats its omitted kinetic energy as thermal energy, affecting velocity- or temperature-dependent stopping times.

## Reproduction and evidence

A native 1D, one-cell harness called the current `DustSources<Probe>::computeDustDrag` with gas and dust densities equal to 1, gas initially at rest, dust speed 1, gas internal energy 2.5, constant reciprocal stopping time 1, `dt=0.1`, heating weight 1, and coefficient iteration disabled. Rotating the initial dust velocity from x to y produced:

```text
direction=0 gas_momentum=0.09065934066 dust_momentum=0.9093406593 internal_energy=2.582440225
direction=1 gas_momentum=0 dust_momentum=1 internal_energy=2.5
```

The rotated uniform states should give identical momentum transfer and heating. This was reproduced with current headers and the local 1D AMReX library, on CPU with one MPI rank. The 2D consequence is established by the same loop bounds, not a separate 2D execution.

## Proposed patch

Use a fixed three-component velocity type throughout `computeDustDrag`:

```cpp
static constexpr int velocity_components = 3;
using Velocity = amrex::GpuArray<amrex::Real, velocity_components>;
```

Replace the velocity and momentum-change arrays currently sized by `AMREX_SPACEDIM` with `Velocity`, including nested species arrays. Change every loop over velocity components in this function to `dir < velocity_components`, covering initial state reads, kinetic energy, relative speeds, stage solves, work/heating, coefficient iteration, and final writes. Keep face-centered field arrays and spatial loops dimension-dependent.

Apply the same three-component state-read/write convention to `computeDustDragAndLorentz` (`:701-706`, `:947-952`) for consistency with its existing `Vec3` algebra; this does not by itself add reduced-dimensional MHD support.

Add a CTest regression that rotates a uniform drag problem through x, y, and z in 1D and 2D, checking identical relaxation, total momentum conservation, and gas-plus-dust energy conservation. Include a common transverse gas/dust boost to check that the stopping-time sound speed is unchanged.

## Duplicate check

Compared with all existing Markdown reports under `issues/` and all 482 open/closed GitHub issues returned on 2026-09-05. No matching report was found; the general dust discussion #1051 does not report this component omission.

mxchen39:
### Description
The current Picard iteration evaluates one set of state-dependent coefficients from the
candidate endpoint and uses it at both GIRK stages. This does not in general satisfy the
nonlinear stage equations and can reduce temporal accuracy. This PR evaluates the drag rate
and charge-to-mass ratio separately from the two implicit stage states and iterates the
resulting coefficients to convergence. The corresponding tests and documentation are updated.

### Related issues
Fixes #2206 

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Dust coefficient iteration supports separate absolute and relative charge tolerances at each integration stage.
  * Drag and Lorentz-force integration supports stage-specific coefficient updates.

* **Bug Fixes**
  * Invalid or non-positive iteration tolerances are rejected during configuration.
  * Dust density profiles now report unscaled density values correctly.

* **Documentation**
  * Updated dust parameter, Picard-iteration, and stopping-time notation documentation.

* **Tests**
  * Expanded gyromotion validation across timestep refinements.

* **Style**
  * Improved diagnostic plots for coefficient trajectories, convergence errors, and dust drift results.


chongchonghe:
**Verdict:** Ship it — math verified, #2206 fix confirmed by direct experiment; one gap, the #2206 fix has no regression test.
**Checked:** 1d + 3d presets, all 9 3D dust targets built; 1D dust suite 6/6 SUCCESS. Per-stage iteration restores design order — GL4 goes 1.008 → **3.998** (error 5.51e-04 → 7.78e-10), TP2025 2.089, Midpoint 2.001; conservation 2.2e-16.

**Non-blocking**
1. **#2206 fix is untested** — every 1D/2D dust problem zeroes `x2/x3*Momentum_index` (`testDustDamping.cpp:130-131,141-142`), so no test covers the code this PR fixed. I hand-rotated `DustDamping` to verify: on `development` the y-oriented run gets *no drag* (dust err **0.518** vs tol 0.03); on this PR y and z are bit-identical to x (1.185482442e-04). Re-introducing `dir < AMREX_SPACEDIM` would be caught by nothing — that rotation is ~10 lines and runs in 0.03 s.
2. **Order gates are tight** — `testDustDampedGyromotion` fails below 3.8 (GL4); measured 3.998. Orders trend up (3.880→3.962→3.990→3.998) so it's asymptotic, not round-off, but a threshold of 3.5 would be less compiler-sensitive.
3. **Tolerances validated even when iteration is off** — `QuokkaSimulation.hpp:842-850` aborts on non-positive `picard_*_tol` outside any `enabled` check, so a 0 in a deck that never reads it is now fatal.
4. **Charge sign-change guard silently replaced** by `atol + rtol·|ξ|`. Correct — a relative-only test is unsatisfiable near ξ=0 — and an O(1) flip is still caught; just not called out in the description.

| Phase | Ran | Observed |
|---|---|---|
| Sanity (base) | `DustDamping` 1d, motion along x | SUCCESS — gas 9.44833995e-05, dust1 1.185482442e-04 |
| **RED** (base) | same, rotated x→y | **FAIL** — dust1 0.5179713552, dust2 0.5292841817 (tol 0.03) |
| **GREEN** (PR) | same, rotated x→y | SUCCESS — bit-identical to x, all 10 digits |
| GREEN (PR) | rotated x→z | SUCCESS — bit-identical |
| Regression (PR) | unrotated x | bit-identical to base |
| EDGE (PR) | 1d `Dust*` | 6/6 SUCCESS |
| EDGE (PR) | 3d dust targets | 9/9 build; `DustDampedGyromotion` SUCCESS |

Block algebra re-derived by hand: `D = block11·block22 − β₁β₂L₁L₂`, `P1 = H11·T1 + H12·T2`, and `X1 = −ε(γ₁H₁₁L₁ + β₂H₁₂L₂)` all match. The closed-form 2×2 block inverse is valid only if the blocks commute — they do, since every `ReducedOperator` shares the axis `b̂` and `operator*` is symmetric in all three coefficients. That is what licenses generalising one `L` to `L₁,L₂`.

Bit-identical pure-drag results are not luck: with `Ω_L=0` every operator has `coeffCross=0`, and `operator*` then keeps `coeffParallel=0`, so the algebra collapses to the old scalar recurrence and `b_hat = 0` is harmless.

*Not a PR issue, recorded only:* on **aarch64**, 4 MHD dust tests abort under ctest's `fpe_trap_*` on this branch but not on `development`. The trap is inside `HydroSystem::ComputeFluxes`, and it **persists with `DustSources.hpp` reverted to `development`'s in a clean build** — so it is not the dust change. Results are correct with traps off, and your CUDA 13.0 / HIP CI is green. Looks like a codegen-sensitive spurious FP exception in the Riemann solver on ARM. GPU compilation rests on your green CUDA 13.0 / HIP CI — my local `1d-cuda` cross-compile (no device here) was too slow to finish.

</details>

chongchonghe (approved):
LGTM

Issue #2210, opened 2026-09-06
Remaking an AMR level discards its fine magnetic field


BenWibking:
## Summary

`AMRSimulation::RemakeLevel` replaces the magnetic field throughout a remade fine level with interpolation from the coarse level. It does not preserve the existing fine field on overlapping valid faces, losing resolved magnetic structure whenever that level is remade.

## Explanation

The cell-centered path in `src/simulation.hpp:2520–2522` fills the replacement array using `FillPatch`, which receives existing fine data. In contrast, the face-centered path at lines 2557–2562 calls `FillCoarsePatchFaceArray` for both replacement arrays, then swaps them into the live state.

The helper at lines 3411–3447 reads only `GetDataFaceArray(lev - 1, ...)` and passes those coarse arrays to `InterpFromCoarseLevel`. It never reads the existing level's magnetic arrays. There is no subsequent overlap copy before those original arrays are destroyed. Using this helper is appropriate for a newly created level; remaking an existing level must retain its resolved data.

For an information-loss example, a divergence-free fine field with `Bx` varying only in `y` can contain opposite deviations on the fine faces that average to a coarse face. Those deviations disappear on restriction and cannot be recovered from coarse interpolation. Remaking overlapping fine coverage therefore erases them without physical evolution. Meanwhile the retained cell-centered total energy can still describe the previous magnetic field, changing the inferred thermal energy.

Validation is a complete first-party data-flow trace through `RemakeLevel`, `FillCoarsePatchFaceArray`, `GetDataFaceArray`, and `AverageDownTo`. No AMR regrid simulation was run and the dependency interpolator implementation was not inspected.

## Proposed patch

Implement a face-array remake operation that preserves old fine valid faces on overlapping coverage and uses divergence-preserving coarse interpolation for newly refined regions. Retain consistent face boundary handling and reconcile faces at the boundary between retained and newly interpolated regions; blindly overwriting an interpolated field can itself violate the divergence constraint. Apply the established old/new timestamp contract consistently.

Add a 3D MHD regression with a nontrivial divergence-free fine field, change the fine BoxArray while retaining an overlap, and verify retained face values, divergence and magnetic/thermal energy before any evolution. Include decomposition-only remapping and expanding coverage cases.

## Duplicate check

[...]

PR #2241, opened 2026-09-07
Fix RemakeLevel: preserve fine mhd face field on regrid

AstroKriel:
### Description

This PR fixes a bug: RemakeLevel discards existing fine face-centred data on a level whenever it is rebuilt. It rebuilt face-centred quantities using `FillCoarsePatchFaceArray`, which only pulls from the coarse level. This is not consistent with how cell-centred quantities are rebuilt: they merge existing fine data with coarse via `FillPatch`. `FillPatchFaceArray` now adopts the same merge strategy, using `amrex::FillPatchTwoLevels` with `face_divfree_interp`, so fine faces already present are preserved on overlapping coverage, falling back to coarse interpolation only where coverage is genuinely new.

The old-time state keeps its existing coarse-only fill, matching how the cell-centred strategy leaves `int_state_old_cc` unfilled after a remake, because it is not used again.

### Validation

`MHDRemakeFaceField` (new problem) sets a divergence-free `Bx(y)` face field on level 1, forces a decomposition-only remake (`SetMaxGridSize` + `regrid()`, unchanged tags so the new BoxArray fully overlaps the old one), and checks the retained field and magnetic energy are unchanged to round-off. Reverting the fix reproduces the issue exactly: the fine structure collapses to the coarse mean and magnetic energy drops ~8%. Confirmed on both CPU and an A100 GPU (Gadi `gpursaa`).

### Related issues

- #2210: this PR implements the fix.

### Checklist

- [x] I have added tests for any new physics that this PR adds to the code: `MHDRemakeFaceField` regression test added.
- [x] I have triggered GPU tests: needs `/azp run` (full suite), not just `/azp run quick` — this touches `simulation.hpp`, shared base code outside `src/problems`.

## Summary by CodeRabbit

- **New Features**
  - Added a three-dimensional MHD face-field simulation configuration with adaptive mesh refinement, periodic boundaries, and standard runtime settings.
  - Added a regression test for preserving magnetic face fields during grid reconfiguration.

- **Bug Fixes**
  - Improved adaptive-mesh regridding to preserve existing fine-resolution magnetic face fields while applying divergence-preserving interpolation to newly covered regions.

### DIFF
--- a/inputs/MHDRemakeFaceField.toml
+++ b/inputs/MHDRemakeFaceField.toml
@@ -1,9 +1,3 @@
-# Regression test for issue #2210: RemakeLevel discards the existing fine mhd face
-# field on regrid instead of preserving it. Level 1 covers the full domain (refineGrid
-# tags everything) so that a decomposition-only remake (see testMHDRemakeFaceField.cpp,
-# triggered via SetMaxGridSize + regrid()) gives 100% overlap between the old and new
-# level-1 BoxArray: any change to the fine field can only come from RemakeLevel.
-
 geometry.prob_lo = [0.0, 0.0, 0.0]
 geometry.prob_hi = [1.0, 1.0, 1.0]
 quokka.bc = ["periodic", "periodic", "periodic"]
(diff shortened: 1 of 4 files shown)


AstroKriel:
The failed test is unrelated flakiness in `ParticleSink`, not an issue in this PR.

BenWibking:
@chongchonghe Can you fix the ParticleSink GPU test issue? I agree with Neco that it does not look related to this PR. I think Aditi's recent PR also had the same issue.

markkrumholz (approved):

BenWibking:
I would like to test this on my machine before merging.

BenWibking:
Can you hold off until I can review this manually?

Issue #2233, opened 2026-09-07
Post-update CFL checks omit magnetic and dust speeds


BenWibking:
> *This was generated by AI during triage.*

## Summary

The post-update CFL validator uses only gas sound speed plus gas velocity, even for MHD and dust simulations. It can accept a timestep that violates the same physical speed constraints used during initial timestep selection.

## Explanation

At commit `0adb99a404c0a65852e8502639f39c98a7c471e9`, `HydroSystem::maxSignalSpeedLocal` (`src/hydro/hydro_system.hpp:351–391`) returns `cs + abs_vel`. It constructs face-field handles but uses them only to obtain thermal sound speed. It never computes a fast magnetosonic speed or reads dust momenta.

By contrast, `ComputeMaxSignalSpeed` in the same file (`:394–470`) computes fast magnetosonic speeds for MHD and incorporates every dust group's velocity. `QuokkaSimulation::computeMaxSignalLocal` calls that function for timestep selection.

`QuokkaSimulation::isCflViolated` (`src/QuokkaSimulation.hpp:2322–2342`) calls the incomplete `maxSignalSpeedLocal`, reduces it over MPI, and compares the completed step against `1.1*cflNumber*dx_min/max_signal`. Its caller uses the result to decide whether to retry the update.

For example, a post-update state with stationary gas, `cs=1`, and dust speed `100` gives `maxSignalSpeedLocal=1`, while the initial-estimation formula gives `101`. With `dx=1`, CFL number `0.4`, and completed step `0.1`, the validator accepts the step (`0.1 <= 0.44`), although the dust-aware threshold is approximately `0.00436`. A magnetically dominated state has the analogous discrepancy. These numbers illustrate the source formulas; no full evolution producing this state was run.

## Proposed patch

Factor the per-cell characteristic-speed calculation out of `ComputeMaxSignalSpeed` and share it with `maxSignalSpeedLocal`. Both paths must include the same gas, magnetic, and dust contributions, EOS handling, dimensional conventions, and finite-value checks. Keep the existing local-versus-MPI reduction ownership and the post-update tolerance factor.

Add a test comparing both speed estimators on identical states for hydro-only, magnetically dominated MHD, and fast dust with multiple groups. Exercise `isCflViolated` with a completed step between the gas-only and full-physics thresholds and verify that it requests a retry. Include an ordinary hydro case to preserve existing behavior.

## Duplicate check

[...]

PR #2242, opened 2026-09-07
Fix maxSignalSpeedLocal: include MHD and dust signal speeds

AstroKriel:
### Description

This PR fixes a bug: the post-update CFL check can accept a timestep that violates the MHD or dust CFL constraint. isCflViolated calls maxSignalSpeedLocal, which only computes gas sound speed plus gas velocity, ignoring the fast magnetosonic speed for MHD and every dust groups velocity that ComputeMaxSignalSpeed already accounts for (see #2233 for the full breakdown).

I factored the shared per-cell speed calculation out of ComputeMaxSignalSpeed into ComputeCellSignalSpeed, and have both callers use it. The two paths cannot drift apart again since there is only one implementation now.

### Related issues

- #2233: this PR implements the fix.

### Checklist

- [x] I have added a link to any related issues (see above).
- [ ] I have added tests for any new physics: no new physics, and the fix removes a duplicate implementation rather than reconciling two, so there is nothing left to compare.
- [x] GPU tests: triggered /azp run (full suite), since this touches hydro_system.hpp shared outside src/problems. All MHD and dust tests passed on CPU, ROCm, and CUDA; the only failure was ParticleSink on ROCm, confirmed unrelated flakiness (see comment).

### DIFF
--- a/src/hydro/hydro_system.hpp
+++ b/src/hydro/hydro_system.hpp
@@ -356,7 +356,7 @@ void HydroSystem<problem_t>::ConservedToPrimitive(amrex::MultiFab const &cons_cc
 template <typename problem_t>
 AMREX_GPU_DEVICE AMREX_FORCE_INLINE auto
 HydroSystem<problem_t>::ComputeCellSignalSpeed(amrex::Array4<const amrex::Real> const &cons_cc, int i, int j, int k,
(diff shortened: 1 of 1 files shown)


AstroKriel:
The failed test is unrelated flakiness in `ParticleSink`, not an issue in this PR.

markkrumholz (approved):

PR #2290, opened 2026-09-17
Fix zero convergence scale passed to BackwardEulerOneVariable

chongchonghe:
#### Problem

`BackwardEulerOneVariable` stops when the residual falls below `rel_tol * compare`, where `compare` is a physical scale supplied by the caller. For the dust-temperature solve in `ComputeDustTemperatureBateKeto` the caller passed

$$\texttt{compare} = N_d \sqrt{T_\mathrm{gas}}\, T_\mathrm{gas},$$

which is the gas-dust collisional term alone. That term vanishes identically when the gas-dust coupling coefficient is zero ($N_d = 0$, reachable by setting `radiation.dust_gas_interaction_coeff = 0`, or in a cell with no hydrogen), so the tolerance becomes zero and the residual test can never be satisfied.

A previous attempt worked around this by falling back to the initial residual, `scale = std::max(compare, std::abs(f0))`, with an early exit `std::abs(f0) < rel_tol * scale`. That condition is unsatisfiable for any non-zero `f0` — it compares the residual against itself — so it never fires, and when `compare` is zero it silently replaces the physical scale with a self-referential one.

The bracketing and bisection guard added earlier still terminates the iteration through the step-size criterion, so this did not show up as a hard failure in the existing tests; the residual-based criterion was simply inoperative.

#### Solution

Remove the `f0` fallback block and use `compare` directly, with `AMREX_ASSERT(compare > 0.0)` so a degenerate scale is caught in a Debug build rather than silently disabling the test.

Every caller now passes a scale that is strictly positive and physically meaningful. For the dust temperature (`ComputeDustTemperatureBateKeto`, `src/radiation/radiation_system.hpp`) the residual balances the radiative term against the collisional term, and each of those can vanish on its own, so the scale is the sum of three contributions, all with the units of a radiation energy density:

$$\Lambda_\mathrm{compare} = \underbrace{N_d \sqrt{T_\mathrm{gas}}\, T_\mathrm{gas}}_{\text{collisional}} \; + \; \underbrace{\hat{c}\, \Delta t\, \rho \sum_g \left( \kappa_{E,g} E_g + \kappa_{P,g}\, \tfrac{4\pi B_g}{c} \right)}_{\text{radiative}} \; + \; \underbrace{\sum_g \tfrac{4\pi B_g}{c}}_{\text{blackbody}},$$

with the opacities and $B_g$ evaluated at the initial dust temperature. The collisional term vanishes at $N_d = 0$ and the radiative term vanishes at $\kappa = 0$, so neither is sufficient alone; the blackbody energy density $\sum_g 4\pi B_g / c$ is positive for any $T_{d,\mathrm{init}} > 0$ and keeps the scale non-zero in every degenerate case. The radiative term is retained because it is the true residual scale in the optically thick limit, where using the blackbody term alone would give a needlessly tight tolerance.

For the `dust_model == 2` gas-energy solve (both call sites in `src/radiation/radiation_dust_system.hpp`), `lambda_gd_times_dt` is signed, so the raw sum could in principle cancel. The terms are now summed by magnitude; `Egas_guess > 0` then keeps the scale positive.

**Existing problems are unaffected to roundoff.** The dust-temperature scale is only enlarged, and in the coupled case the collisional and radiative terms dominate the added blackbody term. `RadDustMG` moves from a relative L1 error norm of `0.0007237950248` on `development` to `0.0007237950194` here, a relative change of $5 \times 10^{-9}$, from the residual test tripping marginally earlier.

#### Testing

Red/green on the degenerate cases, running `RadDustMG` in a `1d-debug` build (where `AMREX_ASSERT` is active) with `radiation.dust_gas_interaction_coeff = 0`, and separately with the problem's opacity constant `chi0` also set to zero (a scratch edit, not part of this PR):

| Scale for the dust solve | $N_d = 0$, $\kappa \neq 0$ | $N_d = 0$, $\kappa = 0$ |
|---|---|---|
| `N_d sqrt(T_g) T_g` only (before) | `Assertion 'compare > 0.0' failed`, exit 6 | `Assertion 'compare > 0.0' failed`, exit 6 |
| `+` radiative term only | runs to completion | `Assertion 'compare > 0.0' failed`, exit 6 |
| `+` radiative `+` blackbody (this PR) | runs to completion | runs to completion |

(The test's own answer check still fails at zero coupling, as expected — its reference solution assumes a coupled dust-gas system. The point of these runs is that the solver no longer receives a degenerate tolerance.)

Regression, `1d` Release, all with the default coupling coefficient and opacity:

| Test | Result | Error norm |
|---|---|---|
| `RadDustMG` | SUCCESS | `0.0007237950194` |
| `RadDust` | SUCCESS | `0.0003586751597` |
| `RadMarshakDust` | SUCCESS | `0.0100814897` |
| `RadMarshakDustPE` | SUCCESS | `0.003104766253` |

`RadDust`, `RadMarshakDust` and `RadMarshakDustPE` are not touched by this PR and exercise the same single-group and multigroup dust paths.

`RadDustMG` also passes in the `1d-debug` build with default settings, confirming the new assertion does not fire on the normal path.

### Related issues

N/A

### Checklist


chongchonghe:
**Verdict:** Ship it after a rebase — no blocking issues; the bug reproduces on `development` and is gone here.
**Checked:** 1d Release + 1d-debug, 5 dust/radiation targets; 1d-cuda compile-only for both `nGroups_` branches. `RadDustMG` 0.0007237950253 → 0.0007237950194; `RadDust`/`RadMarshakDust`/`RadMarshakDustPE` bit-identical.

**Blocking:** none.

**Non-blocking**
1. **Regression table was measured against a base 11 commits stale** — `768f1bcd` forks at `bcea01f2`, missing `fbc37d7d` (#2202, same dust path) and a `tol_p` 0.01→0.02 bump in `testDTypeFront1D.cpp`. Comparing the head directly against `development` shows a spurious ×1.909 in `DTypeFront1D` injected energy; rebased (clean, 3 commits) it collapses to 5e-08.
2. **`DTypeFront1D` already ships the degenerate case** — `inputs/DTypeFront1D.toml:70` sets `dust_gas_interaction_coeff = 0.0`, so `compare == 0` fires in CI today; no hand-edit of `RadDustMG.toml` needed to reproduce.
3. **`AMREX_ASSERT` is a no-op in Release** — `radiation_system.hpp:1635`. `AMREX_ALWAYS_ASSERT` costs one compare and is what caught the original defect for me; the fallback it replaced is gone in the build that actually runs.
4. **Duplicate Planck + opacity evaluation** — `radiation_system.hpp:1789-1797` computes `fourPiBoverC`/`opacity_terms` at `T_d_init`, then `f0 = rhs(x0)` recomputes both at the same temperature (`:1746`). Once per dust solve, per cell, on the device path.
5. **`n_step > 0` divides by `N_d`** — `radiation_system.hpp:1731`, same `N_d == 0` regime this PR hardens. Pre-existing, caught by the failure counter, single-group path only.

| Phase | Ran | Observed |
|---|---|---|
| RED | `development` + scratch `AMREX_ALWAYS_ASSERT(compare > 0.0)`, `DTypeFront1D` | ``Assertion `compare > 0.0' failed ... Msg: RED-PROBE``, exit 1, FAIL |
| RED control | same binary, `RadDustMG` | exit 0, SUCCESS — probe specific to the decoupled case |
| GREEN | rebased PR, **1d-debug** (assert live), `DTypeFront1D` | all 6 checks pass, assert silent |
| GREEN | 1d-debug `RadDustMG`, `RadDust` | SUCCESS |
| EDGE | 1d Release A/B, `development` vs rebased PR | 5/5 SUCCESS both; `DTypeFront1D` p/injected 1.0106684 → 1.010668354 (4.6e-08), reprocessed 0.3683323602 → 0.3683323784 |
| GPU | 1d-cuda, `RadDust` + `RadDustMG` (`-DCMAKE_CUDA_ARCHITECTURES=80`; preset default `native` fails with no device) | both link, exit 0, no device-code diagnostics |

Why the regression surface is safe: with `T_d_init = T_gas` the collisional term vanishes at `x0`, so `f0` is a *difference* of exactly the terms `Lambda_compare` sums as *magnitudes* — hence `Lambda_compare >= |f0|` always, the new scale dominates the old `max(compare, |f0|)`, and the residual test can only trip earlier. `sum(fourPiBoverC) = a*T_gas^4 > 0` at all three call sites, so the scale is provably non-zero.

</details>

PR #2303, opened 2026-09-21
Fix hydrogen network

James471:
### Description

- Photoheating now uses the actual ionizing-band photon energy minus 13.6 eV, instead of a hardcoded `6.4e-12` erg constant
- Add missing collisional ionization cooling term (energy + Jacobian)
- Fix recombination coefficient (drop redundant `+1−1` term) and recombination cooling (based on Krumholz et al. 2007)
- Replace free-free cooling with the Frazer & Heitsch (2019) fit; update `DTypeFront` reference to match
- `ChemBandsHeader_` is now a device-callable `constexpr` function instead of a global array, so photoheating can read band energies on GPU
- Regenerate Julia reference RHS/CSV to match corrected coefficients

### Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved photoionization heating calculations using photon-band energy, with safeguards against negative heating.
  - Updated recombination, collisional-ionization, and free-free cooling calculations, including temperature-dependent behavior.
  - Improved consistency of chemical-band data across photoionization problem configurations.

- **Configuration**
  - Enabled temperature-dependent recombination in the one-zone photoionization setup.

- **Numerical Improvements**
  - Added updated cooling derivatives to improve energy evolution and related calculations.


James471:
pre-commit.ci run

chongchonghe:
**Verdict:** Mechanism is right and RED→GREEN is clean; 2 questions before merge — photoheating is now 24.2 eV/ionization (was 4.0), and neither new term is covered by the tight test.

**Checked:** `1d` and `1d-cuda` builds of `OneZonePhotoionization` (CUDA 163/163, device link clean). Base code against the PR's reference: energy norm **0.822** (tol 1e-6); PR: **2.22e-07**. Species/photon norms sit at ~1e-7 in every run, so the change is confined to `ydot(net_ienuc)`.

**Blocking**

1. **Photoheating 4.00 eV → 24.22 eV per ionization (6.06×)** — `actual_rhs.H:66`. Band mean of [3.29e15, 1.5e16] Hz is 37.8 eV. `DTypeFront`'s enforced ionized-gas $T_\mathrm{eq}$ moves 10,471 K → 15,632 K, and the new one-zone reference ends at **88,847 K** vs 15,094 K. `testDTypeFront.cpp:627` pins the simulation to that analytic value within 5%, but both sides moved together, so nothing catches it. Intended, or an artefact of the 62 eV upper band edge?
2. **Neither new term is tested** — `OneZonePhotoionizationWithEnergy.toml:54-55` sets `ion_ff_cooling_switch=0` and `collisional_ionization_switch=0`, and `fraction.jl` implements neither, so the new `jac(net_ienuc,·)` collisional entries and the whole `get_cle_term` ladder rest on DTypeFront's 5% check. A wrong Jacobian entry costs convergence, not accuracy — it will not surface there.

**Non-blocking**

- **`ion_ff_cooling` is 95% not free-free** — `actual_rhs.H:126`. ff is 5.1% of the total at 1e4 K, so `network.ion_ff_cooling_switch` now gates the dominant cooling channel under a misleading name.
- **Cooling above 2e4 K drops 26×, and 1120× at 1e5 K.** Probably correct — the old `1e-19*exp(-118348/T)` saturated at 1e-19 and never switched off — but worth saying in the description.
- **F&H2019 ladder duplicated 3×** — `actual_rhs.H:102`, `:134` (exponents only), `testDTypeFront.cpp:129`. Also `2.6e-13*(1+x-1)` → `2.6e-13*x` is algebraically a no-op, not a fix.

| Phase | Ran | Observed |
|---|---|---|
| Base sanity | base code + base CSV, `OneZonePhotoionizationWithEnergy.toml` | energy 9.94e-08, species 9.27e-08/3.25e-07/9.27e-08, photon 1.30e-09 → SUCCESS |
| RED | base code + PR's regenerated CSV/TOML | **energy 0.821995697** → `Energy error norm exceeds tolerance!` → FAIL |
| GREEN | PR @ `346aa2f3`, same input | **energy 2.22452361e-07**, species 9.87e-08/3.09e-07/8.64e-08 → SUCCESS |
| EDGE | PR, `OneZonePhotoionization.toml` (`energy_switch=0`) | energy 0, species 1.55e-07/4.23e-07/1.55e-07 → SUCCESS |
| GPU | PR, `1d-cuda`, `-DCMAKE_CUDA_ARCHITECTURES=80` | 163/163, `OneZonePhotoionization SUCCESS` |

$T_\mathrm{eq}$ from `net_energy_ionized`, solving $\alpha_B\Gamma = \Lambda_\mathrm{rec} + \Lambda_\mathrm{ff+cle}$ ($n_e$-independent):

| config | $T_\mathrm{eq}$ |
|---|---|
| base (4.00 eV, old cooling) | 10,471 K |
| heating change only | 12,578 K |
| cooling change only | 7,177 K |
| PR (24.2 eV, new cooling) | **15,632 K** |

The `get_cle_term` ladder is continuous to 3 s.f. at all seven breakpoints, and its derivative uses the same breakpoints with $\alpha\Lambda/T$ — exact for a power law. The three new `jac(net_ienuc,·)` collisional entries are also correct by inspection. `1.380649e-16` → `C::k_B` is exact (`fundamental_constants.H:27`), and `recombination_temperature_dependent = 1` already was the default (`_parameters:7`).

Config notes: `quokka config -d <preset> --delete` fails from the repo root with `Current working directory cannot be established`; use `rm -rf build/<preset>` then a plain `config`. `1d-cuda` needs `-DCMAKE_CUDA_ARCHITECTURES=80` on a device-less host.

</details>

chongchonghe:
@James471 The blocking issue 1 seems real: The one-zone reference solution goes to 8 × 10^4 Kelvin, which is way too high, and this is because the ionizing energy is too high. 

I previously commented that the mean photon energy is not important because it's canceled in the calculation when converting from energy density to number density. However, this is not true because you use it to compute the ionizing heating rate. So we need to get it right.

If you are using the midpoint of the group boundary as the average photon energy, then you could set the boundary to an arbitrary value so that the boundary mean is what you wanted. Note that the group boundaries do not really matter outside of photochemistry, because we do not account for the contribution to LyC from thermal emission.

chongchonghe (changes requested):
see my comments above

James471:
@chongchonghe This should be ready now.

chongchonghe:
**Verdict:** C1 is properly fixed — ship it once you've answered the tolerance question below. $\Gamma_\mathrm{photo}$ is now 6.21 eV and $T_\mathrm{eq}$ lands at 8,543 K.

**Re-ran at `838cb573`:** one-zone energy norm **3.59e-07** (tol 1e-6); `DTypeFront1D` all seven sub-checks pass. Band edge `1.50e16` → `6.29e15` Hz applied consistently across all six `CHEM_BANDS`, `fraction.jl:21` and `testDTypeFront1D.cpp:91`; no stale copies in the repo.

| | base | v1 | **v2** |
|---|---|---|---|
| mean photon energy | — | 37.82 eV | **19.81 eV** |
| $\Gamma_\mathrm{photo}$ | 4.00 eV | 24.22 eV | **6.21 eV** |
| `DTypeFront` $T_\mathrm{eq}$ | 10,471 K | 15,632 K | **8,543 K** |
| one-zone ref final $T$ | 15,094 K | 88,847 K | **23,172 K** |

(23,172 K is fine: that test has `ion_ff_cooling_switch=0`, and 1.54× tracks the 1.55× heating increase.)

**Worth a reply**

1. **`838cb573` "Minor fixes" doubles `tol_p` 0.01 → 0.02 and deletes the measurements that justified the old value** — `testDTypeFront1D.cpp:687`. Loosening it is *necessary* (ratio 1.0107, so 1% now fails), and I re-ran the sweep it documented: **128→1.01067, 256→1.01088, 512→1.01075, 1024→1.01076** — flat to 2e-4, so the excess is converged and 2% is defensible. Please put those numbers back. Separately, it grew 0.38% → 1.08% purely from halving `E_photon` (nothing else in `a19f0c5f` touches the momentum path) — is an `E_photon`-scaling conservation residual understood?

**Still open from v1** (re-checked today, all unchanged)

- **Neither new term is tested** — `OneZonePhotoionizationWithEnergy.toml:54-55` still has both switches at 0 and `fraction.jl:94` implements neither, so the collisional-ionization Jacobian entries and the `get_cle_term` ladder rest on `DTypeFront`'s 5% check. `rocm-quick` does exercise them at default switches.
- **`ion_ff_cooling` is 95% not free-free** at 1e4 K — `actual_rhs.H:126`, `_parameters:28`.
- **F&H2019 ladder duplicated 3×** — `actual_rhs.H:102`, `:134`, `testDTypeFront.cpp:129`.

| Phase | Ran | Observed |
|---|---|---|
| Build | `1d`, `OneZonePhotoionization` + `DTypeFront1D` | 181/181, both SUCCESS |
| Regression | `OneZonePhotoionizationWithEnergy.toml` | energy 3.59e-07, species 9.03e-08/3.19e-07/9.03e-08, photon 1.27e-09 → SUCCESS |
| `DTypeFront1D` | shipped input | ratio 1.0106684, gas share 0.3753, signed total 8.0e-17 of injected → SUCCESS |
| Resolution sweep | `amr.n_cell` 256 / 512 / 1024 | 1.0108819 / 1.0107531 / 1.0107626 |

$T_\mathrm{eq}$ from `net_energy_ionized`, solving $\alpha_B\Gamma = \Lambda_\mathrm{rec} + \Lambda_\mathrm{ff+cle}$ ($n_e$-independent). The `[1000, 1e5]` bracket in `compute_equilibrium_temperature_ionized` still straddles the root.

Not re-checked locally because the two new commits touch neither `actual_rhs.H` nor `network_header.template`, and CI covers the exact head: `CUDA@13.0` build pass, `HIP ROCm 7.14` pass, `rocm-quick` pass. `quick` was still pending at review time — worth confirming green before merge.

</details>

chongchonghe (approved):
LGTM

PR #2305, opened 2026-09-21
James471/thermochemical 3band

James471:
### Description

Adds thermal-band photochemistry: the reaction network's cooling channels now re-emit into the thermal radiation bands. Tests absorption of optical radiation from source by dust and deposition of optical momentum against the KM09 solution.

**New network `thermophotochemistry_handwritten`** — a copy of `photoionization_H_caseB_handwritten` with various heating and cooling processes re-emitting into the thermal bands

**`NUM_THERMAL_BANDS` becomes a registry property**, filled into `network_header.template` and defaulting to 0.

**`computePhotoChemistry's`** unused
`stage`/`energy_update_factor` parameters are dropped — the sole caller passed 1.

**`DTypeFront1D`** rewritten around a numerically integrated thin-shell front solution driven by both ionized-gas pressure and radiation pressure. Two checks: equilibrium gas temperature in the cavity and in the undisturbed neutral gas, and front position against the integrated solution. 

### Checklist


chongchonghe:
**Verdict:** 3 blocking — CUDA build does not compile, the 3D run fails by exactly 64×, and the new front check fails at 2× resolution. Core feature (thermal-band re-emission, `NUM_THERMAL_BANDS` plumbing, `chat/c` scaling) is sound.
**Checked:** 1d/3d/1d-cuda, base `james471/quokka-jaff` @ `3f29b3f4`. 1d `DTypeFront1D` passes all 4 checks (cavity T 11644.32 K vs 11613.24 K; front `-2.769` of 3.0 cells). `DTypeFrontVC` 3d + 6 dust/line-cooling tests 7/7 SUCCESS.

**Blocking**

1. **CUDA build fails** — `radiation_system.hpp:856`. `GetThermalBandQuanta` is `AMREX_GPU_HOST_DEVICE` but indexes the host static member `radBoundaries_`: `error: identifier "RadSystem<    ::DTypeFront1D> ::radBoundaries_" is undefined in device code` (×2). Dropping `AMREX_GPU_HOST_DEVICE` builds clean (verified; all call sites are host-side) — `GetChemBandQuanta` avoids this by calling the `constexpr` `ChemBands()` instead.
2. **`DTypeFront1D` FAILS in 3D** — `testDTypeFront1D.cpp:113`. `compute_effective_length` sums `dx*(1-x_HI)` over every column without dividing by `n_cell[1]*n_cell[2]`, so at `[128,8,8]` it is exactly 64.000× too large: `differs ... (2056.142705 cells)`. `CMakeLists.txt:4` registers the target for SPACEDIM 1 and 3.
3. **Front tolerance is resolution-dependent and 92% consumed** — `testDTypeFront1D.cpp:721`. `tol_cells = 3.0` on a ~6% systematic offset: 128 cells → `-2.769` cells (7.81% vs 8.46% budget); 256 cells → `-4.288` cells, **FAIL** (6.05% vs 4.23%). Absolute offset only drops 4.33e17→3.35e17 cm (order 0.37). A relative tolerance would be resolution-stable. The shell check is fine (`0.052`→`-0.397` cells).

**Non-blocking**

- **Vacuous pass** — `compute_median_and_check` returns on `ntot == 0` without setting `status = 1`, so an empty cavity/neutral selection reports PASS.
- **`network_name` wrong** — `thermophotochemistry_handwritten/actual_network.H:15` still says `"photoionization_H_caseB_handwritten"`.

| Phase | Ran | Observed |
|---|---|---|
| VERIFY | `1d` `DTypeFront1D`, shipped input | exit 0, 4/4 pass; cavity 11644.32 K (0.27%), neutral 41.098 K (1.6%), front `-2.769340214` cells, shell `0.05165080728` cells |
| VERIFY | `1d-cuda` `DTypeFront1D` (compile only, no device) | FAIL, `radiation_system.hpp(856)` device-code error ×2 |
| VERIFY | `1d-cuda` `OneZonePhotoionization` (`NumThermalBands==0`), `RadDustMG` | SUCCESS both |
| VERIFY | `1d-cuda` `DTypeFront1D`, `AMREX_GPU_HOST_DEVICE` removed (probe, reverted) | SUCCESS, 0 errors — only GPU defect in the PR |
| EDGE | `1d` `DTypeFront1D`, `n_cell=[256,8,8]` | FAIL, front `-4.287965711` cells; temps + shell pass |
| EDGE | `3d` `DTypeFront1D` | FAIL, front `2056.142705` cells = 64× ; temps + shell pass |
| EDGE | `3d` `DTypeFrontVC` | SUCCESS, rel. x-momentum error 0, rel. energy error 1.859854176e-08 |
| EDGE | `1d` RadDust, RadDustMG, RadMarshakDust, RadMarshakDustPE, RadLineCooling, RadLineCoolingMG, OneZonePhotoionization | 7/7 SUCCESS — `BackwardEulerOneVariable`/`Lambda_compare` rework causes no regression |

Also noted, not blocking: `GetThermalBandQuanta` hardcodes `C::ev2erg` though it is declared outside `#ifdef PHOTOCHEMISTRY` and `RadDustMG`/`RadMarshakDust`/`RadhydroShockMultigroup` set other `energy_unit`s; VODE never sets `ewt` for thermal rows (`vode_dvode.H:85`) and `vode_dvhin.H:54` gives them `atol_rad_num = 1e-6`, correct today only because `int_state{}` zeroes `ewt` and the input selects Rosenbrock; `prob_hi` is now `[2e19,2e19,2e19]` with `n_cell=[128,8,8]`, so 3D cells are 16:1 while `CMakeLists.txt:1-3` still claims cubic cells; the `chat/c` scaling at `photochemistry.hpp:220` and the un-debited Rydberg emission at `actual_rhs.H:242` were both checked and are correct.

</details>

James471:
**Verdict:** 2 blocking — an unaccounted cooling channel in the new network, and a front check that reports PASS with one of its two metrics 2x outside its own tolerance. Core physics otherwise checks out; GPU unverified here.
**Checked:** `1d-apple`, `DTypeFront1D` built + run at `a401e5db`. Cavity T 7786.55 K vs 7772.49 K analytic (0.18%), neutral T 41.08 K vs 41.77 K (1.66%), shell 4.609e18 cm vs ODE 4.602e18 cm (+0.15%).

**Blocking**
1. **`collisional_ionization_cooling_rate` debited from the gas, emitted into no band** — `actual_rhs.H:237,245-246`. `ydot(net_ienuc)` subtracts it; `ydot(i_ir)`/`ydot(i_optical)` carry only ff+KI and recombination, so energy leaves the system whenever `collisional_ionization_switch = 1` (the default). (`recombination_binding_energy_rate` is the mirror case and is fine — that 13.6 eV was never credited to the gas, since photoheating adds only `eps_ion - 13.6 eV`.)
2. **Check 2 passes with the effective-ionized-length metric at -11.06% against its own 5% tolerance** — `testDTypeFront1D.cpp:740-741,761`. The gate is `!eff_ok && !shell_ok`, so only the shell metric's +0.15% produces the PASS. At `n_cell = 128` one cell is 3.4% of the front position, so the surviving metric is cell-quantized to within a cell of its tolerance. Either tighten/justify the eff metric or make the gate an AND.

**Non-blocking**
- **Deleted coverage** — old Checks 1/3/4/5 (thermal-band energy conservation, exp(-tau) reprocessing, photon budget, M1 `|F| <= cE`) removed with nothing replacing them. Check 5 had a documented history of catching `|F|/(cE) ~ 1e12` from a dropped transparent-group energy source.
- **Unremarked non-conservative floor** — `photochemistry.hpp:215-218`. `a401e5db` removed the gas debit along with the comment explaining it, so the `max(..., small_x)` now creates radiation energy with no compensating debit. Negligible in this problem (bands sit far above `small_x`), but worth a line stating the floor is assumed unreachable.

| Phase | Ran | Observed |
|---|---|---|
| Build | `quokka build -d 1d-apple DTypeFront1D` at `a401e5db` | `DTypeFront1D SUCCESS` |
| Golden | shipped `DTypeFront1D.toml` | exit 0. Cavity 0.18%, neutral 1.66%, shell **+0.15%**, eff **-11.06%** (masked by OR-gate) |
| Edge 1 | `photoionize.flux_optical=0.0` | exit 0, passes. `Xi` is built from the input flux, so reference and sim move together |
| Edge 2 | `photoionize.kappa_optical=0.0` | **exit 1, FAILED** — shell -6.64%, eff -15.58%. Removes momentum deposition while `Xi` is unchanged, so the front check *does* constrain radiation pressure |

Round 1 of this review read `c65c4255`; re-verified at `a401e5db`. `a401e5db4 Minor fixes` fixed the `_pi_num_thermal_bands` CMake loop-variable leak (verified: generated headers give DTypeFront1D 2, OneZonePhotoionization 0), so that finding is dropped.

GPU not built (no CUDA/ROCm toolchain here) — `photochemistry.hpp` and `radiation_system.hpp` are base modules, so `/azp run` + `/azp run rocm-quick` still need to be green.

</details>

🤖 AI-assisted review


**Response**:

**Blocking**

1. It goes into chemical potential.
2. Front has some thickness. If either the effective radius or the location of the max density shell matches, then we call it a pass.

**Non-blocking**

- The test now only checks against the KM09 solution.
- Flooring errors are expected.

chongchonghe:
**Verdict:** 2 blocking: the Microphysics pin has to land first, and DTypeFront1D does not exercise the new thermal-band re-emission.
**Checked:** `1d` + `3d` DTypeFront1D; `1d` RadDust / RadDustMG / RadMarshakDust / OneZonePhotoionization on the PR and on `development`; CUDA/HIP CI green on `a4402fc4`. Shell 4.609e18 vs ODE 4.602e18 cm (+0.15%), cavity T 7802 vs 7772 K. Jacobian fix confirmed: OneZonePhotoionization with `network.energy_switch=1` gives energy norm 1.05e-06 on dev (FAIL, tol 1e-6) and 7.61e-07 on the PR.

**Blocking**
1. **Submodule pinned to an unmerged fork branch.** `extern/Microphysics` points to `94a23219`, the head of chongchonghe/Microphysics-fork#5, but `.gitmodules` tracks `photochemistry` (`d0ec9cc7`). Merge #5 first, then repoint to the merge commit.
2. **Test passes without the feature.** `testDTypeFront1D.cpp`. With `THERMAL_DUST_PHOTOCHEMISTRY` forced undefined, it still passes: shell +0.1529% (identical), cavity 7801.7 K, neutral 41.04 K. It needs a check that fails without re-emission, e.g. the integrated `re_thermal` gain vs `(c_hat/c) × ∫` of the cooling terms in a one-zone run.

**Non-blocking**
- **Dust T solve can exit before iterating** in `radiation_system.hpp:1829`. The bare `+ fourPiBoverC` term in `Lambda_compare`, plus the removed `max(compare, |f0|)` fallback (`:1675`), means that with `dust_gas_interaction_coeff=0` and `c_hat·dt·ρ·κ_P < 1e-8` the first check passes and T_d = T_gas is returned without solving. Found by reading the code, not run.
- **`ydot(5)` / `jac(net_ienuc, 5)` still hard-coded** in `thermophotochemistry_handwritten/actual_rhs.H:239,332` and in the caseB copy. This is the same class as the index bug fixed here; use `i_ngamma`.
- **Front check passes on one of two metrics** (`testDTypeFront1D.cpp:761`). `x_eff` misses by −10.6% in both 1D and 3D; only the grid-quantized shell passes (1 cell = 3.4% of x_ode, tolerance 5%).

| Phase | Ran | Observed |
|---|---|---|
| VERIFY | `1d` DTypeFront1D, PR | exit 0; cavity 7802.0 K vs 7772.5 K; neutral 41.08 vs 41.77 K; shell +0.153%; x_eff −10.64% |
| VERIFY 3D | `3d` DTypeFront1D, PR | exit 0, 489 s; same numbers as 1D (3328 / 3200 cells) |
| Mutation | PR with `#undef THERMAL_DUST_PHOTOCHEMISTRY` via `-include` | exit 0; shell +0.1529%, cavity 7801.7 K, neutral 41.04 K |
| RED | OneZonePhotoionization `energy_switch=1`, dev | exit 1; energy norm 1.0528e-06 |
| GREEN | same, PR | exit 0; energy norm 7.612e-07 |
| EDGE | RadDust / RadDustMG / RadMarshakDust / OneZonePhotoionization, dev vs PR | all pass; norms equal to ≤1e-11 (RadDustMG 7.237950253e-4 vs …194e-4) |

</details>

🤖 AI-assisted review

chongchonghe:
@James471 Now you can point Microphysics to the head of the 'photochemistry' branch. Blocking item 2 and non-blocking item 3 make sense to me. 

Since we are not testing VODE anymore, we should drop its support. When you turn on thermal radiation, we should just abort the code if VODE is used.

James471:
@chongchonghe Regarding non-blocking 3: There are 2 ways we calculate the radius of the front. We calculate the effective ionized radius and the location of the cell with maximum density. Ideally they should coincide, but they don't since the front has a finite thickness. So the test we use is that either of these two radii should be within 5% of the analytical solution, therefore I think we should leave it as it is.

chongchonghe:
> @chongchonghe Regarding non-blocking 3: There are 2 ways we calculate the radius of the front. We calculate the effective ionized radius and the location of the cell with maximum density. Ideally they should coincide, but they don't since the front has a finite thickness. So the test we use is that either of these two radii should be within 5% of the analytical solution, therefore I think we should leave it as it is.

Sounds good. Furthermore, non-block 2 is also worth addressing. @James471

